PR #94, opened 2018-05-10
✨ Initial commit for koa-liveness-ping

tzvipm:
Closes https://github.com/Shopify/quilt/issues/92

@TheMallen what are your thoughts on the API? In particular, how do you feel about the user supplying the ping endpoint? I figure that's one way to make this less tied to our internal deployment stack?

### DIFF
--- a/packages/koa-liveness-ping/README.md
+++ b/packages/koa-liveness-ping/README.md
@@ -14,7 +14,7 @@ $ yarn add @shopify/koa-liveness-ping
 
 ```typescript
 import Koa from 'koa';
-import livenessPing from '@shopify/koa-liveness-ping';
+import ping from '@shopify/koa-liveness-ping';
 
 const app = new Koa();
 
@@ -27,4 +27,4 @@ app.use(ping());
 
 This middleware can also be mounted at any path via `koa-mount` ([Learn more](https://github.com/koajs/mount#mounting-middleware)).
 
-**Note:** The `livenessPing` middleware should be the first routing-related middleware, in order to ensure it can respond quickly and reliably.
+**Note:** Mount the liveness ping middleware before all other routing-related middleware in order to ensure it can respond quickly and reliably.
--- a/packages/koa-liveness-ping/README.md
+++ b/packages/koa-liveness-ping/README.md
@@ -1,0 +1,30 @@
+# `@shopify/koa-liveness-ping`
+
+A package for creating a liveness ping middleware for use with koa
+
+A liveness ping is a URL at which your application will respond with a `200` whenever your server is running. It can be used, for example, for liveness checks in kubernetes deployments.
+
+## Installation
+
+```bash
+$ yarn add @shopify/koa-liveness-ping
+```
+
+## Usage
+
+```typescript
+import Koa from 'koa';
+import livenessPing from '@shopify/koa-liveness-ping';
+
+const app = new Koa();
+
+// set up any error or loggin middlewares
+
+app.use(ping());
+
+// other routing middlewares should go after the liveness ping
+```
+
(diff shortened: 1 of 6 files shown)


tzvipm:
@TheMallen I've updated this PR to include your suggested API changes. I also added a test to confirm that the middleware works with `mount`.

packages/koa-liveness-ping/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+```bash
+$ yarn add @shopify/koa-liveness-ping
+```
+
+## Usage
+
+```typescript
+import Koa from 'koa';
+import livenessPing from '@shopify/koa-liveness-ping';
```
cartogram: `LivenessPing` => `ping`

packages/koa-liveness-ping/README.md
```diff
@@ -0,0 +1,30 @@
...
+// set up any error or loggin middlewares
+
+app.use(ping());
+
+// other routing middlewares should go after the liveness ping
+```
+
+This middleware can also be mounted at any path via `koa-mount` ([Learn more](https://github.com/koajs/mount#mounting-middleware)).
+
+**Note:** The `livenessPing` middleware should be the first routing-related middleware, in order to ensure it can respond quickly and reliably.
```
cartogram: I think this reads a little better:

**Note:** Mount the liveness ping middleware before all other routing-related middleware in order to ensure it can respond quickly and reliably.

packages/koa-liveness-ping/README.md
```diff
@@ -0,0 +1,30 @@
+# `@shopify/koa-liveness-ping`
+
+A package for creating a liveness ping middleware for use with koa
```
cartogram: What do you think about this instead? `A koa middleware package for creating a liveness ping.`

packages/koa-liveness-ping/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "@shopify/koa-liveness-ping",
+  "version": "0.0.0",
+  "license": "MIT",
+  "description": "A package for creating a liveness ping middleware for use with koa",
```
cartogram: If you make the description change above, change here also.

packages/koa-liveness-ping/src/test/index.test.ts
```diff
@@ -0,0 +1,35 @@
...
+    const middleware = ping();
+    const nextFn = jest.fn();
+    const ctx = createMockContext({url: '/ping'});
+    await middleware(ctx, nextFn);
+
+    expect(ctx.status).toBe(200);
+    expect(nextFn).not.toHaveBeenCalled();
+  });
+
+  it('is mountable', async () => {
```
cartogram: Just thinking out loud if it makes sense to include this test. I like that we provide the suggestion in the Readme for using `koa-mount`, but I'm just wondering if we want to support it in this way. It is a third party package that we do not manage, additional dev dependency, etc... I am fine either way on it because koa packages are particularly stable and don't think there is any harm in this, I just wanted to mention it here.

packages/koa-liveness-ping/README.md
tzvipm: I think we need a consistent voice and tone on these descriptions. I like the following in this case, actually:

> Create a liveness ping koa middleware

thoughts?

packages/koa-liveness-ping/README.md
tzvipm: on second thought, I use `ping()` in the tests and it looks good :+1:

packages/koa-liveness-ping/README.md
cartogram: Sounds good to me!

PR #98, opened 2018-05-14
✨ export Style, Script, and DOCTYPE

TheMallen:
We were not exporting the `<Script>` or `<Style>` component, and we also did not expose a DOCTYPE const or mention that a user would need to add the doctype directive to have valid HTML.


packages/react-html/README.md
```diff
@@ -28,6 +29,8 @@ export default (ctx, next) => {
 }
 ```
 
+Due to [limitations in react's implementation of HTML](https://github.com/facebook/react/issues/1035), you still need to append the `<!DOCTYPE html>` directive. To assit with this the module also exports a `DOCTYPE` constant.
```
tzvipm: `append` -> `prepend`.

TheMallen: also assit -> assist :P

packages/react-html/README.md
```diff
@@ -78,3 +81,35 @@ Any serializable data that needs to be available from the DOM when the `synchron
...
+/>
+```
+
+### Script
+
+The `<Script />` component lets you render `<script>` tags in your document dynamically as part of your react app.
+
+```ts
+<Script
+  src={'./some-script.js'}
```
cartogram: Change `{'` to `"` and above aswell

PR #101, opened 2018-05-14
💥💅 standardize auth middleware names to match common koa patterns

TheMallen:
This is a **breaking change 💥** to `@shopify/koa-shopify-auth` that aims to bring our middleware naming more in line with how the community names koa middleware.

#### What?
- `createShopifyAuth` renamed to `shopifyAuth`
- `createVerifyRequest` renamed to `verifyRequest`
- update docs

#### Why?
Currently in our documentation and named exports we're encouraging a style that is not really inline with how the rest of the koa community is naming things.

Most koa middleware (eg. `koa-session`, `koa-mount`, `koa-router`) use a factory type interface `session({config})`, but do not name themselves anything to do with that.

This makes our middleware factories a bit weird and ugly by comparison, and feels a bit frictiony with the ecosystem (even if it's a bit more correct to what's actually going on).

**This will necessitate a major version bump**

### DIFF
--- a/packages/koa-shopify-auth/README.md
+++ b/packages/koa-shopify-auth/README.md
@@ -14,39 +14,39 @@ $ yarn add @shopify/koa-shopify-auth
 
 ## Usage
 
-This package exposes `createShopifyAuth` by default, and `createVerifyRequest` as a named export.
+This package exposes `shopifyAuth` by default, and `verifyRequest` as a named export.
 
 ```js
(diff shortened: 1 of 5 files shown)


TheMallen:
@cartogram yes, I agree. Consistency is nice, and doing the breaking change now means we can add any arbitrary config we need easily without breaking anything later.

Issue #61, opened 2018-04-20
Type definition error in @shopify/enzyme-utilities


michelleyschen:
Using `@shopify/enzyme-utilities`
and doing 
```
const wrapper = mount(<ConnectedAccountCard {...mockDefaultProps} />);
trigger(wrapper.find(Button), 'onClick');
```

I get the following typescript error
```
Argument of type 'ReactWrapp
er<Props, never>' is not assignable to parameter of type 'AnyWrapper'.
  Type 'ReactWrapper<Props, never>' is not assignable to type 'CommonWrapper<any, any>'.
    Types of property 'filterWhere' are incompatible.
      Type '(predicate: (wrapper: ReactWrapper<Props, never>) => boolean) => ReactWrapper<Props, never>' is not assignable to type '(predicate: (wrapper:
CommonWrapper<any, any>) => boolean) => CommonWrapper<any, any>'.
        Types of parameters 'predicate' and 'predicate' are incompatible.
          Types of parameters 'wrapper' and 'wrapper' are incompatible.
            Type 'ReactWrapper<Props, never>' is not assignable to type 'CommonWrapper<any, any>'.
```

upon on research I see the type definition is different from the definition in `web`

in `web`
```
function trigger(wrapper: AnyWrapper, keypath: string, ...args: any[]) 
interface AnyWrapper {
  props: CommonWrapper<any, any>['props'];
  length: number;
  instance(): any;
  update(): void;
}
```

in `@shopify/enzyme-utilities`
```
import { ReactWrapper, CommonWrapper } from 'enzyme';
export declare type AnyWrapper = ReactWrapper<any, any> | CommonWrapper<any, any>;
export declare function trigger(wrapper: AnyWrapper, keypath: string, ...args: any[]): any;
```

----

Also I still need to use '@shopify/enzyme-utilities/dist'

tzvipm:
@michelleyschen do you want to try tackling this one? It sounds like you've already got most of a solution underway.

PR #105, opened 2018-05-14
🐛 update the `AnyWrapper` type to included component that has no state.

michelleyschen:
Close https://github.com/Shopify/quilt/issues/61


michelleyschen:
❤️ would we be updating web to use `enzyme-utilities` as well?

PR #108, opened 2018-05-17
Return middleware from proxy package

cartogram:
In effort to maintain a consistent interface for the quilt koa middleware, this PR wraps the `shopify-graphql-poxy` in a function that returns the actual middleware itself.

**Note** this is a breaking change. Usage of the middleware will change as follows:

```
const app = new Koa()

// old way
app.use(graphQLProxy)

// new way
app.use(graphQLProxy())
```

Open to criticism against this change as well.

cc/ @leo-shopify


leo-shopify:
😄 I personally prefer to remove unnecessary wrappers and nesting, so this change goes against my instincts. But I don't think it is a big deal either way in this case.
I like my bike-sheds pink 🤷‍♂️

cartogram:
@tzvipm readme updated ✍️

PR #109, opened 2018-05-17
Hoist all package config

### LINKED ISSUE #25: Make working scoped to a single library a bit cleaner
Right now linting and testing is a pain if you are only working in the scope of one package folder. We should minimize this friction by getting all the commands working in each package with as little config duplication as we can.

TheMallen:
closes #25 

This PR has two primary purposes:
- Hoist all package level config (tsconfig, eslintrc, eslintignore) into the top of the project using symlinks (and update the generators accordingly)
-  Get development flow for working on one package at a time working cleanly. The symlinked config files make this very easy, all we need to do is add the appropriate scripts to each package folder and make sure the `devDependencies` are accurate.

Testing:
- try running `lint`, `build`, `test` in a package folder and also at the root
- try generating a package and note the symlinks and scripts


TheMallen:
note: looks like I messed up the test commands so I need to fix those

TheMallen:
I can see that argument for testing (since you want to make sure you don't break  anything in a dependent package), but why does it matter if linting is across the repo?

Also note that testing/linting will always be full repo on CI.

CONTRIBUTING.md
```diff
@@ -21,7 +21,13 @@ _Note: replace "your-username" with your Github handle_
 Install the project's dependencies (make sure you first have [yarn](https://yarnpkg.com/) installed):
 
 ```
-yarn
+yarn install
+```
+
+Compile the typescript source so your package(s) can depend on eachother locally.
```
tzvipm: should this prompt end with a colon?

packages/jest-dom-mocks/package.json
```diff
@@ -29,6 +32,8 @@
     "lolex": "^2.3.2"
   },
   "devDependencies": {
+    "eslint": "^4.19.1",
+    "eslint-plugin-shopify": "^22.0.0",
```
tzvipm: why 22 here, but 20 everywhere else?

TheMallen: Should be 22 everywhere prob.

TheMallen:
@tzvipm  Yes we should try to keep them all up to date in all the packages.

packages/jest-dom-mocks/package.json
TheMallen: I'll make it 20 now and then we can fix it for all of them.

TheMallen:
Needs to be rebased again, and it seems like nobody really tries to develop in a particular sub package. I'll shelve it for now.

Issue #110, opened 2018-05-18
Create @shopify/koa-metrics


tzvipm:
Performance metrics implemented with DogStatsD (https://docs.datadoghq.com/developers/dogstatsd/) are used widely across services at Shopify. The standard metrics should be documented and following strong conventions.

This package aims to provide standard middlewares and instrumentation tooling for Koa.

PR #111, opened 2018-05-18
✨ Initial commit for koa-metrics

tzvipm:
Closes https://github.com/Shopify/quilt/issues/110.

### DIFF
--- a/packages/koa-metrics/README.md
+++ b/packages/koa-metrics/README.md
@@ -1,5 +1,8 @@
 # `@shopify/koa-metrics`
 
+[![Build Status](https://travis-ci.org/Shopify/quilt.svg?branch=master)](https://travis-ci.org/Shopify/quilt)
+[![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](LICENSE.md) [![npm version](https://badge.fury.io/js/%40shopify%2Fkoa-metrics.svg)](https://badge.fury.io/js/%40shopify%2Fkoa-metrics.svg)
+
 Opinionated performance metric tracking for Koa, implemented with DogStatsD
 
 ## Installation
@@ -36,33 +39,78 @@ The global StatsD metric name prefix; should be provided in `PascalCase`.
 
 The url for the StatsD host; should be provided in the format: `hostname:port`.
 
-## Metrics
+## API
 
-We rely on the [hot-shots](https://github.com/brightcove/hot-shots) npm package to send the metrics.
+An instance of the `Metrics` object will be available on `ctx.metrics` further dfown in the middleware stack.
 
-The global StatsD metric name prefix is provided through the [options object](#options).
+### `Metrics`
+
+#### `.timing(name: string, value: number, sampleRate?: number, tags?: Tags)`
+
+Sends a timing command with the specified `value` in milliseconds.
+
+#### `.histogram(name: string, value: number, sampleRate?: number, tags?: Tags)`
+
+Sends data for histogram stats.
+
+#### `.initTimer(): Timer`
+
+Returns a new `Timer` started at the current `process.hrtime()`
+
+### `Timer`
+
+#### `.stop(): number`
+
+Returns the time, in milliseconds, since the `Timer` was created.
+
+### `Tags`
+
+tags can be a hash:
+
+```
+{
+  name: value,
+  name2: value2
+}
+```
+
+an array of keys:
+
+```
+['tag1', 'tag2']
+```
+
+or an array of key-value pairs, separated by colons:
+
+```
+['name:value', 'name2:value2']
+```
+
+## Intelligent Defaults
+
+The global metric name prefix is provided through the [options object](#options).
 
 This package automatically provides performance metrics for HTTP requests.
 
 ### Standard tags
 
-* `path`
-* `request_method`
-* `response_code`
-* `response_type` (eg. `2xx`, `3xx`, ...)
+- `path`
+- `request_method`
+- `response_code`
+- `response_type` (eg. `2xx`, `3xx`, ...)
 
-### Histogram metrics
+### Default Metrics
 
-### `request_time`
+#### `request_time`
 
 Time to complete a request, from the application perspective.
 
-### `request_queuing_time`
+#### `request_queuing_time`
 
 Time before a request actually started being processed.
 
 This metric is emitted when the application start processing a request. It relies on the presence of a header `X-Request-Start` set by the first HTTP hop.
 
-### `request_content_length`
+#### `request_content_length`
 
 This metric is based on the response header `Content-Length`. Some responses don't provide this header (chunked encoding); in those cases, this will be `undefined`.
--- a/packages/koa-metrics/README.md
+++ b/packages/koa-metrics/README.md
@@ -113,4 +113,4 @@ This metric is emitted when the application start processing a request. It relie
 
 #### `request_content_length`
 
-This metric is based on the response header `Content-Length`. Some responses don't provide this header (chunked encoding); in those cases, this will be `undefined`.
+This metric is based on the response header `Content-Length`. Some responses don't provide this header (chunked encoding); in those cases, this will not be reported.
--- a/packages/koa-metrics/README.md
+++ b/packages/koa-metrics/README.md
@@ -16,14 +16,6 @@ import metrics from '@shopify/koa-metrics';
 
 const app = new Koa();
 
-app.use(metrics(options));
-```
-
-### Options
-
-The passed in `options` object adheres to the following API:
-
-```javascript
 app.use(
   metrics({
     prefix: 'AppName',
@@ -31,6 +23,10 @@ app.use(
   }),
 );
 ```
+
+### Options
+
+The passed in `options` object adheres to the following API:
 
 #### `prefix` (required)
 
@@ -46,14 +42,14 @@ We rely on the [hot-shots](https://github.com/brightcove/hot-shots) npm package 
 
 The global StatsD metric name prefix is provided through the [options object](#options).
 
-This package automatically provides performance metrics for HTTP requests:
+This package automatically provides performance metrics for HTTP requests.
 
 ### Standard tags
 
 * `path`
 * `request_method`
 * `response_code`
-* `response_type`
+* `response_type` (eg. `2xx`, `3xx`, ...)
 
 ### Histogram metrics
 
--- a/packages/koa-metrics/README.md
+++ b/packages/koa-metrics/README.md
@@ -1,8 +1,6 @@
 # `@shopify/koa-metrics`
 
-Performance metrics implemented with DogStatsD (<https://docs.datadoghq.com/developers/dogstatsd/>) are used widely across services at Shopify. The standard metrics should be documented and following strong conventions.
-
-This package aims to provide standard middlewares and instrumentation tooling for metrics in Koa.
+Opinionated performance metric tracking for Koa, implemented with DogStatsD
 
 ## Installation
 
@@ -71,4 +69,4 @@ This metric is emitted when the application start processing a request. It relie
 
 ### `request_content_length`
 
-This metric is based on the response header `Content-Length`. Some responses don't provide this header (chunked encoding), in this case, this will be deduced from `ctx.body` when possible, or else `undefined`.
+This metric is based on the response header `Content-Length`. Some responses don't provide this header (chunked encoding); in those cases, this will be `undefined`.
--- a/packages/koa-metrics/README.md
+++ b/packages/koa-metrics/README.md
@@ -49,9 +49,9 @@ An instance of the `Metrics` object will be available on `ctx.metrics` further d
 
 Sends a timing command with the specified `value` in milliseconds.
 
-#### `.histogram(name: string, value: number, sampleRate?: number, tags?: Tags)`
+#### `.measure(name: string, value: number, sampleRate?: number, tags?: Tags)`
 
-Sends data for histogram stats.
+Sends data for non-timing stats (measurements). We do this through the distributions feature in DogStatsD. You'll need to ensure that your DogStatsD server supports the distributions feature in order to make use of these metrics.
 
 #### `.initTimer(): Timer`
 
--- a/packages/koa-metrics/README.md
+++ b/packages/koa-metrics/README.md
@@ -65,25 +65,13 @@ Returns the time, in milliseconds, since the `Timer` was created.
 
 ### `Tags`
 
-tags can be a hash:
+tags are an object keyed by the name of the corresponding tag. For example:
 
 ```
 {
   name: value,
   name2: value2
 }
-```
-
-an array of keys:
-
-```
-['tag1', 'tag2']
-```
-
-or an array of key-value pairs, separated by colons:
-
-```
-['name:value', 'name2:value2']
 ```
 
 ## Intelligent Defaults
--- a/packages/koa-metrics/README.md
+++ b/packages/koa-metrics/README.md
@@ -1,0 +1,74 @@
+# `@shopify/koa-metrics`
+
+Performance metrics implemented with DogStatsD (<https://docs.datadoghq.com/developers/dogstatsd/>) are used widely across services at Shopify. The standard metrics should be documented and following strong conventions.
+
+This package aims to provide standard middlewares and instrumentation tooling for metrics in Koa.
+
+## Installation
+
+```bash
+$ yarn add @shopify/koa-metrics
+```
+
+## Usage
+
+```javascript
+import Koa from 'koa';
+import metrics from '@shopify/koa-metrics';
+
+const app = new Koa();
+
+app.use(metrics(options));
+```
+
+### Options
+
+The passed in `options` object adheres to the following API:
+
+```javascript
+app.use(
+  metrics({
+    prefix: 'AppName',
+    host: 'some-statsd-host.com:8125',
+  }),
+);
+```
+
+#### `prefix` (required)
+
+The global StatsD metric name prefix; should be provided in `PascalCase`.
+
+#### `host` (required)
+
+The url for the StatsD host; should be provided in the format: `hostname:port`.
+
+## Metrics
+
+We rely on the [hot-shots](https://github.com/brightcove/hot-shots) npm package to send the metrics.
+
+The global StatsD metric name prefix is provided through the [options object](#options).
+
+This package automatically provides performance metrics for HTTP requests:
+
+### Standard tags
+
+* `path`
+* `request_method`
+* `response_code`
+* `response_type`
+
+### Histogram metrics
+
+### `request_time`
+
+Time to complete a request, from the application perspective.
+
+### `request_queuing_time`
+
+Time before a request actually started being processed.
+
+This metric is emitted when the application start processing a request. It relies on the presence of a header `X-Request-Start` set by the first HTTP hop.
+
+### `request_content_length`
+
+This metric is based on the response header `Content-Length`. Some responses don't provide this header (chunked encoding), in this case, this will be deduced from `ctx.body` when possible, or else `undefined`.
--- a/packages/koa-metrics/documentation/decisions/01 - We use hot-shots as our dogstatsd client.md
+++ b/packages/koa-metrics/documentation/decisions/01 - We use hot-shots as our dogstatsd client.md
@@ -1,0 +1,36 @@
+# We use `hot-shots` as our DogStatsD client
+
+## Date
+
+May 18, 2018
+
+## Contributors
+
+* tzvipm
+
+## Summary
+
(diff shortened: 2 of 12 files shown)


packages/koa-metrics/README.md
```diff
@@ -0,0 +1,74 @@
+# `@shopify/koa-metrics`
+
+Performance metrics implemented with DogStatsD (<https://docs.datadoghq.com/developers/dogstatsd/>) are used widely across services at Shopify. The standard metrics should be documented and following strong conventions.
```
TheMallen: I don't think we want to talk about Shopify here or 'should's, we should instead say what the package offers, so maybe something like 

'Opinionated performance metric tracking for Koa, implemented with DogStatsD'

packages/koa-metrics/src/test/index.test.ts
```diff
@@ -0,0 +1,5 @@
+describe('koa-metrics', () => {
+  it('works', () => {
```
TheMallen: I don't like leaving this empty D:

I'd really like to see us at least leave empty its defining the tests we intend to write if we're going to do them in a subsequent PR.

packages/koa-metrics/src/index.ts
```diff
@@ -0,0 +1,66 @@
...
+  host: string;
+}
+
+export default function metrics({prefix, host}: Options) {
+  const statsdHost = host.split(':')[0];
+  const statsdPort = parseInt(host.split(':')[1], 10);
+
+  return async function statsdMiddleware(ctx: Context, next: Function) {
+    const requestStartTime = process.hrtime();
+    // @ts-ignore the types here are correct. TS complains about method
```
TheMallen: What method is potentially undefined? How can it be true that the types are correct and also we need this @ts-ignore?

Can w add a fix for this type definition to our `config/typescript` and open a PR on definitely typed instead?

packages/koa-metrics/src/index.ts
```diff
@@ -0,0 +1,66 @@
...
+      const responseClient = client.childClient({
+        globalTags: {
+          response_code: `${ctx.response.status}`,
+          // example: 2xx, 3xx, ...
+          response_type: `${Math.floor(ctx.response.status / 100)}xx`,
+        },
+      });
+
+      const duration = process.hrtime(requestStartTime);
+      const ms = duration[0] * 1000 + duration[1] / 1e6;
```
TheMallen: nit: 
This seems a little hard to follow without context, I think it reads a lot nicer to use some consts here to make things more readable.
```js
const timeSince = process.hrtime(requestStartTime);
const [seconds, nanoseconds]  = timeSince;

const milliseconds = 
  seconds * MILLIS_PER_SECOND + nanoseconds / NANOS_PER_MILLI
```

packages/koa-metrics/src/index.ts
```diff
@@ -0,0 +1,66 @@
...
+      responseClient.timing('request_time', Math.round(ms));
+
+      responseClient.histogram('request_content_length', ctx.response.length);
+
+      // @ts-ignore According to the hot-shots documentation, callback
```
TheMallen: Same comment here, I'd rather we fix this type locally and open a PR like we do in other places in our codebases than use `@ts-ignore`.

packages/koa-metrics/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Usage
+
+```javascript
+import Koa from 'koa';
+import metrics from '@shopify/koa-metrics';
+
+const app = new Koa();
+
+app.use(metrics(options));
```
lemonmade: I always find it annoying when a usage example does not actually work as-is (options not defined)

packages/koa-metrics/README.md
```diff
@@ -0,0 +1,74 @@
...
+```
+
+### Options
+
+The passed in `options` object adheres to the following API:
+
+```javascript
```
lemonmade: Don't need this part.

packages/koa-metrics/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+The global StatsD metric name prefix; should be provided in `PascalCase`.
+
+#### `host` (required)
+
+The url for the StatsD host; should be provided in the format: `hostname:port`.
+
+## Metrics
+
+We rely on the [hot-shots](https://github.com/brightcove/hot-shots) npm package to send the metrics.
```
lemonmade: Do they need to know this?

packages/koa-metrics/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+The global StatsD metric name prefix is provided through the [options object](#options).
+
+This package automatically provides performance metrics for HTTP requests:
```
lemonmade: Weird to have a colon followed by a heading

packages/koa-metrics/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Standard tags
+
+* `path`
+* `request_method`
+* `response_code`
+* `response_type`
```
lemonmade: I think this could use a little description

packages/koa-metrics/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+### `request_queuing_time`
+
+Time before a request actually started being processed.
+
+This metric is emitted when the application start processing a request. It relies on the presence of a header `X-Request-Start` set by the first HTTP hop.
+
+### `request_content_length`
+
+This metric is based on the response header `Content-Length`. Some responses don't provide this header (chunked encoding), in this case, this will be deduced from `ctx.body` when possible, or else `undefined`.
```
lemonmade: Mentioned in slack, I would not recommend measuring the body as a fallback

packages/koa-metrics/documentation/decisions/01 - We use hot-shots as our dogstatsd client.md
```diff
@@ -0,0 +1,36 @@
...
+Among those, there are 2 libraries listed under `Node.js` with support for `DogStatsD`:
+
+* `hot-shots`
+* `node-dogstatsd`
+
+`hot-shots` appears to be the clear choice between these 2 for the following reasons:
+
+1.  security - It is actively maintained
+2.  reliability - It has a test suite
+3.  performance - It has performance tests
```
lemonmade: Love that you added this 👍

packages/koa-metrics/package.json
```diff
@@ -0,0 +1,32 @@
...
+  },
+  "author": "Shopify Inc.",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/Shopify/quilt.git"
+  },
+  "bugs": {
+    "url": "https://github.com/shopify/quilt/issues"
+  },
+  "homepage": "https://github.com/shopify/quilt/packages/koa-metrics/README.md",
```
lemonmade: This (and all the other ones) are not actual URLs

packages/koa-metrics/src/index.ts
```diff
@@ -0,0 +1,66 @@
...
+    ctx.state.statsd = client;
+
+    const requestQueuingTime = ctx.request.get('X-Request-Start');
+    if (requestQueuingTime) {
+      try {
+        const timeAsInt = parseInt(requestQueuingTime, 10);
+        // @ts-ignore see description of tags in the hot-shots documentation
+        client.timing('request_queuing_time', timeAsInt, {});
+      } catch (err) {
+        // ignore errors here
```
lemonmade: Don't describe what ("ignoring errors"), describe why

packages/koa-metrics/src/index.ts
lemonmade: I don't get it either, I'd rather one cast than losing strong typing on the entire statement

packages/koa-metrics/src/index.ts
```diff
@@ -0,0 +1,66 @@
...
+      port: statsdPort,
+      prefix,
+      globalTags: {
+        path: ctx.path,
+        request_method: ctx.request.method,
+      },
+    });
+
+    // allow later middleware to add metrics
+    ctx.state.statsd = client;
```
lemonmade: It kind of bugs me when we do so much in each middleware. Could we break this up into the three/ four discrete things it is doing (set up client, timing for queuing, request details, maybe split between timing and status)

packages/koa-metrics/src/index.ts
lemonmade: Agreed

packages/koa-metrics/README.md
tzvipm: I think so, since the `hot-shots` package provides all the mechanisms for custom timing, etc.

packages/koa-metrics/README.md
tzvipm: I do for my editor to do proper code highlighting

packages/koa-metrics/src/index.ts
tzvipm: I'd rather the users don't need to call `app.use` on 3/4 different middlewares just for metrics. Any idea what a suitable middleground might be? I'm thinking something like what we have now, but we can add a `skipDefaultMetrics` option?

lemonmade: You can have one middleware that just composes a bunch of functions together

tzvipm: To confirm, then, you're talking more about code style, rather than the middleware's functionality?

lemonmade: Yes

packages/koa-metrics/src/index.ts
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    // allow later middleware to add metrics
+    ctx.state.statsd = client;
+
+    try {
+      await next();
+    } finally {
+      const responseClient = client.childClient({
+        globalTags: tagsForResponse(ctx),
```
tzvipm: I wonder if these tags should be somehow added to metrics logged within the `await next()` call? thoughts @lemonmade ? I could perhaps make a wrapper around the statsd interface that looks at ctx?

lemonmade: I don't understand how you think that would be possible since these tags are probably not set until the very end of the middleware chain for the most part

packages/koa-metrics/src/tags.ts
```diff
@@ -0,0 +1,17 @@
...
+  return {
+    path: ctx.path,
+    request_method: ctx.request.method,
+  };
+}
+
+export function tagsForResponse(ctx: Context) {
+  return {
+    response_code: `${ctx.response.status}`,
+    response_type: `${Math.floor(ctx.response.status / 100)}xx`,
```
lemonmade: Can we have an enum for the different tags you are adding?

packages/koa-metrics/src/index.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  return async function statsdMiddleware(ctx: Context, next: Function) {
+    const client = new StatsD({
+      host: statsdHost,
+      port: statsdPort,
+      prefix,
+      globalTags: tagsForRequest(ctx),
+    });
+
+    let timer;
```
lemonmade: 👎 implicit `any`

packages/koa-metrics/src/timing.ts
```diff
@@ -0,0 +1,25 @@
+import {StatsD} from 'hot-shots';
+import {Context} from 'koa';
+
+export function initTimer(client: StatsD, ctx: Context) {
```
lemonmade: Can you explain how this would be used for timing other than the full response time? Would it be?

packages/koa-metrics/src/timing.ts
```diff
@@ -0,0 +1,29 @@
+import {StatsD} from 'hot-shots';
+import {Context} from 'koa';
+
+export interface Timer {
+  close(responseClient?: StatsD): void;
+}
+
+export function initTimer(client: StatsD, ctx: Context): Timer {
```
TheMallen: While this factory is pretty nice compared to before, I wonder if there are a couple separate concerns here? 
- a generic `Timer` that tells us the high res time in milliseconds
- getting the queue time from the header
- sending the queue time to statsD 
- sending the end time to statsD 

I picture something a bit simpler than this that has the same api

```ts
  () => {
    stop(): number
  }
```

and then the logic to send the request start and end as a separate call

and then you would do

```ts
const requestQueueTime = ctx.request.get(Headers.QueueTime);
client.timing(Tags.QueueTime, requestQueueTime);

const timer = initTimer()

// ...rest of thing

const duration = timer.stop();
responseClient.timing(Tags.EndTime, duration)
```

I feel like this split makes things a bit clearer (and means once we have our own logger that supports timers we can pull timer out and use that).

This would also make it easier for other things in the app to pull in the timer logic and use it to time whatever they care about.

packages/koa-metrics/src/tags.ts
```diff
@@ -0,0 +1,22 @@
+import {Context} from 'koa';
+
+export enum Tags {
+  Path = 'path',
+  RequestMethod = 'request_method',
+  ResponseCode = 'response_code',
+  ResponseType = 'response_type',
```
TheMallen: shouldn't this enum include the queue time and such?

packages/koa-metrics/src/index.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export {Tags} from './tags';
+
+export interface Options {
+  prefix: string;
+  host: string;
+  skipInstrumentation?: boolean;
+}
+
+export default function metrics({
```
TheMallen: should we maybe take a basic `logger` interface (defaulting to console.log) and also output the stats there?

packages/koa-metrics/src/timing.ts
```diff
@@ -0,0 +1,29 @@
+import {StatsD} from 'hot-shots';
+import {Context} from 'koa';
+
+export interface Timer {
+  close(responseClient?: StatsD): void;
```
TheMallen: I think close is a bit weird of a name. If I was going to time something on a stopwatch or my phone I would probably say `end` or `stop` instead of `close` to describe finishing timing.

packages/koa-metrics/src/tags.ts
tzvipm: Those are metrics, whereas these are tags. The idea is that each metric is allowed to have tags associated with it.

packages/koa-metrics/src/index.ts
tzvipm: I think we should do that for development, yeah.

packages/koa-metrics/src/Metrics.ts
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  public closeClient() {
+    // @ts-ignore According to the hot-shots documentation, callback
+    // is not required.
+    this.rootClient.close();
+  }
+
+  /* eslint-disable no-undefined */
+  private parseArgs(args: any[]): [number | undefined, Tags | undefined] {
```
tzvipm: I'm open to a better approach to parsing these args. The issue is that I like the `hot-shots` API, which allows for 2 optional arguments, and I need to parse them and then pass them through tot he underlying `hot-shots` and `log` APIs.

packages/koa-metrics/src/Metrics.ts
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  public histogram(name: string, value: number);
+  public histogram(name: string, value: number, sampleRate: number);
+  public histogram(name: string, value: number, tags: Tags);
+  public histogram(name: string, value: number, sampleRate: number, tags: Tags);
+  public histogram(name: string, value: number, ...args) {
+    const [sampleRate, tags] = this.parseArgs(args);
+    this.log(MetricType.Histogram, name, value, sampleRate, tags);
+    this.client.timing(name, value, sampleRate, tags);
```
tzvipm: timing -> histogram

packages/koa-metrics/src/Metrics.ts
tzvipm: follow-up question: do we even need to support a `sampleRate`? Maybe I should just :fire: this and only support tags?

TheMallen: Not sure if we need to support sample rate. If not we should burn it.

packages/koa-metrics/src/Metrics.ts
```diff
@@ -0,0 +1,134 @@
...
+  }
+
+  public closeClient() {
+    // @ts-ignore According to the hot-shots documentation, callback
+    // is not required.
+    this.rootClient.close();
+  }
+
+  /* eslint-disable no-undefined */
+  private parseArgs(args: any[]): [number | undefined, Tags | undefined] {
```
TheMallen: Are we able to use method overloading and [tuples](https://www.typescriptlang.org/docs/handbook/basic-types.html#tuple) to type this better?

packages/koa-metrics/src/test/Metrics.test.ts
```diff
@@ -0,0 +1,33 @@
+describe('Metrics', () => {
```
TheMallen: Let's write these tests :)

packages/koa-metrics/src/test/index.test.ts
```diff
@@ -0,0 +1,63 @@
+describe('koa-metrics', () => {
+  it.skip('provides a Metrics client as ctx.metrics');
+
+  it.skip('uses the provided prefix for metrics metrics');
```
TheMallen: metrics metrics?

packages/koa-metrics/src/test/Metrics.test.ts
tzvipm: done :)

packages/koa-metrics/src/test/Metrics.test.ts
```diff
@@ -0,0 +1,165 @@
...
+    });
+
+    it('logs tags with timing metrics to the logger in development', () => {
+      withEnv('development', () => {
+        const logger = jest.fn();
+        const metrics = new Metrics(defaultOptions, logger);
+        metrics.timing('foo', 123, {tag: 'value'});
+
+        expect(logger).toHaveBeenCalledTimes(1);
+        expect(logger).toHaveBeenCalledWith('timing foo:123 #tag:value');
```
tzvipm: should we be including the global tags in these logs?

tzvipm:
@TheMallen this should be ready for another review now. Note, also, the use of distributions for `measure` instead of `histogram`. This is our recommended way to send dogstatsd metrics, although it is currently only supported via a private beta with DataDog.

packages/koa-metrics/src/test/Metrics.test.ts
```diff
@@ -0,0 +1,165 @@
...
+      const closeFn = stats.close;
+      expect(closeFn).toHaveBeenCalledTimes(1);
+    });
+  });
+
+  describe('Timer', () => {
+    it('initTimer returns a Timer whose stop method returns an integer', () => {
+      const originalHrTime = process.hrtime;
+      process.hrtime = () => {
+        return [3, 18000000];
```
TheMallen: Can we use some consts for this to make it a bit more readable?

packages/koa-metrics/src/test/Metrics.test.ts
```diff
@@ -0,0 +1,165 @@
...
+      };
+
+      const metrics = new Metrics(defaultOptions);
+      const timer = metrics.initTimer();
+      const elapsedTime = timer.stop();
+
+      expect(elapsedTime).toBe(3018);
```
TheMallen: can we derive this value rather than have the magic number?

packages/koa-metrics/src/test/index.test.ts
```diff
@@ -0,0 +1,296 @@
...
+    it('tags the initial metrics client with the path', async () => {
+      const metricsMiddleware = metrics(defaultOptions);
+      const ctx = createMockContext({
+        url: '/some/path/',
+      });
+
+      await metricsMiddleware(ctx, () => {});
+      expect(MetricsMock.mock.calls[0][0]).toMatchObject({
+        globalTags: {
+          [Tags.Path]: '/some/path/',
```
TheMallen: can we get this value from a const?

packages/koa-metrics/src/test/index.test.ts
```diff
@@ -0,0 +1,296 @@
...
+      const ctx = createMockContext({
+        headers: {
+          'X-Request-Start': '100',
+        },
+      });
+
+      await metricsMiddleware(ctx, () => {
+        expect(MetricsMock.mock.instances[0].timing).toHaveBeenCalledWith(
+          CustomMetrics.QueuingTime,
+          100,
```
TheMallen: same here

packages/koa-metrics/src/test/Metrics.test.ts
TheMallen: I don't *think* so

packages/koa-metrics/src/test/index.test.ts
tzvipm: oops; missed this one

PR #121, opened 2018-05-18
✨ create @shopify/react-shortcuts

### LINKED ISSUE #13: Create @shopify/react-shortcuts
We should create a package to house our shortcut manager library from web.

TheMallen:
closes #13 

This PR adds ` @shopify/react-shortcuts` to quilt. Some small changes were made from the internal version to remove dependencies, restructure tests, and resolve linting issues.


packages/react-shortcuts/src/ShortcutManager/ShortcutManager.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+  return (
+    target.tagName === 'INPUT' ||
+    target.tagName === 'SELECT' ||
+    target.tagName === 'TEXTAREA' ||
+    target.hasAttribute('contenteditable')
+  );
+}
+
+function arraysMatch<T>(first: T[], second: T[]) {
```
TheMallen: This function + using `array.splice` let us cut `isEqual` and `take` from lodash and drop that dependency.

packages/react-shortcuts/src/tests/ShortcutWithRef.tsx
```diff
@@ -0,0 +1,43 @@
...
+    const {node} = this.state;
+    return (
+      <div className="app">
+        <button type="button" ref={this.setRef} />
+        <Shortcut keys={['z']} onMatch={spy} node={node} />
+      </div>
+    );
+  }
+
+  private setRef = (node: HTMLElement | null) => {
```
TheMallen: Using arrow function properties let us drop `@shopify/react-utilities` as a dependency for this package

packages/react-shortcuts/README.md
```diff
@@ -0,0 +1,156 @@
+# `@shopify/react-shortcuts`
+
+![bundle size badge](https://img.shields.io/bundlephobia/minzip/@shopify/react-shortcuts.svg)
```
TheMallen: Adding this badge to our libraries lets developers easily understand the cost of adding our library to their project.

packages/react-shortcuts/README.md
```diff
@@ -0,0 +1,156 @@
...
+
+Declaratively and efficiently match shortcut combinations in your react application.
```
tzvipm: This describes why to use the package, or what its benefit is, but not what *it* is. Perhaps we should establish a clear convention here too. I like the following sentence structure:

> Declarative and efficient library for matching shortcut combinations in React applications.

packages/react-shortcuts/README.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Installation
+
+```bash
+$ yarn add @shopify/react-shortcuts
+```
+
+## Usage
+
+The library exposes three main parts, `<ShortcutProvider />`, `<Shortcut />` and `ShortcutManager`.
```
fandy: Since `ShortcutManager` is only used internally and not by the consumer, should we just not mention it?

packages/react-shortcuts/README.md
```diff
@@ -0,0 +1,156 @@
...
+
+describe('my-component', () => {
+  it('renders a shortcut for f,o,o', () => {
+    const component = mount(<MyComponent />, {
+      context: {shortcutManager: new ShortcutManager()},
+    });
+
+    const shortcut = component.find(Shortcut);
+
+    expect(shortcut.exists()).toBe(true);
```
tzvipm: We generally prefer using `toHaveLength` for better test error messages. Alternatively, I wonder if this test is necessary at all, since the `prop` lookup below will fail if the component doesn't exist.

packages/react-shortcuts/src/Shortcut/Shortcut.tsx
```diff
@@ -0,0 +1,48 @@
...
+    if (node != null) {
+      return;
+    }
+
+    const {shortcutManager} = this.context;
+    this.subscription = shortcutManager.subscribe(this.data);
+  }
+
+  componentWillUnmount() {
+    if (this.subscription) {
```
tzvipm: I'd test for `this.subscription != null` here as per our usual style conventions.

packages/react-shortcuts/src/ShortcutManager/ShortcutManager.tsx
```diff
@@ -0,0 +1,124 @@
...
+    document.addEventListener('keydown', this.handleKeyDown);
+  }
+
+  subscribe(data: Data) {
+    this.shortcuts.push(data);
+    const self = this;
+
+    return {
+      unsubscribe() {
+        const unsubscribeIndex = self.shortcuts.findIndex(
```
tzvipm: :+1:

packages/react-shortcuts/src/ShortcutManager/ShortcutManager.tsx
```diff
@@ -0,0 +1,124 @@
...
+      case 0:
+        this.resetKeys();
+        break;
+      case 1:
+        this.callMatchedShortcut(event);
+        break;
+      default:
+        this.timer = window.setTimeout(() => {
+          this.callMatchedShortcut(event);
+        }, 500);
```
tzvipm: :fire: magic number

packages/react-shortcuts/README.md
```diff
@@ -0,0 +1,156 @@
...
+  */
+  ignoreInput?: boolean;
+  /*
+    a boolean that lets you opt out of swalloing the key event and let it propagate
+  */
+  allowDefault?: boolean;
+}
+```
+
+#### Basic example
```
fandy: Can we also add an example for modifier keys? E.g. `<Shortcut key={['ctrlKey', 't', 'o']} onMatch={() => console.log('foo')} />`, and mention that internally, we are checking the `event.key`?

Also will need to write a matching test for it, and update the `ModifierKeys` enum to contain the updated key names.

packages/react-shortcuts/src/ShortcutManager/ShortcutManager.tsx
tzvipm: nice!

packages/react-shortcuts/src/ShortcutManager/ShortcutManager.tsx
```diff
@@ -0,0 +1,124 @@
...
+  private shortcutsMatched: Data[] = [];
+  private timer!: number;
+
+  setup() {
+    document.addEventListener('keydown', this.handleKeyDown);
+  }
+
+  subscribe(data: Data) {
+    this.shortcuts.push(data);
+    const self = this;
```
tzvipm: I'm not a fan of this pattern, although it is quite common. I'd rather see us desructure `this.shortcuts` above and rely on that being in the closure, rather than suing the `self` keyword.

fandy:
One thing I missed was that this will not actually work for certain modifier keys. Only catching it now
 after working with `event.key` some more in the Vault editor project 😬

packages/react-shortcuts/README.md
TheMallen: Yeah I think having both in both places is the best. Keep in mind a lot of people will probably get to these readmes from the NPM link, which makes it higher value to have them in the README for the package itself.

packages/react-shortcuts/README.md
TheMallen: (see web's app-context mock)

packages/react-shortcuts/src/ShortcutManager/ShortcutManager.tsx
TheMallen: Me neither, looked over it while porting.

fandy:
Sounds good, it's mainly the `ctrlKey`, `altKey`, `metaKey`, `shiftKey`. If you changed the enum to use those values instead, then it would fix the problem because people will be able to pass in the right keys.

TheMallen:
@tzvipm  👍?

Issue #30, opened 2018-04-06
Improve test coverage for jest-mock-apollo


tzvipm:
We need to properly test things like `resolveAllGraphQL`, query resolution, etc. We currently just test that the context doesn't screw things up.

PR #125, opened 2018-05-22
 🏗 changing API for jest-mock-apollo to use factory 

### LINKED ISSUE #123: Consider better approach to jest-mock-apollo schema config
The schema is currently determined by having some default path, then having the user overwrite that path, then lazily instantiating the schema when first called.

Is there a better way to do this? Globally mutable state is pretty much always bad news. Also, we'd like to allow applications to provide the schema itself, rather than a path to a schema. Not all use cases will have the schema in a file.

Perhaps we are better off having a factory here, where the library vends functions that can create a GraphQL client, with all the instantiation logic configurable by the user. For the schema in particular, we could just make them pass a function that we call once lazily, since that will allow for schemas in a variety of shapes.

### LINKED ISSUE #124: Allow passing through cache options for jest-mock-apollo
We need access to basically all the cache options (as shown here: https://github.com/Shopify/web/pull/4402/files#diff-79c4ff104d28ec7dab79444410e619c6R245)

tzvipm:
Closes #123 
Closes #124 

~~One thing I'd like to call out here is the API for `create`. I don't like that we end up wrapping it to hide the implementation. I'd rather expose a function on factory that returns a `createGraphQLClient` function. Maybe something like `koa`'s `app.callback`? Or else we can always export a bound version of the `create` function?~~

Edit: I fixed this by binding it in the constructor.


packages/jest-mock-apollo/README.md
```diff
@@ -10,16 +10,24 @@ $ yarn add @shopify/jest-mock-apollo
 
 ## Setup
 
-This package provides a method, `registerSchema`, which should be called in the jest `setup` file. For example:
+This package provides a factory class, which should be instantiated in a `utilities` file for your project. For example:
```
TheMallen: I find "in a `utilities` file" a bit over prescriptive and also a little vaguely named

packages/jest-mock-apollo/README.md
```diff
@@ -10,16 +10,24 @@ $ yarn add @shopify/jest-mock-apollo
...
 
 ```ts
-registerSchema({
-  schemaBuildPath: './fixtures/schema.graphql',
-  unionOrIntersectionTypes,
+// tests/utilities
+const schemaSrc = fs.readFileSync(
+  path.resolve(__dirname, '../build/schema.graphql'),
+);
+const graphQLClientFactory = new GraphQLClientFactory({
```
TheMallen: Why expose this as a class? Why not just export a factory function that just returns a `createGraphQLClient` function? Then the user wouldn't need to bother with pulling off `create` and we can remove one step from the usage guide.

I also like shipping an example that works out of the box, so we could just inline a simple schema for this.

```ts
// tests/utilities/graphql.ts

import configureMockApollo from '@shopify/jest-mock-apollo';

// You likely want to import this from somewhere  in real production code
const schema = `
  type Product {
    id: Int!
    title: String
  }

  type Query {
    products: [Product]
  }
`;

export const createGraphQLClient = configureMockApollo(schema);
```

packages/jest-mock-apollo/src/GraphQLClientFactory.ts
```diff
@@ -0,0 +1,95 @@
...
+      throw new Error('schema or schemaSrc is required.');
+    }
+    this.schema = proposedSchema;
+
+    this.cacheOptions = cacheOptions;
+
+    this.create = this.create.bind(this);
+  }
+
+  public create(
```
TheMallen: Yeah, seems like we never actually mutate anything on this class or otherwise use it's instances, don't see why we need a class

packages/jest-mock-apollo/src/test/index.test.tsx
```diff
@@ -2,19 +2,26 @@ import * as React from 'react';
...
-  schemaBuildPath: './fixtures/schema.graphql',
+const graphQLClientFactory = new GraphQLClientFactory({
+  schemaSrc: readFileSync(
+    path.resolve(__dirname, './fixtures/schema.graphql'),
+    'utf8',
+  ),
   unionOrIntersectionTypes,
 });
 
+const createGraphQLClient = graphQLClientFactory.create;
```
TheMallen: Should we have separate tests for `graphQLClientFactory` (or whatever it ends up called) then the classes that it outputs? That way we could have one test suite that makes sure the factory configures things correctly, and another that makes sure the mock link functions correctly.

packages/jest-mock-apollo/src/GraphQLClientFactory.ts
```diff
@@ -0,0 +1,95 @@
...
+
+export default class GraphQLClientFactory {
+  private unionOrIntersectionTypes: any[];
+  private schema: GraphQLSchema;
+  private cacheOptions: ApolloReducerConfig;
+
+  constructor({
+    unionOrIntersectionTypes = [],
+    schema,
+    schemaSrc,
```
lemonmade: Why take both?

tzvipm: I thought that we want to allow for passing in an arbitrary schema? Are you saying we should make every consumer build their schema and pass it in?

packages/jest-mock-apollo/src/test/index.test.tsx
tzvipm: Ideally, yes. There's already an issue for improving these tests, but I believe that should be done separately from this PR: https://github.com/Shopify/quilt/issues/30

packages/jest-mock-apollo/README.md
```diff
@@ -10,16 +10,32 @@ $ yarn add @shopify/jest-mock-apollo
 
 ## Setup
 
-This package provides a method, `registerSchema`, which should be called in the jest `setup` file. For example:
+This package provides a function `createGraphQLClientFactory` that can be used to instantiate a `createGraphQLClient` function specific to your project:
```
lemonmade: This sentence reads real weird

packages/jest-mock-apollo/src/factory.ts
```diff
@@ -0,0 +1,81 @@
+import {buildSchema} from 'graphql';
```
lemonmade: `factory` is not a useful name for a file — factory of what? IMO the file name should represent what its purpose is, not how it accomplishes that

packages/jest-mock-apollo/src/factory.ts
```diff
@@ -0,0 +1,81 @@
...
+import {
+  InMemoryCache,
+  IntrospectionFragmentMatcher,
+} from 'apollo-cache-inmemory';
+import ApolloClient from 'apollo-client';
+import {GraphQLMock, GraphQLClientConfig, GraphQLClientOptions} from './types';
+import MockApolloLink from './MockApolloLink';
+import Requests from './Requests';
+
+export type MockGraphQLClient = ApolloClient<any> & {
```
lemonmade: For future discussion, but should we just have a custom GraphQL client class that has the additional fields?

packages/jest-mock-apollo/src/types.ts
```diff
@@ -0,0 +1,19 @@
+import {GraphQLRequest} from 'apollo-link';
+import {GraphQLSchema} from 'graphql';
+import {ApolloReducerConfig} from 'apollo-cache-inmemory';
+
+export type MockGraphQLResponse = Error | object;
+export type GraphQLMock =
+  | {[key: string]: MockGraphQLResponse}
+  | ((request: GraphQLRequest) => MockGraphQLResponse);
+
+export interface GraphQLClientConfig {
```
lemonmade: Why is this in a `types` file instead of alongside where it is actually needed?

packages/jest-mock-apollo/src/types.ts
```diff
@@ -0,0 +1,19 @@
...
+  unionOrIntersectionTypes?: any[];
+  schema?: GraphQLSchema;
+  schemaSrc?: string;
```
lemonmade: IMO it is not good API design, especially initially, to expose two ways of accomplishing identical things. We should pick the thing we know addresses the use cases we consider in scope (in this case, some projects have just the JSON from introspection, some have the schema .graphql file, so the `schema: GraphQLSchema` option is the one they can both use).

tzvipm: fair enough. I was trying to make the user of our package do less work, but I suppose calling `buildSchema` is pretty trivial.

packages/jest-mock-apollo/src/test/index.test.tsx
```diff
@@ -2,19 +2,24 @@ import * as React from 'react';
...
-import {registerSchema, createGraphQLClient} from '..';
+import configureClient from '..';
 import unionOrIntersectionTypes from './fixtures/schema-unions-and-interfaces.json';
 import petQuery from './fixtures/PetQuery.graphql';
 
 // setup
-registerSchema({
-  schemaBuildPath: './fixtures/schema.graphql',
+const createGraphQLClient = configureClient({
+  schemaSrc: readFileSync(
```
tzvipm: hmm... not sure how this is passing tests :(

packages/jest-mock-apollo/README.md
```diff
@@ -13,16 +13,32 @@ $ yarn add @shopify/jest-mock-apollo
...
+    title: String
+  }
+
+  type Query {
+    products: [Product]
+  }
+`;
+const schema = buildSchema(schemaSrc);
+
+const createGraphQLClient = configureClient({schema});
```
TheMallen: `configureClient` is a bit of a weird name 😬, not sure what to call this without being super verbose though.

PR #127, opened 2018-05-22
✨ Adds mount tracking to enzyme-utilities

tzvipm:
* Adding custom `mount` function
* Adding tracking of wrappers via `addMountedWrapper` and `unmountAllWrappers`.
* Tests, refactoring
* docs


packages/enzyme-utilities/src/mount.ts
```diff
@@ -0,0 +1,12 @@
+import {ReactElement} from 'react';
+import {mount as enzymeMount, ReactWrapper, MountRendererProps} from 'enzyme';
+import {addMountedWrapper} from './wrappers';
```
tzvipm: Note: by making the `wrappers` file separate from `mount`, we facilitate testable code via jest mocks: https://github.com/Shopify/quilt/pull/127/files#diff-b57d56e2c24aea42642caf55dc0610b9R8

packages/enzyme-utilities/README.md
```diff
@@ -53,3 +53,24 @@ describe('<MyComponent />', () => {
...
+
+### `function addMountedWrapper(wrapper: ReactWrapper<any, any>): void`
+
+Adds an enzyme wrapper to a tracked list of mounted wrappers. These wrappers can later be unmounted via `unmountAllWrappers` (see below).
+
+### `function unmountAllWrappers(): void`
+
+Unmounts all wrappers previously registered via `addMountedWrapper`. This can be used in a global `afterEach` hook with the following caveat:
+
+> Usually, you can
```
lemonmade: Usually only need one >

packages/enzyme-utilities/package.json
```diff
@@ -23,10 +23,12 @@
   },
   "homepage": "https://github.com/shopify/quilt/packages/enzyme-utilities/README.md",
   "dependencies": {
-    "@shopify/javascript-utilities": "^2.1.0",
     "lodash": "^4.17.5"
   },
   "devDependencies": {
+    "@shopify/javascript-utilities": "^2.1.0",
+    "enzyme": "^3.3.0",
+    "react": "^16.3.2",
```
lemonmade: Need the types as a dependency?

packages/enzyme-utilities/src/mount.ts
lemonmade: Not usually a great reason to do this kind of separation

packages/enzyme-utilities/src/test/trigger.test.tsx
```diff
@@ -0,0 +1,35 @@
+import * as React from 'react';
+import {mount} from 'enzyme';
+import {trigger} from '..';
+import {Toggle} from './components/Toggle';
+import {ActionList, Action} from './components/Actions';
+
+describe('trigger', () => {
+  it('calls functions on props with arguments', () => {
```
lemonmade: I know this is mostly just moved, but this feels super under tested

packages/enzyme-utilities/src/trigger.ts
```diff
@@ -0,0 +1,48 @@
+import {ReactWrapper, CommonWrapper} from 'enzyme';
+import {get} from 'lodash';
+
+export type AnyWrapper =
+  | ReactWrapper<any, any>
+  | ReactWrapper<any, never>
```
lemonmade: Why do you need both types?

packages/enzyme-utilities/src/wrappers.ts
```diff
@@ -0,0 +1,25 @@
+import {ReactWrapper} from 'enzyme';
+
+const mountedWrappers: ReactWrapper<any, any>[] = [];
```
lemonmade: IMO, we should be using a factory that bends this stuff and keeps the mounted wrappers in the closure, I feel this is a bad pattern for library code and it would make your tests more meaningful

packages/enzyme-utilities/src/test/trigger.test.tsx
tzvipm: https://github.com/Shopify/quilt/issues/129

packages/enzyme-utilities/src/trigger.ts
tzvipm: https://github.com/Shopify/quilt/issues/61

packages/enzyme-utilities/README.md
```diff
@@ -56,3 +60,29 @@ describe('<MyComponent />', () => {
   });
 });
 ```
+
+### `function setupMount(): { mount, addMountedWrapper, unmountAllWrappers }
```
TheMallen: If this takes no parameters why do we want it as a factory?

TheMallen: Also, I feel like it's weird to have this here with no explanation of why we feel it's necessary. I know there's a bit of an explanation below about the manual usages, but if I was looking at using this package I don't think I'd be able to tell if / why I needed this.

packages/enzyme-utilities/README.md
```diff
@@ -56,3 +60,29 @@ describe('<MyComponent />', () => {
...
+
+### `function mount<P, S = any>(node: ReactElement<P>, options?: MountRendererProps | undefined): ReactWrapper<P, S>`
+
+A custom `mount` function mirorring `enzyme`'s `mount` function, designed to assist with automatic tracking of mounted nodes. These can later be unmounted via the `unmountAllWrappers` function described below.
+
+#### `function unmountAllWrappers(): void`
+
+Unmounts all wrappers previously registered via `mount` or `addMountedWrapper`. This can be used in a global `afterEach` hook with the following caveat:
+
+> Usually, you can rely on running this once for the entire suite after each test (and its nested hooks) have been run. However, in some cases you may need to run this manually. This is most commonly needed when your component uses timeouts/ animation frames that you mock out for tests, and waiting until after you have restored those globals will lead to their stored identifiers for the timeout or animation frame to be invalid.
```
TheMallen: I sort of feel like the "right" thing to do here is to try to contribute this functionality to enzyme.

packages/enzyme-utilities/src/mount.ts
```diff
@@ -0,0 +1,39 @@
...
+import {mount as enzymeMount, ReactWrapper, MountRendererProps} from 'enzyme';
+
+export function setupMount() {
+  const mountedWrappers: ReactWrapper<any, any>[] = [];
+
+  function addMountedWrapper(wrapper: ReactWrapper<any, any>) {
+    mountedWrappers.push(wrapper);
+  }
+
+  /**
```
TheMallen: Do we need the full explanation in both places?

packages/enzyme-utilities/README.md
lemonmade: There is nothing enzyme can do in these cases; it has no way of knowing when to unmount or what parts of the environment may have changed in such a way that unmounting will have negative side effects

tzvipm: https://github.com/airbnb/enzyme/issues/1689

tzvipm:
I've decided to make an issue in the `enzyme` repo for this functionality, instead of contributing it to `@shopify/enzyme-utilities` (see https://github.com/airbnb/enzyme/issues/1689). I'll offer them to make a PR, assuming they approve of the suggested changes.

PR #128, opened 2018-05-22
Fix graphql-proxy homepage link

cartogram:
If you go [here](https://www.npmjs.com/package/@shopify/koa-shopify-graphql-proxy) and click 'homepage' it doesn't work. This should fix that.


cartogram:
@mallen So actually just noticed a lot of the links are wrong, I'll open another PR that takes care of the rest.

TheMallen:
@cartogram  while you're at it can you change it in the template too? :)

PR #132, opened 2018-05-24
✅ test package name validity

### LINKED ISSUE #37: CI should check if package names match their path
We had a build fail due to this already, so it's probably best to catch before we ever get to shipit.
https://shipit.shopify.io/shopify/quilt/production/deploys/549563.txt

TheMallen:
closes #37 

This PR adds a test to make sure that package names in the repo match their directory name. This should prevent our releases from failing by stopping people from merging packages with malformed setups.

### DIFF
--- a/test/consistent-package-names.ts
+++ b/test/consistent-package-names.ts
@@ -1,0 +1,23 @@
+import {readdirSync, readFileSync} from 'fs';
+import * as path from 'path';
+
+describe('package name', () => {
+  it('matches the path name', () => {
+    const packagesPath = path.resolve(__dirname, '..', 'packages');
+    const packageNames = readdirSync(packagesPath);
+
+    for (const packageName of packageNames) {
+      const packageJSONPath = path.join(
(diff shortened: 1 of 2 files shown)


test/consistent-package-names.ts
```diff
@@ -0,0 +1,23 @@
+import {readdirSync, readFileSync} from 'fs';
+import * as path from 'path';
+
+describe('package name', () => {
+  it('matches the path name', () => {
+    const packagesPath = path.resolve(__dirname, '..', 'packages');
+    const packageNames = readdirSync(packagesPath);
+
+    for (const packageName of packageNames) {
```
tzvipm: this will need to change below too. We need to account for the case where the folder is left over from a different branch, locally.

tzvipm: see the change here: https://github.com/Shopify/quilt/pull/127/files#diff-08f9f004f75cebe0e8288463aea4b843R19

PR #140, opened 2018-05-29
Add troubleshooting note for withRouter type issues

macdonaldr93:
`withRouter` is overly strict about its type definition for wrapped component. Just adding a note in the README to work around this. Hopefully, we can get a fix in `react-router` to loosen the type on the wrapped component.

### DIFF
--- a/packages/react-shopify-app-route-propagator/README.md
+++ b/packages/react-shopify-app-route-propagator/README.md
@@ -72,6 +72,14 @@ export default function() {
 }
 ```
 
+#### `react-router` typescript issues
+
+`withRouter` has a strict type definition for the component it wraps. The component is expected to require the entire WithRouterProps prop definition. If you're getting a type error on the `ShopifyRoutePropagator` component, you can cast it to `any` when passing it in to withRouter.
+
+```javascript
+const Propagator = withRouter(ShopifyRoutePropagator as any);
+```
+
 ### With React Router (`<Route>`)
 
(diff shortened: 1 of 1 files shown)


packages/react-shopify-app-route-propagator/README.md
```diff
@@ -100,3 +100,13 @@ export default function() {
   );
 }
 ```
+
+### Troubleshooting
+
+#### `react-router` typescript issues
+
+`withRouter` has a strict type definition for the component it wraps. The component has to implement the entire `withRouter` interface, rather than a partial implementation. If you're getting a type error on the `ShopifyRoutePropagator` component, you can define it as `any`.
```
TheMallen: > has to implement the entire `withRouter` interface
I think a more accurate version would be `is expected to require the entire WithRouterProps prop definition`

> you can define it as `any`

Technically what we have to do here is casting, so we can say something like 'You can cast it to any when passing it in to withRouter`

macdonaldr93:
@TheMallen updated!

packages/react-shopify-app-route-propagator/README.md
```diff
@@ -100,3 +100,13 @@ export default function() {
   );
 }
 ```
+
+### Troubleshooting
```
tzvipm: Why is this a 3rd level heading? Can we :fire: this and just nest the level 4 heading under the `withRouter` section above?

TheMallen:
@macdonaldr93, could you rebase and fix the heading thing Tzvi mentioned? If so we're good to merge.

macdonaldr93:
Updated!

PR #145, opened 2018-05-31
🐛 ensure relevant types are exported from the package

tzvipm:


tzvipm:
Thanks @kvendrik, I did want to export the other types, just in case they're needed by any other libraries, although they are indeed not needed for that PR in `web`. I like the export style you suggested; I didn't even know you could do that for a default export. I've updated the PR with your suggested changes.

PR #146, opened 2018-05-31
📝 Auto-generate root readme file

tzvipm:
Closes https://github.com/Shopify/quilt/issues/149

Just run `yarn generate docs` to generate the readme file.

- [x] merge this into the plopfile and :fire: ejs dependency.


package.json
```diff
@@ -8,6 +8,7 @@
   },
   "scripts": {
     "publish": "lerna publish",
+    "precommit": "yarn plop docs",
```
tzvipm: @TheMallen what are your thoughts about doing this in a precommit hook? Should we instead just run a test for it in CI? Or just make people run `yarn plop docs` when they think of it?

TheMallen: `yarn plop docs` + test is nicer I think

tzvipm:
I found the easiest way to test this is just to generate the docs in CI, then assert there are no git changes.

PR #147, opened 2018-05-31
Fix display: none styling on HTML component

lemonmade:
Will fix https://github.com/Shopify/web/issues/4527. When this was brought over, the condition on when to hide the body was incorrectly set to be based on blocking scripts. This PR fixes this by having an explicit prop for it instead, which makes this a little less magical and a little more flexible for consumers.

cc/ @ismail-syed when I applied this locally it fixed the issue /shrug

### DIFF
--- a/packages/react-html/src/HTML.tsx
+++ b/packages/react-html/src/HTML.tsx
@@ -32,6 +32,7 @@ export interface Props {
   scripts?: Asset[];
   headData?: {[id: string]: any};
   data?: {[id: string]: any};
+  hideForInitialLoad?: boolean;
 }
 
 export default function HTML({
@@ -41,6 +42,7 @@ export default function HTML({
   styles = [],
   data = {},
   headData = {},
+  hideForInitialLoad,
 }: Props) {
   const markup = renderToString(children);
   const helmet = Helmet.renderStatic();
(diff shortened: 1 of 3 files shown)


packages/react-html/src/HTML.tsx
```diff
@@ -82,12 +84,9 @@ export default function HTML({
...
   });
 
-  /* eslint-disable no-process-env, no-undefined */
-  const bodyStyles =
-    process.env.NODE_ENV === 'development' && blockingScripts.length > 0
-      ? {display: 'none'}
-      : undefined;
-  /* eslint-enable no-process-env, no-undefined */
+  /* eslint-disable no-undefined */
+  const bodyStyles = hideForInitialLoad ? {display: 'none'} : undefined;
```
ismail-syed: Glad it worked for you! When I tested I just replaced `blockingScripts` with `scripts` in the ternary check straight from `node_modules`. 

Why do we need consumers to explicitly set `hideForInitialLoad`. When would someone every want that flash of unstyled content in dev?

lemonmade: It's less about them wanting the flash, and more about when to do this being figured out correctly. There is nothing wrong with putting all assets as blocking assets in development or production, but this only works if you put them in as non-blocking ones. It's also not always needed in dev, since some things do not load any styles, but do load scripts.

PR #156, opened 2018-06-12
AddressFormatter

nimzco:
Extract core functionality of [`Address`](https://github.com/Shopify/web/blob/master/packages/@shopify/react-address/src/Address/Address.tsx) and [`AddressForm`](https://github.com/Shopify/web/blob/master/packages/@shopify/react-address/src/AddressForm/AddressForm.tsx) components into an `AddressFormatter` class that exposes the following methods:

```javascript
class AddressFormatter(private locale: string)

async getCountry(countryCode: string): Promise<Country>
async getCountries(): Promise<Country[]>
updateLocale(locale: string): FieldName[][]
async getOrderedFields(countryCode): FieldName[][]
async format(address: Address): string[]
async getTranslationKey(countryCode: string, key: FieldName): string
```

See [README](https://github.com/Shopify/quilt/pull/156/files#diff-65aaf21f001e24a1891ade0241122aff) for full details.


packages/address/README.md
```diff
@@ -0,0 +1,112 @@
+# `@shopify/address`
+
+[![Build Status](https://travis-ci.org/Shopify/quilt.svg?branch=master)](https://travis-ci.org/Shopify/quilt)
+[![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](LICENSE.md) [![npm version](https://badge.fury.io/js/%40shopify%2Faddress.svg)](https://badge.fury.io/js/%40shopify%2Faddress)
```
nimzco: Not sure what I have to do about this? cc @TheMallen

nimzco:
@clauderic @TheMallen, if none of you have the time to look at this PR, it's totally fine, but can you redirect me to someone who has the time to look at it?

packages/address/README.md
nimzco: 😍

packages/address/src/tests/AddressFormatter.test.ts
```diff
@@ -0,0 +1,203 @@
+import {fetch} from '@shopify/jest-dom-mocks';
+import {Address, FieldName} from '../types';
+import AddressFormatter from '..';
+import {JapanJa, FranceFr, CanadaEn, CanadaFr} from './fixtures';
+
+const countries = [JapanJa, CanadaFr, FranceFr];
+
+beforeEach(() => {
+  fetch.mock(
```
tzvipm: there shouldn't need to be any fetch mocking here if these are unit tests, in which case we could just mock out the loader utility. If you are writing e2e tests, which could also be nice, then do we need to test the loader utility independently at all?

packages/address/src/tests/utilities.test.ts
```diff
@@ -0,0 +1,5 @@
+import {renderLineTemplate} from '../utilities';
+
+describe('renderLineTemplate()', () => {
+  it('TODO', () => {});
```
tzvipm: I'd rather see a list of what you plan to test via `it.skip`, then we can decide if it makes sense to write the tests in this PR or defer it for another one. We're trying not to include any untested library code though.

nimzco: Oh shoe, I'm sorry, I completely missed that one... 🤦‍♂️

packages/address/src/utilities.ts
```diff
@@ -0,0 +1,76 @@
...
+  let lineIsEmpty = true;
+  result.forEach(key => {
+    const addressKey = key.replace('{', '').replace('}', '') as FieldName;
+
+    if (address[addressKey]) {
+      lineIsEmpty = false;
+    }
+
+    switch (addressKey) {
+      // Country
```
tzvipm: can we 🔥 these comments?

packages/address/README.md
```diff
@@ -0,0 +1,118 @@
...
+![npm bundle size (minified + gzip)](https://img.shields.io/bundlephobia/minzip/%40shopify/shopify%2address.svg)
```
tzvipm: ❤️

tzvipm:
97.19% of diff target is 🌶

packages/address/src/types.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export interface Address {
+  company?: string;
+  firstName?: string;
+  lastName?: string;
+  address1: string;
+  address2: string;
+  city: string;
+  // Province code
```
tzvipm: do we need this comment?

packages/address/src/types.ts
```diff
@@ -0,0 +1,48 @@
...
+  province?: string;
+  zip: string;
+  // Country code of format: ISO 3166-1 alpha-2
```
tzvipm: do we really care what the address format is? Maybe just specify this in the readme?

packages/address/src/utilities.ts
```diff
@@ -0,0 +1,74 @@
...
+  '{phone}': FieldName.Phone, // eslint-disable-line id-length
+  '{company}': FieldName.Company, // eslint-disable-line id-length
+};
+
+/*
+ * Given a line template (eg. "{city} {zip}, {province}")
+ * Returns a string with the correct replacement
+ * If all replacement fields are empty, an empty string will be returned
+ */
+export function renderLineTemplate(
```
tzvipm: I'm not sure how I feel about the inline documentation. cc @TheMallen. My thoughts are that we should do some combination of:
1. clarifying the names of the functions to more accurately describe what they do
2. adding documentation for any user-facing functions tot he readme file.

packages/address/src/types.ts
nimzco: Yes, it's to specify that it's not the name of the province but the code following the [ISO_3166-2 format](https://en.wikipedia.org/wiki/ISO_3166-2).
I'll update the code mentioning this.

packages/address/src/types.ts
nimzco: True, that really needs to be in the README, but I thought it was cool to have it in the code. 😄

tzvipm: I think I'd rather just call the variable `countryCode` or type is as an alias: `type CountryISOCode = string`

packages/address/src/types.ts
nimzco: I actually just realized that it's not entirely true.
I've opened an [issue on countryDB](https://github.com/Shopify/country_db/issues/169) for validation.

packages/address/src/utilities.ts
TheMallen: I don't actually think this comment is helpful, the typescript code is pretty self explanatory. We can see just from the types that it's got the interface

```typescript
function renderLineTemplate(country: Country, template: string,  address: Address): string
```
The only non-obvious thing is that the return value is an empty string when the replacements are empty.

All that said, in general I'd be open to discussing autogenerating real docs from tsdoc comments in quilt libraries in the future if we thought it was really helpful.

packages/address/src/AddressFormatter.ts
```diff
@@ -0,0 +1,140 @@
...
+    this.locale = locale;
+    COUNTRIES_CACHE[this.locale] = {};
+  }
+
+  updateLocale(locale: string) {
+    this.locale = locale;
+    COUNTRIES_CACHE[this.locale] = {};
+  }
+
+  async getCountry(countryCode: string): Promise<Country> {
```
TheMallen: I suppose we can't know what all  the country codes are ahead of time?

It would be nice if we could use a type the caches and these functions a bit stronger.

packages/address/src/loader.ts
```diff
@@ -0,0 +1,39 @@
+import {Country} from './types';
+
+const API_VERSION = '1';
+const API_HEADERS = {
+  Accept: 'application/json',
+  'Content-Type': 'application/json',
+  'API-VERSION': API_VERSION,
+};
+const COUNTRY_SERVICE_URL = 'https://country-service.shopifycloud.com';
```
TheMallen: Is this publicly available?

nimzco:
On behalf of @Shopify/i18n, @Mangoov and I are 👍 to have this and the `country-service` open source. 🚀

nimzco:
I'm not sure why the tests fail.
Locally I get
```
    Cannot find module '@shopify/jest-dom-mocks' from 'AddressFormatter.test.ts'
```

Even though it is in the package.json. Any clue @tzvipm @TheMallen?

tzvipm:
For internal reference: https://github.com/Shopify/web/issues/4605

nimzco:
Oh well, tests are green now! 😄

TheMallen:
The tests were likely failing for you locally because you hadnt run build.

PR #157, opened 2018-06-14
✨Add react-import-remote

9at8:
Adds import remote, which is used for asynchronously loading external scripts.


packages/react-import-remote/README.md
```diff
@@ -0,0 +1,135 @@
...
+}
+```
+
+#### Props
+
+**source**
+Source of the script to load the global from
+
+**preload**
+Generates a preload link tag for the source’s domain using [`@shopify/react-preconnect`](https://github.com/Shopify/quilt/tree/master/packages/react-preconnect)
```
9at8: This is false. It generates a `preconnect` link, and not a `preload` link. https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types

tzvipm:
Is this PR ready for review? I notice CI is failing and there seems to be an outstanding comment. Were you waiting for some feedback first?

TheMallen:
@9at8 should we review this? Not sure if it's waiting for something or what

9at8:
Yes this is done. Needs :eyes:

packages/react-import-remote/README.md
tzvipm: @9at8 did you mean to fix this?

9at8: I don't know if we're supposed to add preconnect links or preload links. I feel that we should do preload, but I'm not sure.

TheMallen: both seem good tbh

lets make this reflect the code we have now though and we can do actual testing to see which is better diwn the line (which we should open issues for)

packages/react-import-remote/src/load.ts
```diff
@@ -0,0 +1,59 @@
...
+
+function scriptTagFor(source: string) {
+  const node = document.createElement('script');
+  node.setAttribute('type', 'text/javascript');
+  node.setAttribute('src', source);
+  return node;
+}
+
+function appendScriptTag(scriptTag: HTMLScriptElement) {
+  document.head.appendChild(scriptTag);
```
TheMallen: It feels really weird that we do regular old DOM manipulation for this but we also render `<preconnect>` which depends on `helmet`

packages/react-import-remote/src/ImportRemote.tsx
```diff
@@ -0,0 +1,35 @@
...
+  getImport(window: Window): Imported;
+  onImported(imported: Imported): void;
+}
+
+export default class ImportRemote extends React.PureComponent<Props, never> {
+  async componentDidMount() {
+    const {source, getImport, onError, onImported} = this.props;
+
+    try {
+      const imported = await load(source, getImport);
```
TheMallen: what about if this is ever rerendered with a different url? usually with these kinds of side-effecty components you need to handle `componentWillUpdate` or `componentDidUpdate`

TheMallen: note that render would get called if that happened so we would add a new preconnect but not the script tag. If we used helmet for both we would be able to make them both work the same

9at8: Yeah that's true. Right now, if this rerenders, then the old Preconnect link is removed (should be?) and the new one is added. At the same point, even if we manually insert script tags, it won't remove the old ones.

9at8:
@tzvipm idk why this is failing 🙈

packages/react-import-remote/README.md
```diff
@@ -0,0 +1,135 @@
+# `@shopify/react-import-remote`
+
+[![Build Status](https://travis-ci.org/Shopify/quilt.svg?branch=master)](https://travis-ci.org/Shopify/quilt)
+[![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](LICENSE.md) [![npm version](https://badge.fury.io/js/%40shopify%2Freact-import-remote.svg)](https://badge.fury.io/js/%40shopify%2Freact-import-remote.svg) [![npm bundle size (minified + gzip)](https://img.shields.io/bundlephobia/minzip/@shopify/react-import-remote.svg)](https://img.shields.io/bundlephobia/minzip/@shopify/react-import-remote.svg)
+
+Provides a utility function and React component to asynchronously load scripts and extract the newly-added global
```
tzvipm: I'm not 💯 on this description. Perhaps something like this:

> Asynchronous script loading for React

Also, you'd need to update in the `package.json` too

packages/react-import-remote/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "@shopify/react-import-remote",
+  "version": "0.0.0",
```
tzvipm: :+1:

packages/react-import-remote/package.json
```diff
@@ -0,0 +1,36 @@
...
+  },
+  "bugs": {
+    "url": "https://github.com/shopify/quilt/issues"
+  },
+  "homepage": "https://github.com/Shopify/quilt/blob/master/packages/react-import-remote/README.md",
+  "dependencies": {
+    "@shopify/react-preconnect": "^0.1.0"
+  },
+  "devDependencies": {
+    "@shopify/javascript-utilities": "^2.1.0",
```
tzvipm: @TheMallen what is our plan around this? I noticed you used `lodash-decorators` directly in `react-form-state`.

packages/react-import-remote/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "enzyme": "^3.3.0",
+    "typescript": "~2.7.2"
```
tzvipm: ⬆️

packages/react-import-remote/package.json
```diff
@@ -0,0 +1,36 @@
...
+  },
+  "peerDependencies": {
+    "react": "^16.3.2"
```
tzvipm: ⬆️

TheMallen:
that just means you need to run the doc generation command

packages/react-import-remote/src/load.ts
TheMallen: that's fair

TheMallen: So I guess the question becomes whether we want to expose `load` at all. A quick grep of our codebase shows us only using the component version, so we should probably not export `load`.

With `load` as an implementation detail of `import-remote` we can safely change the implementation to use `<Helmet />` if we want down the line.

PR #158, opened 2018-06-15
Upgrade GraphQL dependencies

lemonmade:
This PR updates some GraphQL dependencies for Quilt in order to work with the new GraphQL stuff we are doing in graphql-tools-web. It also moves some things to peer dependencies that a project would have to manually install in order for things to work at all, mostly to avoid unnecessary nesting.

### DIFF
--- a/packages/jest-mock-apollo/package.json
+++ b/packages/jest-mock-apollo/package.json
@@ -23,16 +23,20 @@
   },
   "homepage": "https://github.com/Shopify/quilt/blob/master/packages/jest-mock-apollo/README.md",
   "dependencies": {
-    "apollo-cache-inmemory": "^1.1.12",
-    "apollo-client": "^2.2.8",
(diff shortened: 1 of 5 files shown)


packages/jest-mock-apollo/package.json
```diff
@@ -23,16 +23,20 @@
...
   "devDependencies": {
     "@shopify/react-apollo": "^2.0.7",
+    "apollo-cache-inmemory": "^1.2.4",
+    "apollo-client": "^2.3.4",
     "react": "^16.2.0",
     "react-dom": "^16.2.0",
     "typescript": "~2.7.2"
+  },
+  "peerDependencies": {
+    "apollo-cache-inmemory": "^1.2.4",
```
tzvipm: :+1:

lemonmade:
I did 🎩 in combination with an updated sewing-kit and it seemed to work fine 👍

Issue #90, opened 2018-05-08
Create @shopify/logger


tzvipm:
## Problem Summary
Logging is an important tool in a developer’s arsenal. It can be used to debug applications both in development and production to better diagnose problems. It can be used, along with statistical analysis, to gain insight into the activity of even the most complex distributed systems.

As Shopify becomes increasingly distributed and complex, it will become imperative that logging is something that we do right.

We eventually want to build a great experience with a nice API, filtering, and a slick UI. By getting the API right and creating a nice separation of concerns, we allow for ease of continuous improvement for things like filtering, etc.

In development, for web application, there are concerns around simultaneous, asynchronous requests. We want immediate feedback, but, when many requests are happening at the same time, the UI is hard to get right.

In production web applications, stdout is our interface into splunk (via fluentd). We want our logs to have context about our application, and perhaps we want to aggregate these logs, so that we send only a single JSON payload to splunk per request.

For CLI applications, we’d like to avoid duplicating work in creating a nice UI for logging. It would be nice if our logger was not specifically tied to server applications.

## Proposed Solution
* Logging is split into 2 stages:
    * Logging (API emits intermediate data)
    * Formatting (converts intermediate data into desired format)
* We use JSON as the intermediate logging format (pino / bunyan / bole)
* The core logging API will stay as close as possible to https://console.spec.whatwg.org
    * If API exists, we’ll use it:
        * `.time`, `.timeEnd`
    * If API doesn’t exist, we’ll use idiomatic naming:
        * `.tag`, `.tagEnd`
* Buffering for production is handled by `@shopify/koa-logger`

## Why?
* By splitting logging into 2 stages, we conceptually decouple the concept of an API from the concept of a logging UI. This allows us to focus on perfecting the 2 independently of each other.
* By logging to a JSON format, we can easily create aggregation or filtering logic that gets applied before the logs get formatted. Again, we can do all of this independently of the logger API mechanism.
* By staying close to the console specification for our API, we reduce friction for new developers into our code base. We also defer to the W3C arguments or differences of opinion for how things like tagging, grouping, or timing should be done. Sometimes it’s nice to have a decision made for you in a consistent way, even if it’s not your favorite outcome.

## References
Inspiration was taken from many sources, but the primary influences were:
http://getpino.io/#/
https://github.com/klauscfhq/signale
https://console.spec.whatwg.org/

tzvipm:
## On separation of logging and formatting

There are several possible solutions here:
[...]

PR #163, opened 2018-06-26
🎉  [New Package] @shopify/logger

tzvipm:
Closes https://github.com/Shopify/quilt/issues/90.

I think this is ready for a first round of review now.


tzvipm:
Aww yeah, 100% coverage baby 🎉 📔

packages/logger/README.md
```diff
@@ -0,0 +1,53 @@
...
+  name: 'my-logger',
+  formatter: new ConsoleFormatter(),
+});
+```
+
+**Notice** the two required arguments: `name` and `formatter`. The `Logger` constructor adheres to the following interface:
+
+```typescript
+interface LoggerOptions {
+  formatter: Formatter;
```
TheMallen: Should this have a default?

packages/logger/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+a `Formatter` is simply an object that implements the `Formatter` interface:
+
+```typescript
+export interface Formatter {
+  next(entry: FormatEntry): any;
+}
+```
+
+In addition to writing your own formatter, See [Fromatters](#formatters) below for a list of provided formatters.
```
TheMallen: Fromatters -> Formatters

packages/logger/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+### Logging
```
TheMallen: There is no list of formatters 😅

packages/logger/README.md
```diff
@@ -0,0 +1,53 @@
...
+## Usage
+
+### Instantiation
+
+A logger is created by passing in an `options` object to the `Logger` constructor.
+For example:
+
+```typescript
+const logger = new Logger({
+  name: 'my-logger',
```
TheMallen: What is name for?

packages/logger/src/formatters/ConsoleFormatter.ts
```diff
@@ -0,0 +1,77 @@
...
+import {LogLevel, FormatEntry, Formatter} from '../types';
+
+export class ConsoleFormatter implements Formatter {
+  private startTime: [number, number];
+
+  constructor() {
+    this.startTime = process.hrtime();
+  }
+
+  next(entry: FormatEntry) {
```
TheMallen: This interface feels a bit weird, I get that it's formatting the next log, but `next` almost feels like it should be yielding the  next data not parsing the next thing.

Maybe we should go with something more like `format`?

packages/logger/src/Logger.ts
```diff
@@ -0,0 +1,34 @@
+import {LogEntry, LogLevel, Value, Formatter} from './types';
+
+export interface LoggerOptions {
+  formatter: Formatter;
+  name: string;
+}
+
+export class Logger {
```
TheMallen: this class looks super clean! 💖

packages/logger/src/formatters/ConsoleFormatter.ts
```diff
@@ -0,0 +1,77 @@
...
+    const timeDiff = process.hrtime(this.startTime);
+    const msDiff = timeDiff[0] * 1000 + timeDiff[1] / 1000;
```
TheMallen: This feels like a totally different level of abstraction from the rest of the method to me. I'd probably at least kill the literals in favour of named constants. It's fine for now but once we have a couple different formatters we might also want to pull it out into a utility function that we can use in all of them. 

```
const timeDiff = msDiff(this.startTime);
```

or even

```
const timestamp = timestampFrom(this.startTime);
```

packages/logger/src/formatters/ConsoleFormatter.ts
```diff
@@ -0,0 +1,77 @@
...
+
+function configForLogLevel(logLevel: LogLevel) {
+  if (logLevel === LogLevel.Warn) {
+    return {
+      // eslint-disable-next-line no-console
+      logFn: console.warn,
+      icon: warning,
+      chalkColor: chalk.yellow,
+    };
+  } else if (logLevel === LogLevel.Critical) {
```
TheMallen: `else if` is really redundant here when we are returning in our forks, it might be more readable if we just have a second if.

packages/logger/src/formatters/ConsoleFormatter.ts
```diff
@@ -0,0 +1,77 @@
...
+    const timeDiff = process.hrtime(this.startTime);
+    const msDiff = timeDiff[0] * 1000 + timeDiff[1] / 1000;
+    const timestamp = chalk.magenta(`+${prettyMs(msDiff)}`);
+
+    const {level, payload, scopes} = entry;
+
+    let logMsg = `${payload} ${timestamp}`;
+
+    if (level === LogLevel.Critical) {
+      const {message, stack} = payload as Error;
```
TheMallen: Why do we need to cast here?  Can we use a discriminator instead?

packages/logger/src/formatters/ConsoleFormatter.ts
```diff
@@ -0,0 +1,77 @@
...
+      spaces(8 - level.length),
+      logMsg,
+    );
+  }
+}
+
+function spaces(num: number) {
+  return (spaces as any).memoize[num] || new Array(num).fill(' ').join('');
+}
+(spaces as any).memoize = {};
```
TheMallen: A few things we can change here:
- Use [`string.repeat()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat) over this array stuff
- I don't think this memoization is working right now
- We can kill the `any` and `as` pretty easily by using `Map`

```typescript
const spacesCache = new Map<number, string>();

function spaces(number: number) {
  const cachedString = spacesCache.get(number);

  if (cachedString) {
    return cachedString;
  }

  const newString = ' '.repeat(number);
  spacesCache.set(number, newString);
  return newString;
}
```

packages/logger/src/formatters/ConsoleFormatter.ts
```diff
@@ -0,0 +1,77 @@
...
+    const msDiff = timeDiff[0] * 1000 + timeDiff[1] / 1000;
+    const timestamp = chalk.magenta(`+${prettyMs(msDiff)}`);
+
+    const {level, payload, scopes} = entry;
+
+    let logMsg = `${payload} ${timestamp}`;
+
+    if (level === LogLevel.Critical) {
+      const {message, stack} = payload as Error;
+      logMsg = message;
```
TheMallen: We don't want the timestamp in this case?

packages/logger/src/formatters/ConsoleFormatter.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    const prefix = `[${scopes.join(':')}] `;
+    const {logFn, icon, chalkColor} = configForLogLevel(level);
+    logFn(
+      prefix,
+      icon,
+      spaces(1),
+      chalkColor.bold.underline(level),
+      spaces(8 - level.length),
+      logMsg,
```
TheMallen: Should we just put `timestamp` here with all of these instead? It seems like we basically should always have it right?

packages/logger/src/formatters/ConsoleFormatter.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    let logMsg = `${payload} ${timestamp}`;
+
+    if (level === LogLevel.Critical) {
+      const {message, stack} = payload as Error;
+      logMsg = message;
+
+      if (stack != null) {
+        const [name, ...lines] = stack.split('\n');
+        logMsg = `${name} ${timestamp}\n${lines
```
TheMallen: So if there's a stack we overwrite the whole log message with the stack and add a timestamp? I'm not sure the intention here.

packages/logger/src/formatters/test/ConsoleFormatter.test.ts
```diff
@@ -0,0 +1,114 @@
...
+    const args = errorSpy.mock.calls[0];
+
+    expect(consoleCallInludes(args, errorMsg)).toBe(true);
+    expect(consoleCallInludes(args, LogLevel.Critical)).toBe(true);
+
+    errorSpy.mockReset();
+    errorSpy.mockRestore();
+  });
+
+  it('logs Critical entries using Error.message when Error.stack is null', () => {
```
TheMallen: We don't seem to have a test for the case where we check the formatting when stack isn't null

packages/logger/src/formatters/test/ConsoleFormatter.test.ts
```diff
@@ -0,0 +1,114 @@
...
+    formatter.next({
+      level: LogLevel.Warn,
+      payload: message,
+      scopes: [],
+    });
+
+    expect(warnSpy).toHaveBeenCalledTimes(1);
+    const args = warnSpy.mock.calls[0];
+
+    expect(consoleCallInludes(args, message)).toBe(true);
```
TheMallen: `consoleCallInludes` -> `consoleCallIncludes` :)

packages/logger/src/formatters/test/ConsoleFormatter.test.ts
```diff
@@ -0,0 +1,114 @@
...
+    formatter.next({
+      level: LogLevel.Critical,
+      payload: new Error(errorMsg),
+      scopes: [],
+    });
+
+    expect(errorSpy).toHaveBeenCalledTimes(1);
+    const args = errorSpy.mock.calls[0];
+
+    expect(consoleCallInludes(args, errorMsg)).toBe(true);
```
TheMallen: I'm not sure if this is a big deal but I came here looking to see how things should be outputted, since I found the code for compiling the messages a bit confusing, and not being able to look anywhere and see examples of the output makes that somewhat difficult.

Other than having fixture files or using snapshots I'm not sure of a good way to address that though.

packages/logger/src/test/Logger.test.ts
```diff
@@ -0,0 +1,108 @@
+import {Logger, LogLevel} from '..';
+
+describe('Logger', () => {
+  describe('info()', () => {
+    it('passes the name as the root scope', () => {
```
TheMallen: ✨ V nice tests in this file :)

packages/logger/src/formatters/ConsoleFormatter.ts
TheMallen: (ie we won't see the error's `message`)

packages/logger/README.md
tzvipm: Oh, snap 😂

packages/logger/README.md
tzvipm: Maybe it should be optional too 🤔

packages/logger/src/formatters/ConsoleFormatter.ts
tzvipm: Sorry, I agree thsi is a bit confusing. `name` is probably the wrong identifier here. The first line of the stack trace contains the error type and message. Also, I wanted the timestamp to appear ont he first line, alongside the message, rather than after the stack trace.

packages/logger/src/formatters/test/ConsoleFormatter.test.ts
tzvipm: Yeah, I didn't want to tie the tests to the formatting too much, especially since we use chalk, so we can't perform strict equality checks on just the content without duplicating a lot of the logic. I was trying to sidestep that by checking that pieces of the output are included, since we don't care much about the colors, etc. in this alpha release.

For example, we want the message to be there, but if we later decide to move the `scopes` to after the text, or change the concept of `scopes` to `tags`, I didn't want the tests to be too brittle.

I think that testing the formatting more strictly makes a lot of sense after the 1.0 release, though, where code changes should be purely architectural. Perhaps snapshot testing would be good at that point?

packages/logger/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "url": "https://github.com/shopify/quilt/issues"
+  },
+  "homepage": "https://github.com/Shopify/quilt/blob/master/packages/logger/README.md",
+  "dependencies": {
+    "chalk": "^2.4.1",
+    "log-symbols": "^2.2.0",
+    "pretty-ms": "^3.2.0"
+  },
+  "devDependencies": {
+    "typescript": "~2.7.2"
```
tzvipm: ⬆️

packages/logger/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+- `ConsoleFormatter`
+
+### `ConsoleFormatter`
+
+Formats and prints logs via `console.log`, `console.warn`, and `console.error`.
+
+#### Screenshots
+
+![`ConsoleFormatter` sample error output](./docs/images/ConsoleFormatter__Error.png)
```
TheMallen: oooooooh

packages/logger/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+![`ConsoleFormatter` sample log output on a dark background](./docs/images/ConsoleFormatter__DarkBG.png)
+
+![`ConsoleFormatter` sample log output on a light background](./docs/images/ConsoleFormatter__LightBG.png)
```
TheMallen: aaaaaaaah

packages/logger/src/Logger.ts
```diff
@@ -0,0 +1,35 @@
...
+  name?: string;
+}
+
+export class Logger {
+  private formatter: Formatter;
+  private scopes: string[];
+
+  constructor(options?: LoggerOptions) {
+    this.formatter = (options && options.formatter) || new ConsoleFormatter();
+    this.scopes = options && options.name ? [options.name] : [];
```
TheMallen: ✅

packages/logger/src/formatters/ConsoleFormatter.ts
TheMallen: though tbh I am not sure this isnt a premature optimization

packages/logger/src/formatters/test/ConsoleFormatter.test.ts
TheMallen: 👍

PR #166, opened 2018-06-26
🎉 [New Package] create @shopify/react-form-state

### LINKED ISSUE #17: Create @shopify/react-form
We should pull out our form handling HoC into a package.

TheMallen:
closes #17 

## Create `@shopify/react-form-state`

This PR is absolutely ginormous and I'm sorry. Part a port of `with-form`, part a rewrite of it, part a set of guidelines for how to use it properly. This thing should serve as a solid base for our react forms going forward. 

One nice thing is the vast majority of the PR is actually tests and documentation, which makes me feel all warm and tingly inside.

The PR is already huge, so I really don't want to make any additive changes to it. We also already had some pseudo review cycles on the RFC and the demo PR on our internal repos.

OFC subtractive changes and neutral changes are good :).

For an understanding of all the things this lets you do, check out the README. It's real big and thorough.

@loic-d  @michelleyschen  @9at8  you all wanted to take a look at the final PR for this too.

## Future / Questions
- currently the setup causes some react warnings due to polaris inputs passing stuff like `dirty`, `initialValue` down. I'm not *sure* we actually need to pass those down, so maybe I should filter them off before passing them into the render prop?
- probably we need to change our validation strategy to validate all fields whenever one field is cahnged, @loic-d  did some work on that we can port to here
- we may want to pull out the validator helpers into a second module so that the minimum size for this goes down
- maybe we can make smaller versions of the lodash functions I used here

### DIFF
--- a/packages/react-form-state/README.md
+++ b/packages/react-form-state/README.md
@@ -1,0 +1,54 @@
+# `@shopify/react-form-state`
+
+[![Build Status](https://travis-ci.org/Shopify/quilt.svg?branch=master)](https://travis-ci.org/Shopify/quilt)
+[![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](LICENSE.md) [![npm version](https://badge.fury.io/js/%40shopify%2Freact-formwell.svg)](https://badge.fury.io/js/%40shopify%2Freact-formwell.svg) [![npm bundle size (minified + gzip)](https://img.shields.io/bundlephobia/minzip/@shopify/react-formwell.svg)](https://img.shields.io/bundlephobia/minzip/@shopify/react-formwell.svg)
+
+Manage react forms tersely and safely-typed with no magic.
+
+## Installation
+
+```bash
+$ yarn add @shopify/react-form-state
+```
+
+## Usage
+
+### `<FormState />`
+
+The default component exported by this library is `<FormState />`.
+
+```typescript
+import FormState from '@shopify/react-form-state';
+```
+
+```typescript
+// Fields here refers to the inferred type of your initialValues object
+interface Props<Fields> {
+  initialValues: Fields;
+  validators?: Partial<ValidatorDictionary<Fields>>;
+  onSubmit?: SubmitHandler<Fields>;
+  children(form: FormDetails<Fields>): React.ReactNode;
+}
+```
+
+Its only mandatory props are `initialValues` and `children`. The `initialValues` prop is used to infer all the types for the rest of the component, and to generate handlers and field state objects. The `children` prop expects a function of the current state of the form, which is represented by a `FormDetails` object.
+
+```typescript
+<FormState initialValues={myInitialValues}>
+  {({fields, dirty, valid, submitting, errors, reset, submit}) => {
+    return /* some cool ui */;
+  }}
+</FormState>
+```
+
+For detailed explanations of how to use `<FormState />` check out [the guide](/docs/building-forms.md).
+
+### `validators`
+
+The library also makes a number of validation factory functions available out of the box that should help with common use cases, as well as some tools to make building reusable custom validators easy.
+
+```typescript
+import {validate, validators} from '@shopify/react-form-state';
+```
+
+For detailed explanations of the validation uitilities, check out [the validation docs](/docs/validation.md).
--- a/packages/react-form-state/docs/building-forms.md
+++ b/packages/react-form-state/docs/building-forms.md
@@ -1,0 +1,710 @@
+# Building forms with FormState
+
+## Basic
+
+When using `<FormState>`, you express your form's component tree as a function of the generated field state objects provided by `<FormState />` into its `children` function as `formDetails`.
+
+```typescript
+interface FormDetails<Fields> {
+  fields: FieldDescriptors<Fields>;
+  dirty: boolean;
+  valid: boolean;
+  submitting: boolean;
+  errors: ClientError[];
+  reset(): void;
+  submit(): void;
+}
+```
+
+The `formDetails` object passed into `children` contains a `fields` dictionary, which includes all the state and handlers you need to render your form.
+
+```typescript
+interface FieldState<Value> {
+  name: string;
+  initialValue: Value;
+  value: Value;
+  dirty: boolean;
+  error?: any;
+  onBlur(): void;
+}
+```
+
+The simplest way to use `<FormState />` is to pass it some `initialValues` and render some `input`s in its `children` render prop.
+
+```typescript
+import FormState from '@shopify/react-form-state';
+
+function MyComponent() {
+  return (
+    <FormState
+      initialValues={{
+        title: 'Cool title',
+        description: 'Cool product',
+      }}
+    >
+      {formDetails => {
+        const {fields} = formDetails;
+        const {title, description} = fields;
+
+        return (
+          <form>
+            <label htmlFor={title.name}>Title</label>
+            <input
+              {...title}
+              id={title.name}
+              onChange={({currentTarget}) => {
+                // our onChange expects just a value, no event needed
+                title.onChange(currentTarget.value);
+              }}
+            />
+
+            <label htmlFor={description.name}>Title</label>
+            <input
+              {...description}
+              id={title.name}
+              onChange={({currentTarget}) => {
+                description.onChange(currentTarget.value);
+              }}
+            />
+          </form>
+        );
+      }}
+    </FormState>
+  );
+}
+```
+
+## Reducing boilerplate with custom inputs
+
+The previous example has a fair amount of repetition around passing the right handlers to the right inputs. It also doesn't handle rendering errors. We can eliminate this boilerplate by using a custom component with props which match our field descriptor objects.
+
+[@shopify/polaris](https://polaris.shopify.com/components/forms/text-field)'s `<TextField />` component expects props that match our API. So, using `<TextField />` and ES6 [spread](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) syntax we can greatly simplify our forms. The Polaris inputs also support `error` props, allowing us to surface error messages to our users at the field level once we set up validation.
+
+```typescript
+import {TextField} from '@shopify/polaris';
+import FormState from '@shopify/react-form-state';
+
+function MyComponent() {
+  return (
+    <FormState
+      initialValues={{
+        title: 'Cool title',
+        description: 'Cool product',
+      }}
+    >
+      {formDetails => {
+        const {fields} = formDetails;
+
+        return (
+          <form>
+            <TextField label="Title" {...fields.title} />
+            <TextField label="Description" {...fields.description} />
+          </form>
+        );
+      }}
+    </FormState>
+  );
+}
+```
+
+There is also nothing to stop you from using your own custom inputs.
+
+## Submission
+
+Forms usually need some mechanism of submission. `<FormState />` is not opinionated about how this is done, but does provide a built-in mechanism for managing loading state and updating based on errors from the submission process. Most of this is done through the `onSubmit` property.
+
+```typescript
+  onSubmit={async (fields) => {
+    const result = await updateProduct(fields);
+
+    if (isErrorResult(result)) {
+      return result.errors;
+    }
+  }}
+```
+
+When given an [async](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function) function (or one that returns a [promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function), the `children` function will be rerendered with `submitting: true` until the promise resolves or rejects. Once it has returned the function will be re-rendered with `submitting: false` and `errors` either `[]` (if the promise resolved with no value), or with an array of `{field?: string, error: string}` if the promise resolved with an error list.
+
+```typescript
+// api.ts
+
+const fakeErrorResult = {error: [{message: 'api is not real'}]};
+
+// simple async function that fakes a failing api call
+export function fakeApiCall(data: any) {
+  return new Promise(resolve => {
+    setTimeout(() => {
+      resolve(fakeErrorResult);
+    }, 2000);
+  });
+}
+```
+
+```typescript
+// MyPage.ts
+import {fakeApiCall} from './api';
+import {TextField, Button, Banner, FormLayout, Spinner} from '@shopify/polaris';
+import FormState from '@shopify/react-form-state';
+
+export function MyPage() {
+  return (
+    <FormState
+      initialValues={{email: '', password: ''}}
+      onSubmit={async () => {
+        const results = await fakeApiCall();
+        if (results.error) {
+          return results.error;
+        }
+      }}
+    >
+      {({fields, submitting, errors, submit, valid}) => {
+        return (
+          <form onSubmit={submit}>
+            {submitting && <Spinner size="large" color="teal" />}
+
+            {!valid && (
+              <Banner status="critical">
+                <ul>
+                  {errors.map(error => (
+                    <li key={`${error.message}${error.field}`}>
+                      {error.message}
+                    </li>
+                  ))}
+                </ul>
+              </Banner>
+            )}
+
+            <FormLayout>
+              <TextField label="Email" {...fields.email} />
+              <TextField label="Password" {...fields.password} />
+            </FormLayout>
+
+            <Button primary type="submit" disabled={submitting}>
+              Save
+            </Button>
+          </form>
+        );
+      }}
+    </FormState>
+  );
+}
+```
+
+## Resetting
+
+Most forms need to expose a way to reset all fields to their initial values. The `reset()` function passed into `children()` allows you to easily make this available to a user.
+
+```typescript
+// MyPage.ts
+import {TextField, Button} from '@shopify/polaris';
+import FormState from '@shopify/react-form-state';
+
+export function MyPage() {
+  return (
+    <FormState
+      initialValues={{email: '', password: ''}}
+    >
+      {({fields, reset, dirty}) => {
+        return (
+          <form>
+            <FormLayout>
+              <TextField label="Email" {...fields.email} />
+              <TextField label="Password" {...fields.password} />
+            </FormLayout>
+
+            <Button
+              onClick={reset}
+              disabled={!dirty}
+            >
+              Reset
+            </Button>
+          </form>
+        );
+      }}
+    </FormState>
+  );
+```
+
+## Dirty Tracking
+
+Often you want to change some UI behavior based on whether a form has changed from its initial values. `<FormState />` supports this both on the form and individual field level.
+
+```typescript
+// MyPage.ts
+import {TextField, Button} from '@shopify/polaris';
+import FormState from '@shopify/react-form-state';
+
+export function MyPage() {
+  return (
+    <FormState
+      initialValues={{email: '', password: ''}}
+    >
+      {({fields, dirty}) => {
+        const dirtyMarkup = fields.password.dirty && (
+          <p>you have edited this</p>
+        );
+
+        return (
+          <form>
+            <FormLayout>
+              <TextField label="Email" {...fields.email} />
+              <TextField label="Password" {...fields.password} />
+              {dirtyMarkup}
+            </FormLayout>
+
+            <Button
+              primary
+              type="submit"
+              disabled={!dirty}
+            >
+              Save
+            </Button>
+          </form>
+        );
+      }}
+    </FormState>
+  );
+```
+
+# Validation
+
+`<FormState />` supports client-side validation through the `validators` prop.
+
+```typescript
+export type ValidatorDictionary<Fields> = {
+  [FieldPath in keyof Fields]: MaybeArray<
+    ValidationFunction<Fields[FieldPath], Fields>
+  >
+};
+
+interface ValidationFunction<Value, Fields> {
+  (value: Value, fields: FieldStates<Fields>): any;
+}
+```
+
+To define validators use the `validators` prop and pass it a partial dictionary of validation functions.
+
+```typescript
+  validators={{
+    title(input) {
+      if (input === 'foo') {
+        return 'title cannot be foo';
+      }
+    },
+  }}
+```
+
+To run multiple functions on the same field you can also use an array of functions.
+
+```typescript
+  validators={{
+    password: [
+      (input) => {
+        if (input.length > 20) {
+          return 'too big';
+        }
+      },
+      (input) => {
+        if (input.length < 3) {
+          return 'too small';
+        }
+      },
+  ]}}
+```
+
+`<FormState />` generates handlers that follow the [Polaris form validation guidelines](https://polaris.shopify.com/patterns/error-messages#section-form-validation). When you blur an input, or when you change it and it already has an error, any defined validators for that field will be invoked.
+
+```typescript
+import {TextField} from '@shopify/polaris';
+import FormState, {validators} from '@shopify/react-form-state';
+
+function MyComponent() {
+  return (
+    <FormState
+      initialValues={{
+        title: 'Cool title',
+        description: 'Cool product',
+        quantity: 0,
+      }}
+      validators={{
+        title(input) {
+          if (input.length > 10) {
+            return 'That title is too long';
+          }
+        },
+        quantity() {
+          if (!/[^0-9.,]/g.test(input)) {
+            return 'Quantity must be a number';
+          }
+        },
+      }}
+    >
+      {formDetails => {
+        const {fields} = formDetails;
+
+        return (
+          <form>
+            <TextField label="Title" {...fields.title} />
+            <TextField label="Description" {...fields.description} />
+            <TextField label="Quantity" {...fields.quantity} />
+          </form>
+        );
+      }}
+    </FormState>
+  );
+}
+```
+
+To learn more about building validators, and the built in functions exposed by this package, check out the [validators guide](/validators.md).
+
+## Compound fields
+
+The default API for generating handlers with `<FormState />` is very simple, but real world apps are often very complex. You will often have to represent editing many resources with one form. This can mean needing support for arrays of fields, fields that are nested objects, or both. `<FormState />` allows passing these complex values into `initialValues` and having your own special sub-components that use the handlers, but for most cases it's recommended to use the sub-components `<FormState.List />` and `<FormState.Nested />`.
+
+### `<FormState.Nested />`
+
+Sometimes your data might be best represented using a sub-object.
+
+```typescript
+initialValues={{
+  firstVariant: {
+    option: 'size',
+    value: 'enormous',
+    price: '2.00',
+    sku: 'mc23',
+  }
+  /* other fields */
+}}
+```
+
+Passing this directly into `<FormState />` will generate one field description object for the entire object, just like any other field.
+
+This might work for some usecases if you have some kind of custom component that you just want to pass an updater to, but this can make collection values back for submission, validation, and other subtleties a pain to manage.
+
+In cases like this you can use `<FormState.Nested />` to generate individual field descriptors for the sub-fields.
+
+```typescript
+// ProductPage.ts
+import {TextField} from '@shopify/polaris';
+import FormState from '@shopify/react-form-state';
+
+export function ProductPage() {
+  return (
+    <FormState
+      initialValues={{
+        name: '',
+        description: '',
+        firstVariant: {
+          option: 'size',
+          value: 'enormous',
+          price: '2.00',
+          sku: 'mc23',
+        }
+      }}
+    >
+    {({fields}) => {
+      return (
+        <form>
+          <div>
+            <TextField label="Name" {...fields.name} />
+            <TextField label="Description" {...fields.description} />
+
+            <FormState.Nested field={fields.firstVariant}>
+              {(fields) => {
+                <TextField label="Option" {...fields.option} />
+                <TextField label="Value" {...fields.value} />
+                <TextField label="SKU" {...fields.sku} />
+                <TextField label="Price" {...fields.price} />
+              }}
+            </FormState.Nested>
+          </div>
+        </form>
+      )
+    }}
+    </FormState>
+  );
+}
+```
+
+### `<FormState.List />`
+
+Sometimes your data might be best represented using an array of objects.
+
+```typescript
+initialValues={{
+  variants: [{
+    option: 'size',
+    value: 'enormous',
+    price: '2.00',
+    sku: 'mc23',
+  }, {
+    option: 'size',
+    value: 'teensy',
+    price: '0.20',
+    sku: 'mc25',
+  }]
+  /* other fields */
+}}
+```
+
+Passing this directly into `<FormState />` will generate one field description object for the entire array, just like any other field.
+
+This might work for some usecases if you have some kind of custom component that you just want to pass an updater to, but this can make collecting values back for submission, validation, and other subtleties a pain to manage. It can also make adding and removing fields quite code intensive.
+
+In cases like this you can use `<FormState.List />` to generate individual field descriptors for the sub-fields of each object in the array. `<FormState.List />` will render the `children` function you give it once for each item in the `field` array you give it, passing in a `fields` object with the same type of signature as `<FormState.Nested />`.
+
+```typescript
+// ProductPage.ts
+import {TextField} from '@shopify/polaris';
+import FormState from '@shopify/react-form-state';
+
+export function ProductPage() {
+  return (
+    <FormState
+      initialValues={{
+        name: '',
+        description: '',
+        variants: [{
+          option: 'size',
+          value: 'enormous',
+          price: '2.00',
+          sku: 'mc23',
+        }, {
+          option: 'size',
+          value: 'teensy',
+          price: '0.20',
+          sku: 'mc25',
+        }],
+      }}
+    >
+    {({fields}) => {
+      return (
+        <form>
+          <div>
+            <TextField label="Name" {...fields.name} />
+            <TextField label="Description" {...fields.description} />
+
+            <FormState.List field={fields.variants}>
+              {(fields) => {
+                <TextField label="Option" {...fields.option} />
+                <TextField label="Value" {...fields.value} />
+                <TextField label="SKU" {...fields.sku} />
+                <TextField label="Price" {...fields.price} />
+              }}
+            </FormState.List>
+          </div>
+        </form>
+      )
+    }}
+    </FormState>
+  );
+}
+```
+
+## Putting it all together
+
+The following example shows how you can use everything this documentation covered to build a full [Polaris](https://polaris.shopify.com) styled page.'s
+
+```typescript
+import * as React from 'react';
+import {
+  AppProvider,
+  Page,
+  TextField,
+  FormLayout,
+  PageActions,
+  Spinner,
+  Banner,
+  Stack,
+  Card,
+  Layout,
+  ButtonGroup,
+  Button,
+} from '@shopify/polaris';
+import FormState, {
+  validators,
+  validateArray,
+  validateObject,
+  arrayUtils,
+} from '@shopify/react-form-state';
+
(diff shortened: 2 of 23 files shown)


TheMallen:
Looks like I have some linter troubles, taking a look.

packages/react-form-state/README.md
```diff
@@ -0,0 +1,947 @@
+# `@shopify/react-form-state`
+
+[![Build Status](https://travis-ci.org/Shopify/quilt.svg?branch=master)](https://travis-ci.org/Shopify/quilt)
+
```
tzvipm: :fire: extra newline here?

packages/react-form-state/README.md
```diff
@@ -0,0 +1,947 @@
...
+[![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](LICENSE.md) [![npm version](https://badge.fury.io/js/%40shopify%2Freact-formwell.svg)](https://badge.fury.io/js/%40shopify%2Freact-formwell.svg) [![npm bundle size (minified + gzip)](https://img.shields.io/bundlephobia/minzip/@shopify/react-formwell.svg)](https://img.shields.io/bundlephobia/minzip/@shopify/react-formwell.svg)
```
tzvipm: I'd prefer each badge on a separate line

packages/react-form-state/README.md
```diff
@@ -0,0 +1,947 @@
...
+
+Manage react forms tersely and type-safe with no magic.
```
tzvipm: `tersely and safely-typed`?

packages/react-form-state/README.md
```diff
@@ -0,0 +1,947 @@
...
+
+## Another form library?
+
+Good forms in React are surprisingly tough. A good api needs to support complex usecases with large numbers of fields split up across multiple components and explicit validation. It also needs to be quick to get started with and work for the simplest of forms cleanly. In meeting these goals, a package needs to balance developer experience, flexibility, and performance.
```
tzvipm: What are `Good` forms?

packages/react-form-state/README.md
```diff
@@ -0,0 +1,947 @@
...
+
+Several popular community packages sit at different places on the spectrum for these requirements, and all are solid choices for many projects.
+
+**Examples**
+
+- [formik](https://github.com/jaredpalmer/formik) offers a terse but relatively unmagical way of managing form state with minimal dependencies.
+- [redux-form](https://redux-form.com/7.4.2/) offers a redux integrated solution that can be quite magical and give you the power of the redux devtools for debugging broken states.
+
+At Shopify we wanted our form solution to be true to the Web Foundations team's core values. We wanted it to be **explicit**, **type-safe**, and to **prioritize user experience** while **keeping code quality and developer happiness high**.
```
tzvipm: should we link these to our web foundations repo? In particular:

1. Should we explicitly address https://github.com/Shopify/web-foundation/blob/master/Principles/5%20-%20Community%20over%20ownership.md and why the advantages of writing our own forms library outweigh the costs?
2. Should we link to https://github.com/Shopify/web-foundation/blob/master/Principles/3%20-%20Explicit%20over%20automatic.md ?
3. See my suggestion via https://github.com/Shopify/web-foundation/issues/32 to add docs about our preferences for type safety.

packages/react-form-state/README.md
```diff
@@ -0,0 +1,947 @@
...
+
+As such the main difference in our solution is the total lack of magic. Form fields are given explicit handlers generated individually for each field and made available through a [render prop](https://reactjs.org/docs/render-props.html). To reduce boilerplate we've kept the generated `field` objects as easy to use as possible. Usually you can just splat them onto your inputs. We'll go into this in more detail in the [Usage](/#Usage) section.
```
tzvipm: I like the word magic in the description, but I think it makes sense to use a more accurate term throughout the text. Maybe `implicit behaviour` or `automatic behaviour`?

tzvipm: `/#Usage` -> `#usage`, no? does it make a difference?

packages/react-form-state/README.md
```diff
@@ -0,0 +1,947 @@
...
+
+Creating handlers for each field instead of having one that you share or inject using a higher order component means we are fully type safe, with editor autocomplete able to show a developer exactly what fields are available, and compilation breaking when someone tries to use a field that doesn't exist.
```
tzvipm: ...one that *you* share... means *we* are fully...to show *a developer*... when *someone*...

I feel like more consistency in voice and tone here makes sense.

packages/react-form-state/README.md
```diff
@@ -0,0 +1,947 @@
...
+
+## Installation
+
+```bash
+$ yarn add @shopify/react-formwell
```
tzvipm: `react-form-state` ?

packages/react-form-state/README.md
```diff
@@ -0,0 +1,947 @@
...
+  {({fields, dirty, valid, submitting, errors, reset, submit}) => {
+    return /* some cool ui */;
+  }}
+</FormState>
+```
+
+When using `<FormState>`, you express your form's component tree as a function of the generated field state objects provided by `<FormState />` into it's `children` function as `formDetails`.
+
+```typescript
+export interface FormDetails<Fields> {
```
tzvipm: :fire: export?

packages/react-form-state/README.md
```diff
@@ -0,0 +1,947 @@
...
+  errors: ClientError[];
+  reset(): void;
+  submit(): void;
+}
+```
+
+The `formDetails` object passed into `children` contains a `fields` dictionary, which includes all the state and handlers you need to render your form.
+
+```typescript
+export interface FieldState<Value> {
```
tzvipm: here too. I don't think we need to word `export`, do we?

packages/react-form-state/README.md
```diff
@@ -0,0 +1,947 @@
...
+    </FormState>
+  );
+}
+```
+
+#### Reducing boilerplate with custom inputs
+
+The previous example has a fair amount of repetition around passing the right handlers to the right inputs. It also doesn't handle rendering errors. We can eliminate this boilerplate by using a custom component that's props match our field descriptor objects.
+
+It just so happens that [@shopify/polaris](https://polaris.shopify.com/components/forms/text-field)'s `<TextField />` component does this. So, using `<TextField />` and ES6 [spread](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) syntax we can greatly simplify our forms. The Polaris inputs also support `error` props, which we'll take advantage of later.
```
tzvipm: should we use backticks in the link text? for `@shopify/polaris`?

packages/react-form-state/README.md
```diff
@@ -0,0 +1,947 @@
...
+
+function MyComponent() {
+  return (
+    <FormState
+      initialValues={{
+        title: 'Cool title',
+        description: 'Cool product',
+      }}
+    >
+      {formDetails => {
```
tzvipm: I'd rather see `fields` destructured here.

packages/react-form-state/README.md
```diff
@@ -0,0 +1,947 @@
...
+            <TextField label="Description" {...fields.description} />
+          </form>
+        );
+      }}
+    </FormState>
+  );
+}
+```
+
+There is also nothing to stop you from using your own custom inputs.
```
tzvipm: I'd :fire: this line. They can infer that from the above sentences, and we probably do want to push polaris here, for consistency.

packages/react-form-state/README.md
```diff
@@ -0,0 +1,947 @@
...
+  onSubmit={async (fields) => {
+    const result = await updateProduct(fields);
+
+    if (isErrorResult(result)) {
+      return result.errors;
+    }
+  }}
+```
+
+When given an [async](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function) function (or one that returns a [promise](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function), the `children` function will be rerendered with `submitting: true` until the promise resolves or rejects. Once it has returned the function will be rerendered with `submitting: false` and `errors` either `[]` (if the promise resolved with no value), or with an array of `{field?: string, error: string}` if the promise resolved with an error list.
```
tzvipm: I think we missed a closing parenthesis after the `promise` link. The one there simply closes the link, but doesn't visibly show.

packages/react-form-state/README.md
```diff
@@ -0,0 +1,947 @@
...
+```typescript
+// api.ts
+
+const fakeErrorResult = {error: [{message: 'api is not real'} };
+
+// simple async function that fakes a failing api call
+export function fakeApiCall(data: any) {
+  return new Promise(resolve => {
+    setTimeout(() => {
+      resolve(]);
```
tzvipm: what is this supposed to resolve with?

packages/react-form-state/README.md
```diff
@@ -0,0 +1,947 @@
...
+    }, 2000);
+  });
+}
+```
+
+```typescript
+// MyPage.ts
+import {fakeApiCall} from './api';
```
tzvipm: according to shopify's eslint rules, this relative import should be below the package imports.

packages/react-form-state/README.md
```diff
@@ -0,0 +1,947 @@
...
+import {TextField, Button, Banner, FormLayout, Spinner} from '@shopify/polaris';
+import FormState from '@shopify/react-form-state';
+
+export function MyPage() {
+  return (
+    <FormState
+      initialValues={{email: '', password: ''}}
```
tzvipm: Interesting, so I guess we need these empty strings here in order to properly infer types?

packages/react-form-state/README.md
```diff
@@ -0,0 +1,947 @@
...
+          return results.error;
+        }
+      }}
+    >
+      {({fields, submitting, error, submit}) => {
+        return (
+          <form onSubmit={submit}>
+            {submitting && <Spinner size="large" color="teal" />}
+
+            {errors.length > 0 && (
```
tzvipm: should we provide a `hasErrors` prop here? I don't really like the boolean check inside of the jsx like this.

packages/react-form-state/README.md
```diff
@@ -0,0 +1,947 @@
...
+                </ul>
+              </Banner>
+            )}
+
+            <FormLayout>
+              <Textfield label="Email" {...fields.email} />
+              <Textfield label="Password" {...fields.password} />
+            </FormLayout>
+
+            <Button type="submit" disabled={submitting}>
```
tzvipm: loading instead of disabled? also, primary?

packages/react-form-state/README.md
```diff
@@ -0,0 +1,947 @@
...
+        );
+      }}
+    </FormState>
+  );
+}
+```
+
+### Dirty Tracking
+
+Often you want to change some UI behaviour based on whether a form has changed form it's initial values. `<FormState />` supports this both on the form and individual field level.
```
tzvipm: changed form -> changed from

packages/react-form-state/README.md
```diff
@@ -0,0 +1,947 @@
...
+    <FormState
+      initialValues={{email: '', password: ''}}
+    >
+      {({fields, dirty}) => {
+        return (
+          <form>
+            <FormLayout>
+              <Textfield label="Email" {...fields.email} />
+              <Textfield label="Password" {...fields.password} />
+              {fields.pasword.dirty && <p>you have editied this</p>}
```
tzvipm: can we make this a temporary constant outside of the return value? Also, `editied` -> `edited`

packages/react-form-state/README.md
TheMallen: I think the rest of this paragraph says that :p

packages/react-form-state/README.md
TheMallen: nooooooooo

packages/react-form-state/README.md
TheMallen: dont think so but maaaaaybe im wrong

packages/react-form-state/README.md
TheMallen: Yes, though you can also use the magic of casting the initialFieldValues or using generics syntax to give it a specific interface.

packages/react-form-state/README.md
TheMallen: oh lol there totally is one `valid`

packages/react-form-state/README.md
TheMallen: ya sure

packages/react-form-state/README.md
```diff
@@ -0,0 +1,979 @@
...
+
+The `formDetails` object passed into `children` contains a `fields` dictionary, which includes all the state and handlers you need to render your form.
+
+```typescript
+export interface FieldState<Value> {
+  name: string;
+  initialValue: Value;
+  value: Value;
+  dirty: boolean;
+  changed: boolean;
```
loic-d: 👍 Happy to see you added the `changed` prop :)

packages/react-form-state/README.md
```diff
@@ -0,0 +1,979 @@
...
+      {({fields, submitting, error, submit, valid}) => {
+        return (
+          <form onSubmit={submit}>
+            {submitting && <Spinner size="large" color="teal" />}
+
+            {!valid && (
+              <Banner status="critical">
+                <ul>
+                  {errors.map(error => (
+                    <li key={error.message + error.field}>{error.message}</li>
```
loic-d: I would use a template literal here, no?

packages/react-form-state/README.md
```diff
@@ -0,0 +1,979 @@
...
+      }}
+    >
+    {({fields}) => {
+      return (
+        <form>
+          <div>
+            <TextField label="Name" {...fields.name} />
+            <TextField label="Description" {...fields.description} />
+
+            <FormState.Nested field={fields.firstVariant}>
```
loic-d: Really nice!!

packages/react-form-state/src/FormState.ts
```diff
@@ -0,0 +1,351 @@
...
+  message: string;
+}
+
+export type FormError = GraphQLError | ClientError;
+
+export type FieldStates<Fields> = {
+  [FieldPath in keyof Fields]: FieldState<Fields[FieldPath]>
+};
+
+type MaybeArray<T> = T | T[];
```
loic-d: Why are these types required?

packages/react-form-state/src/FormState.ts
```diff
@@ -0,0 +1,351 @@
...
+        }),
+      };
+    });
+  }
+
+  private updateErrors(newErrors: FormError[]) {
+    this.setState(({fields}: State<Fields>) => {
+      const newFields = {...(fields as any)};
+
+      const errors: ClientError[] = newErrors.map(({message, field}) => {
```
loic-d: Only updating client errors here?

packages/react-form-state/README.md
TheMallen: This is actually out of date I had removed it because it's not yet useful (we're not doing full form validation on each blur / update yet)

packages/react-form-state/README.md
TheMallen: 👍

packages/react-form-state/src/FormState.ts
TheMallen: you can pass an array of validators
```
validators={{
  price: [
    validators.numericString(numericStringErrorMessage),
    validators.lengthLessthan(10, tooLongErrorMessage),
  ],
}}
```

packages/react-form-state/src/FormState.ts
TheMallen: Yes, the polaris guidelines tell us we should only handle client validation at the field level and only handle remote validation at the banner level.

packages/react-form-state/README.md
TheMallen: :)

packages/react-form-state/README.md
TheMallen: typo

packages/react-form-state/package.json
```diff
@@ -0,0 +1,37 @@
...
+  },
+  "author": "Shopify Inc.",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/Shopify/quilt.git"
+  },
+  "bugs": {
+    "url": "https://github.com/shopify/quilt/issues"
+  },
+  "homepage": "https://github.com/Shopify/quilt/blob/master/packages/react-formwell/README.md",
```
lemonmade: Needs update

packages/react-form-state/src/FormState.ts
```diff
@@ -0,0 +1,351 @@
...
+  dirty: boolean;
+  valid: boolean;
+  submitting: boolean;
+  errors: ClientError[];
+  reset(): void;
+  submit(): void;
+}
+
+interface Props<Fields> {
+  initialValues: Fields;
```
lemonmade: What do you think about this just being `values`?

packages/react-form-state/src/FormState.ts
```diff
@@ -0,0 +1,351 @@
...
+  fields: FieldStates<Fields>;
+}
+
+export default class FormState<
+  Fields extends Object
+> extends React.PureComponent<Props<Fields>, State<Fields>> {
+  static List = List;
+  static Nested = Nested;
+
+  static specialize<FormFields>() {
```
lemonmade: 🔥 ?

packages/react-form-state/src/FormState.ts
```diff
@@ -0,0 +1,351 @@
...
+
+    const oldInitialValues = initialValuesFromFields(oldState.fields);
+    const shouldReinitialize = !isEqual(oldInitialValues, newInitialValues);
+
+    if (shouldReinitialize) {
+      const newState = createFormState(newInitialValues);
+      return newState;
+    }
+
+    return oldState;
```
lemonmade: Can return `null` for no update

packages/react-form-state/src/FormState.ts
```diff
@@ -0,0 +1,351 @@
...
+  static specialize<FormFields>() {
+    return (FormState as any) as new () => FormState<FormFields>;
+  }
+
+  static getDerivedStateFromProps<T>(newProps: Props<T>, oldState?: State<T>) {
+    const newInitialValues = newProps.initialValues;
+
+    if (oldState == null) {
+      const newState = createFormState(newInitialValues);
+      return newState;
```
lemonmade: I never get creating intermediate variables, then returning — why not just return?

packages/react-form-state/src/FormState.ts
```diff
@@ -0,0 +1,351 @@
...
+    }
+
+    this.setState({submitting: false});
+  };
+
+  private reset = () => {
+    this.setState(this.cleanFormState);
+  };
+
+  private get cleanFormState() {
```
lemonmade: Why have a getter for this?

packages/react-form-state/src/FormState.ts
```diff
@@ -0,0 +1,351 @@
...
+      return newState;
+    }
+
+    return oldState;
+  }
+
+  state = this.cleanFormState;
+  private mounted = false;
+
+  private getFieldUpdater = memoize(<K extends keyof Fields>(fieldPath: K) => {
```
lemonmade: Why not use the decorators from `javascript-utilities`?

packages/react-form-state/src/FormState.ts
```diff
@@ -0,0 +1,351 @@
...
+    return this.updateField.bind(this, fieldPath);
+  });
+
+  private getFieldBlurHandler = memoize(
```
lemonmade: I find the mix of methods and properties really distracting (also the order is wonky)

packages/react-form-state/src/FormState.ts
```diff
@@ -0,0 +1,351 @@
+import * as React from 'react';
+import isEqual from 'lodash/isEqual';
+import isArray from 'lodash/isArray';
+import memoize from 'lodash/memoize';
+
+import {mapObject} from './utilities';
+import {FieldDescriptors, ClientError, FieldState} from './types';
+import {List, Nested} from './components';
+
+const ERROR_SEPARATOR = ', ';
```
lemonmade: This is probably a no-no for i18n...

packages/react-form-state/src/FormState.ts
```diff
@@ -0,0 +1,351 @@
...
+
+interface GraphQLError {
```
lemonmade: Can we name it a little more generically?

packages/react-form-state/src/FormState.ts
```diff
@@ -0,0 +1,351 @@
...
+      const shouldUpdateError = error != null;
+
+      const newError = shouldUpdateError
+        ? this.validateFieldValue(fieldPath, newFieldData)
+        : // eslint-disable-next-line no-undefined
+          undefined;
+
+      return {
+        dirtyFields: [...Array.from(newDirtyFields)],
+        fields: Object.assign({}, fields, {
```
lemonmade: Why no spread?

packages/react-form-state/src/components/List.tsx
```diff
@@ -0,0 +1,71 @@
...
+  never
+> {
+  render() {
+    const {
+      field: {value, initialValue, error, onBlur},
+      children,
+    } = this.props;
+
+    return (
+      <>
```
lemonmade: Why?

packages/react-form-state/src/components/Nested.tsx
```diff
@@ -0,0 +1,55 @@
...
+          name: `${name}.${fieldPath}`,
+          initialValue: initialFieldValue,
+          dirty: value !== initialFieldValue,
+          error: get(error, [fieldPath]),
+          onChange: this.handleChange(fieldPath),
+        };
+      },
+    );
+
+    return <>{children(innerFields)}</>;
```
lemonmade: Same here, why the fragment?

packages/react-form-state/src/components/Nested.tsx
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  private handleChange = <Key extends keyof Fields>(key: Key) => {
```
lemonmade: Not memoized?

packages/react-form-state/src/components/List.tsx
```diff
@@ -0,0 +1,71 @@
...
+            <React.Fragment key={index}>
+              {children(innerFields, index)}
+            </React.Fragment>
+          );
+        })}
+      </>
+    );
+  }
+
+  private handleChange = <Key extends keyof Fields>(
```
lemonmade: Not memoized?

packages/react-form-state/src/components/tests/Input.tsx
```diff
@@ -0,0 +1,18 @@
+import React, {InputHTMLAttributes} from 'react';
```
lemonmade: Can you match our typical input style?

lemonmade: IMO it is really messy to have what amount to fixtures directly beside the actual test files.

packages/react-form-state/src/validators.ts
```diff
@@ -0,0 +1,132 @@
...
+      // eslint-disable-next-line consistent-return
+      return errorContent(toString(input));
+    }
+
+    // eslint-disable-next-line consistent-return
+    return errorContent;
+  };
+}
+
+const validators = {
```
lemonmade: Why export as an object and not just named exports?

packages/react-form-state/README.md
```diff
@@ -0,0 +1,980 @@
...
+
+At Shopify we wanted our form solution to be true to the Web Foundations team's core values. We wanted it to be [explicit, type-safe](https://github.com/Shopify/web-foundation/blob/master/Principles/3%20-%20Explicit%20over%20automatic.md), and to [prioritize user experience](https://github.com/Shopify/web-foundation/blob/master/Principles/1%20-%20User%20over%20team%20over%20self.md) while keeping code quality and developer happiness high
+
+Several popular community packages sit at different places on the spectrum for these requirements, and all are solid choices for many projects. While we find it valuable to [use and contribute to community libraries,](https://github.com/Shopify/web-foundation/blob/master/Principles/5%20-%20Community%20over%20ownership.md) in this case we felt that we were better off going our own way.
+
+- [formik](https://github.com/jaredpalmer/formik) offers a declarative way of managing form state with minimal dependencies. Unfortunately since it uses a single set of handlers across all a form's fields we found it insufficiently type safe.
```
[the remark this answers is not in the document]
TheMallen: I think its good as an open source lib that is in a super saturated space to have these kind of justifications surfaced in the docs... I guess a decision doc would still be public though

packages/react-form-state/src/validators.ts
TheMallen: I think its a cleaner api to namespace them rather than have a million `validateFoo` methods exported. I was thinking this file might want to be a separate package though in which case that package could function as the namespace.

packages/react-form-state/README.md
```diff
@@ -0,0 +1,980 @@
...
+- [redux-form](https://redux-form.com/7.4.2/) offers a redux integrated solution that can be quite magical and gives you the power of the redux dev tools for debugging broken states. Not all react apps need or want redux, and the library is fairly heavy, so we found it unsuitable for our uses.
+
+As such the main difference in our solution is the explicit, declarative api. Form fields are given explicit handlers generated individually for each field and made available through a [render prop](https://reactjs.org/docs/render-props.html). To reduce boilerplate we've kept the generated `field` objects as easy to use as possible. Usually you can just splat them onto your inputs. We'll go into this in more detail in the [usage](#Usage) section.
+
+Creating handlers for each field instead of having one that you share or inject using a higher order component means we are fully type safe, with editor autocomplete able to show a developer exactly what fields are available, and compilation breaking when someone tries to use a field that doesn't exist.
+
+## Installation
+
+```bash
+$ yarn add @shopify/react-form-state
```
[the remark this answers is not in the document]
TheMallen: if we are going to move the form component here that name makes sense but I dont think it makes sense for what the main component here does. If it was called `<Form />` you'd expect it to actually render a form tag no?

I think this package should just have the formstate stuff.

packages/react-form-state/README.md
```diff
@@ -0,0 +1,980 @@
...
+    >
+      {formDetails => {
+        const {fields} = formDetails;
+        const {title, description} = fields;
+
+        return (
+          <form>
+            <label htmlFor={title.name}>Title</label>
+            <input
+              {...title}
```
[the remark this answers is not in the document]
TheMallen: Good call it only has name. Suppose we could have it autogenerate ids too, or just change the example.

packages/react-form-state/README.md
```diff
@@ -0,0 +1,980 @@
...
+    }, 2000);
+  });
+}
+```
+
+```typescript
+// MyPage.ts
+import {fakeApiCall} from './api';
+import {TextField, Button, Banner, FormLayout, Spinner} from '@shopify/polaris';
+import FormState from '@shopify/react-form-state';
```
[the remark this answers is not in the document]
TheMallen: I don't think anything else should go in this package, it's likely already too big including the validators. If anything I'd want to see those pulled into their own package.

packages/react-form-state/README.md
```diff
@@ -0,0 +1,980 @@
...
+        );
+      }}
+    </FormState>
+  );
+}
+```
+
+### Resetting
+
+Most forms want to expose a way to reset all fields to their initial values. The `reset()` function passed into `children()` allows you to easily make this available to a user.
```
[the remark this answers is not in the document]
TheMallen: Thats deep

packages/react-form-state/src/FormState.ts
TheMallen: not sure how we can make this work nicely out of the box then

packages/react-form-state/src/FormState.ts
TheMallen: RemoteError?

packages/react-form-state/src/FormState.ts
TheMallen: because its typed grossly

m-ux:
@TheMallen this is great 🎉 - we've been playing with this in Marketing Tech land, and are big fans of the approach (this also helps us to solve a number of problems consistently w/the rest of core).  

One piece of feedback is around the form submit (server side validation) handler.  

We recommend something I would consider to be a legacy pattern in the [Polaris error messaging](https://polaris.shopify.com/patterns/error-messages) recommendations (which is a list of field error messages displayed within a banner element at the top of the page).  

If the server error field keys were consistent with the client side keys, it would be nice if we could move these error messages inline.  It would be nice if the solution could automatically piggy-back on the client-side validation handler so that the implementing developer doesn't have to write any extra boilerplate to achieve this.

packages/react-form-state/src/FormState.ts
m-ux: As I mentioned in my other comment, I feel like this is a legacy recommendation based on the way we handle things in rails.  

While looking to maintain consistency w/existing rails forms makes sense, I would argue that this is not the best UX for this problem.  We've had plenty of teams talk about the problems with scrolling down on long forms and not being able to see the error message beside the flagged field 

(while maintaining consistency might end up trumping improvements to our UX, I figured it was still worth having the conversation).

tzvipm: I'm pretty sure it's not a legacy recommendation, but rather an issue of accessibility and usability. Imagine if the banner updated while you type; then the entire page would reflow while you're trying to enter text which would be super annoying. Furthermore, we can't put server validation at the field level for performance reasons, because we can't do a server round trip quickly and reliably enough to make the UX nice there.

TheMallen:
@m-ux [here's the real PR](https://github.com/Shopify/quilt/pull/166)

Have you spoken about doing errors like that with the polaris folks? I had originally been trying to do that but they suggested I don't. Personally, I think it actually doesn't make sense to mix the errors like that because the user doesn't necessarily understand why some errors that appear under their inputs dissapear while they type and some don't. I think it usually looks pretty broken.

packages/react-form-state/src/FormState.ts
m-ux: I'm only suggesting that field-specific error messages should be displayed inline with their input element, regardless of whether the error came from the server or the client.  

I'm not suggesting the server side validation should be triggered on field-level interaction, or that the banner content should change dynamically. 

In regards to the a11y callout, what about this pattern do you think makes it good from an accessibility standpoint?  

From a [blog post](https://cantina.co/form-errors-screen-readers-can-access/#screen-reader-error-messaging-block) on the topic:
> ...If the form contained several errors it could be difficult for a user to retain all the messages at once and then fix them. **It would be better to give the user an overview of how many errors there are and the option to go and fix them one at a time**. This can be done by making the overview message a link. The overview message link would focus the first input error message, then the input error message link would directly focus the user to the invalid input...

TheMallen: Right but if you have no means of knowing when to remove the server side errors then it is just confusing for the user I think. I'd talk to polaris about it, as I cleared this with them.

TheMallen: Seems like we will discuss this in slack for a bit, but imo shouldn't be a blocker for V1 since it can always be added without an api change in the future.

packages/react-form-state/src/components/tests/Nested.tsx
```diff
@@ -0,0 +1,156 @@
+import React from 'react';
```
9at8: Filename: Nested.test.tsx

packages/react-form-state/src/FormState.ts
```diff
@@ -0,0 +1,331 @@
...
+
+  @memoize
+  private getHandlersForField(path: keyof FieldStates<Fields>) {
+    return {
+      onChange: this.updateField.bind(this, path),
+      onBlur: this.blurField.bind(this, path),
+    };
+  }
+  private updateField<K extends keyof Fields>(fieldPath: K, value: Fields[K]) {
+    this.setState<any>(({fields, dirtyFields}: State<Fields>) => {
```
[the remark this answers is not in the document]
TheMallen: Because TS hates splats over generics

packages/react-form-state/src/components/List.tsx
```diff
@@ -0,0 +1,66 @@
+import * as React from 'react';
+import get from 'lodash/get';
+import {memoize, bind} from 'lodash-decorators';
+
+import {FieldDescriptor, FieldDescriptors} from '../types';
+import {mapObject, replace} from '../utilities';
+
+interface Props<Fields> {
+  field: FieldDescriptor<Fields[]>;
+  children(fields: FieldDescriptors<Fields>, index: number): React.ReactNode;
```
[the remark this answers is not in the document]
TheMallen: `index` isn't optional, it's always supplied. You can always supply a function that only cares about the first subset of positional args.

packages/react-form-state/docs/building-forms.md
```diff
@@ -0,0 +1,710 @@
...
+  }]
+  /* other fields */
+}}
+```
+
+Passing this directly into `<FormState />` will generate one field description object for the entire array, just like any other field.
+
+This might work for some usecases if you have some kind of custom component that you just want to pass an updater to, but this can make collecting values back for submission, validation, and other subtleties a pain to manage. It can also make adding and removing fields quite code intensive.
+
+In cases like this you can use `<FormState.List />` to generate individual field descriptors for the sub-fields of each object in the array. `<FormState.List />` will render the `children` function you give it once for each item in the `field` array you give it, passing in a `fields` object with the same type of signature as `<FormState.Nested />`.
```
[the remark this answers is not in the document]
TheMallen: why did it trip you up? What was your expectation?

packages/react-form-state/docs/building-forms.md
```diff
@@ -0,0 +1,710 @@
...
+  }
+  /* other fields */
+}}
+```
+
+Passing this directly into `<FormState />` will generate one field description object for the entire object, just like any other field.
+
+This might work for some usecases if you have some kind of custom component that you just want to pass an updater to, but this can make collection values back for submission, validation, and other subtleties a pain to manage.
+
+In cases like this you can use `<FormState.Nested />` to generate individual field descriptors for the sub-fields.
```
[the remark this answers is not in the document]
TheMallen: in theory it can go arbitrarily deep though I'd question whether that is a good idea

packages/react-form-state/docs/building-forms.md
```diff
@@ -0,0 +1,710 @@
...
+    >
+      {({fields, submitting, error, submit, valid}) => {
+        return (
+          <form onSubmit={submit}>
+            {submitting && <Spinner size="large" color="teal" />}
+
+            {!valid && (
+              <Banner status="critical">
+                <ul>
+                  {errors.map(error => (
```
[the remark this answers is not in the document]
TheMallen: We are already pulling off `errors`

TheMallen:
@lemonmade  dis good now?

packages/react-form-state/docs/building-forms.md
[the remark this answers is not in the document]
9at8: I got confused between the the object from `FormState` and this one. Thoughts on `renderItem` render prop vs children as a render prop?

packages/react-form-state/src/components/List.tsx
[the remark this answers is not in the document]
9at8: Wouldn't that cause problems with typescript?

I tried this in the [playground](http://www.typescriptlang.org/play/), and it gives me an error. 🤔 

```ts
function helloWorld(stuff: string, index: number) {
    return;
}

helloWorld("a");  // error here
helloWorld("a", 1);  // no error here
```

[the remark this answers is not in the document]
9at8: Oh but we're supplying a function and not calling it, my bad. 👍

packages/react-form-state/docs/building-forms.md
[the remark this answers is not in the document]
9at8: prolly not 😅

packages/react-form-state/docs/building-forms.md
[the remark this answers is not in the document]
9at8: But we are not!

PR #168, opened 2018-06-28
Add findById utility

lemonmade:
Does what it says on the tin, sourced from https://github.com/Shopify/web/pull/4893#pullrequestreview-132623908.


packages/enzyme-utilities/src/test/index.test.tsx
```diff
@@ -1,8 +1,10 @@
 import * as React from 'react';
 import {mount} from 'enzyme';
-import {trigger} from '..';
-import {Toggle} from './Toggle';
-import {ActionList, Action} from './Actions';
+
+import {trigger, findById} from '..';
+
+import {Toggle} from './fixtures/Toggle';
+import {ActionList, Action} from './fixtures/Actions';
```
lemonmade: I don't love having these in separate files anyways, but IMO it was quite messy to have them alongside the test file like this.

packages/enzyme-utilities/src/index.ts
```diff
@@ -46,3 +46,7 @@ export function trigger(wrapper: AnyWrapper, keypath: string, ...args: any[]) {
 function updateRoot(wrapper: AnyWrapper) {
   (wrapper as any).root().update();
 }
+
+export function findById(root: ReactWrapper<any, any>, id: string) {
```
tzvipm: does this need to be the root node?

packages/enzyme-utilities/src/test/index.test.tsx
```diff
@@ -34,4 +36,31 @@ describe('enzyme-utilities', () => {
...
+            <div id="bar" />
+          </div>,
+        ),
+        'foo',
+      );
+
+      expect(found.getElement()).toEqual(matchedNode);
+    });
+
+    it('only returns an outer component when it renders its own children with the same ID', () => {
```
TheMallen: I think this wants to be two tests, this one and one that is expected not to find.

lemonmade: I don't think I understand, not sure what the second test would look like? This is just meant to make sure we don't get the rendered `input`s when this utility finds a `TextField`/ `Checkbox`. I can do a test for no matches but IMO it's not a super meaningful test

PR #170, opened 2018-06-28
⚡[FormState] remove needless renders

### LINKED ISSUE #183: [react-form-state] optimize rerenders
Right now the whole form rerenders whenever you change anything. We should be able to cache field objects intelligently such that this doesn't happen (assuming the fields are being rendered as pure components)

TheMallen:
closes #183 
Adds some memoization and other refactors to remove some needless re-renders, both on the field and whole render prop level.


packages/react-form-state/src/FormState.ts
```diff
@@ -186,52 +180,103 @@ export default class FormState<
...
-    return {
-      onChange: this.updateField.bind(this, path),
-      onBlur: this.blurField.bind(this, path),
-    };
+  @bind()
+  private fieldWithHandlers<Key extends keyof Fields>(
+    field: FieldStates<Fields>[Key],
+    fieldPath: Key,
+  ) {
+    return Object.assign({}, field, {
```
loic-d: Can you use the spread operator here?

packages/react-form-state/src/FormState.ts
```diff
@@ -186,52 +180,103 @@ export default class FormState<
...
+  }) {
+    const dirtyFieldsSet = new Set(dirtyFields);
+
+    if (dirty) {
+      dirtyFieldsSet.add(fieldPath);
+    } else {
+      dirtyFieldsSet.delete(fieldPath);
+    }
+
+    const newDirtyFields = Array.from(dirtyFieldsSet);
```
loic-d: Same here, can you use the spread operator `[...dirtyFieldSet]`?

packages/react-form-state/src/FormState.ts
```diff
@@ -186,52 +180,103 @@ export default class FormState<
...
+    dirty,
+    dirtyFields,
+  }: {
+    fieldPath: Key;
+    dirty: boolean;
+    dirtyFields: (keyof Fields)[];
+  }) {
+    const dirtyFieldsSet = new Set(dirtyFields);
+
+    if (dirty) {
```
loic-d: Nice to have this extracted into it's own function.

packages/react-form-state/src/FormState.ts
```diff
@@ -186,52 +180,103 @@ export default class FormState<
...
+    });
+
+    /*
+    We only want to update errors as the user types if they
+    already have an error.
+    https://polaris.shopify.com/patterns/error-messages#section-form-validation
+    */
+    const shouldUpdateError = field.error != null;
+    newField.error = shouldUpdateError
+      ? this.validateFieldValue(fieldPath, newField)
```
loic-d: Would it work with `newField.error = shouldUpdateError && this.validateFieldValue(fieldPath, newField)`?

packages/react-form-state/src/components/tests/Nested.tsx
```diff
@@ -0,0 +1,156 @@
...
+describe('<Nested />', () => {
+  beforeEach(() => {
+    faker.seed(ARBITRARY_SEED);
+  });
+
+  it('passes field state into child function', () => {
+    const renderPropSpy = jest.fn(() => null);
+
+    const product = {
+      title: faker.commerce.productName(),
```
loic-d: First time that I see `faker`, do we have access to it in `web`?

packages/react-form-state/src/components/tests/Nested.tsx
```diff
@@ -0,0 +1,156 @@
...
+              {renderPropSpy}
+            </FormState.Nested>
+          );
+        }}
+      </FormState>,
+    );
+
+    expect(renderPropSpy).toBeCalledWith({
+      title: {
+        name: 'nodejs.title',
```
loic-d: Is this coming from `faker`?

packages/react-form-state/src/FormState.ts
lemonmade: Agreed, even if you need to cast it is better to use the feature you want to use (also, apparently spread will be faster than object assign somehow /shrug)

packages/react-form-state/src/components/tests/Nested.tsx
```diff
@@ -0,0 +1,156 @@
...
+    mount(<FormState.Nested field={field}>{renderSpy}</FormState.Nested>);
+
+    const renderPropArgs = lastCallArgs(renderSpy);
+
+    expect(renderPropArgs.title.error).toBe(field.error.title);
+    expect(renderPropArgs.department.error).toBe(field.error.department);
+  });
+});
+
+function lastCallArgs(spy: jest.Mock) {
```
loic-d: Should this function be extracted to a shared file since it is also used by `FormState`?

packages/react-form-state/src/FormState.ts
```diff
@@ -186,52 +180,103 @@ export default class FormState<
...
+    const dirtyFieldsSet = new Set(dirtyFields);
+
+    if (dirty) {
+      dirtyFieldsSet.add(fieldPath);
+    } else {
+      dirtyFieldsSet.delete(fieldPath);
+    }
+
+    const newDirtyFields = Array.from(dirtyFieldsSet);
+    return isEqual(dirtyFields, newDirtyFields) ? dirtyFields : newDirtyFields;
```
lemonmade: I think you can cheat and just use length here instead of a full comparison

packages/react-form-state/src/FormState.ts
```diff
@@ -186,52 +180,103 @@ export default class FormState<
...
+    field: FieldStates<Fields>[Key];
+    value: Fields[Key];
+    dirty: boolean;
+  }) {
+    const newField = Object.assign({}, field, {
+      value,
+      dirty,
+    });
+
+    /*
```
lemonmade: This format of comments feels really odd — IMO line comments are better here

packages/react-form-state/src/FormState.ts
lemonmade: Not really sure why you are mutating the object instead of making this a property this in the third argument to `newField`?

packages/react-form-state/src/components/tests/Nested.tsx
```diff
@@ -0,0 +1,156 @@
...
+import {mount} from 'enzyme';
+
+import {Input} from './components';
+import FormState from '../..';
+
+const ARBITRARY_SEED = 1337;
+
+describe('<Nested />', () => {
+  beforeEach(() => {
+    faker.seed(ARBITRARY_SEED);
```
lemonmade: Why do you need to seed this?

packages/react-form-state/src/components/tests/Nested.tsx
lemonmade: Where does this come from?

packages/react-form-state/src/components/tests/Nested.tsx
```diff
@@ -0,0 +1,156 @@
...
+    });
+  });
+
+  it("updates the top level FormState's array when an inner field is updated", () => {
+    const product = {title: faker.commerce.productName()};
+    const newTitle = faker.commerce.productName();
+
+    const renderPropSpy = jest.fn(({fields: {product}}: any) => {
+      return (
+        <>
```
lemonmade: Don't need this?

packages/react-form-state/src/components/tests/Nested.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+    const form = mount(
+      <FormState initialValues={{product}}>{renderPropSpy}</FormState>,
+    );
+
+    form
+      .find(Input)
+      .props()
+      .onChange(newTitle);
+    form.update();
```
lemonmade: Why is this part needed?

packages/react-form-state/src/components/tests/Nested.tsx
```diff
@@ -0,0 +1,156 @@
...
+    });
+
+    const form = mount(
+      <FormState initialValues={{product}}>{renderPropSpy}</FormState>,
+    );
+
+    form
+      .find(Input)
+      .props()
+      .onChange(newTitle);
```
lemonmade: Bring in `trigger`?

packages/react-form-state/src/components/tests/Nested.tsx
TheMallen: yes!

packages/react-form-state/src/components/tests/Nested.tsx
```diff
@@ -0,0 +1,156 @@
...
+    form.update();
+
+    const {fields} = lastCallArgs(renderPropSpy);
+
+    expect(fields.product.value.title).toBe(newTitle);
```
lemonmade: `expect(renderPropSpy).toHaveBeenLastCalledWith(expect.objectContaining(...))`

packages/react-form-state/src/components/tests/Nested.tsx
```diff
@@ -0,0 +1,156 @@
...
+        error: undefined,
+        initialValue: product.adjective,
+        value: product.adjective,
+        onChange: expect.any(Function),
+        onBlur: expect.any(Function),
+      },
+    });
+  });
+
+  it("updates the top level FormState's array when an inner field is updated", () => {
```
lemonmade: Please use real apostrophes so you don't need to change the quote style

packages/react-form-state/src/components/tests/Nested.tsx
```diff
@@ -0,0 +1,156 @@
...
+    expect(renderPropArgs.department.error).toBe(field.error.department);
+  });
+});
+
+function lastCallArgs(spy: jest.Mock) {
+  const calls = spy.mock.calls;
+  return calls[calls.length - 1][0];
+}
+
+function TextField({onChange, ...props}: any) {
```
lemonmade: Need this and `Input`?

packages/react-form-state/src/components/tests/Nested.tsx
TheMallen: No, I see why you're confused though. This is coming from the keypath of the `title` field. `nodejs` seems to be an artifact of the test environment.

packages/react-form-state/src/test/FormState.test.tsx
```diff
@@ -124,6 +124,36 @@ describe('<FormState />', () => {
...
+      reset();
+      expect(fields).toMatchObject({
+        product: {
+          value: product,
+          initialValue: product,
+          dirty: false,
+        },
+      });
+    });
+  });
```
lemonmade: Are any of the above tests actually related to the changes you are making?

packages/react-form-state/src/components/tests/Nested.tsx
TheMallen: possibly

packages/react-form-state/src/test/FormState.test.tsx
```diff
@@ -652,8 +682,75 @@ describe('<FormState />', () => {
       expect(descriptionError.message).toEqual(submitErrors[1].message);
     });
   });
+
+  describe('performance', () => {
+    it('does not re-render form when state after onchange is identical', () => {
```
lemonmade: `onChange`

packages/react-form-state/src/FormState.ts
TheMallen: We can use spread here I think, though I'm not sure which reads nicer.

packages/react-form-state/src/FormState.ts
TheMallen: I avoided spread in cases where I would need to cast to satisfy typescript.

packages/react-form-state/src/test/FormState.test.tsx
```diff
@@ -652,8 +682,75 @@ describe('<FormState />', () => {
...
+      const renderPropSpy = jest.fn(() => null);
+
+      mount(
+        <FormState initialValues={{product, description}}>
+          {renderPropSpy}
+        </FormState>,
+      );
+
+      const formDetails = lastCallArgs(renderPropSpy);
+      formDetails.fields.product.onChange(product);
```
lemonmade: I think it would be nicer in the tests to have them simulate things a little better, so spread this field into an input and then trigger that input `onChange`

packages/react-form-state/src/test/FormState.test.tsx
```diff
@@ -652,8 +682,75 @@ describe('<FormState />', () => {
...
+
+  describe('performance', () => {
+    it('does not re-render form when state after onchange is identical', () => {
+      const product = faker.commerce.productName();
+      const description = faker.lorem.words();
+
+      const renderPropSpy = jest.fn(() => null);
+
+      mount(
+        <FormState initialValues={{product, description}}>
```
lemonmade: You don't need description, just distracts from the purpose of the test

packages/react-form-state/src/test/FormState.test.tsx
```diff
@@ -652,8 +682,75 @@ describe('<FormState />', () => {
...
+        </FormState>,
+      );
+
+      const formDetails = lastCallArgs(renderPropSpy);
+      formDetails.fields.product.onChange(product);
+
+      expect(renderPropSpy).toHaveBeenCalledTimes(1);
+    });
+
+    it('only renders renderProp once on mount', () => {
```
lemonmade: Not a sentence

packages/react-form-state/src/test/FormState.test.tsx
```diff
@@ -652,8 +682,75 @@ describe('<FormState />', () => {
...
+      mount(
+        <FormState initialValues={{product, description}}>
+          {renderPropSpy}
+        </FormState>,
+      );
+
+      expect(renderPropSpy).toHaveBeenCalledTimes(1);
+    });
+
+    it('does not cause pureComponent re-renders when a different field is changed', () => {
```
lemonmade: Your casing is all over the place :S also not a sentence

packages/react-form-state/src/FormState.ts
TheMallen: I don't think so because `shouldUpdateError` would be false not undefined.

packages/react-form-state/src/components/tests/Nested.tsx
TheMallen: It's pretty handy.

packages/react-form-state/src/FormState.ts
TheMallen: what if a new dirty field was added and one was removed?

packages/react-form-state/src/test/FormState.test.tsx
TheMallen: probably we actually want both cases

packages/react-form-state/src/FormState.ts
lemonmade: Unless I am mistaken that is impossible in this case given that you only operate on a single keypath?

packages/react-form-state/src/components/tests/Nested.tsx
TheMallen: we can probably burn this

packages/react-form-state/src/components/tests/Nested.tsx
TheMallen: the name field is generated from the keypath, but for some reason in the test env it always includes this as the first segment

packages/react-form-state/src/test/FormState.test.tsx
TheMallen: they're just tests that were missing

packages/react-form-state/src/test/FormState.test.tsx
TheMallen: same

packages/react-form-state/src/FormState.ts
loic-d: Yeah I had the same issues with the spread operator and casting

packages/react-form-state/src/FormState.ts
loic-d: In Discounts we tried to use the `[...dirtyFieldSet]` syntax we we could, but it's a personal preference.

packages/react-form-state/src/FormState.ts
TheMallen: I guess we can just do it with spread, I just really don't like having casts all over the place as it feels messy and error prone.

packages/react-form-state/src/FormState.ts
TheMallen: 🤔 true

packages/react-form-state/src/FormState.ts
TheMallen: I actually don't remember why I mutated here, I'll take a look when I go over this this afternoon

TheMallen: It was because of some awkwardness comparing the field object... Found a much nicer way to structure this though

packages/react-form-state/src/components/tests/Nested.tsx
TheMallen: I don't like that for single field comparisons when we already need `lastCallArgs` in this file anyway.

TheMallen: That said I should change some of the other places to do that.

packages/react-form-state/src/test/FormState.test.tsx
TheMallen: Actually I think they were partially some kind of rebase issue

packages/react-form-state/src/test/FormState.test.tsx
TheMallen: I do need it in this object since we are testing whether the description field rerenders. I can make this value just an empty string though.

packages/react-form-state/src/FormState.ts
TheMallen: I'm leaving this for now, seems like there's a missing transpilation issue or something, and I kind of like how this reads anyway.

packages/react-form-state/src/tests/FormState.test.tsx
```diff
@@ -105,15 +101,37 @@ describe('<FormState />', () => {
...
       formDetails.fields.product.onChange(faker.commerce.productName());
-      form.update();
 
       const {reset} = lastCallArgs(renderPropSpy);
       reset();
 
-      form.update();
-
       const {fields} = lastCallArgs(renderPropSpy);
+      expect(fields).toMatchObject({
```
TheMallen: I left a lot of these using `lastCallArgs` because we only actually care about a few of the fields, and doing  `expect.objectContaining` with `toHaveLastBeenCalledWith` adds too much noise and stuff we don't care about IMO. 

I did use that strategy in places we do actually care about every field though.

packages/react-form-state/src/tests/FormState.test.tsx
```diff
@@ -156,17 +174,14 @@ describe('<FormState />', () => {
       const formDetails = lastCallArgs(renderPropSpy);
       formDetails.fields.product.onChange(faker.commerce.productName());
       formDetails.fields.color.onChange(faker.commerce.color());
-      form.update();
```
TheMallen: found and replaced all of these because they weren't needed

packages/react-form-state/src/FormState.ts
```diff
@@ -186,52 +180,106 @@ export default class FormState<
...
-  private getHandlersForField(path: keyof FieldStates<Fields>) {
+  @bind()
+  private fieldWithHandlers<Key extends keyof Fields>(
+    field: FieldStates<Fields>[Key],
+    fieldPath: Key,
+  ) {
     return {
-      onChange: this.updateField.bind(this, path),
-      onBlur: this.blurField.bind(this, path),
+      ...(field as FieldState<Fields[Key]>),
```
tzvipm: do we not need the `name: path` piece any more?

TheMallen: Added back

PR #171, opened 2018-06-29
🎨 [react-shortcuts] Add better key types

9at8:
This adds more (if not all) key types from https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values


packages/react-shortcuts/src/Shortcut/Shortcut.tsx
```diff
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import {contextTypes} from '../ShortcutProvider';
-import {Key} from '../types';
+import {Key} from '../keys';
```
TheMallen: We should make `Key` the `default` export and keep the more specific one named exports I think.

packages/react-shortcuts/src/keys.ts
```diff
@@ -0,0 +1,408 @@
...
+  | 'AudioTrebleUp'
+  | 'AudioVolumeDown'
+  | 'AudioVolumeMute'
+  | 'AudioVolumeUp'
+  | 'MicrophoneVolumeDown'
+  | 'MicrophoneVolumeMute'
+  | 'MicrophoneVolumeUp';
+
+export type TVControlKey =
+  | 'TV'
```
TheMallen: wow

9at8:
@tzvipm Probably a good idea 😅

PR #179, opened 2018-07-09
Add type definition for phoneNumberPrefix

koshini:
Add type definition for `phoneNumberPrefix` for `Country` to reflect [changes in the country_service API](https://github.com/Shopify/country-service/pull/42/files)

### DIFF
--- a/packages/address/src/tests/fixtures/CA_en.ts
+++ b/packages/address/src/tests/fixtures/CA_en.ts
@@ -5,6 +5,7 @@ const data = {
     code: 'CA',
     zoneKey: 'province',
     zipKey: 'postalCode',
+    phoneNumberPrefix: 1,
     format: {
       edit:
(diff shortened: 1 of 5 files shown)


packages/address/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@shopify/address",
-  "version": "1.0.1",
+  "version": "1.0.2",
```
nimzco: @TheMallen, do we have to keep the version bump in a separate PR / commit?

TheMallen: yes

TheMallen: The `release` line tool will find all changed packages and let the person who runs it choose new versions

koshini:
Reverted the version bump! 🔙  @nimzco @TheMallen

PR #194, opened 2018-07-11
⬆️ bump ts-jest

TheMallen:
The greenkeeper PR failed so I'm seeing if this works.


tzvipm:
@TheMallen how is this different from what greenkeeper did?

TheMallen:
CI passes on it

PR #195, opened 2018-07-12
🐛 Make modifier keys work in react-shortcuts

### LINKED ISSUE #133: Shortcuts module modifier keys could use some refactoring
Meta is cmd on Mac and Alt on Windows. We're also checking the modifier keys just like any other keys, which would make the shortcut not work if you didn't start the shortcut with the modifier key. Ideally this module should check the event.metaKey or event.shiftKey, etc properties on the last key that was pressed 🤔

Even though right now this module works for the most part, if I were to rewrite it that's something I'd fix

9at8:
Fixes #133 

We weren't checking for modifier keys properly using `getModifierState`, but just checking `event.key` instead. This doesn't always work, as the user would need to press the modifier keys in that particular order. (which shouldn't be the case, example: `Cmd + Option + c` equals `Option + Cmd + c`)

Doing what we did also meant that the user couldn't perform an `onMatch` by holding the modifier keys and then repeatedly pressing a normal key. (example: `Cmd (hold) + v + v + v` pastes thrice, but it would work only once if we use the current version of react-shortcuts)

Considering all these problems, I think that separating the `keys` prop into `modifierKeys` prop and `keys` prop made most sense.

Also, I'll push this and #171 together, and then release.

### DIFF
--- a/packages/react-shortcuts/src/Shortcut/Shortcut.tsx
+++ b/packages/react-shortcuts/src/Shortcut/Shortcut.tsx
@@ -3,8 +3,8 @@ import {contextTypes} from '../ShortcutProvider';
 import Key, {ModifierKey} from '../keys';
 
 export interface Props {
-  keys: Key[];
-  modifierKeys?: ModifierKey[];
+  ordered: Key[];
+  held?: ModifierKey[];
   node?: HTMLElement | null;
   ignoreInput?: boolean;
   onMatch(keys: Key[]): void;
@@ -19,8 +19,8 @@ export default class Shortcut extends React.Component<Props, never> {
   static contextTypes = contextTypes;
   public data = {
     node: this.props.node,
-    keys: this.props.keys,
-    modifierKeys: this.props.modifierKeys,
+    ordered: this.props.ordered,
+    held: this.props.held,
     ignoreInput: this.props.ignoreInput || false,
     onMatch: this.props.onMatch,
     allowDefault: this.props.allowDefault,
--- a/packages/react-shortcuts/src/Shortcut/Shortcut.tsx
+++ b/packages/react-shortcuts/src/Shortcut/Shortcut.tsx
@@ -1,9 +1,10 @@
 import * as React from 'react';
 import {contextTypes} from '../ShortcutProvider';
-import Key from '../keys';
+import Key, {ModifierKey} from '../keys';
 
 export interface Props {
   keys: Key[];
+  modifierKeys?: ModifierKey[];
   node?: HTMLElement | null;
   ignoreInput?: boolean;
   onMatch(keys: Key[]): void;
@@ -19,6 +20,7 @@ export default class Shortcut extends React.Component<Props, never> {
   public data = {
     node: this.props.node,
     keys: this.props.keys,
+    modifierKeys: this.props.modifierKeys,
     ignoreInput: this.props.ignoreInput || false,
     onMatch: this.props.onMatch,
     allowDefault: this.props.allowDefault,
--- a/packages/react-shortcuts/src/Shortcut/Shortcut.tsx
+++ b/packages/react-shortcuts/src/Shortcut/Shortcut.tsx
@@ -7,7 +7,7 @@ export interface Props {
   held?: ModifierKey[];
   node?: HTMLElement | null;
   ignoreInput?: boolean;
-  onMatch(keys: Key[]): void;
+  onMatch(matched: {ordered: Key[]; held?: ModifierKey[]}): void;
   allowDefault?: boolean;
 }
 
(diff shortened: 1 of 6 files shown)


packages/react-shortcuts/src/tests/ShortcutManager.test.tsx
```diff
@@ -235,18 +231,18 @@ describe('ShortcutManager', () => {
 
     mount(
       <ShortcutProvider>
-        <Shortcut keys={['shiftKey']} onMatch={spy} />
+        <Shortcut keys={['a']} onMatch={spy} />
```
tzvipm: I guess modifier keys can't be their own shortcut any more. Makes sense, but worth noting.

9at8: yeah, it doesn't really make sense to make them their own shortcut tbh

9at8:
@tzvipm there's two tests for the new behaviour now. I don't think we need more, but if you think of a particular case that should be in here, lmk and I'll add it. 👍

packages/react-shortcuts/src/tests/ShortcutManager.test.tsx
```diff
@@ -235,18 +214,65 @@ describe('ShortcutManager', () => {
...
+      );
+
+      keydown('/', document, {
+        getModifierState: key => modifierKeys.includes(key),
+      });
+
+      expect(fooSpy).toHaveBeenCalled();
+    });
+
+    it('doesn’t match shortcut when all all modifier keys not pressed', () => {
```
TheMallen: all all?

packages/react-shortcuts/src/ShortcutManager/ShortcutManager.tsx
```diff
@@ -60,27 +61,36 @@ export default class ShortcutManager {
...
+        const partiallyMatching = arraysMatch(
+          this.keysPressed,
+          keys.slice(0, this.keysPressed.length),
+        );
 
-      if (node) {
-        const onFocusedNode = document.activeElement === node;
-        return partiallyMatching && onFocusedNode;
-      }
+        if (node) {
```
TheMallen: Weird spacing

packages/react-shortcuts/src/Shortcut/Shortcut.tsx
```diff
@@ -1,9 +1,10 @@
 import * as React from 'react';
 import {contextTypes} from '../ShortcutProvider';
-import {Key} from '../types';
+import Key, {ModifierKey} from '../keys';
 
 export interface Props {
   keys: Key[];
+  modifierKeys?: ModifierKey[];
```
TheMallen: Are we sure we never want to allow ordered key combos to be modifier keys or non-modifier-keys to be held?

I we don't have any shortcuts like that not then it's probably a good assumption. If we do we should consider changing the api to let you use all the keys for either and look something more like:

```TSX
type AnyKey = Key | ModifierKey;

interface ShortcutProps {
  ordered: AnyKey[];
  held: AnyKey[];
}

...

<Shortcut ordered={['f', 'Alt']} held={['Shift']} />
```

packages/react-shortcuts/src/keys.ts
```diff
@@ -0,0 +1,445 @@
+type Key =
```
TheMallen: Looks like we should close the other PR that has this stuff?

packages/react-shortcuts/src/ShortcutManager/ShortcutManager.tsx
9at8: What is the expected spacing?

packages/react-shortcuts/src/Shortcut/Shortcut.tsx
9at8: I like the `ordered` and the `held` props, but I don't think we should allow any key to be held, because:

- it doesn't really make sense,
- could be a hacky solution, as `event.getModifierState` is only meant for checking if a modifier key is held down.

packages/react-shortcuts/src/keys.ts
9at8: yes, I'll merge that PR in just before this one, just don't wanna do two separate releases :sweat_smile:

packages/react-shortcuts/src/ShortcutManager/ShortcutManager.tsx
TheMallen: I mean in github it's showing it as like a different character or something. Probably running prettier on it will fix?

packages/react-shortcuts/src/Shortcut/Shortcut.tsx
TheMallen: 👍

PR #199, opened 2018-07-13
🐛 render helmet scripts

### LINKED ISSUE #198: [react-html] Support for <script>some code here</script>
My use case: adding Google Analytics to a small app directly in the `<head>`.

Instead I'm having to do this, which doesn't appear in the `<head>` but in the body instead:

```jsx
        <script
          async
          src="https://www.googletagmanager.com/gtag/js?id=UA-49178120-48"
        />
        <script
          dangerouslySetInnerHTML={{
            __html: `
          window.dataLayer = window.dataLayer || [];
          function gtag(){dataLayer.push(arguments);}
          gtag('js', new Date());

          gtag('config', 'UA-49178120-48');
          `,
          }}
        />
```

See https://github.com/Shopify/polaris-react/commit/7b9d125369f6b798cd9f2c12815d16b390180e82 where I sadly had to get rid of react-html just so I could add Google Analytics to the app.

Is there something I'm missing / doing wrong? (otherwise: please consider this as a feature request :) )

TheMallen:
closes #198 

This PR fixes an oversight where we rendered links, meta, and html attributes from helmet, but not scripts.

We don't want programmatic script tags to be commonly used, but for things like tracking script we make helmet available as an escape hatch.

### DIFF
--- a/packages/react-html/src/HTML.tsx
+++ b/packages/react-html/src/HTML.tsx
@@ -101,10 +101,11 @@ export default function HTML({
       <head>
         {helmet.title.toComponent()}
         {helmet.meta.toComponent()}
+        {helmet.script.toComponent()}
         {helmet.link.toComponent()}
+
         {stylesMarkup}
         {headDataMarkup}
-
         {blockingScriptsMarkup}
       </head>
 
--- a/packages/react-html/src/tests/HTML.test.tsx
+++ b/packages/react-html/src/tests/HTML.test.tsx
@@ -139,6 +139,7 @@ describe('<HTML />', () => {
       const html = mount(<HTML {...mockProps} />);
       expect(html.find('head').contains(link)).toBe(true);
     });
+
     it('includes the script component', () => {
       const script = (
         <script
--- a/packages/react-html/src/tests/HTML.test.tsx
+++ b/packages/react-html/src/tests/HTML.test.tsx
@@ -142,10 +142,7 @@ describe('<HTML />', () => {
 
     it('includes the script component', () => {
       const script = (
-        <script
-          type="text/javascript"
-          dangerouslySetInnerHTML={{__html: 'alert("hi")'}}
-        />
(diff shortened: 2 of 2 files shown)


packages/react-html/src/HTML.tsx
```diff
@@ -101,10 +101,11 @@ export default function HTML({
       <head>
         {helmet.title.toComponent()}
         {helmet.meta.toComponent()}
+        {helmet.script.toComponent()}
```
TheMallen: since this is mainly for tracking scripts we put it right before links.

packages/react-html/src/tests/HTML.test.tsx
```diff
@@ -140,6 +140,22 @@ describe('<HTML />', () => {
       expect(html.find('head').contains(link)).toBe(true);
     });
 
+    it('includes the script component', () => {
+      const script = (
+        <script
+          type="text/javascript"
```
kaelig: This line isn't useful

packages/react-html/src/tests/HTML.test.tsx
```diff
@@ -140,6 +140,22 @@ describe('<HTML />', () => {
...
+          dangerouslySetInnerHTML={{__html: 'alert("hi")'}}
```
kaelig: Would it also make sense to test this API?


```jsx
<Helmet script={[{innerHtml: 'some js'}]} />
```

TheMallen: I think that might be more testing how Helmet works than testing our api. It might be a good idea to change the tests to just check that the right Helmet render methods were called though.

TheMallen: Similarly, we might want to document that doing regular helmet stuff will work with this out of the box.

TheMallen:
🎉

packages/react-html/src/tests/HTML.test.tsx
TheMallen: true

PR #200, opened 2018-07-13
[react-form-state] Adds support for arrays of validators to validateNested

iainmcampbell:
was just trying out this great new library and noticed that this:

```js
validateNested({
  field: [validator1, validator2],
});
```

was throwing a TypeError, so thought I'd put up a quick PR to add support for arrays of validators to `validateNested`. I basically just copied the functionality from [validateFieldValue](https://github.com/Shopify/quilt/blob/2e4c911cbb74391225c6d5ca61f16b9f0d1406e5/packages/react-form-state/src/FormState.ts#L306).

### DIFF
--- a/packages/react-form-state/src/validators.ts
+++ b/packages/react-form-state/src/validators.ts
@@ -48,27 +48,30 @@ export function validateNested<Input extends Object, Fields>(
   // eslint-disable-next-line consistent-return
   return (input: Input, fields: Fields) => {
     const errors = mapObject<Input, any>(input, (value, field) => {
-      if (validatorDictionary[field]) {
-        if (typeof validatorDictionary[field] === 'function') {
-          return validatorDictionary[field](value, fields);
-        }
-        if (!isArray(validatorDictionary[field])) {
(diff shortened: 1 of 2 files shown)


iainmcampbell:
will fix ci errors, guess I do have to use `eslint-ignore` instead of `return null`

packages/react-form-state/src/validators.ts
```diff
@@ -48,9 +48,27 @@ export function validateNested<Input extends Object, Fields>(
   // eslint-disable-next-line consistent-return
   return (input: Input, fields: Fields) => {
     const errors = mapObject<Input, any>(input, (value, field) => {
-      return (
-        validatorDictionary[field] && validatorDictionary[field](value, fields)
-      );
+      if (validatorDictionary[field]) {
```
TheMallen: doing an intermediary const and an early return instead of wrapping the whole function in an if is more idiomatic to the repo's code style.

```
const validate = validatorDictionary[field];

if (validate == null) {
  return null;
}

if (typeof validate === 'function') {
```

PR #202, opened 2018-07-13
✨ [react-form-state] propagate errors down

### LINKED ISSUE #181: [react-form-state] Pass submit errors down to fields
We should be passing submit errors down to fields when they have a path that matches an existing field.

When a field with these errors is revalidated, we should discard the server errors.

TheMallen:
closes #181 

This PR adds the ability for submit errors to be passed down to the appropriate fields. They are cleared the next time the user types by the extant validation logic.

It also cleans up some of the types around those errors and removes logic around generating keypaths (the arrays that come from graphql work fine).

## 🎩 instructions
- pull this PR down
- run `yarn link` in the `packages/react-form-state` directory
- spin up a CRA or a Playground in any sewing-kit project
- run `yarn link @shopify/react-form-state` in your new app directory
- add the following code to App.js

```jsx
// eslint-disable
import {
  AppProvider,
  Page,
  TextField,
  FormLayout,
  PageActions,
  Spinner,
  Banner,
  Stack,
  Card,
  Layout,
  ButtonGroup,
  Button,
} from '@shopify/polaris';
import FormState, {
  validators,
  validateList,
  validateNested,
  arrayUtils,
} from '@shopify/react-form-state';

const {required, numericString, nonNumericString, lengthMoreThan} = validators;

export default class Playground extends React.Component {
  render() {
    return (
      <AppProvider>
        <FormState
          initialValues={{
            title: 'Cool title',
            description: 'Cool product',
            sku: 'abc',
            quantity: '0',
            firstVariant: {
              option: 'color',
              value: 'orange',
              price: '123123',
            },
            variants: [],
          }}
          validators={{
            title: required('Required'),
            quantity: numericString('Must be a number'),
            sku: lengthMoreThan(3, 'Must  be longer than 3 characters'),
            firstVariant: validateNested({
              option: required('required'),
              price: numericString('value must be numeric'),
            }),
            variants: validateList({
              option: nonNumericString('option must be nonNumeric'),
              price: numericString('value must be numeric'),
            }),
          }}
          onSubmit={someAsyncThing}
        >
          {formDetails => {
            const {
              fields,
              dirty,
              reset,
              submit,
              valid,
              submitting,
              errors,
            } = formDetails;
            const {
              title,
              description,
              sku,
              quantity,
              variants,
              firstVariant,
            } = fields;

            const submitAction = {
              content: 'Save',
              onAction: submit,
              loading: submitting,
            };

            const resetAction = {
              content: 'Reset',
              onAction: reset,
              disabled: !dirty || submitting,
            };

            const errorBanner = errors.length > 0 && (
              <Banner status="critical">
                <ul>
                  {errors.map(error => (
                    <li key={error.message + error.field}>{error.message}</li>
                  ))}
                </ul>
              </Banner>
            );

            function addVariant() {
              variants.onChange(
                arrayUtils.push(variants.value, {
                  option: '',
                  value: '',
                  price: '0',
                }),
              );
            }

            return (
              <Page
                title="Products"
                primaryAction={submitAction}
                secondaryActions={[resetAction]}
              >
              <Layout>
                  <Layout.Section>
                    {errorBanner}
                  </Layout.Section>
                  <Layout.Section>
                    <Stack distribution="center">
                      {submitting && <Spinner />}
                    </Stack>
                    <Card>
                      <Card.Section>
                        <FormLayout>
                          <TextField label="Title" {...title} />
                          <TextField label="Description" {...description} />
                          <TextField label="SKU" {...sku} />
                          <TextField
                            type="number"
                            label="Quantity"
                            {...quantity}
                          />
                        </FormLayout>
                      </Card.Section>
                    </Card>
                  </Layout.Section>
                  <Layout.Section>
                    <Card title="Variants">
                      <Card.Section title="Default">
                        <FormLayout>
                          <FormState.Nested field={firstVariant}>
                            {({option, value, price}) => (
                              <React.Fragment>
                                <TextField label="option" {...option} />
                                <TextField label="value" {...value} />
                                <TextField label="price" {...price} />
                              </React.Fragment>
                            )}
                          </FormState.Nested>
                        </FormLayout>
                      </Card.Section>
                      <Card.Section>
                        <FormLayout>
                          <FormState.List field={variants}>
                            {renderVariants}
                          </FormState.List>
                          <ButtonGroup>
                            <Button onClick={addVariant} primary>
                              Add variant
                            </Button>
                          </ButtonGroup>
                        </FormLayout>
                      </Card.Section>
                    </Card>
                  </Layout.Section>
                  <Layout.Section>
                    <PageActions
                      primaryAction={submitAction}
                      secondaryActions={[resetAction]}
                    />
                  </Layout.Section>
                </Layout>
              </Page>
            );
          }}
        </FormState>
      </AppProvider>
    );
  }
}

function renderVariants({option, value, price}) {
  return (
    <React.Fragment>
      <TextField label="option" {...option} />
      <TextField label="value" {...value} />
      <TextField label="price" {...price} />
    </React.Fragment>
  );
}

function someAsyncThing() {
  return new Promise(resolve => {
    setTimeout(() => {
      resolve([
        {message: 'everything is broken, go home'},
        {field: 'title', message: 'title sounds dumb'},
        {
          field: ['variants', '0', 'option'],
          message: 'variants option is not real',
        },
        {
          field: ['firstVariant', 'option'],
          message: 'default variants option is not real',
        },
      ]);
    }, 2000);
  });
}
```

### DIFF
--- a/packages/react-form-state/src/FormState.ts
+++ b/packages/react-form-state/src/FormState.ts
@@ -1,18 +1,17 @@
 import * as React from 'react';
 import isEqual from 'lodash/isEqual';
 import isArray from 'lodash/isArray';
+import set from 'lodash/set';
 import {memoize, bind} from 'lodash-decorators';
 
 import {mapObject} from './utilities';
-import {FieldDescriptors, ClientError, FieldState} from './types';
+import {FieldDescriptors, FieldState} from './types';
 import {List, Nested} from './components';
 
 interface RemoteError {
-  field: string[] | null;
+  field?: string[] | null;
   message: string;
 }
-
-export type FormError = RemoteError | ClientError;
 
 export type FieldStates<Fields> = {
   [FieldPath in keyof Fields]: FieldState<Fields[FieldPath]>
@@ -23,7 +22,7 @@ type MaybePromise<T> = T | Promise<T>;
 
 interface SubmitHandler<Fields> {
   (formDetails: FormData<Fields>):
-    | MaybePromise<FormError[]>
+    | MaybePromise<RemoteError[]>
     | MaybePromise<void>;
 }
 
@@ -41,7 +40,7 @@ export interface FormData<Fields> {
   fields: FieldDescriptors<Fields>;
   dirty: boolean;
   valid: boolean;
-  errors: ClientError[];
+  errors: RemoteError[];
 }
 
 export interface FormDetails<Fields> extends FormData<Fields> {
@@ -59,9 +58,9 @@ interface Props<Fields> {
 
 interface State<Fields> {
   submitting: boolean;
-  errors: ClientError[];
-  dirtyFields: (keyof Fields)[];
   fields: FieldStates<Fields>;
+  dirtyFields: (keyof Fields)[];
+  errors: RemoteError[];
 }
 
 export default class FormState<
@@ -128,7 +127,7 @@ export default class FormState<
   }
 
   private get valid() {
-    return this.allErrors.length === 0;
+    return this.errorCount === 0;
   }
 
   private get fields() {
@@ -142,7 +141,7 @@ export default class FormState<
     return fieldDescriptors;
   }
 
-  private get allErrors(): ClientError[] {
+  private get errorCount(): number {
     const {errors, fields} = this.state;
 
     const fieldErrors = Object.keys(fields)
@@ -152,9 +151,9 @@ export default class FormState<
       })
       .filter(error => {
         return error.field != null && error.message != null;
-      }) as ClientError[];
+      });
 
-    return [...errors, ...fieldErrors];
+    return fieldErrors.length + errors.length;
   }
 
   @bind()
@@ -178,7 +177,7 @@ export default class FormState<
 
     const result = await onSubmit(formData);
     if (result) {
-      this.updateErrors(result);
+      this.updateRemoteErrors(result);
     }
 
     this.setState({submitting: false});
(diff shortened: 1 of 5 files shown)


TheMallen:
`Nested` and `List`'s current tests check that they handle nested error objects correctly. `<FormState />` is just now able to generate those from submit errors.

packages/react-form-state/src/components/Nested.tsx
```diff
@@ -30,7 +30,7 @@ export default class Nested<Fields> extends React.PureComponent<
           name: `${name}.${fieldPath}`,
           initialValue: initialFieldValue,
           dirty: value !== initialFieldValue,
-          error: get(error, [fieldPath]),
```
TheMallen: wrapping this in an array was unnecessary, tests remain green.

packages/react-form-state/src/FormState.ts
```diff
@@ -152,9 +151,9 @@ export default class FormState<
       })
       .filter(error => {
         return error.field != null && error.message != null;
-      }) as ClientError[];
+      });
 
-    return [...errors, ...fieldErrors];
+    return fieldErrors.length + errors.length;
```
tzvipm: Is it possible to have an error counted twice here?

packages/react-form-state/src/FormState.ts
```diff
@@ -1,19 +1,18 @@
...
 import isArray from 'lodash/isArray';
+import set from 'lodash/set';
 import {memoize, bind} from 'lodash-decorators';
 
 import {mapObject} from './utilities';
-import {FieldDescriptors, ClientError, FieldState} from './types';
+import {FieldDescriptors, FieldState} from './types';
 import {List, Nested} from './components';
 
 interface RemoteError {
```
tzvipm: export this?

packages/react-form-state/src/FormState.ts
TheMallen: doesn't actually matter, all we care about is that it is greater than 0

packages/react-form-state/src/FormState.ts
TheMallen: We have a separate PR for that

packages/react-form-state/src/FormState.ts
TheMallen: Can maybe just roll this logic into `valid` so it's more obvious that that's what matters

packages/react-form-state/src/FormState.ts
```diff
@@ -346,25 +336,27 @@ export default class FormState<
...
+            return accumulator;
+          }
 
-        if (isArray(field)) {
-          return {message, field: field.join('.')};
-        }
-
-        return {message, field};
-      });
+          return set(accumulator, field, message);
```
loic-d: You are using `set` from lodash because the `field` can be a path?

TheMallen: yep it can be an array of path segments (that's what it looks like when they come from graphql)

PR #206, opened 2018-07-19
🐛 Fixes undefined name in FormState.Nested

darrenhebner:
This PR fixes a bug with `FormState.Nested` where we are referencing an undefined value.

If you pass this object to `initialValues` with `FormState`:
```js
{
  budget: {
    amount: '25'
  }
}
```

The `name` for the nested field should become `budget.amount`.

### DIFF
--- a/packages/react-form-state/src/components/tests/Nested.test.tsx
+++ b/packages/react-form-state/src/components/tests/Nested.test.tsx
@@ -30,7 +30,7 @@ describe('<Nested />', () => {
 
     expect(renderPropSpy).toBeCalledWith({
       title: {
-        name: 'nodejs.title',
+        name: 'product.title',
         dirty: false,
         // eslint-disable-next-line no-undefined
         error: undefined,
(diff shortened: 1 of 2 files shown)


darrenhebner:
I'm a bit confused by the CI failures. Looks like some other open PRs are failing for the same reasons.

TheMallen:
this is just failing because you need to fix the test as well 
image

darrenhebner:
Oh ok. I thought I was seeing a bunch of unrelated stuff. I'll fix 😄

packages/react-form-state/src/components/tests/Nested.test.tsx
```diff
@@ -30,7 +30,7 @@ describe('<Nested />', () => {
 
     expect(renderPropSpy).toBeCalledWith({
       title: {
-        name: 'nodejs.title',
+        name: 'product.title',
```
darrenhebner: Just want to double check that it makes sense to update the test. I didn't understand where `nodejs` was coming from.

TheMallen: yeah this makes sense 👍 that was coming form the bug you found probably

Issue #182, opened 2018-07-09
[react-form-state] Lighten library weight


TheMallen:
We should see if we can lighten the weight of the library by cutting some lodash functions or something. Right now it's pretty big. (19.2kb minzipped)

TheMallen:
it's a pretty acceptable size now that we cut out most of lodash. Shelving for now.

PR #207, opened 2018-07-19
⚡ remove unnecessary lodash functions for tiny local implementations

TheMallen:
part of #182 

This PR removes two of our lodash function imports from `<FormState />` in favour of less heavy implementations. In the case of `isArray` we just use `Array.isArray`. In the case of `isEqual` we implement a less crufty version of the algorithm ourselves.

### DIFF
--- a/packages/react-form-state/src/utilities.ts
+++ b/packages/react-form-state/src/utilities.ts
@@ -1,3 +1,61 @@
+export function deepEqual(first: any, second: any, maxDepth = 20, depth = 0) {
+  const newDepth = depth + 1;
+
+  if (newDepth > maxDepth) {
+    throw new Error('Maximum comparison depth reached.');
+  }
+
+  if (first === second) {
+    return true;
+  }
+
+  if (typeof first !== typeof second) {
+    return false;
+  }
+
+  if (first === null || second === null) {
+    return false;
+  }
+
+  if (Array.isArray(first)) {
+    return arrayEqual(first, second, depth, maxDepth);
+  }
+
+  if (typeof first === 'object') {
+    return objectEqual(first, second, depth, maxDepth);
+  }
+
+  return first.valueOf() === second.valueOf();
+}
+
+function arrayEqual(
+  first: any[],
+  second: any[],
+  depth: number,
+  maxDepth: number,
+) {
+  if (first.length !== second.length) return false;
+
+  return first.every((valueOfFirst, index) =>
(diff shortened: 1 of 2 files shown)


packages/react-form-state/src/utilities.ts
```diff
@@ -1,3 +1,61 @@
+export function deepEqual(first: any, second: any, maxDepth = 2000, depth = 0) {
```
tzvipm: @TheMallen did you just come up with this code, or did you copy it from somewhere?

TheMallen: Came up with it, though I've seen many versions of this written before.

tzvipm: fair enough. I just wonder what the nuanced differences are in behavior here. Also, what is the performance tradeoff over using `lodash.isequal`? Is it worth the risk of deviating from an implementation that is industry-trusted and battle tested? Perhaps we will encounter bugs with this implementation that are unexpected?

TheMallen: I don't think runtime microperformance is the right thing to optimize for. We should worry about bundle size, since it's already a problem for us, and causing too many rerenders (which this does not effect), and only worry about microperf if we encounter a real issue. That being said, even if they had some black magic going on to speed up deep equality operations that makes it ten times faster I don't think that would actually be worth the weight. The only two places we use this are comparing new initialValues coming in and comparing fields to their initial values on update. the former will very seldomly happen (only after a save or on first mount), and the latter is 90 percent of the time going to be on scalar values. 

As for edgecases we have to remember that a generic library version of something like this needs to worry about a much wider variety of objects than we do. We should not be worrying about things like objects with weird crazy prototype setups, or even just ones with getters and setters. In fact we shouldn't even care about objects with methods. The only things these are being applied to are form data which *should always* be a plain old javascript object.

TheMallen: It also might actually be a good idea to drop `deepEquals/isEqual` from the updater check entirely and just use `Object.is` there, since we expect immutable updates to those anyway.

TheMallen:
Yeah, IMO the best argument for using `lodash/isEqual` is just that many projects will have it anyway, in which case a good build setup (like we have in Web) can make it actually end up slightly cheaper. IMO the cost of +50 lines  on pages that already have lodash equals versus several kilobytes for pages that don't is a better balance though.

TheMallen:
after chatting with chris and gord, deferring this for a bit.

PR #217, opened 2018-08-01
✨ optimistically regenerate form initial values on successful submit

### LINKED ISSUE #216: [react-form-state] Optimistically update initialValues when submit is successful
We should immediately de-dirty the form and update initialValues when a submit is successful. This will make things snappier and also cover cases where the submit is to a rest api or a graphQL mutation that won't immediately update the cache.

TheMallen:
fixes #216 

## what this
When a submit handler is successful we always want the form to be marked clean and have it's initial values updated as soon as possible. In some cases this might not be able to happen implicitly if the submit function does not also update

## worries
Weird stuff could happen if you do the update and then somehow the parent component is re-rendered and gives us new initialValues that don't match (as we would reinitialize the form again). I can't think of a state where that's wrong though.

## how 2 test
- checkout repo and branch
- yarn build && yarn link
- go to your consuming project
- yarn link `@shopify/react-form-state`

### DIFF
--- a/packages/react-form-state/src/FormState.ts
+++ b/packages/react-form-state/src/FormState.ts
@@ -176,7 +176,6 @@ export default class FormState<
     }
   }
 
-  @bind
   private makeClean() {
     this.setState(({fields}) => createFormState(valuesFromFields(fields)));
   }
--- a/packages/react-form-state/src/FormState.ts
+++ b/packages/react-form-state/src/FormState.ts
@@ -166,12 +166,19 @@ export default class FormState<
 
     this.setState({submitting: true});
 
-    const result = await onSubmit(formData);
-    if (result) {
-      this.updateRemoteErrors(result);
-    }
+    const errors = await onSubmit(formData);
 
-    this.setState({submitting: false});
+    if (errors) {
+      this.setState({submitting: false});
+      this.updateRemoteErrors(errors);
+    } else {
+      this.makeClean();
+    }
+  }
+
+  @bind
+  private makeClean() {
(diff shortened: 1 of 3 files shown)


packages/react-form-state/src/FormState.ts
```diff
@@ -166,12 +166,18 @@ export default class FormState<
 
     this.setState({submitting: true});
 
-    const result = await onSubmit(formData);
-    if (result) {
-      this.updateRemoteErrors(result);
+    const errors = await onSubmit(formData);
```
tzvipm: wait, so `onSubmit` now returns errors or soemthing falsey, whereas before it returned a result?

packages/react-form-state/src/FormState.ts
```diff
@@ -166,12 +166,18 @@ export default class FormState<
...
+
+    if (errors) {
+      this.setState({submitting: false});
+      this.updateRemoteErrors(errors);
+    } else {
+      this.makeClean();
```
tzvipm: how can this fix the issue mentioned in the PR if we're still waiting for the submit before getting here?

packages/react-form-state/src/FormState.ts
TheMallen: it never returned a result it was just poorly named

packages/react-form-state/src/FormState.ts
TheMallen: submit wasn't stalling or anything the cache just wasn't updating properly

TheMallen:
I'm going to close this for now after a chat with @lemonmade about not exposing ways to do this that aren't updating the props.

PR #219, opened 2018-08-02
add private app support to `shopify-graphql-proxy` and fix the contex…

michelleyschen:
…t typing

This PR does two things.

1. Fix type error that come with extending `Context`
example of this type error:
while using 'koa-session'
```
import * as session from 'koa-session';
```

the typing of koa Context was then redeclare by koa-session
```
declare module "koa" {
    interface Context {
        session: session.Session | null;
        readonly sessionOptions: session.opts | undefined;
    }
}
```
which resolved in typing error for `SessionContext` as follow

```
Interface 'SessionContext' incorrectly extends interface 'Context'.
  Types of property 'session' are incompatible.
    Type '{ accessToken?: string | undefined; shop?: string | undefined; } | undefined' is not assignable to type 'Session | null'.
      Type 'undefined' is not assignable to type 'Session | null'.
```

2. Adding private app support.
`shopify-graphql-proxy` now takes in `shop` &  `password` as options,
if both option exist, use password as accessToken.


packages/koa-shopify-graphql-proxy/src/shopify-graphql-proxy.ts
```diff
@@ -1,24 +1,39 @@
...
 
-export interface SessionContext extends Context {
-  session?: {accessToken?: string; shop?: string};
-}
-
 export const PROXY_BASE_PATH = '/graphql';
 export const GRAPHQL_PATH = '/admin/api/graphql.json';
 
-export default function shopifyGraphQLProxy() {
+export interface ProxyOptions {
```
TheMallen: We need both or nothing right? I don't think these should be optional, rather the whole object should be.

packages/koa-shopify-graphql-proxy/src/shopify-graphql-proxy.ts
```diff
@@ -1,24 +1,39 @@
...
+  password: privateAppPassword,
+}: ProxyOptions = {}) {
   return async function shopifyGraphQLProxyMiddleware(
-    ctx: SessionContext,
+    ctx: Context,
     next: () => Promise<any>,
   ) {
-    const {session = {}} = ctx;
-    const {accessToken, shop} = session;
+    let accessToken: string | null = null;
```
TheMallen: IMO a bit nicer to do something like 

```
const shop = privateAppShop || session.shop
const accessToken = privateAppPassword || session.accessToken
```

and then just burn the if else.

packages/koa-shopify-graphql-proxy/src/shopify-graphql-proxy.ts
michelleyschen: I was future proofing that if there are going to be other config in the future, they can be easily combined. But I agree for now it is both or nothing.

packages/koa-shopify-graphql-proxy/src/shopify-graphql-proxy.ts
```diff
@@ -1,24 +1,39 @@
...
-      await next();
-      return;
+    if (privateAppShop && privateAppPassword) {
+      accessToken = privateAppPassword;
+      shop = privateAppShop;
+    } else {
+      const {session = {}} = ctx;
+      const {accessToken: publicAppAccessToken, shop: publicAppShop} = session;
+
+      if (ctx.path !== PROXY_BASE_PATH || ctx.method !== 'POST') {
```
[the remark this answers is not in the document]
michelleyschen: I think I completely miss-understood what this check is for. I will move it outside the else.

packages/koa-shopify-graphql-proxy/src/shopify-graphql-proxy.ts
```diff
@@ -1,31 +1,36 @@
...
     next: () => Promise<any>,
   ) {
     const {session = {}} = ctx;
-    const {accessToken, shop} = session;
 
-    if (ctx.path !== PROXY_BASE_PATH || ctx.method !== 'POST') {
-      await next();
-      return;
-    }
+    const shop = proxyOptions ? proxyOptions.shop : session.shop;
```
TheMallen: 👏 much cleaner

packages/koa-shopify-graphql-proxy/src/test/proxy.test.ts
```diff
@@ -132,6 +128,35 @@ describe('koa-shopify-graphql-proxy', () => {
     });
   });
 
+  it('configures a custom koa-better-http-proxy with private app credentials from the options', async () => {
+    const password = 'sdfghsdghsh';
+    const shop = 'i-sell-things.myshopify.com';
```
TheMallen: good store name :P

PR #224, opened 2018-08-07
Edit getTranslationKey

koshini:
Part of https://github.com/Shopify/internationalization-issues/issues/128
- Change the `getTranslationKey` function to return the appropriate key for address2
- Update README


packages/address/README.md
```diff
@@ -74,6 +74,8 @@ await getTranslationKey('US', 'province'); // => "state"
 await getTranslationKey('JA', 'province'); // => "prefecture"
 await getTranslationKey('CA', 'zip'); // => "postalCode"
 await getTranslationKey('US', 'zip'); // => "zipCode"
+await getTranslationKey('CA', 'address2'); // => "aptUnitNumber"
+await getTranslationKey('CA', 'address2'); // => "aptSuiteEtc"
```
nimzco: You forgot to change the country code. One should be `CA`, the other one could be `FR` for instance.

koshini: 🔥

koshini: Edited

packages/address/README.md
```diff
@@ -74,6 +74,8 @@ await getTranslationKey('US', 'province'); // => "state"
 await getTranslationKey('JA', 'province'); // => "prefecture"
 await getTranslationKey('CA', 'zip'); // => "postalCode"
 await getTranslationKey('US', 'zip'); // => "zipCode"
+await getTranslationKey('CA', 'address2'); // => "aptUnitNumber"
+await getTranslationKey('JA', 'address2'); // => "aptSuiteEtc"
```
nimzco: Can you actually add all of the possible values that we already list in [country-service](https://github.com/Shopify/country-service#translation-keys)?

koshini: Currently, we only have two values in country-service

nimzco: I know but we don't say explicitly all of the possible values neither for `zip` or `province`.
I would basically copy paste the block of doc we have there.

koshini: Oh I see what you mean now!

PR #229, opened 2018-08-08
[WIP] Add @shopify/async-chunks

ismail-syed:


packages/async-chunks/src/server/async-chunks.ts
```diff
@@ -0,0 +1,99 @@
...
+  scripts: Set<Asset>;
+  styles: Set<Asset>;
+}
+
+const root = resolve(__dirname, '../../../../');
+export const defaultManifest = resolve(
+  root,
+  'build',
+  'client',
+  'react-loadable.json',
```
ismail-syed: Use a better name for the default manifest. perhaps `async-chunks.json`?

ismail-syed: Updated manifest name to `async-chunks.json`

packages/async-chunks/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+- Initializers to preload chunks on the server & client
+
+## Installation
+
+```bash
+$ yarn add @shopify/async-chunks
+```
+
+## Usage
```
ismail-syed: Add step about webpack and babel plugins. Should we abstract this away and assume the consumers are using sewing-kit with the experiments flag enabled.

packages/async-chunks/src/lib/components/AsyncChunk.ts
```diff
@@ -0,0 +1,210 @@
...
+  return new Promise(resolve => {
+    // Will always resolve, errors should be handled within loading UIs.
+    // eslint-disable-next-line promise/catch-or-return
+    flushInitializers(READY_INITIALIZERS).then(resolve, resolve);
+  });
+}
+
+export default function asyncChunk(opts) {
+  return asyncChunkGenerator(chunkLoader, opts);
+}
```
lemonmade: OK, so I imagine this mostly just came from loadable, but IMO it does what it does in a bad way. Here is the API I think would be more appropriate:

```ts
// This will be used to collect the async chunks, and to extract them at the end for the
// HTML render.
class AsyncComponentManager {}

// This just provides the async manager to the rest of the tree
export class AsyncComponentProvider extends React.Component<{manager: AsyncComponentManager}> {}

// This component manages the actual async loading. It is roughly equivalent to the
// LoadableComponent here, but can operate independently from the factory function,
// so that someone can easily do <AsyncComponent loader={() => import('./Foo')} />
// anywhere in the app, without needing to construct a special component for it. It would use
// AsyncComponentManager for managing the chunks, and ideally would also use something
// like [this](https://github.com/lemonmade/perfect-app/blob/master/packages/%40shopify/react-extract/server.ts) to use a method for SSR extraction that we could
// reuse in various other SSR things
export class AsyncComponent<AllTheLoaderErrorRenderEtcProps> {}

// A trivial wrapper function that just creates a new component that async loads the
// options passed in to it
export function createAsyncComponent(options: AllTheLoaderErrorRenderEtcProps) {
  return () => <AsyncComponent {...options} />;
}
```

packages/async-chunks/src/babel.ts
```diff
@@ -0,0 +1 @@
+export * from './lib/babel';
```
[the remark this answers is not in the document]
ismail-syed: It's to expose the webpack and babel entry points so consumers can use `@shopify/async-chunks/babel`, `@shopify/async-chunks/webpack`

PR #231, opened 2018-08-09
[react-form-state] 🎨 rename FieldDescriptors to Fields

### LINKED ISSUE #228: [react-form-state] simplify exported type names
`FieldDescriptor` -> `Field`

TheMallen:
closes #228 
This PR changes the name of our exported type for the field data passed into the render prop from `FieldDescriptors` to `Fields`.

This is a breaking (for types only) change.


TheMallen:
@darrenhebner yeah it will break types in web when we bump if they reference `FieldDescriptors`. CI would stop that change until we fix the types though, should be an easy fix.

PR #233, opened 2018-08-09
[react-form-state] 🐛 fix unmounting FormState during submission throwing errors

### LINKED ISSUE #215: [react-form-state] Unmounting FormState on submit handler throws a warning
### Problem
Triggering an action that unmounts `<FormState>` in passed the submit handler triggers the following React warning: 

`Warning: Can't call setState (or forceUpdate) on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.`

This is because the [submit handler is executed](https://github.com/Shopify/quilt/blob/master/packages/react-form-state/src/FormState.ts#L169) before calling `setState` to set `submitting: false`, so if the handler unmounts `<FormState>` there's no point in continuing.

### Use case
In my case, I want to redirect the user to a different page after a successful submit:

```tsx
@autobind
  private async handleSubmit({fields}: FormData<Fields>) {
    const {mutate, history} = this.props;

    const {
      data: {updateTeam: payload},
    } = await mutate({
      variables: {
        id: this.props.team.id,
        team: {
          description: fields.description.value,
          missionStatement: fields.missionStatement.value,
        },
      },
    });

    if (payload) {
      if (payload.userErrors.length) {
        // do something with errors
        return;
      }
      
      // successful save
      if (payload.team) {
        history.push('/details');
      }
    }
  }
```

This will get executed before FormState finishes the submit process.

This is also true for anything that triggers a re-render which unmounts `<FormState>`. For example, I also tried setting the state to something like `formSubmitted` and then rendering a `<Redirect>` component.

> ☝️ The above would have worked (or so I thought) in conjunction with FormState's exposed `submitting` prop, but I had no luck.

CodeSandbox: https://codesandbox.io/s/jlljj34ylv

### Proposed solution
What did work for me, and thought it might be a good practice, is to set a guard check in FormState's  `submit()`, after the user's submit handler gets executed, checking if it is still `mounted`. 

Something like this [around here](https://github.com/Shopify/quilt/blob/master/packages/react-form-state/src/FormState.ts#L169):

```tsx
private async submit(event?: Event) {
    const {onSubmit} = this.props;
    const {formData, mounted} = this;

    if (!mounted) {
      return;
    }

    if (event && !event.defaultPrevented) {
      event.preventDefault();
    }

    if (onSubmit == null) {
      return;
    }

    const result = await onSubmit(formData);
    
    if (!mounted) {
      return;
    }

    if (result) {
      this.updateRemoteErrors(result);
    }

    this.setState({submitting: false});
  }
```

_Surely there's a more elegant way to implement what I'm proposing._ 😅 

I'd love to hear your thoughts on this! And alternative solutions to this problem, of course. I'm happy to send a PR if you're okay with it. 🙏 Hopefully the problem makes sense, I struggled a while until I found what was going on.

TheMallen:
closes #215 

This PR fixes cases where unmounting during a submission handler would cause the component to throw.

### DIFF
--- a/packages/react-form-state/src/FormState.ts
+++ b/packages/react-form-state/src/FormState.ts
@@ -166,9 +166,14 @@ export default class FormState<
 
     this.setState({submitting: true});
 
-    const result = await onSubmit(formData);
-    if (result) {
-      this.updateRemoteErrors(result);
+    const errors = await onSubmit(formData);
+
+    if (!mounted) {
+      return;
+    }
+
+    if (errors) {
+      this.updateRemoteErrors(errors);
     }
 
     this.setState({submitting: false});
--- a/packages/react-form-state/src/tests/FormState.test.tsx
+++ b/packages/react-form-state/src/tests/FormState.test.tsx
@@ -600,7 +600,7 @@ describe('<FormState />', () => {
         return Promise.resolve(submitErrors);
       }
 
-      const form = mount(
+      mount(
         <FormState initialValues={{product}} onSubmit={onSubmit}>
           {renderPropSpy}
         </FormState>,
@@ -628,7 +628,7 @@ describe('<FormState />', () => {
         return Promise.resolve(submitErrors);
       }
 
-      const form = mount(
+      mount(
         <FormState
           initialValues={{product: faker.commerce.productName()}}
           onSubmit={onSubmit}
@@ -659,7 +659,7 @@ describe('<FormState />', () => {
         return Promise.resolve(submitErrors);
       }
 
-      const form = mount(
+      mount(
         <FormState
           initialValues={{product: faker.commerce.productName()}}
           onSubmit={onSubmit}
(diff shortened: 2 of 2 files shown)


packages/react-form-state/src/tests/FormState.test.tsx
```diff
@@ -600,7 +600,7 @@ describe('<FormState />', () => {
         return Promise.resolve(submitErrors);
       }
 
-      const form = mount(
+      mount(
```
TheMallen: these consts were unused

packages/react-form-state/src/FormState.ts
```diff
@@ -166,9 +166,14 @@ export default class FormState<
 
     this.setState({submitting: true});
 
-    const result = await onSubmit(formData);
-    if (result) {
-      this.updateRemoteErrors(result);
+    const errors = await onSubmit(formData);
```
TheMallen: errors is more accurate to what this result is

PR #234, opened 2018-08-09
[react-form-state] never validate empty fields unless using the required validator

### LINKED ISSUE #211: [react-form-state] Only validate non-empty inputs

### LINKED ISSUE #218: [react-form-state] required validators should work differently from non required ones
We should bail early in all validators that are not `required`

### LINKED ISSUE #297: Add a CHANGELOG.md for react-form-state before next release
We should add a CHANGELOG.md to FormState before it's next release and add the unreleased functionality to it.

TheMallen:
closes #218 
closes #211 
closes #297 

This PR makes it so that any validator created using `validate` will automatically pass if the given input is empty. It also rewrites `required` and `requiredString` to skip using the helper, allowing them to continue to work as expected.

**Edit** with the validators handling bailing on empty inputs we are free to remove the `if  (!dirty)` check form our `validateFieldValue` method so that cases like @darrenhebner pointed out work as expected.

### DIFF
--- a/packages/react-form-state/src/validators.ts
+++ b/packages/react-form-state/src/validators.ts
@@ -110,7 +110,10 @@ export function validate<Input, Fields = never>(
   return (input: Input, fields: Fields) => {
     const matches = matcher(input, fields);
 
-    // always mark empty fields valid if they are strings
+    /*
+      always mark empty fields valid to match Polaris guidelines
+      https://polaris.shopify.com/patterns/error-messages#section-form-validation
+    */
     if (isEmpty(input)) {
       return;
     }
--- a/packages/react-form-state/src/validators.ts
+++ b/packages/react-form-state/src/validators.ts
@@ -110,6 +110,11 @@ export function validate<Input, Fields = never>(
   return (input: Input, fields: Fields) => {
     const matches = matcher(input, fields);
 
+    // always mark empty fields valid if they are strings
+    if (isEmpty(input)) {
+      return;
+    }
+
     if (matches) {
       return;
     }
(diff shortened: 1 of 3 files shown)


packages/react-form-state/src/validators.ts
```diff
@@ -137,16 +145,40 @@ const validators = {
...
-    return validate(not(isEmptyString), errorContent);
-  },
-
   nonNumericString(errorContent: ErrorContent) {
     return validate(not(isNumericString), errorContent);
   },
 
+  requiredString(errorContent: ErrorContent) {
+    return (input: string) => {
+      if (not(isEmptyString)(input)) {
```
loic-d: Really nice.

packages/react-form-state/src/validators.ts
```diff
@@ -137,16 +145,40 @@ const validators = {
...
+      }
+
+      // eslint-disable-next-line consistent-return
+      return errorContent;
+    };
+  },
+
   required(errorContent: ErrorContent) {
-    return validate(not(isEmpty), errorContent);
+    return (input: any) => {
```
loic-d: Wondering if `any` should be `string` here. An input field will always be a string, no?

TheMallen: possibly makes sense, I was defaulting to `input` `any` to let validators be used even on object / array fields, but maybe that doesn't really make sense

darrenhebner:
Just to double check, this change will fix this problem with `requiredString`? The empty field is caught on submit, but if the user blurs from the field when it is empty I'd expect to see the error.

required

TheMallen:
@darrenhebner it actually won't because I didn't remove the dirty check in https://github.com/Shopify/quilt/blob/master/packages/react-form-state/src/FormState.ts#L315

....but! It should. I'll update the PR now.

TheMallen:
@darrenhebner  @debchn  I rolled in the change to validation on non-dirty fields we talked about to this PR (since what it adds doesn't do much without it anyway) if you could tophat this I would be grateful :)

PR #238, opened 2018-08-14
✨Move `@shopify/react-i18n` into Quilt from another repo

andrewlo:
Move `@shopify/react-i18n` library from here: https://github.com/lemonmade/react-i18n to Quilt.

Tested by importing this & using it in a consuming Shopify app.

Some changes were made to the source repo to get it to work with Quilt's build, packages & test set up, such as:
- removing use of `@shopify/react-utilities` since it didn't work with quilt's current packages
- updating `hoist-non-react-statics` so it could be imported
- adjusting tests to use quilt's jest set up & packages
- re-writing some code due to a different TS compile set up

However, most of the code is from the original repo.


config/typescript/global.d.ts
```diff
@@ -0,0 +1,7 @@
+declare namespace Intl {
```
tzvipm: why isn't this added to the `i18n` package directly?

config/typescript/modules.d.ts
```diff
@@ -5,3 +5,27 @@ declare module 'url' {
 
   export function parse(url: string): URLDetails;
 }
+
+declare module 'enzyme-adapter-react-16' {
```
tzvipm: Why is this needed?

package.json
```diff
@@ -54,6 +54,9 @@
     "ts-jest": "^23.0.0",
     "typescript": "~2.9.2"
   },
+  "dependencies": {
```
tzvipm: why was this added to the global package.json?

packages/react-i18n/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Installation
+
+```bash
+$ yarn add @shopify/react-i18n
+```
+
+## Usage
+
+Set up the top-level provider & locale manager:
```
tzvipm: you need to introduce the parts of this library before referring to them. Assume people don't know what this p[ackage is.

packages/react-compose/package.json
```diff
@@ -23,7 +23,7 @@
   },
   "homepage": "https://github.com/Shopify/quilt/blob/master/packages/react-compose-enhancers/README.md",
   "dependencies": {
-    "hoist-non-react-statics": "^2.5.0"
+    "hoist-non-react-statics": "^3.0.1"
```
tzvipm: why was this dependency updated? Can that be done in a separate PR?

packages/react-i18n/README.md
```diff
@@ -0,0 +1,80 @@
+# `@shopify/react-i18n`
+
+[![Build Status](https://travis-ci.org/Shopify/quilt.svg?branch=master)](https://travis-ci.org/Shopify/quilt)
+[![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](LICENSE.md) [![npm version](https://badge.fury.io/js/%40shopify%2Freact-i18n.svg)](https://badge.fury.io/js/%40shopify%2Freact-i18n.svg) [![npm bundle size (minified + gzip)](https://img.shields.io/bundlephobia/minzip/@shopify/react-i18n.svg)](https://img.shields.io/bundlephobia/minzip/@shopify/react-i18n.svg)
+
+I18n library for React
```
tzvipm: why should we use this package over https://github.com/i18next/react-i18next? What additional benefits does this provide? I'd use that as a basis for the description here.

packages/react-i18n/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "@shopify/react-i18n",
+  "version": "0.0.0",
+  "license": "MIT",
+  "description": "I18n library for React",
```
tzvipm: this would need to be updated too as part of any description changes.

packages/react-i18n/src/e2e/server.test.tsx
```diff
@@ -0,0 +1,117 @@
+import '../test/matchers';
```
tzvipm: I haven't seen this `e2e` convention used in our other packages. We usually write tests for the external API of our packages anyhow, so I'd just put this in the regular `tests` folder.

config/typescript/global.d.ts
tzvipm: Or is this part of the w3c spec?

packages/react-i18n/README.md
andrewlo: @lemonmade @TheMallen may be able to provide more context

config/typescript/global.d.ts
andrewlo: It seems to be referring to this https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl

packages/react-compose/package.json
TheMallen: It was updated because of some typing issues and the problems that come from having mismatched types in different sub-packages. IMO this is fine. Especially given how simple `hoist` is.

packages/react-i18n/README.md
TheMallen: Picking a specific third party library to compare to in particular is a little weird, there are 100s of libraries for everything in the react ecosystem. I think it can be useful to do comparison / contrast to popular libraries.

We're going to want to revisit the readme when Chris returns regardless since he has the most context on it.

cartogram: There was some discussion at the beginning of the year around custom vs third-party libraries: https://github.com/Shopify/web-foundation/pull/3#issuecomment-364538838, but agree Chris is going to have a lot of input on this Readme either way.

.vscode/settings.json
```diff
@@ -10,9 +10,8 @@
     "**/.git": true,
     "**/.svn": true,
     "**/.hg": true,
-    "**/.DS_Store": true,
     "**/node_modules": true,
-    "build": true,
```
tzvipm: why do we need to include the `build` output in vs code?

config/typescript/modules.d.ts
tzvipm: what's wrong with the community one? https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/enzyme-adapter-react-16/index.d.ts

.vscode/settings.json
lemonmade: My bad, leftover from debugging stuff.

lemonmade:
@tzvipm @TheMallen can I get another look at this?

packages/react-i18n/src/index.ts
```diff
@@ -0,0 +1,7 @@
+export {default as Provider, Props as ProviderProps} from './Provider';
+export {default as Manager} from './manager';
```
tzvipm: why is `Provider.tsx` capitalized, but `manager.ts` is lowercase? Is it just because one is react and one isn't? Do we document that convention anywhere?

packages/react-i18n/src/utilities.tsx
```diff
@@ -0,0 +1,209 @@
...
+}
+
+function normalizeIdentifier(id: string, scope?: string | string[]) {
+  if (scope == null) {
+    return id;
+  }
+
+  return `${
+    typeof scope === 'string' ? scope : scope.join(SEPARATOR)
+  }${SEPARATOR}${id}`;
```
tzvipm: ❤️ I'm so glad this isn't relying on lodash :D

tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "compilerOptions": {
     "moduleResolution": "node",
     "target": "es5",
+    "downlevelIteration": true,
```
tzvipm: why do we need this? Do we need to mention the following point from the TS docs?

> Please note that this requires a native Symbol.iterator or Symbol.iterator shim at runtime for any non-array values

Should we include such a shim as part of this package?

config/typescript/modules.d.ts
```diff
@@ -5,3 +5,22 @@ declare module 'url' {
 
   export function parse(url: string): URLDetails;
 }
+
+declare module 'react-tree-walker' {
```
TheMallen: Can we make a PR to add this to DefinitelyTyped, or this library itself (later)?

packages/react-i18n/README.md
```diff
@@ -0,0 +1,226 @@
...
+- The `i18n` object supplied to a given component can reference translations at any level of depth using a keypath (for example, `NotFound.heading` in the code above). It can also reference translations in parent components; use this to include common translations around a component that contains most of your application.
+- When `translate` is called, it looks up translations in the following order: explicit translations provided by the component’s `translations` function, then translations from the `fallback` for the component, then the same process in every parent component, from bottom to top, that are also connected with `i18n`.
+- In the case of asynchronous translations, your component will only be able to look up translations from its (and ancestors’) `fallback` translation dictionaries until the translations have loaded.
+
+##### Replacements
+
+Replacements can be provided as key-value pairs following the translation key. Your translations should reference the relevant key names, surrounded by a single set of curly braces:
+
+```ts
+// Assuming a dictionary like:
```
TheMallen: it's weird that this code sample is entirely comments, why not just comment the first line?

packages/react-i18n/src/Provider.tsx
```diff
@@ -0,0 +1,28 @@
...
+}
+
+export interface Props {
+  manager: Manager;
+}
+
+export default class Provider extends React.PureComponent<Props, never> {
+  static childContextTypes = contextTypes;
+
+  getChildContext(): Context {
```
TheMallen: We should open an issue to move this to the modern syntax for providers

packages/react-i18n/src/connection.ts
```diff
@@ -0,0 +1,34 @@
+import {TranslationDictionary} from './types';
+import {noop} from './utilities';
+
+export interface TranslationGetter {
+  (locale: string):
+    | TranslationDictionary
+    | Promise<TranslationDictionary | undefined>
```
TheMallen: Tiny nit: can we reorder this so that the promise is last, so that it's a bit more obvious that this is `MaybePromise<TranslationDictionary | undefined>`

packages/react-i18n/README.md
```diff
@@ -0,0 +1,226 @@
+# `@shopify/react-i18n`
+
+[![Build Status](https://travis-ci.org/Shopify/quilt.svg?branch=master)](https://travis-ci.org/Shopify/quilt)
+[![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](LICENSE.md) [![npm version](https://badge.fury.io/js/%40shopify%2Freact-i18n.svg)](https://badge.fury.io/js/%40shopify%2Freact-i18n.svg) [![npm bundle size (minified + gzip)](https://img.shields.io/bundlephobia/minzip/@shopify/react-i18n.svg)](https://img.shields.io/bundlephobia/minzip/@shopify/react-i18n.svg)
+
+i18n utilities for React handling translations, formatting, and more.
```
TheMallen: We might want something in the docs (FAQ?) that amounts to "Why another translation library?" since we get that question a lot it would be nice to be able to link to an answer.

packages/react-i18n/src/index.ts
lemonmade: Yes that is why they are cased the way they are. Opened an issue to document file naming conventions: https://github.com/Shopify/web-foundation/issues/59

tsconfig.json
lemonmade: This is because the target is ES5 but we are using iterators in one or two spots (via for of loops). I don't think we need to include anything here as our supported browsers all have `Symbol`.

config/typescript/modules.d.ts
lemonmade: https://github.com/Shopify/quilt/issues/250

PR #240, opened 2018-08-22
🐛 [react-shopify-app-route-propagator] add polyfill for Edge

macdonaldr93:
Currently in Edge we get this bug:

Bugsnag link: https://app.bugsnag.com/shopify/facebook-ads-client/errors/5b7dc350bb7eb4001a7d33e7?filters[event.since][0]=30d&filters[error.status][0]=open
```
Unable to get property 'delete' of undefined or null reference
```

reason being that Edge doesn't have the property `searchParams` on an instance `URL` by default.

`url-polyfill` is a 5.3K polyfill (1.8K) gzipped. So I think it's worth adding this in this library package instead of projects adding it.


tzvipm:
I don't like this being added in our route propagation library. I'd rather just document that edge needs to have a polyfill. @TheMallen what are your thoughts? Is there a way we can do this whole thing without using `URL.prototype.searchParams`?

macdonaldr93:
@tzvipm at first I was thinking the same but I also don't like the idea that right now none of the apps using this package work in Edge and I think it's a pretty low impact polyfill to include.

It's kinda weird that I would include a package and then need to look at the README to catch that if I don't add this polyfill to my application, the package will cause my app to stop running in Edge. It's not easy to debug or grok that info.

macdonaldr93:
We could use a regex to remove the HMAC instead of `normalizedLocation.searchParams.delete('hmac');` and then all the other attributes are supported by edge... Just throwing that out there as another idea.

macdonaldr93:
I can't seem to get the polyfill to work in our app... it needs to be in the vendor folder but then it doesn't work because I can import it... I think this needs to be here.

kaelig:
👋 There are many different polyfilling strategies, each with a different impact on performance. For this reason, I'm leaning towards @tzvipm’s solution, and delegate to consumers of this library to decide how they want to polyfill instead polyfilling out of the box (1.8 KB seems negligible, but those things add up).

And if we're going for @macdonaldr93’s polyfill solution, would there be a middle ground possible where a consumer could opt out of the polyfill (so they can apply a different polyfill strategy)?

macdonaldr93:
Yeah that makes sense to me. I think I'm having trouble polyfilling this due to the way vendors works in sewing-kit too. What I ended up doing was taking the polyfill and this package out of the vendors bundle, then the polyfill worked fine. It feels like a lot of gotchas though.

TheMallen:
I actually like just doing this without a polyfill but manually (ie. with a regex). It isn't much more code and we don't need to do any extra documentation.

Issue #243, opened 2018-08-27
Improve Type-checking experience (projects)


tzvipm:
Some contributors reported they had troubles getting up and running due to TS conflicts or dependency conflicts, etc.

We think that using TS projects can help with this cc @TheMallen

PR #257, opened 2018-08-29
⬆️ 🏗️ [RFC] Use TS 3 project references

### LINKED ISSUE #248: Upgrade to TS 3.x

TheMallen:
closes #248 
helps with #243  (hopefully, maybe)

## Summary
This PR ups us to TS 3.0.1, sets up [project references](https://www.typescriptlang.org/docs/handbook/project-references.html) and fixes up our config a bit.

### TS 3.0
The changes I had to make to get TS 3.0 working were very minimal, just updating it in all the `package.json`s and the `yarn.lock`, and changing one spot that used an uninitialized class property.

#### pros
-  `yarn build` no longer relies on lerna, just `tsc` itself. 
- `yarn build` is now guaranteed to build things in the right order (previously it could in theory have failed with one package depending on another that is not built)
- `yarn link`ing a single package while working on multiple will now work better, even if you are making changes to packages it depends on, since `tsc -b` will also rebuild those.

#### cons
- `ts-jest` [does not support composite projects](https://github.com/kulshekhar/ts-jest/issues/669) yet, so we still have the same clunkiness for running tests that depend on a package that needs to be built
- the output from `tsc -b packages` is slightly less useful than `lerna run yarn build` was, since we no longer get output sorted by package
- we now have to add an entry to `references` for any in-repo (non-dev) dependency of a package or `yarn build` will fail
- we have one more config file we need to regenerate whenever we add  a new package to the repo (we can probably just get `yarn generate package` to update these though)

### basics
- `g checkout TS-3`
- `yarn run build`
- `yarn test`

### linking
- `g checkout TS-3`
- `cd packages/react-html`
- `yarn link`
- (in another tab) `dev cd web && yarn link @shopify/react-html`
- (in your quilt tab) add `console.log('foo bar baz')` to `packages/react-serialize/src/index.ts`
- (in your quilt tab) run `yarn run build` inside `packages/react-html`
- (in web tab) run `dev run dev`, **you should see your console log when compilation finishes**

## Concerns
I'm not sure how much value this actually adds without working for tests. We may want to hold off on it until it does since it does add some extra complexity to our config. On the other hand, it does bring us a bit closer to not actually needing `lerna`. It's possible that at some point we will be able to just use TS and yarn workspaces to handle the monorepo on their own.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "react-helmet": "^5.2.0",
     "rimraf": "^2.6.2",
     "ts-jest": "^23.0.0",
-    "typescript": "~2.9.2"
+    "typescript": "~3.0.1"
   },
   "dependencies": {},
   "jest": {
--- a/package.json
+++ b/package.json
@@ -8,10 +8,10 @@
   },
   "scripts": {
     "publish": "lerna publish",
-    "prebuild": "yarn run clean",
-    "build": "lerna run build",
+    "build": "tsc -b packages",
     "lint": "yarn eslint '**/*.{ts,tsx}'",
     "ci:lint-docs": "yarn generate docs && test -z \"$(git status --porcelain)\" || echo 'The root README has not been updated. Run `yarn generate docs` in the root of your quilt directory and try again.'",
+    "ci:lint-tsconfig": "yarn generate tsconfig && test -z \"$(git status --porcelain)\" || echo 'packages/tsconfig.json has not been updated. Run `yarn generate tsconfig` in the root of your quilt directory and try again.'",
     "test": "jest --maxWorkers=3",
     "check": "lerna run check",
     "release": "lerna publish && git push --follow-tags",
@@ -75,6 +75,11 @@
       "url": "http://localhost:3000/"
     },
     "coverageDirectory": "./coverage/",
-    "collectCoverage": true
+    "collectCoverage": true,
+    "globals": {
+      "ts-jest": {
+        "tsConfigFile": "tsconfig.test.json"
+      }
+    }
   }
 }
--- a/packages/address/tsconfig.json
+++ b/packages/address/tsconfig.json
@@ -1,12 +1,9 @@
 {
-  "extends": "../../tsconfig.json",
-  "compileOnSave": false,
+  "extends": "../../tsconfig.base.json",
   "compilerOptions": {
     "outDir": "./dist",
     "baseUrl": "./src",
     "rootDir": "./src"
   },
-  "include": [
-    "src/**/*.ts"
-  ]
+  "include": ["src/**/*.ts", "src/**/*.tsx"]
 }
--- a/packages/jest-dom-mocks/src/match-media.ts
+++ b/packages/jest-dom-mocks/src/match-media.ts
@@ -6,7 +6,8 @@ export interface MediaMatching {
 
 export default class MatchMedia {
   private isUsingMockMatchMedia = false;
-  originalMatchMedia: (mediaQuery: string) => MediaQueryList;
+  originalMatchMedia: (mediaQuery: string) => MediaQueryList =
+    window.matchMedia;
 
   mock(media: MediaMatching = defaultMatcher) {
     if (this.isUsingMockMatchMedia) {
--- a/packages/react-html/tsconfig.json
+++ b/packages/react-html/tsconfig.json
@@ -1,12 +1,10 @@
 {
-  "extends": "../../tsconfig.json",
-  "compileOnSave": false,
+  "extends": "../../tsconfig.base.json",
   "compilerOptions": {
     "outDir": "./dist",
     "baseUrl": "./src",
     "rootDir": "./src"
   },
-  "include": [
-    "src/**/*.ts"
-  ]
+  "include": ["src/**/*.ts", "src/**/*.tsx"],
+  "references": [{"path": "../react-serialize"}]
 }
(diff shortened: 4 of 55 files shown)


package.json
```diff
@@ -8,10 +8,10 @@
   },
   "scripts": {
     "publish": "lerna publish",
-    "prebuild": "yarn run clean",
-    "build": "lerna run build",
+    "build": "tsc -b packages",
```
TheMallen: `tsc -b` lets us use `references` and also automatically builds incrementally.

packages/address/tsconfig.json
```diff
@@ -1,12 +1,9 @@
 {
-  "extends": "../../tsconfig.json",
-  "compileOnSave": false,
+  "extends": "../../tsconfig.base.json",
```
TheMallen: renaming this to better express it's meaning

packages/jest-dom-mocks/src/match-media.ts
```diff
@@ -6,7 +6,8 @@ export interface MediaMatching {
 
 export default class MatchMedia {
   private isUsingMockMatchMedia = false;
-  originalMatchMedia: (mediaQuery: string) => MediaQueryList;
+  originalMatchMedia: (mediaQuery: string) => MediaQueryList =
```
TheMallen: had to do this to get `strict` working (which I had to do to get TS3 working)

packages/react-html/tsconfig.json
```diff
@@ -1,12 +1,10 @@
...
+  "extends": "../../tsconfig.base.json",
   "compilerOptions": {
     "outDir": "./dist",
     "baseUrl": "./src",
     "rootDir": "./src"
   },
-  "include": [
-    "src/**/*.ts"
-  ]
+  "include": ["src/**/*.ts", "src/**/*.tsx"],
```
TheMallen: not sure how this was ever working without the `.tsx` files included, but `-b` seems to have broken it  ¯\_(ツ)_/¯

packages/react-html/tsconfig.json
```diff
@@ -1,12 +1,10 @@
...
+  "references": [{"path": "../react-serialize"}]
```
TheMallen: Adding these lets `tsc -b` understand how to traverse our in-repo dependencies and build them in the right order, and lets you build from within the package's sub-folder and still rebuild projects it depends on

packages/tsconfig.json
```diff
@@ -0,0 +1,28 @@
+{
+  "extends": "../tsconfig.base.json",
+  "files": [],
+  "references": [
```
TheMallen: this root tsconfig is what lets us do `tsc -b packages` and have everything build correctly according to their own tsconfigs

packages/tsconfig.json
```diff
@@ -0,0 +1,28 @@
+{
+  "extends": "../tsconfig.base.json",
+  "files": [],
```
TheMallen: for some reason not having this field here made `tsc -b packages` not respect inner tsconfigs anymore ¯\_(ツ)_/¯

plopfile.js
```diff
@@ -82,4 +65,37 @@ module.exports = function(plop) {
       ];
     },
   });
+
+  plop.setGenerator('tsconfig', {
```
TheMallen: probably we should change this and the docs command to be "custom actions" and then have the new package generator consume them

tsconfig.test.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": "./tsconfig.base.json",
+  "compilerOptions": {
+    "composite": false
```
TheMallen: :( ts-jest can't handle this being true yet

plopfile.js
```diff
@@ -82,4 +65,37 @@ module.exports = function(plop) {
...
+    description: 'Generate package level tsconfig',
```
tzvipm: Isn't this a root-level ts config. To me, `package level` means for an individual package.

plopfile.js
```diff
@@ -82,4 +65,37 @@ module.exports = function(plop) {
...
+    prompts: [],
+    actions(data) {
+      const packageNames = getPackageNames();
+
+      return [
+        {
```
tzvipm: can we also make this inject the cross-package dependencies? We should be able to simply compile, for each package, a list of `@shopify/<pkg>` dependencies, then filter that by the `packageNames` variable and generate the `references` section from that data.

test/typescript-version.test.ts
```diff
@@ -10,6 +10,10 @@ describe('typescript version', () => {
     const packageNames = readdirSync(packagesPath);
 
     for (const packageName of packageNames) {
+      if (packageName.includes('.')) {
```
tzvipm: why would a `packageName` include a `.`? I'd rather see the following:

```
const packageFolderContents = readdirSync(packagesPath);
const packageNames = packageFolderContents.filter(name => !name.includes('.'));

...
```

plopfile.js
TheMallen: Yeah I meant "packages" level since that's the folder it's in, but maybe root makes sense anyway.

plopfile.js
TheMallen: That does seem like it would make this less cumbersome, I'll see if I can think of any downsides to that approach

TheMallen:
Going forward:
- I'll pull out the TS3 bump into it's own PR
- I'll keep working on this, I think I have some ways to mitigate the cons for this

TheMallen:
closing this for now, will revisit in the future (possibly when ts-jest supports references)

PR #258, opened 2018-08-30
Update service.yml

tzvipm:
**Internal use only** see https://services.shopify.io/services/quilt/production

### DIFF
--- a/service.yml
+++ b/service.yml
@@ -1,5 +1,9 @@
-director: Delaney Manders
+org_line: Shopify Core
 owners:
-  - themallen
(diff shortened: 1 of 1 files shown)


service.yml
```diff
@@ -1,5 +1,10 @@
-director: Delaney Manders
```
ismail-syed: Why are we removing this? Also, it should've been a github account linked with the director

tzvipm: It's just not needed. I think director ownership is deprecated anyhow.

PR #261, opened 2018-08-30
🔥 custom dev command

TheMallen:
This PR fixes the `dev up` flow to not give misleading error messages. We do this simply by removing our custom dev command. It seems that we didn't need to run`lerna bootstrap` anyway since we are using yarn workspaces.

## 🎩 instructions
- `dev clone quilt`
- `dev up`
- `dev build && dev test`

### DIFF
--- a/dev.yml
+++ b/dev.yml
@@ -4,11 +4,6 @@ up:
   - node:
       version: v8.10.0
       yarn: true
-  - custom:
-      name: lerna bootstrap
-      met?: 'false'
-      meet: yarnpkg run lerna bootstrap
-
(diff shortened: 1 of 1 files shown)


tzvipm:
Is this more about the confusing errors, or more about running lerna bootstrap? If the former, the EASDK team had a nice solution.

TheMallen:
@tzvipm we don't actually need `bootstrap` since we use yarn workspaces. Make a fresh clone, checkout this branch and try `dev up`, `yarn build`, `yarn test`, etc. to confirm

PR #268, opened 2018-09-04
ITP2 fixes

ragalie:
This PR contains the necessary changes to work around Safari 12's ITP 2.0 implementation. In short, it implements this flow:

image

To tophat:

- `yarn link` in this package and then perform `yarn link "@shopify/koa-shopify-auth"` in `unite-react-node-app-workshop`. Use the following `server/index.js`:

```
import 'isomorphic-fetch';

import dotenv from 'dotenv';
import Koa from 'koa';
import session from 'koa-session';
import shopifyAuth, {verifyRequest} from '@shopify/koa-shopify-auth';

dotenv.config();

const {SHOPIFY_API_KEY, SHOPIFY_SECRET} = process.env;

const app = new Koa();
app.keys = [SHOPIFY_SECRET];

app
// sets up secure session data on each request
  .use(session(app))

  // sets up shopify auth
  .use(
    shopifyAuth({
      apiKey: SHOPIFY_API_KEY,
      secret: SHOPIFY_SECRET,
      baseUrl: 'your url',
      scopes: ['write_orders, write_products'],
      prefix: '/shopify',
      afterAuth(ctx) {
        const {shop, accessToken} = ctx.session;

        console.log('We did it!', accessToken);

        ctx.redirect('/');
      },
    }),
  )

  // everything after this point will require authentication
  .use(verifyRequest({authRoute: '/shopify/auth', fallbackRoute: '/shopify/auth'}))

  // application code
  .use((ctx) => {
    ctx.body = `
    <script src="https://cdn.shopify.com/s/assets/external/app.js"></script>

    <script type="text/javascript">
      ShopifyApp.init({
        apiKey: "${SHOPIFY_API_KEY}",
        shopOrigin: "https://${ctx.session.shop}",
        debug: true,
        forceRedirect: true
      });
    </script>
    
    🎉
    `;
  });

export default app;
```


packages/koa-shopify-auth/src/auth/create-enable-cookies.ts
```diff
@@ -0,0 +1,425 @@
...
+
+    const {query} = ctx;
+    const {shop} = query;
+
+    if (shop == null) {
+      ctx.throw(400, Error.ShopParamMissing);
+      return;
+    }
+
+    ctx.body = `
```
ragalie: Is there a better way/place to do/put this HTML?

packages/koa-shopify-auth/src/auth/create-enable-cookies.ts
```diff
@@ -0,0 +1,425 @@
+import {Context} from 'koa';
+
+import Error from './errors';
+import {OAuthStartOptions} from "../types";
+
+export default function createEnableCookies({apiKey}: OAuthStartOptions) {
+  return function enableCookies(ctx: Context) {
+    const HEADING = 'Enable cookies';
```
ragalie: In the [shopify_app PR](https://github.com/Shopify/shopify_app/pull/617) we use the `i18n` gem to provide translations for these. Do we have something similar here? If we don't already have something in this project I'm planning on leaving that for later in the interest of time.

packages/koa-shopify-auth/src/auth/test/index.test.ts
```diff
@@ -0,0 +1,158 @@
+import {createMockContext} from '@shopify/jest-koa-mocks';
+
+import createShopifyAuth from '../index';
+
+import createTopLevelRedirect from '../create-top-level-redirect';
+jest.mock('../create-top-level-redirect');
```
ragalie: This mock doesn't seem to work and I can't figure out why. It's the cause of the test failures. Any ideas?

packages/koa-shopify-auth/src/auth/index.ts
```diff
@@ -3,36 +3,77 @@ import {Context} from 'koa';
...
   return async function shopifyAuth(ctx: Context, next: NextFunction) {
-    if (ctx.path === oAuthStartPath) {
+    if (ctx.path === oAuthStartPath && !hasCookieAccess(ctx)) {
+      // This is to avoid a redirect loop if the app doesn't use verifyRequest or set the test cookie elsewhere.
+      ctx.cookies.set(TEST_COOKIE_NAME, '1');
+      await redirectToEnableCookies(ctx);
+      return;
+    }
+
+    if (ctx.path == inlineOAuthPath || (ctx.path === oAuthStartPath && shouldPerformInlineOAuth(ctx))) {
```
TheMallen: why double equals instead of triple?

packages/koa-shopify-auth/src/auth/index.ts
```diff
@@ -3,36 +3,77 @@ import {Context} from 'koa';
...
+      ctx.cookies.set(TOP_LEVEL_OAUTH_COOKIE_NAME);
       await oAuthStart(ctx);
       return;
     }
 
+    if (ctx.path === oAuthStartPath) {
+      ctx.cookies.set(TOP_LEVEL_OAUTH_COOKIE_NAME, '1');
```
TheMallen: it feels kind of weird to set cookies in this function since otherwise it's only concern is routing.

packages/koa-shopify-auth/src/auth/test/index.test.ts
TheMallen: What does 'doesn't seem to work' mean? Have you tried manually mocking like you do in the other test?

packages/koa-shopify-auth/src/auth/create-enable-cookies.ts
TheMallen: We don't have something like that in this project. We have it for react but we don't render any strings in server side node anywhere else so haven't needed it.

packages/koa-shopify-auth/src/auth/create-enable-cookies.ts
TheMallen: There's no better way to do this than a template literal if we don't want to depend on some kind of templating library (or ejs), and need interpolation. Similar to the routing being a bunch of dumb ifs we opted for minimizing the dependencies in this package since it's relatively simple.

If it was just a static page we could have koa just serve the file directly or something.

packages/koa-shopify-auth/src/auth/index.ts
```diff
@@ -3,36 +3,77 @@ import {Context} from 'koa';
...
+      await redirectToTopLevelOAuth(ctx);
+      return;
+    }
+
     if (ctx.path === oAuthCallbackPath) {
       await oAuthCallback(ctx);
       return;
     }
 
+    if (ctx.path == enableCookiesPath) {
```
TheMallen: At some point we may want to consider bringing in `koa-router` since this is starting to be a lot of routes.

packages/koa-shopify-auth/src/auth/test/index.test.ts
```diff
@@ -0,0 +1,158 @@
...
+
+        await shopifyAuth(ctx, nextFunction);
+
+        expect(ctx.cookies.set).toBeCalledWith('shopifyTopLevelOAuth');
+      });
+    });
+  });
+
+  describe('with the /auth/inline path', () => {
+    it('performs inline oauth', async () => {
```
TheMallen: I'm curious how you found working with `createMockContext`, since it hasn't gotten much use outside of quilt yet.

I think we probably also want some E2E tests on this at some point, but this changeset is already quite large so don't worry yet.

packages/koa-shopify-auth/src/auth/test/oauth-query-string.test.ts
```diff
@@ -0,0 +1,72 @@
...
+const queryData = {
+  state: fakeNonce,
+  scope: 'write_orders, write_products',
+  // eslint-disable-next-line camelcase
+  client_id: baseConfig.apiKey,
+  // eslint-disable-next-line camelcase
+  redirect_uri: `https://${baseUrl}/callback`,
+};
+
+describe('oAuthQueryString', () => {
```
TheMallen: I like this having it's own test file :)

packages/koa-shopify-auth/src/auth/create-enable-cookies.ts
TheMallen: I'm not sure there's a good way to share the templates that would be consumable by both projects and not unnecessarily complex.

packages/koa-shopify-auth/src/auth/test/index.test.ts
ragalie: Within some of the tests I do something like:

```
const mockTopLevelRedirect = jest.fn();
createTopLevelRedirect.mockImplementation(mockTopLevelRedirect);
```

You can see the resultant test failures: when my code attempts to call the function returned by `createTopLevelRedirect` (which should be `jest.fn()` in this case) I get `[function name] is not a function`.

TheMallen: ohhh, try doing `require.requireMock('../create-top-level-redirect')` instead of the import in your test file.

ragalie:
@TheMallen @tylerball I updated based on your suggestions, would love another look! Thanks!

packages/koa-shopify-auth/src/auth/create-enable-cookies.ts
```diff
@@ -0,0 +1,440 @@
+import {Context} from 'koa';
+
+import Error from './errors';
+import {OAuthStartOptions} from '../types';
+
+const HEADING = 'Enable cookies';
```
francinen: Does quilt have a mechanism to support i18n?

ragalie: Seems like no: https://github.com/Shopify/quilt/pull/268#discussion_r215017108

I think it's worth asking the dev tools team to consider this as part of their backlog though!

PR #278, opened 2018-09-07
🐛 [jest-koa-mocks] Set default response status to be 404

BPScott:
The node mocks default to a 200 status, but the koa default is 404. We
should follow koa's default as per https://github.com/koajs/koa/blob/master/docs/api/response.md#responsestatus


BPScott:
This makes testing the liveness ping a bit more obvious as prior to this the response mock returned a 200 status by default. Thus you couldn't really say "adding the liveness-ping middleware now makes the status endpoint return 200"  given that the response was 200 even if that middleware was never called.

Issue #221, opened 2018-08-06
[I18n] Share the implementation of pseudotranslation to i18n


nimzco:
Make the `pseudolocalize` method shipped in this PR: https://github.com/Shopify/web/pull/5394 part of the intl.
See: https://github.com/Shopify/web/pull/5394/files#r204060231

Improve the function for it to be able to spit longer or shorter strings, eg: Smartling Sketch plugin:
![image](https://user-images.githubusercontent.com/445045/43321237-36e5407c-9179-11e8-814b-08ff24d81bce.png)

PR #279, opened 2018-09-09
Add pseudolocalization to react-i18n

lemonmade:
Follow up from https://github.com/Shopify/quilt/pull/273. Fixes https://github.com/Shopify/quilt/issues/221.

This PR uses the new pseudotranslation stuff from the above PR in `react-i18n`. Users of that library can now enable pseudotranslation by setting the `pseudolocalize` option on the `Manager` class to `true`, or to a string representing the locale to localize for.


packages/react-i18n/src/i18n.ts
```diff
@@ -65,17 +78,14 @@ export default class I18n {
...
-  get countryCode(): string | undefined {
-    return this.locale.split('-')[1];
-  }
-
   constructor(
     public translations: TranslationDictionary[],
-    {locale, currency, timezone}: I18nDetails,
+    {locale, currency, timezone, pseudolocalize = false}: I18nDetails,
   ) {
-    this.locale = locale.toLowerCase();
```
TheMallen: we don't need to lowercase it anymore?

packages/react-i18n/src/i18n.ts
```diff
@@ -43,12 +44,24 @@ const RTL_LANGUAGES = [
...
+  }
+
+  get region() {
+    return regionFromLocale(this.locale);
+  }
+
+  /**
+   * @deprecated Use I18n#region instead.
+   */
+  get countryCode() {
```
TheMallen: this feels a bit weird to do in this PR

packages/react-i18n/src/i18n.ts
lemonmade: No, this was basically a bug/ incorrect choice in the previous version, an appropriate locale representation has lowercase language and uppercase region

packages/react-i18n/src/i18n.ts
lemonmade: I definitely see that, but it was a choice between keeping the incorrect language for pseudo localization as well (where we use the region), or updating them both.

Issue #150, opened 2018-06-01
Add an `addEventListener` mock to `jest-dom-mocks`


TheMallen:
We should expose a handy way of mocking out the document level `addEventListener` and `removeEventListener`, as well as add this to our between test checks for having restored them.

lemonmade:
I think generally speaking, there are a whole bunch of document mocks that we could use (`createElement`, spy on particular methods of DOM nodes for testing refs, etc)

donnielrt:
+1 :) would especially love a mock for `scrollWidth`, `offsetWidth`, etc.

PR #281, opened 2018-09-11
[WIP] add/remove eventListener mock

hughxie:
PR adds addEventListener and removeEventListener mock to [jest-dom-mock](https://github.com/Shopify/quilt/tree/master/packages/jest-dom-mocks).

### DIFF
--- a/packages/jest-dom-mocks/src/document.ts
+++ b/packages/jest-dom-mocks/src/document.ts
@@ -1,0 +1,42 @@
+export interface MockWindow extends Window {
+  addEventListener: jest.Mock<{}> & typeof window.addEventListener;
+  removeEventListener: jest.Mock<{}> & typeof window.removeEventListener;
+}
+
+export default class Document {
+  private isUsingMockWindow: boolean = false;
+  private window?: MockWindow;
+
+  mockWindow() {
+    window.addEventListener = jest.fn();
+    window.removeEventListener = jest.fn();
+    return window as MockWindow;
+  }
+
+  mock() {
+    if (this.isUsingMockWindow) {
+      throw new Error('You tried to mock window when it was already mocked.');
+    }
+
+    this.window = this.mockWindow();
+    this.window.addEventListener('load', () => {});
+    this.isUsingMockWindow = true;
+  }
+
+  restore() {
+    if (!this.isUsingMockWindow) {
+      throw new Error(
+        'You tried to restore window when it was already restored',
+      );
+    }
+
+    if (this.window) {
+      this.window.addEventListener.mockClear();
+    }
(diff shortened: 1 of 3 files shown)


hughxie:
@lemonmade is this what #150 is looking for?

lemonmade:
This is in line with what I was thinking, yes. Maybe @tzvipm / @TheMallen / @GoodForOneFare can give some initial feedback on this approach?

packages/jest-dom-mocks/src/document.ts
```diff
@@ -0,0 +1,42 @@
+export interface MockWindow extends Window {
+  addEventListener: jest.Mock<{}> & typeof window.addEventListener;
+  removeEventListener: jest.Mock<{}> & typeof window.removeEventListener;
+}
+
+export default class Document {
```
TheMallen: It's weird to call this `Document` but internally it's doing all it's mocking on `window`

packages/jest-dom-mocks/src/document.ts
```diff
@@ -0,0 +1,42 @@
...
+  private isUsingMockWindow: boolean = false;
+  private window?: MockWindow;
+
+  mockWindow() {
+    window.addEventListener = jest.fn();
```
TheMallen: I'd also note that most of the other stuff in this package is also mocking properties of `window`, so it seems a little disingenuous to represent whether we have mocked the two functions this file cares about as `usingMockWindow` / `mockWindow`

packages/jest-dom-mocks/src/test/document.test.ts
```diff
@@ -0,0 +1,40 @@
+import Document from '../document';
+
+describe('Document', () => {
+  describe('mock', () => {
```
TheMallen: Can we test that when it's `mocked` calling `addEventListener` and `removeEventListener` doesn't do anything / that you can assert on them?

packages/jest-dom-mocks/src/document.ts
```diff
@@ -0,0 +1,42 @@
...
+    return window as MockWindow;
+  }
+
+  mock() {
+    if (this.isUsingMockWindow) {
+      throw new Error('You tried to mock window when it was already mocked.');
+    }
+
+    this.window = this.mockWindow();
+    this.window.addEventListener('load', () => {});
```
TheMallen: Overriding `window.addEventListener` does not mean you are also overriding `document.addEventListener` (since you are just overriding a reference, not the function itself). Considering we use `document.addEventListener` *and* `window.addEventListener` in our codebases, we probably want to override both of these.

**example**
image

Issue #284, opened 2018-09-13
Add @shopify/react-google-analytics


antoinegrant:
## Overview
Extract out `@shopify/react-google-analytics` from Shopify/web

## Type

- [x] New feature
- [ ] Changes to existing features

## Motivation
To avoid code duplication since Shopify/online-store-web needs this package.

## Labels

- [x] Add the `Type: Feature Request` label to this issue.

### Area

- [ ] Add any relevant `Area: <area>` labels to this issue

### Scope

- Is this issue related to a specific package?

  - [ ] Tag it with the `Package: <package_name>` label.

- Is it related to a new package?
  - [ ] Tag it with the `New package` Label

---

## Checklist

- [ ] Please delete the labels section before submitting your issue
- [ ] I have described this issue in a way that is actionable (if possible)

PR #285, opened 2018-09-13
Add react-google-analytics to quilt

antoinegrant:
Resolves https://github.com/Shopify/quilt/issues/284

### DIFF
--- a/packages/react-google-analytics/README.md
+++ b/packages/react-google-analytics/README.md
@@ -1,0 +1,138 @@
+# `@shopify/react-google-analytics`
+
+[![Build Status](https://travis-ci.org/Shopify/quilt.svg?branch=master)](https://travis-ci.org/Shopify/quilt)
+[![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](LICENSE.md) [![npm version](https://badge.fury.io/js/%40shopify%2Freact-google-analytics.svg)](https://badge.fury.io/js/%40shopify%2Freact-google-analytics.svg) [![npm bundle size (minified + gzip)](https://img.shields.io/bundlephobia/minzip/@shopify/react-google-analytics.svg)](https://img.shields.io/bundlephobia/minzip/@shopify/react-google-analytics.svg)
+
+Allows React apps to easily embed tracking pixel iframes
+
+## Installation
+
+```bash
+$ yarn add @shopify/react-google-analytics
+```
+
+## Usage
+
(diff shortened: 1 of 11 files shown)


packages/react-google-analytics/src/GaJS.tsx
```diff
@@ -0,0 +1,80 @@
...
+  Props,
+  never
+> {
+  render() {
+    const {account, debug} = this.props;
+
+    return (
+      <>
+        <script
+          type="text/javascript"
```
kaelig: You can safely delete this line

packages/react-google-analytics/src/Universal.tsx
```diff
@@ -0,0 +1,97 @@
...
+  never
+> {
+  render() {
+    const {debug} = this.props;
+
+    return (
+      <>
+        <script
+          id="google-analytics-universal-script"
+          type="text/javascript"
```
kaelig: This line can also safely be removed.

kaelig:
Did you mean to leave the README empty?

antoinegrant:
🏓 for another round of reviews. 🙏

antoinegrant:
Comments addressed! 👍

packages/react-google-analytics/README.md
```diff
@@ -0,0 +1,68 @@
...
+```jsx
+import {
+  Universal,
+  GaJS,
+} from '@shopify/react-google-analytics';
+
+const UNIVERSAL_GA_ACCOUNT_ID = 'UA-xxxx-xx';
+const GA_JS_ACCOUNT_ID = 'UA-xxxx-xx';
+
+<Universal
```
kaelig: This is unclear - would one ever use both GaJS and Universal at the same time in an app? Or should users only import one that fits their need?

Which one is the recommended way?

antoinegrant: Good point, I guess I was focused on how we use it. We have to load both scripts in order to be on par with the way we do analytics in core.

I'll update the docs to just use or the other.

Thanks

antoinegrant:
@kaelig I've made some changes to the docs.

antoinegrant:
@kaelig I've added a little info https://github.com/Shopify/quilt/pull/285/commits/1ebe4e2e8e7f070cde36677e8b871e21ab64df34

antoinegrant:
Merging this PR. @TheMallen are you able to run the release script?

Issue #289, opened 2018-09-14
Add package.json#sideEffects to all Quilt packages


BlazingBBQ:
## Overview

Add `package.json#sideEffects` to all existing Quilt packages and modify the generator template to include this change in new packages.

## Motivation

This change would enable webpack to tree shake any unused code. 

## Checklist

- [x] Please delete the labels section before submitting your issue
- [x] I have described this issue in a way that is actionable (if possible)

BlazingBBQ:
Added the `sideEffect` to all Quilt packages, but did not add it to the generator template.

GoodForOneFare:
Because these are public-facing libraries, and not everyone has our safeguards against accidental treeshaking.  `sideEffects: false` should be an informed decision.

PR #290, opened 2018-09-14
[WIP] Added side effects config options to tracking pixel and preconnect

BlazingBBQ:
Added `sideeffects` config option to tracking pixel and preconnect packages (See issue #289 ). 
If this is the correct way to add side effects to our packages, add this config to other Quilt packages, then update generator to add this config to new packages.

### DIFF
--- a/templates/package.hbs.json
+++ b/templates/package.hbs.json
@@ -26,6 +26,5 @@
   "devDependencies": {
     "typescript": "~3.0.1"
   },
-  "files": ["dist/*"],
-  "sideEffects": false
+  "files": ["dist/*"]
 }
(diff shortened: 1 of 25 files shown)


templates/package.hbs.json
```diff
@@ -26,5 +26,6 @@
   "devDependencies": {
     "typescript": "~3.0.1"
   },
-  "files": ["dist/*"]
+  "files": ["dist/*"],
+  "sideEffects": false
```
GoodForOneFare: I'm optimistic about this being a useful thing, but it *could* cause unintended treeshaking if a Quilt package does something like, `import './polyfills';`.

cc @TheMallen @tzvipm - default to optimization 👍 or default to safety 👎 ?

Issue #172, opened 2018-06-29
react-shortcuts missing dependency error


tzvipm:
<img width="650" alt="screen shot 2018-06-29 at 1 50 34 pm" src="https://user-images.githubusercontent.com/1950680/42106871-7c1dfe9a-7ba3-11e8-98c4-af5b5881ef1b.png">

PR #291, opened 2018-09-14
Added `react` & `prop-types` as dependency to `react-i18n`

### LINKED ISSUE #288: react-i18n minzipped size badge missing dependency error
Similar issue as #172.

image

hughxie:
This pr adds missing dependencies from [`react-i18n`](https://github.com/Shopify/quilt/tree/master/packages/react-i18n).

- [prop-types](https://yarnpkg.com/en/package/prop-types)
- [react](https://yarnpkg.com/en/package/react)

Should also fix #288. 

Here is the current badge error: 
image

Screenshot of error from [Bundlephobia](https://bundlephobia.com/result?p=@shopify/react-i18n):
image

### DIFF
--- a/packages/react-i18n/package.json
+++ b/packages/react-i18n/package.json
@@ -30,6 +30,8 @@
   "dependencies": {
     "@types/hoist-non-react-statics": "^3.0.1",
     "hoist-non-react-statics": "^3.0.1",
+    "prop-types": "^15.6.2",
+    "react": "^16.2.0",
     "react-tree-walker": "^4.3.0"
   },
   "sideEffects": false


hughxie:
#172 was fixed by same approach.

packages/react-i18n/package.json
```diff
@@ -30,6 +30,8 @@
   "dependencies": {
     "@types/hoist-non-react-statics": "^3.0.1",
     "hoist-non-react-statics": "^3.0.1",
+    "prop-types": "^15.6.2",
```
GoodForOneFare: I thought the new context API didn't even need explicit `childContextTypes`? 🤔

PR #292, opened 2018-09-17
Changed lodash version for tracking pixel

BlazingBBQ:
Changed lodash version for `@shopify/react-tracking-pixel` to match lodash version used in Web.


kaelig:
I'm not sure I understand this change - both semver ranges may resolve to the same version, no?

GoodForOneFare:
[`react-apollo`](https://github.com/Shopify/react-apollo/blob/master/package.json#L122) has a hard dependency on `lodash@4.17.4`.  To avoid pulling in two versions of lodash in `web`, we're relaxing the constraint here.

PR #301, opened 2018-09-20
Revert "[react-form-state] 🎨 rename FieldDescriptors to Fields"

lemonmade:
Reverts Shopify/quilt#231


TheMallen:
closed so we can ship other stuff, will revert the revert before we ship 1.0 of formstate

PR #306, opened 2018-09-21
remove style attribute entirely

danrosenthal:
As a follow-up to this PR: https://github.com/Shopify/quilt/pull/304, @lemonmade and I discussed in 🍖 👾 and reached the conclusion that we should remove the style attribute entirely. This will still be a breaking change requiring a 4.0.0 release of `react-html`.


danrosenthal:
cc: @tzvipm

Issue #299, opened 2018-09-20
[react-form-state] validate form programatically


macdonaldr93:
## Overview

The same way we get `submit` as a render prop from FormState, it would be great if we could get `validate` as a prop from FormState. We have a few use cases now where we need to run validation on initial render or hooked into other events than dirtying the form. Facebook Ads, Facebook Marketplace(@michelleyschen) and Marketing section (@darrenhebner).

## Motivation

> What inspired this enhancement? What makes you think this should be in `quilt`?

For us, we need to run validation right away because the data gets pre-populated from the backend so the merchant needs to know what fields are invalid according to our app.

---

## Checklist

- [x] Please delete the labels section before submitting your issue
- [x] I have described this issue in a way that is actionable (if possible)

TheMallen:
@macdonaldr93 do you think it would make sense / be enough to just have a `validateOnMount` or `validateOnInitialize` prop?

michelleyschen:
Not quite, the form fields can also hide / show based on a state change. I would also need to re-validate when those state change.

TheMallen:
I'm not really sure I follow, if fields can hide and show based on a state change but the form was validated at the start wouldn't that be good enough? They'd still have their initial errors they would just be hidden.

macdonaldr93:
Yeah I think in this particular case, a given field selection may drastically change the fields required in the form and therefore the validations needed. So when the field value changes, we'd want to re-run validation since some of the field requirements will change.

More concretely, we have a field that is product category. So if the product category field is set to Apparel, that means the MPN, Size, and Field X are required. But if the product category gets changed to something else, that won't be the case. Does that make sense?

michelleyschen:
The use case I am running into:
1. Form start with 5 fields, validate on init => 5 errors count is shown on screen
2. User trigger a state change 
3. The form now show 3 fields, (but the form it self still have 5)
4. I have two choice at this point. 
1️⃣ re-run validation (where my validation are based on state)
2️⃣ count the number of error only for the form fields that's being shown.

I would prefer 1️⃣ because I have a few components that will count / show text acoordingly
but 2️⃣ is do able

TheMallen:
Yeah, but what I'm asking is would that usecase not be handled by having all fields validate when any one is updated + the ability to have fields validate even when not dirty?

michelleyschen:
cc @laranoell also have the same use case

TheMallen:
[...]

Issue #286, opened 2018-09-13
[react-form-state] Provide ability to validate entire form at once


debchn:
## Overview

Provide the ability to validate _all_ form fields at once (e.g. onSubmit).

## Type

New feature

## Motivation

Right now, the validator functions only run automatically on `blur`. However, some users may click submit without having touched invalid fields (e.g. blank required fields).

## Thoughts on what to do

From Mal:
- Add a `validateBeforeSubmit` boolean prop that makes it automatic?
- Probably, call [`validateFieldValue`](https://github.com/Shopify/quilt/blob/master/packages/react-form-state/src/FormState.ts#L311) in a loop
- Probably [here](https://github.com/Shopify/quilt/blob/master/packages/react-form-state/src/FormState.ts#L168)

cbolton97:
+1 to this. My use-case is similar:

Context: A user is creating a new "project" using a form. 

The form is blank when they reach the page. They may fill out some information, but not other fields, and hit submit. Since some of the fields aren't "dirty" (the initial values are blank) they don't get validated (even if they click into the field and the a blur event does trigger, as long as the field isn't dirty it doesn't seem to validate - see gif below). 

When they hit submit, the backend will return the errors about some fields being blank, however it would be nice to run validation beforehand to give the user an error without making a trip to the backend.

The backend will always have validation so the impact isn't huge, but it would be nice to leverage the frontend validation for this case as well :)

![26](https://user-images.githubusercontent.com/1896810/45718301-8d848580-bb6a-11e8-8b91-0389fba4c722.gif)

michelleyschen:
similar to https://github.com/Shopify/quilt/issues/185

TheMallen:
Validating the entire form at once is now supported as of 0.3.0, if you update to the newest version of the library and check out the docs for guidance you should be good. Let me know / open an issue if it's still a problem.

PR #309, opened 2018-09-24
✨ Validate whole form on submit, expose `validateForm`

TheMallen:
#299 
#286 

This PR adds the `validateOnSubmit` prop to `<FormState>`. This prop allows you to run all your validators at submit and bail without hitting the server if any fail.

This PR also adds the public instance method `validateForm`, giving consumers the ability to validate the entire form at once, regardless of dirty state. This should only be used as an escape hatch when doing non-standard form flows.

### 🎩  instructions
- clone branch
- `yarn install && yarn build`
- `cd packages/react-form-state`
- `yarn link`
- go to your own project
- `yarn link @shopify/react-form-state`
- try it out :)

### DIFF
--- a/packages/react-form-state/docs/building-forms.md
+++ b/packages/react-form-state/docs/building-forms.md
@@ -355,6 +355,47 @@ function MyComponent() {
 }
 ```
 
+### validateOnSubmit
+
+You can configure `<FormState />` to run all validators on the form before executing `onSubmit` by passing it the `validateOnSubmit` prop. If any of the validators return an error, the submit is cancelled and `onSubmit` is not run.
+
+```typescript
+import {TextField, Form, Button} from '@shopify/polaris';
+import FormState, {validators} from '@shopify/react-form-state';
+
+function MyComponent() {
+  return (
+    <FormState
+      validateOnSubmit
+      initialValues={{
+        title: 'Cool title',
+      }}
+      validators={{
+        title(input) {
+          if (input.length > 10) {
+            return 'That title is too long';
+          }
+        },
+      }}
+      onSubmit={() => {
+        console.log('I will not be run if title is too long');
+      }}
+    >
+      {formDetails => {
+        const {fields, submit} = formDetails;
+
+        return (
+          <Form onSubmit={submit}>
+            <TextField label="Title" {...fields.title} />
+            <Button type="submit">Submit</Button>
+          </Form>
+        );
+      }}
+    </FormState>
+  );
+}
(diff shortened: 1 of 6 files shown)


packages/react-form-state/src/FormState.ts
```diff
@@ -122,19 +122,29 @@ export default class FormState<
...
+    const {errors} = this.state;
 
-    const fieldsWithErrors = Object.keys(fields).filter(fieldPath => {
-      const {error} = fields[fieldPath];
-      return error != null;
-    });
+    return !this.hasClientErrors && errors.length === 0;
+  }
+
+  private get hasClientErrors() {
```
TheMallen: should this be public as well?

michelleyschen:
🎩 and this work well for me!

packages/react-form-state/src/FormState.ts
darrenhebner: I can't think of a case where I'd need this instead of `valid`.

TheMallen:
just an update on this: Going to make it an optional prop to enable this behaviour so I can release as a feature instead of breaking change, and vet how much use this gets.

packages/react-form-state/src/FormState.ts
```diff
@@ -166,6 +177,14 @@ export default class FormState<
 
     this.setState({submitting: true});
 
+    if (validateOnSubmit) {
```
TheMallen: this is the main change, and the new docs

TheMallen:
@michelleyschen @darrenhebner  slightly new direction so I can get this merged earlier without extensive tophats (since it's no longer a breaking api change). I'd appreciate you two taking another look.

packages/react-form-state/docs/building-forms.md
```diff
@@ -355,6 +355,46 @@ function MyComponent() {
...
+
+You can configure `<FormState />` to run all validators on the form before executing `onSubmit` by passing it the `validateOnSubmit` prop. If any of the validators return an error, the submit is cancelled and `onSubmit` is not run.
+
+```typescript
+import {TextField, Form, Button} from '@shopify/polaris';
+import FormState, {validators} from '@shopify/react-form-state';
+
+function MyComponent() {
+  return (
+    <FormState
```
darrenhebner: Do you need to actually include the `validateOnSubmit` prop here?

TheMallen: lol, yes.

PR #310, opened 2018-09-25
Revert "Revert "[react-form-state] 🎨 rename FieldDescriptors to Fields""

TheMallen:
Reverts Shopify/quilt#301

We want to ship this as part of formstate 1.0, we should avoid merging for a bit to avoid blocking other package releases


TheMallen:
closing this for now, clutter

PR #311, opened 2018-09-25
Add SupportedLocale and SupportedCountry types

aishad:
Since the `country-service` graphql endpoint is typed, we want to type the library as well to let the users know upfront what are the supported locales and country codes.

**How is this accomplished?**

- Created SupportedLocale and SupportedCountry types
- Replaced locale type from `locale: string` to `locale: SupportedLocale`
- Replaced countryCode type from `countryCode: string`  to `countryCode: SupportedCountry`
- Updated tests
- Updated README and CHANGELOG


packages/address/src/loader.ts
```diff
@@ -51,8 +54,8 @@ export async function loadCountry(
       query,
       operationName: GRAPHAL_OPERATION_NAMES.country,
       variables: {
-        countryCode,
-        locale,
+        countryCode: countryCode.toUpperCase(),
```
nimzco: This shouldn't be needed since all of the country codes in `SupportedCountry` are upper case already.

packages/address/src/tests/AddressFormatter.test.ts
```diff
@@ -143,7 +143,7 @@ describe('getCountries()', () => {
   it('should not call the API again for the countries if the locale is the same.', async () => {
     mockAPICall('countries', countriesEn, 'YY');
 
-    const addressFormatter = new AddressFormatter('YY');
+    const addressFormatter = new AddressFormatter('yy' as SupportedLocale);
```
nimzco: Can you just add a comment on the line above saying `// Bypass the cache by using a non existant locale`?

nimzco: Same for the other instances as well.

packages/address/src/tests/AddressFormatter.test.ts
```diff
@@ -265,3 +265,23 @@ describe('getTranslationKey()', () => {
     expect(result).toBe('country');
   });
 });
+
+describe('toSupportedLocale', () => {
+  it('changes the lowercase locale to uppercase', async () => {
+    mockAPICall('country', countryJpJa);
```
nimzco: To make the test more explicit, I would pass explicitly `"ja"` as third argument here.

PR #312, opened 2018-09-26
[react-form-state] Stop Unnecessary Re-Renders of List and Nested

jfangrad:
This PR aims to stop `FormState.Nested` and `FormState.List` from re-rendering when none of its children have been changed. This will significantly reduce the amount of work that must be done when a component inside the form is update. (Primarily for complex forms that use multiple `Nested` and `List` components.

#### Performance
From the limited comparisons I have run, I was able to see an average of about 2-3 times better update times _on my test form_ compared to before these changes. The test form I used can be seen below. (I will admit these probably were not the most robust tests and were run on a fairly simple form)
<details>
<summary>My test form:</summary>

screen shot 2018-09-26 at 3 11 00 pm

</details>
<br>


jfangrad:
@TheMallen Thats a really good point! I completely missed that, thanks!

packages/react-form-state/src/components/List.tsx
```diff
@@ -10,10 +10,21 @@ interface Props<Fields> {
...
 }
 
-export default class List<Fields> extends React.PureComponent<
+export default class List<Fields> extends React.Component<
   Props<Fields>,
   never
 > {
+  shouldComponentUpdate(nextProps) {
+    const {
+      field: {value: nextValue},
```
TheMallen: changes to `error` should also rerender right?

packages/react-form-state/src/components/List.tsx
```diff
@@ -10,10 +10,21 @@ interface Props<Fields> {
   children(fields: FieldDescriptors<Fields>, index: number): React.ReactNode;
 }
 
-export default class List<Fields> extends React.PureComponent<
+export default class List<Fields> extends React.Component<
   Props<Fields>,
   never
 > {
+  shouldComponentUpdate(nextProps) {
```
michelleyschen: is it possible for `initialValue` to change but value & error to stay the same?
mostly thinking if something is displayed based on the `initialValue`

TheMallen: it *is* possible but it's a pretty weird edgecase

TheMallen:
@jfangrad any chance you can get to this? Should I close the PR?

jfangrad:
@TheMallen sorry for the delay! I pushed the changes to have both `Nested` and `List` re-render when their initial values change. Let me know if there is anything else you need me to do!

PR #317, opened 2018-09-28
✨Adds the ability to disable tracking

antoinegrant:
### Debug vs disableTracking
- The `debug` prop will load `analytics_debug.js`, disable tracking and output verbose logs.
- The `disableTracking` prop will simply disable tracking. This is intended to be used in `development` where we want to run the same GA code in `development` and `production` but don't want to send data to GA.

### DIFF
--- a/packages/react-google-analytics/src/GaJS.tsx
+++ b/packages/react-google-analytics/src/GaJS.tsx
@@ -12,7 +12,7 @@ export interface Props {
   allowHash?: boolean;
   set?: any[][];
   onLoad?(analytics: GaJSAnalytics): void;
-  debug?: boolean;
+  disableTracking?: boolean;
 }
 
 export const SETUP_SCRIPT = `
@@ -31,14 +31,16 @@ export default class GaJSGoogleAnalytics extends React.PureComponent<
   never
 > {
   render() {
-    const {account, debug} = this.props;
+    const {account, disableTracking} = this.props;
 
     return (
       <>
         <script
           id="google-analytics-gtag-script"
           dangerouslySetInnerHTML={{
-            __html: debug ? setupWithDebugScript(account) : SETUP_SCRIPT,
+            __html: disableTracking
+              ? setupWithDebugScript(account)
+              : SETUP_SCRIPT,
           }}
         />
         <ImportRemote
--- a/packages/react-google-analytics/src/Universal.tsx
+++ b/packages/react-google-analytics/src/Universal.tsx
@@ -10,6 +10,7 @@ export interface Props {
   set?: {[key: string]: any};
   onLoad?(analytics: UniversalAnalytics): void;
   debug?: boolean;
+  disableTracking?: boolean;
 }
 
 export const SETUP_SCRIPT = `
@@ -57,6 +58,7 @@ export default class UniversalGoogleAnalytics extends React.PureComponent<
       set: setVariables = {},
       onLoad,
       debug = false,
+      disableTracking = false,
     } = this.props;
 
     const normalizedDomain = getRootDomain(domain);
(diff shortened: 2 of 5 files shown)


antoinegrant:
👋  🏓  👍

packages/react-google-analytics/src/Universal.tsx
```diff
@@ -66,7 +68,7 @@ export default class UniversalGoogleAnalytics extends React.PureComponent<
       allowLinker: true,
     };
 
-    if (debug) {
+    if (debug || disableTracking) {
       // The debug version of the analytics.js library
       // https://developers.google.com/analytics/devguides/collection/analyticsjs/debugging
```
owain: Comments probably need to be updated to something like this (with a link to the specific doc heading)
```ts
// Prevent data being sent to Google
// https://developers.google.com/analytics/devguides/collection/analyticsjs/debugging#testing_your_implementation_without_sending_hits
```

packages/react-google-analytics/src/GaJS.tsx
```diff
@@ -12,7 +12,7 @@ export interface Props {
   allowHash?: boolean;
   set?: any[][];
   onLoad?(analytics: GaJSAnalytics): void;
-  debug?: boolean;
```
TheMallen: So this would be a breaking change (major bump) I guess?

packages/react-google-analytics/src/tests/GaJS.test.tsx
```diff
@@ -206,6 +192,32 @@ describe('<GaJS />', () => {
...
+  describe('disableTracking', () => {
+    it('not setting disableTracking should load script without restrictions', () => {
+      const gajs = mount(<GaJS {...mockProps} />);
+      expect(gajs.find(ImportRemote).prop('source')).toBe(GA_JS_SCRIPT);
+      expect(
+        gajs.find('script').prop('dangerouslySetInnerHTML'),
+      ).toHaveProperty('__html', SETUP_SCRIPT);
+    });
+
+    it('true', () => {
```
TheMallen: `it` blocks should be full sentences, If this failed in CI it would be hard to tell what it had been testing.

packages/react-google-analytics/src/tests/Universal.test.tsx
```diff
@@ -138,12 +139,59 @@ describe('<Universal />', () => {
   });
 
   describe('debug', () => {
-    it('sets the debug mode', () => {
+    it('not setting debug should load script without restrictions', () => {
```
TheMallen: This should read as a sentence.
`it loads scripts without restrictions when debug is not set`

packages/react-google-analytics/src/tests/Universal.test.tsx
```diff
@@ -138,12 +139,59 @@ describe('<Universal />', () => {
...
+      const universal = mount(<Universal {...mockProps} debug={false} />);
+      expect(universal.find(ImportRemote).prop('source')).toBe(
+        UNIVERSAL_GA_SCRIPT,
+      );
+      trigger(universal.find(ImportRemote), 'onImported', analytics);
+      expect(analytics).not.toHaveBeenCalledWith('set', 'sendHitTask', null);
+    });
+  });
+
+  describe('disableTracking', () => {
```
TheMallen: this block too

packages/react-google-analytics/src/GaJS.tsx
antoinegrant: I think that makes sense. 👍

antoinegrant:
@TheMallen Could you do another review round? Thanks!

PR #320, opened 2018-10-01
Fix Nested and List race condition

jfangrad:
I encountered a race condition when using two components within a `FormState.Nested` where one depended on the other and they would both have their state updated when a certain action occurred. Because the updated `Nested` `field item` was created outside of the `setState` function, when the two `onChange's` were called from within the same `Nested` component the second `field` object didn't include the updated values from the first `onChange` because `setState` is asynchronous.

To solve this, I moved the creation of the `Nested` `field item` object into the setState function to ensure it is done atomically. That way if consecutive `onChange` function calls are made, the `Nested` `field item` object will include the previous changes.

The `Nested` component will call the `onChange` function with the new optional `nested` parameter with an object containing the `key` to the field that was updated.

This same issue also occurs in the `FormState.List` component for exactly the same reasons and was fixed in a similar manner. `List` calls the `onChange` function with an object containing the `key` and `index` for the update.

Note: I'm kinda new to TypeScript so if my typings need to be better let me know.

### DIFF
--- a/packages/react-form-state/CHANGELOG.md
+++ b/packages/react-form-state/CHANGELOG.md
@@ -8,3 +8,4 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 - Validators based on `validate` now always succeed for an empty input. The `required` and `requiredString` validators continue to behave the same way they used to.
+- Fixed Nested and List component race condition. Nested and List now pass a function to their `onChange` prop instead of an object so that the data object will be created within `setState`.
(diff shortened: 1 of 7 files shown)


jfangrad:
@TheMallen thats a cool idea! I implemented what you suggested and its a lot cleaner, thanks!

TheMallen:
Cool cool cool, I'll give an in-depth review later today

jfangrad:
@TheMallen I did top-hat it locally. I just yarn linked it to a local test project and built a simple form with normal, nested and list fields. I also tested it with the component that I first discovered the race condition issue with and they all worked fine and the state models looked good as well.

TheMallen:
If you rebase this we can merge it and I'll do a patch release :) Thanks for the contribution! ❤️

TheMallen:
Oh, one more thing, please add an entry to the changelog (relatively new thing)

PR #325, opened 2018-10-02
✨ Add the ability to set a nonce for comply with script-src CSP

antoinegrant:
Adds the ability to set the [`nonce`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src) attribute to the script tag.

```jsx
<ImportRemote source={...} nonce="1a2b3c"  />
```


antoinegrant:
@TheMallen any chance we can cut a release for this?

antoinegrant:
🙏  thx

PR #329, opened 2018-10-03
Revert "✨ Add the ability to set a nonce for comply with script-src CSP"

antoinegrant:
Reverts Shopify/quilt#325

It is not possible to set the `nonce` at the `script` node creation time. As it would be to easy to simply find the `nonce` and inject a script with that `nonce` which would make an evil script trusted.


antoinegrant:
False alarm. This will work with the use of the CSP header config `strict-dynamic`.

PR #335, opened 2018-10-08
✅ Add more tests for trigger in enzyme-utilities

### LINKED ISSUE #129: Improve testing of `trigger` function in `enzyme-utilities`

andrewlo:
fixes #129

Add tests for the two error cases for trigger in enzyme-utilities, bringing test coverage up to 100%.

Before:
screen shot 2018-10-08 at 12 10 08 pm

After:
screen shot 2018-10-08 at 12 41 02 pm

### DIFF
--- a/packages/enzyme-utilities/src/test/index.test.tsx
+++ b/packages/enzyme-utilities/src/test/index.test.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import {mount} from 'enzyme';
+import {noop} from '@shopify/javascript-utilities/other';
 
 import {trigger, findById} from '..';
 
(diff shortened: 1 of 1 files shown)


packages/enzyme-utilities/src/test/index.test.tsx
```diff
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import {mount} from 'enzyme';
+import {noop} from '@shopify/javascript-utilities/other';
```
tzvipm: @TheMallen when did `javascript-utilities` get added to `quilt`? I think we should avoid using that directly, especially for something as simple as a `noop`. The plan is to deprecate `javascript-utilities` in favour of smaller packages in `quilt` AFAIK.

PR #338, opened 2018-10-09
Revert Fix Nested and List race condition [#320]

jfangrad:
Reverting #320 
#320 made it so that using a Nested within a Nested no longer works


TheMallen:
we can fix forward with your other PR.

PR #339, opened 2018-10-10
Allow per-operation functions for Apollo mock

lemonmade:
This PR allows you to specify a function for an individual operations in our GraphQL mocks. We already supported a function for all operations (in place of an object keyed to the operation name with values as the raw fixtures), but it might be more desirable to just use a function per-operation (which could allow us to more easily handle cases where there is an operation they did not supply a fixture for in the future).


lemonmade:
@cartogram the docs just point to the TypeScript definitions for how to use it :P

PR #340, opened 2018-10-10
Fix race condition when using multiple levels of Nested and List

jfangrad:
This PR is built off of #320 with changes to ensure that it still works when using multiple levels / combinations of Nested and List components. (#320 fixed the race condition when using one level deep but broke all functionality of multiple levels)

**New tests cases:**
- Nested -> Nested
- Nested -> List
- List -> List
- List -> Nested

**Tophat testing:**
Tested it on a form that uses:
- Individual field
- Nested
- Nested -> Nested
- Nested -> Nested -> List -> Nested

All behaved normally and as expected.


TheMallen:
(make sure to add to the changelog :) )

PR #341, opened 2018-10-10
✅jest-mock-apollo tests

### LINKED ISSUE #30: Improve test coverage for jest-mock-apollo

andrewlo:
fixes #30

Add tests for jest-mock-apollo package. Fix one bug in the package

Before:
screen shot 2018-10-08 at 2 29 29 pm

After:
screen shot 2018-10-10 at 1 14 29 pm

### DIFF
--- a/packages/jest-mock-apollo/src/Requests.ts
+++ b/packages/jest-mock-apollo/src/Requests.ts
@@ -13,7 +13,7 @@ export default class Requests {
 
   nth(index: number) {
     return index < 0
-      ? this.requests[this.requests.length - index]
+      ? this.requests[this.requests.length - 1 + index]
       : this.requests[index];
   }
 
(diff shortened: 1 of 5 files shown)


packages/jest-mock-apollo/src/Requests.ts
```diff
@@ -13,7 +13,7 @@ export default class Requests {
 
   nth(index: number) {
     return index < 0
-      ? this.requests[this.requests.length - index]
+      ? this.requests[this.requests.length - 1 + index]
```
andrewlo: I think this is a bug. If index is negative, subtracting it from length would go outside the bounds of the array right?

I assumed the intention was passing -N would be Nth from the end, ex. -1 would be 2nd last

TheMallen:
You can just rebase and merge this at your leisure :)

TheMallen:
@andrewlo  looks like you have a conflict now, if you can fix that we can get this merged

PR #346, opened 2018-10-10
🎨 Use reduceRight for react-compose

sambostock:
This updates `react-compose` to use `reduceRight` to successively call the enhancers, rather than build a chain of functions that all call each other.

This also removes the need to check the length of `wrappingFunctions`, as we can just pass `OriginalComponent` in as the initial value.

I was reading through the code to understand how `react-compose` works, and I was a little confused off my the implementation. 😅 **I'm not sure if there is a reason for the original implementation**, but as far as I can tell they behave identically.

It should be noted here that [`reduceRight`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/ReduceRight) is used, not [`reduce`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce), as otherwise the order of enhancement would be reversed.

### DIFF
--- a/packages/react-compose/src/index.ts
+++ b/packages/react-compose/src/index.ts
@@ -14,18 +14,11 @@ export default function compose<Props>(
   return function wrapComponent<ComposedProps, C>(
     OriginalComponent: ReactComponent<ComposedProps> & C,
   ): ReactComponent<Props> & C {
-    let result: ReactComponent<ComposedProps>;
-
-    if (wrappingFunctions.length === 0) {
-      result = OriginalComponent;
-    } else {
-      result = wrappingFunctions.reduce(
-        (wrappingFunctionA, wrappingFunctionB) => {
-          return (WrappingComponent: ReactComponent<any>) =>
-            wrappingFunctionA(wrappingFunctionB(WrappingComponent));
-        },
-      )(OriginalComponent);
-    }
+    const result: ReactComponent<ComposedProps> = wrappingFunctions.reduceRight(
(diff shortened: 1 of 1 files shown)


TheMallen:
@nickjs you might want to take a look at this since the quilt library was copied from your work in web :)

sambostock:
Thanks for the 👀, @nickjs!

> **The reason for the initial short circuit is just to save some logic in case no enhancers are provided. You may still want to include that.**

From MDN
> if `initialValue` is provided but the array is empty, the solo value would be returned without calling callback.

```javascript
[].reduceRight
// ƒ reduceRight() { [native code] }
```

`reduceRight` is implemented as native code, so it is probably faster for us to let it do the `length` check, rather than do it ourselves.

I could _technically_ extract

```typescript
function wrap(component: ReactComponent<any>, wrappingFunction: WrappingFunction): ReactComponent<any> {
  return wrappingFunction(component);
}
```

into an external named function, so it isn't created every time `compose` is called, and then call
```typescript
wrappingFunctions.reduceRight(wrap, OriginalComponent)
```

but that might be overkill.

> **Changing the order of the enhancers doesn't break anything**

I tested this out, and while `reduce` would reverse the order the functions wrap the component in, `reduceRight` preserves the current order. 👍

nickjs:
Sounds good, works for me!

TheMallen:
@sambostock just rebase and merge once your CI is green and I can cut a release for you. I'll handle version bumping.

PR #350, opened 2018-10-11
✨ [react-form-state] add validateRequired helper

TheMallen:
These PR adds another validation factory function to `<FormState />`'s validation DSL, `validateRequired`. This new function works the same way as `validate`, except that it does not bail on empty input. See the docs and changelog for more information.


packages/react-form-state/src/validators.ts
```diff
@@ -133,6 +133,32 @@ export function validate<Input, Fields = never>(
   };
 }
 
+export function validateRequired<Input>(
+  matcher: Matcher<Input>,
+  errorContent: ErrorContent,
+): (input: Input) => ErrorContent | void;
+
+export function validateRequired<Input, Fields = never>(
```
michelleyschen: I run into the same type of typing issue as https://github.com/Shopify/quilt/pull/349

Otherwise this looks good!

packages/react-form-state/src/validators.ts
```diff
@@ -97,12 +96,12 @@ export function validateList<Input extends Object, Fields>(
 export function validate<Input>(
   matcher: Matcher<Input, any>,
   errorContent: ErrorContent,
-): (input: Input) => ValidationFunction<Input, never>;
```
TheMallen: this type was just wrong (probably wasn't breaking anything because validationFunction can return `any` anyway)

PR #351, opened 2018-10-12
🐛 fix submitting status when submit validation fails

TheMallen:
This PR fixes [submission status getting stuck](https://codesandbox.io/s/1v2zlzm0lj) when `validateOnSubmit` is true and a validator fails.

### DIFF
--- a/packages/react-form-state/CHANGELOG.md
+++ b/packages/react-form-state/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
+## [0.3.2]
+
+### Fixed
+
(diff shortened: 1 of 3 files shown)


packages/react-form-state/CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
+## [0.3.2]
```
cbolton97: I'm guessing there is a 0.3.1 coming first?

TheMallen: yeah 0.3.1 is actually already deployed, changelog entry is just missing

TheMallen: I'll fix that separately

PR #354, opened 2018-10-12
🔥 code using @shopify/address-mocks

nimzco:
## What

🔥code using by replacing it with that lives in `@shopify/address-mocks`


packages/address/src/index.ts
```diff
@@ -1,3 +1,6 @@
+import * as testUtilities from './testUtilities';
```
TheMallen: So the problem with adding this to your index file is that it will be added to the main bundle of your app, which might have some weird side effects if you import any jest related stuff, and will add to the size.

packages/address/README.md
```diff
@@ -147,9 +147,9 @@ If your component uses this package and you want to test it with mock API calls
 
 ```
 import {fetch} from '@shopify/jest-dom-mocks';
-import {mockCountryRequests} from '@shopify/address/tests';
+import {testUtilities} from '@shopify/address';
```
TheMallen: is it possible you could have a custom tsconfig just for your test utilities that just compiles straight to the folder? then you could import this like

```
import testUtilities from '@shopify/address/test-utilities';
```

packages/address/src/tests/AddressFormatter.test.ts
```diff
@@ -1,7 +1,7 @@
 import {fetch} from '@shopify/jest-dom-mocks';
 import {Address, FieldName} from '../types';
 import AddressFormatter from '..';
-import {mockCountryRequests} from './utilities';
+import {mockCountryRequests} from '../testUtilities';
```
TheMallen: the other thing you could do is add the test utilities as a separate package.

packages/address/src/index.ts
nimzco: Hmm, I wasn't sure about this one. I thought it would be ok as long as we don't do imports other than in our test files.

packages/address/src/tests/AddressFormatter.test.ts
nimzco: Love that idea 👍 
I've extracted everything in https://github.com/Shopify/quilt/pull/355 and will reuse it here once shipped published.

packages/address/src/tests/AddressFormatter.test.ts
```diff
@@ -1,7 +1,7 @@
 import {fetch} from '@shopify/jest-dom-mocks';
+import {mockCountryRequests} from '../../../address-mocks/src';
```
nimzco: For a reason I don't understand:
```js
import {mockCountryRequests} from '@shopify/address-mocks';
```
does not work...

packages/address/src/tests/AddressFormatter.test.ts
```diff
@@ -125,7 +125,7 @@ describe('format()', () => {
       '日本',
       '〒100-8994東京都目黒区八重洲1-5-3',
       'Shopify',
-      '田中恵子様',
+      '田中 恵子様',
```
nimzco: That's because of fixture update. Now, we put a space between `firstName` and `lastName` in `ja`

packages/address/src/tests/AddressFormatter.test.ts
```diff
@@ -21,14 +21,14 @@ beforeEach(mockCountryRequests);
 afterEach(fetch.restore);
 
 describe('updateLocale()', () => {
-  it('returns the country in the correct locale', async () => {
+  it.only('returns the country in the correct locale', async () => {
```
BlazingBBQ: Is `.only` necessary here?

nimzco: Oh shoe, I forgot to remove it! 🙈

PR #356, opened 2018-10-12
Updated countries fixtures

nimzco:
- Some countries where out of date. Just updating the fixtures from what `country-service` returns.
- Added typing.


nimzco:
Thanks for the follow up review! 😄

PR #360, opened 2018-10-15
upgrading graphql-tool-utilities to 0.9.0

patsissons:
Upgrades graphql tooling to the latest release. `graphql-tool-utilities` is current in beta pending confirmation that the upgrade works properly in `@shopify/jest-mock-apollo`.

We can remove the `graphql` dependency as `graphql-tool-utilities` will now provide the correct version.

There are an expectedly large number of changes to the `yarn.lock` file because of the dependency upgrades that occurred in `graphql-tool-utilities`. This should be a drop in replacement for the old `0.7.x` release, everything should be backwards compatible with only additions and upgrades changing in `0.8.x`. `graphql-tool-utilities` received a minor version bump just because many of the dependencies were upgraded significantly, so there is a risk of some behaviour changing. In practice, no issues have surfaced.

### DIFF
--- a/packages/jest-mock-apollo/package.json
+++ b/packages/jest-mock-apollo/package.json
@@ -24,8 +24,7 @@
   "homepage": "https://github.com/Shopify/quilt/blob/master/packages/jest-mock-apollo/README.md",
   "dependencies": {
     "apollo-link": "^1.2.2",
-    "graphql": "^0.13.2",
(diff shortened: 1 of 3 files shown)


patsissons:
to anyone arriving after the merge, the branch name was created before `0.9.0` was released. `0.9.0` is the expected upgrade version 👍

PR #362, opened 2018-10-16
Ensure tests pass typescript compilation and bump jest and ts-jest

BPScott:
The newest ts-jest compiles your tests and will fail by default if your
tests don't pass typescript compilation.

So that upgrade is smooth whenever you want to do that, lets ensure the tests compile successfully.


TheMallen:
Rebase and merge :)

PR #363, opened 2018-10-16
[address] Improve cache by memoizing async functions

nimzco:
### Problem

Right now, if `.getCountries` is called one after the other before the request has finished, the cache is useless. This is something that will happen often when used with components. We already have the problem in [`CustomerNew`](https://github.com/Shopify/web/blob/master/app/sections/Customers/CustomerNew/CustomerNew.tsx#L205) where we have both `PhoneField` and `AddressForm` that fetch all countries.

### Solution

- Memoize async functions
- Remove cache for `.getCountry` since now the fetch call is memoized
- Don't remove it when calling `.getCountries` because it will still be used if we do something like:

```tsx
addressFormatter.getCountries()
addressFormatter.getCountry("CA") // Will retrieve from `ORDERED_COUNTRIES_CACHE`
```

### DIFF
--- a/packages/address/src/loader.ts
+++ b/packages/address/src/loader.ts
@@ -18,7 +18,9 @@ const HEADERS = {
   'Access-Control-Allow-Origin': '*',
 };
 
-export async function loadCountries(locale: string): Promise<Country[]> {
+export const loadCountries: (
+  locale: string,
+) => Promise<Country[]> = memoizeAsync(async (locale: string) => {
   const response = await fetch(GRAPHQL_ENDPOINT, {
     method: 'POST',
     headers: HEADERS,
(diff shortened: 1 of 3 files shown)


packages/address/src/loader.ts
```diff
@@ -18,7 +18,9 @@ const HEADERS = {
   'Access-Control-Allow-Origin': '*',
 };
 
-export async function loadCountries(locale: string): Promise<Country[]> {
+export const loadCountries: (
+  locale: string,
+) => Promise<Country[]> = memoizeAsync(async (locale: string) => {
```
loic-d: Do you need an explicit return type?

packages/address/src/loader.ts
```diff
@@ -94,3 +98,20 @@ export function toSupportedLocale(locale: string) {
...
+
+type AsyncFunc = (...args: any[]) => Promise<any>;
+interface Cache {
+  [key: string]: Promise<any>;
+}
+
+function memoizeAsync(asyncFunction: AsyncFunc) {
+  const cache: Cache = {};
+
+  return (..._arguments: any[]) => {
```
loic-d: Why do you need to prefix `arguments` with `_`. The param is used in your function.

packages/address/src/tests/AddressFormatter.test.ts
```diff
@@ -52,20 +52,21 @@ describe('getCountry()', () => {
   });
 
   it('should not call the API again for the same country if the locale is the same', async () => {
-    const addressFormatter = new AddressFormatter('ja');
+    const addressFormatter = new AddressFormatter('pt-br');
+
     await addressFormatter.getCountry('CA');
     await addressFormatter.getCountry('CA');
 
     expect(fetch.calls()).toHaveLength(1);
```
loic-d: Nice!

packages/address/src/loader.ts
```diff
@@ -94,3 +98,20 @@ export function toSupportedLocale(locale: string) {
     return DEFAULT_LOCALE;
   }
 }
+
+type AsyncFunc = (...args: any[]) => Promise<any>;
+interface Cache {
+  [key: string]: Promise<any>;
+}
+
+function memoizeAsync(asyncFunction: AsyncFunc) {
```
loic-d: Don't we have a JS utility already?

packages/address/src/loader.ts
nimzco: @clauderic says it's nice to have one in case I change my function's signature 👅

packages/address/src/loader.ts
```diff
@@ -38,31 +40,33 @@ export async function loadCountries(locale: string): Promise<Country[]> {
...
+        variables: {
+          countryCode,
+          locale: toSupportedLocale(locale),
+        },
+      }),
+    });
+    const country: LoadCountryResponse = await response.json();
+    if ('errors' in country) {
+      throw new CountryLoaderError(country);
+    } else {
```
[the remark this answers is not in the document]
nimzco: Good point! 👍

packages/address/src/loader.ts
nimzco: I did that because `arguments` was a used keywords and `args` was already taken.
But I've just renamed it 👍

packages/address/src/loader.ts
nimzco: I didn't find one for `memoizeAsync` 🤔

PR #373, opened 2018-10-22
[DO_NOT_MERGE] replacing @shopify/react-apollo with react-apollo

patsissons:
`@shopify/jest-mock-apollo` should match `web` once it switches to `react-apollo`.

### DIFF
--- a/packages/jest-mock-apollo/src/test/index.test.tsx
+++ b/packages/jest-mock-apollo/src/test/index.test.tsx
@@ -6,6 +6,7 @@ import {mount} from 'enzyme';
 import {readFileSync} from 'fs';
 import {buildSchema} from 'graphql';
 import * as path from 'path';
+import PropTypes from 'prop-types';
 
 import configureClient from '..';
 import unionOrIntersectionTypes from './fixtures/schema-unions-and-interfaces.json';
--- a/packages/react-i18n/src/decorator.tsx
+++ b/packages/react-i18n/src/decorator.tsx
@@ -150,11 +150,10 @@ export function withI18n({id, fallback, translations}: WithI18nOptions = {}) {
       }
     }
 
-    const FinalComponent = hoistStatics(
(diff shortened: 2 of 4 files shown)


packages/jest-mock-apollo/src/test/index.test.tsx
```diff
@@ -40,6 +41,9 @@ describe('jest-mock-apollo', () => {
       context: {
         client: createGraphQLClient(),
       },
+      childContextTypes: {
+        client: PropTypes.object,
```
patsissons: this was required to get the context pushed into mounted components.

patsissons: @lemonmade do you have insight on this? Feels like we should probably abstract away the requirement to specify the `childContextTypes` when mounting.

lemonmade: I'm surprised this didn't come up before. We usually create a custom mount function that abstracts this away (as in `mountWithAppContext` for web)

patsissons: i suspect our internal `react-apollo` was wrapping with child context for us previously.

packages/react-i18n/src/decorator.tsx
```diff
@@ -138,7 +138,7 @@ export function withI18n({id, fallback, translations}: WithI18nOptions = {}) {
       WithTranslation,
       WrappedComponent as React.ComponentClass<any>,
     );
-    return FinalComponent as React.ComponentClass<any> & C;
+    return FinalComponent as any;
```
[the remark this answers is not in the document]
patsissons: this was no longer necessary, we can properly type the composed component now.

patsissons:
I just rebased this to see if there was anything to keep, it looks like just some broken upgrades. I think i will probably close this PR out if no objections.

Issue #337, opened 2018-10-09
Lint against importing all of lodash


TheMallen:
ref:  #336

Set up https://github.com/Shopify/eslint-plugin-shopify/blob/ff227bbe0e13d0aa057e1baa0e232869e944eff1/docs/rules/restrict-full-import.md for `lodash` in this repo so that we do not accidentlaly import all of lodash.

PR #376, opened 2018-10-24
Add linting for full lodash package

theodoretan:
Closes https://github.com/Shopify/quilt/issues/337

We want to restrict importing the entire lodash package.

Errors:
screen shot 2018-10-24 at 5 04 00 pm

No Errors:
screen shot 2018-10-24 at 5 07 52 pm

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,5 +2,8 @@
   "extends": [
     "plugin:shopify/typescript-react",
     "plugin:shopify/typescript-prettier"
-  ]
+  ],
+  "rules": {
+    "shopify/restrict-full-import": ["error", "lodash"]
+  }
 }
--- a/packages/dates/src/parse-date-string.ts
+++ b/packages/dates/src/parse-date-string.ts
@@ -18,7 +18,7 @@ export function parseDateString(dateString: string, timeZone?: string) {
   }
 
   const [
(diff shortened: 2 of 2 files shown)


TheMallen:
Hmmm, seems like this is exploding in CI?

theodoretan:
Yeah 🤔

```
Cannot read property 'type' of null
TypeError: Cannot read property 'type' of null
    at left.type.Identifier.left.type.ArrayPattern.left.elements.some (/home/travis/build/Shopify/quilt/node_modules/eslint-plugin-shopify/lib/rules/restrict-full-import.js:42:26)
```

Did some digging and it looks like it’s coming from here because first value is blank.
https://github.com/Shopify/quilt/blob/56106636699ef57fed8eac9403ecbe7c9a86b78a/packages/dates/src/parse-date-string.ts#L20-L32

TheMallen:
Might as well change that here, it won't effect any functionality :)

TheMallen:
Still seeing an error in CI, I think you can fix it by changing it from `_` to any `_` prefixed name like `_unused`. If that doesn't work you can just put a `// @ts-ignore` right above the offending line.

theodoretan:
🙌 Thanks!

PR #391, opened 2018-10-31
✨ update the propagator to use app-bridge action instead of post mess…

michelleyschen:
…age.

In version one, this library actually will fail if the user was not using easdk and have not initialize the app.

This update add the dependency with app bridge, and call the app bridge action when the url should be updated.


michelleyschen:
cc @tmlayton

packages/react-shopify-app-route-propagator/package.json
```diff
@@ -28,5 +28,8 @@
   "devDependencies": {
     "typescript": "~3.0.1"
   },
-  "sideEffects": false
+  "sideEffects": false,
+  "dependencies": {
+    "@shopify/app-bridge": "^0.8.0"
```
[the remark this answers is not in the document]
michelleyschen: We would expect them to already have it installed. And that would be `peerDependency`?

michelleyschen:
so what I realized is, the original lib actually do have a dependency on the old easdk. It assumed that easdk initialized method had been called before this library would work.

I am un-sure how the app-bridge team feel about handling this in their app. My feelings is that it will be 👎 . But lets ask.

cc @vividviolet @gf3

gf3:
i like the idea! i think it probably belongs with the polaris embedded components

vividviolet:
I really like @gf3's suggestion. Make this route propagator a component that users can import from `@shopify/polaris/embedded`

michelleyschen:
cc @tmlayton join the discussion on where this component belong.

tmlayton:
We have removed the embedded bundle in Polaris and now all components are imported from the main bundle. The goal is to not have a concept of embedded vs non-embedded components and props. There are only a few props which are for embedded apps only, otherwise the only offending component is resource picker which we are considering adding a non-embedded version of the component. So I don't think this belongs in Polaris.

michelleyschen:
IMO we should do `Peer dependency`, even in the current version, we do have a peer dependency on easdk just un-documented.

michelleyschen:
@TheMallen I update to have `app-bridge` as a peer dependency, and the README a bit more.

@andrewlo @macdonaldr93 @jgodson @jennaleeb 
I can use some 👁 on if this README is clear plz 🙏

packages/react-shopify-app-route-propagator/README.md
```diff
@@ -3,16 +3,54 @@
 [![Build Status](https://travis-ci.org/Shopify/quilt.svg?branch=master)](https://travis-ci.org/Shopify/quilt)
 [![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](LICENSE.md) [![npm version](https://badge.fury.io/js/%40shopify%2Freact-shopify-app-route-propagator.svg)](https://badge.fury.io/js/%40shopify%2Freact-shopify-app-route-propagator) ![npm bundle size (minified + gzip)](https://img.shields.io/bundlephobia/minzip/@shopify/react-shopify-app-route-propagator.svg)
 
-`<ShopifyRoutePropagator />` is a simple component to synchronize a Shopify embedded app's client side routing with the outer iframe host.
+`<ShopifyRoutePropagator />` is a simple component to synchronize a Shopify embedded app's client side routing with the outer iframe host. It assume the embedded app is using [Shopify's App Bridge Library](https://help.shopify.com/en/api/embedded-apps/app-bridge)
```
jennaleeb: Nit `assumes`.

Is there any way to have an embedded app without app bridge?

packages/react-shopify-app-route-propagator/README.md
```diff
@@ -3,16 +3,54 @@
...
 
 The component is quite small and can be used with any routing solution.
 
 ## Installation
 
 ```bash
+$ yarn add @shopify/app-bridge
```
jennaleeb: Would specify only if you do not have it already otherwise people mindlessly c/p.

michelleyschen: if they have it already, yarn will not install it again.

packages/react-shopify-app-route-propagator/README.md
```diff
@@ -3,14 +3,51 @@
 [![Build Status](https://travis-ci.org/Shopify/quilt.svg?branch=master)](https://travis-ci.org/Shopify/quilt)
 [![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](LICENSE.md) [![npm version](https://badge.fury.io/js/%40shopify%2Freact-shopify-app-route-propagator.svg)](https://badge.fury.io/js/%40shopify%2Freact-shopify-app-route-propagator) ![npm bundle size (minified + gzip)](https://img.shields.io/bundlephobia/minzip/@shopify/react-shopify-app-route-propagator.svg)
 
-`<ShopifyRoutePropagator />` is a simple component to synchronize a Shopify embedded app's client side routing with the outer iframe host.
+`<ShopifyRoutePropagator />` is a simple component to synchronize a Shopify embedded app's client side routing with the outer iframe host. It assumes the embedded app is using [Shopify's App Bridge Library](https://help.shopify.com/en/api/embedded-apps/app-bridge)
```
jgodson: or `Polaris V3`?

michelleyschen: 😂 lol yes

tmlayton:
FWIW I’m proposing a [pilot project to extend the Polaris library](https://unicorn.shopify.io/projects/6624) where we could move components like this. I could also see moving the Polaris components which wrap App Bridge, [behavior components](https://github.com/orgs/Shopify/teams/polaris/discussions/10), etc to that project.

PR #393, opened 2018-11-01
React-i18n / Add support for `getCurrencySymbol`

loic-d:
This PR adds support for `getCurrencySymbol`.

This method will return the currency symbol according to the currency code and locale.

The implementation has been duplicated from `withIntl` in `web`.

Also 🔥 the `utilities.tsx` to organize all utilities under a folder.

🎩 instructions to come... 

No luck with `yarn link`. You'll have to copy the generated `dist` folder in `web` under `node_modules/@shopify/react-i18n`

You can then checkout `discounts/localize-currency` in `web` and run `dev run dev`

### DIFF
--- a/packages/react-i18n/src/i18n.ts
+++ b/packages/react-i18n/src/i18n.ts
@@ -1,4 +1,5 @@
 import {languageFromLocale, regionFromLocale} from '@shopify/i18n';
+import {memoize, autobind} from '@shopify/javascript-utilities/decorators';
 import {
   I18nDetails,
   PrimitiveReplacementDictionary,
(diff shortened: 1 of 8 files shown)


packages/react-i18n/src/i18n.ts
```diff
@@ -169,6 +174,22 @@ export default class I18n {
 
     return WEEK_START_DAYS.get(country) || DEFAULT_WEEK_START_DAY;
   }
+
+  @autobind
+  getCurrencySymbol(currencyCode?: string) {
+    const currency = currencyCode || this.defaultCurrency;
+    if (currency == null) {
+      throw new Error(
```
nimzco: Can you create a custom error here: https://github.com/Shopify/quilt/blob/master/packages/react-i18n/src/errors.ts to follow the other errors? 🙏

loic-d: I can re-use `MissingCurrencyCodeError`

nimzco: Awesome 👍

PR #396, opened 2018-11-02
adding formatPercentage utility function to simplify legacy i18n migration

patsissons:
this is a utility _passthru_ function to simplify migration from legacy `i18n` solutions that simply calls back into `formatNumber` with `{as: 'currency'}` automatically added to the format options.

relates to #395


nimzco:
I forgot to ask, but can you update the README?

patsissons:
ah readme, i knew i was forgetting something 👍

PR #397, opened 2018-11-02
add DateStyle date formatting helpers to react-i18n

patsissons:
Provides `DateStyle` helpers to `formatDate` (`Long`, `Short`, `Humanize`, `Time`) to simplify migration from legacy `i18n` solutions.

`Humanize` adds additional formatting for `Today` and `Yesterday`.

This change is the first pass of adding date formats, with #398 covering the remaining formats.

### DIFF
--- a/packages/react-i18n/src/i18n.ts
+++ b/packages/react-i18n/src/i18n.ts
@@ -8,6 +8,8 @@ import {
   LanguageDirection,
 } from './types';
 import {
+  dateStyle,
+  DateStyle,
   DEFAULT_WEEK_START_DAY,
   WEEK_START_DAYS,
   RTL_LANGUAGES,
@@ -156,7 +158,10 @@ export default class I18n {
     return this.formatNumber(amount, {as: 'percent', ...options});
   }
 
-  formatDate(date: Date, options?: Intl.DateTimeFormatOptions) {
+  formatDate(
+    date: Date,
+    options?: Intl.DateTimeFormatOptions & {style?: DateStyle},
+  ): string {
     const {locale, defaultTimezone: timezone} = this;
 
     if (timezone == null && (options == null || options.timeZone == null)) {
@@ -165,9 +170,17 @@ export default class I18n {
       );
     }
 
+    const {style = undefined, ...formatOptions} = options || {};
+
+    if (style) {
+      return style === DateStyle.Humanize
+        ? this.humanizeDate(date, formatOptions)
+        : this.formatDate(date, {...formatOptions, ...dateStyle[style]});
+    }
+
     return new Intl.DateTimeFormat(locale, {
       timeZone: timezone,
-      ...options,
+      ...formatOptions,
     }).format(date);
   }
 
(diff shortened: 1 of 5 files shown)


packages/react-i18n/src/i18n.ts
```diff
@@ -157,9 +162,17 @@ export default class I18n {
       );
     }
 
+    const {style = undefined, ...formatOptions} = options || {};
+
+    if (style) {
+      return style === DateStyle.Humanize
+        ? this.humanizeDate(date, formatOptions)
+        : this.formatDate(date, {...formatOptions, ...dateStyle[style]});
```
patsissons: should the options expansion be in the other order here?

packages/react-i18n/src/i18n.ts
```diff
@@ -190,6 +203,27 @@ export default class I18n {
...
+    } else if (isYesterday(date)) {
+      humanizedDate = this.translate('yesterday');
+    } else {
+      humanizedDate = this.formatDate(date, {
+        ...options,
+        ...dateStyle[DateStyle.Humanize],
+      });
+    }
+
+    return humanizedDate;
```
patsissons: we don't need to store the `humanizedDate` to return it here, we can just return it immediately.

packages/react-i18n/src/i18n.ts
```diff
@@ -200,3 +234,21 @@ function isTranslateOptions(
...
+  );
+}
+
+function isYesterday(date: Date) {
+  const yesterday = new Date();
+  yesterday.setDate(yesterday.getDate() - 1);
+
+  return (
+    yesterday.getFullYear() === date.getFullYear() &&
+    yesterday.getMonth() === date.getMonth() &&
```
patsissons: call `isCurrentMonthAndYear` here instead?

packages/react-i18n/src/test/i18n.test.ts
```diff
@@ -387,6 +395,89 @@ describe('I18n', () => {
 
       expect(i18n.formatDate(date, {timeZone: timezone})).toBe(expected);
     });
+
+    it('formats a date using DateStyle.Long', () => {
+      const date = new Date('2012-12-20T00:00:00-00:00');
+      const i18n = new I18n(defaultTranslations, {
+        ...defaultDetails,
+        timezone,
```
patsissons: `timezone` for static dates should be in `UTC`

packages/react-i18n/src/i18n.ts
nimzco: Hmmm, in this way, we can't overwrite dateStyle keys, but at the same time, if you pass a key, you want it to follow our set of styles. So I'd say it's good this way. 🤔

packages/react-i18n/src/test/i18n.test.ts
```diff
@@ -411,6 +419,89 @@ describe('I18n', () => {
...
+
+    it('formats a date using DateStyle.Short', () => {
+      const date = new Date('2012-12-20T00:00:00-00:00');
+      const i18n = new I18n(defaultTranslations, {
+        ...defaultDetails,
+        timezone,
+      });
+      const expected = 'Dec 20, 2012';
+
+      expect(i18n.formatDate(date, {style: DateStyle.Short})).toBe(expected);
```
nimzco: 👌 
🚲 : I would even get rid of the constant here

PR #399, opened 2018-11-02
Update koa-metrics to use distribution

JakeHendrick:
### Purpose

Update `timing` -> `distribution` since datadog removed support for `timing`.


GoodForOneFare:
Tagging in @ismail-syed because Mal is out of office today.

PR #400, opened 2018-11-06
Update locale support to follow iso format

nimzco:
Currently if the translations are stored in file named `pt-BR.json`, they
won't be loaded because it is looking for `pt-br.json` lowercase.
We follow the ISO standard that has first part of the locale lowercase
and the second part uppercase.

Before:
`getPossibleLocales('pt-BR') => ["pt-br", "pt"]`

After:
`getPossibleLocales('pt-BR') => ["pt-BR", "pt-br", ""pt"]`

### DIFF
--- a/packages/react-i18n/src/test/manager.test.ts
+++ b/packages/react-i18n/src/test/manager.test.ts
@@ -22,11 +22,11 @@ describe('Manager', () => {
     switch (locale) {
       case 'en':
         return enParent;
-      case 'en-us':
+      case 'en-US':
         return enUSParent;
       case 'fr':
         return frParent;
-      case 'fr-ca':
+      case 'fr-CA':
         return frCAParent;
     }
 
@@ -37,11 +37,11 @@ describe('Manager', () => {
     switch (locale) {
       case 'en':
         return en;
-      case 'en-us':
+      case 'en-US':
         return enUS;
       case 'fr':
         return fr;
-      case 'fr-ca':
+      case 'fr-CA':
         return frCA;
     }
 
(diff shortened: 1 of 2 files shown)


packages/react-i18n/src/test/manager.test.ts
```diff
@@ -22,11 +22,11 @@ describe('Manager', () => {
     switch (locale) {
       case 'en':
         return enParent;
-      case 'en-us':
+      case 'en-US':
```
GoodForOneFare: Isn't this going to break the `requests translations for the (lowercase) full locale and language when the country code is provided` test?

nimzco: I can actually leave it as is, both should work. I've updated the function for it to return:
`["pt-BR", "pt-br", ""pt"]` so everything would work.

nimzco: I've updated some tests after all.

GoodForOneFare:
Why aren't we renaming the file in `shopify/web`?

nimzco:
I thought about this, but I think it should actually allow every possibility to work or favor the standard which is `lowercase-UPPERCASE` (`pt-BR`).
Also, when looking at the README, I see:

> Your translations function can be called several times for a given locale. If, for example, the locale is en-CA, your function will be called with en-CA and en, which allows you to load country-specific variations for translations.

Which makes me think as a consumer, that my translations should be in `fr.json` and `fr-CA.json`.

But I can leave all the tests to be lowercase as well, both will work.

lemonmade:
This was a mistake in my original implementation, I agree with this PR that we should fix it.

PR #405, opened 2018-11-13
✨Update `DateStyle.Humanize` to match Polaris guidelines

rexmac:
Update `humanizeDate` to adhere to [Polaris guidelines](https://polaris.shopify.com/content/grammar-and-mechanics#section-dates-numbers-and-addresses).

screen shot 2018-11-16 at 3 54 41 pm


packages/react-i18n/src/constants.ts
```diff
@@ -1,11 +1,19 @@
 export enum DateStyle {
+  ExtraLong = 'ExtraLong',
```
nimzco: Should that be called `LongWithTime` to indicate it shows the time as well?

packages/react-i18n/src/test/i18n.test.ts
```diff
@@ -420,6 +420,18 @@ describe('I18n', () => {
...
 
+    it('formats a date using DateStyle.ExtraLong', () => {
+      const date = new Date('2012-12-20T00:00:00-00:00');
+      const i18n = new I18n(defaultTranslations, {
+        ...defaultDetails,
+        timezone,
+      });
+
+      expect(i18n.formatDate(date, {style: DateStyle.ExtraLong})).toBe(
+        'December 20, 2012, 11:00 AM',
```
nimzco: Shouldn't you have the weekday here? Like `Thursday, December 20, 2012 00:00` that you have in the doc?

packages/react-i18n/src/i18n.ts
```diff
@@ -250,7 +248,12 @@ function isSameDate(source: Date, target: Date) {
   );
 }
 
-function isYesterday(date: Date) {
+export function isToday(date: Date) {
```
nimzco: Can you group all date related functions in a separate `date` file?
I see we have `isToday` `isYesterday`, `isSameMonthAndYear`. 
That would also be a place where we can put more utilities.
I'm thinking about https://github.com/Shopify/javascript-utilities/blob/master/src/dates.ts right now, but we'll add as we might need them.

packages/react-i18n/src/constants.ts
rexmac: Ya, I feel like the name could be better.

packages/react-i18n/src/test/i18n.test.ts
rexmac: The doc example is a mistake. The weekday should not be included.

nimzco: Oh ok then! 👍

packages/react-i18n/src/i18n.ts
rexmac: https://github.com/Shopify/javascript-utilities/pull/44

packages/react-i18n/src/constants.ts
patsissons: Yea i would vote for `LongWithTime` as well

rexmac: OK, I've gone with `LongWithTime` for now, though it bothers me that `Long` includes the weekday and `LongWithTime` does not.

packages/react-i18n/src/i18n.ts
rexmac: Done.

packages/react-i18n/src/constants.ts
kyledurand: It bothers me as well since it doesn't exactly follow the polaris guidelines. Maybe we should remove this and leave it up to the consumer to provide custom formatting

packages/react-i18n/README.md
```diff
@@ -70,6 +70,7 @@ The provided `i18n` object exposes many useful methods for internationalizing yo
   - `DateStyle.Long`: e.g., `Thursday, December 20, 2012`
   - `DateStyle.Short`: e.g., `Dec 20, 2012`
   - `DateStyle.Humanize`: e.g., `December 20, 2012`, `Today`, or `Yesterday`
+  - `DateStyle.HumanizeWithTime`: e.g., `December 20, 2012 00:00`, `Just now`, `3 minutes ago`, `Today`, or `Yesterday`
```
qq99: not `Today at 3:15pm` or `Yesterday at 2:23:am`?

packages/react-i18n/src/i18n.ts
```diff
@@ -237,22 +314,35 @@ function isTranslateOptions(
   return 'scope' in object;
 }
 
-function isSameMonthAndYear(source: Date, target: Date) {
-  return (
-    source.getFullYear() === target.getFullYear() &&
-    source.getMonth() === target.getMonth()
-  );
+enum TimeUnit {
```
rexmac: I plan to find a proper a home for this enum and the `isLessThan*` functions below. `@shopify/javascript-utilities/dates` perhaps?

patsissons: yea that probably makes sense

packages/react-i18n/README.md
```diff
@@ -70,6 +70,7 @@ The provided `i18n` object exposes many useful methods for internationalizing yo
   - `DateStyle.Long`: e.g., `Thursday, December 20, 2012`
   - `DateStyle.Short`: e.g., `Dec 20, 2012`
   - `DateStyle.Humanize`: e.g., `December 20, 2012`, `Today`, or `Yesterday`
+  - `DateStyle.HumanizeWithTime`: e.g., `December 20, 2012 00:00`, `Just now`, `3 minutes ago`, `4 hours ago`, `Today`, or `Yesterday`
```
nimzco: You forgot to replace `Yesterday` by `Yesterday at xx:xx`

nimzco: `December 20, 2012 00:00` should also be `Dec 20 at 00:00` or `Dec 20, 2012`

packages/react-i18n/src/i18n.ts
```diff
@@ -226,6 +230,79 @@ export default class I18n {
...
+        return this.translate('lessThanOneMinuteAgo');
+      }
+      if (isLessThanOneHourAgo(date)) {
+        // m minutes ago
+        return this.translate('lessThanOneHourAgo', {
+          minutes: getDateDiff(TimeUnit.Minute, date),
+        });
+      }
+      if (isLessThanOneDayAgo(date)) {
+        // hh:mm AM
```
kyledurand: lowercase am?

packages/react-i18n/src/constants.ts
```diff
@@ -22,6 +23,13 @@ export const dateStyle = {
     day: '2-digit',
     year: 'numeric',
   },
+  [DateStyle.HumanizeWithTime]: {
+    month: 'long',
+    day: '2-digit',
```
kyledurand: Can we make sure all days in this file use numeric vs 2-digit? Trying to get rid of those leading zeroes

packages/react-i18n/src/i18n.ts
```diff
@@ -173,9 +174,14 @@ export default class I18n {
     const {style = undefined, ...formatOptions} = options || {};
 
     if (style) {
-      return style === DateStyle.Humanize
-        ? this.humanizeDate(date, formatOptions)
-        : this.formatDate(date, {...formatOptions, ...dateStyle[style]});
+      if (style === DateStyle.HumanizeWithTime) {
```
kyledurand: I got rid of humanize with time in web. Humanize covers it all now: https://github.com/Shopify/web/blob/master/packages/@shopify/legacy-i18n/src/Intl.ts#L177

packages/react-i18n/src/test/i18n.test.ts
```diff
@@ -459,52 +459,153 @@ describe('I18n', () => {
       );
     });
 
-    it('formats a date using DateStyle.Humanize', () => {
-      const date = new Date('2012-12-20T00:00:00-00:00');
-      const i18n = new I18n(defaultTranslations, {
-        ...defaultDetails,
-        timezone,
+    describe('with DateSTyle.Humanize', () => {
```
beefchimi: `T` is capitalized. Should be `DateStyle`

packages/react-i18n/src/i18n.ts
```diff
@@ -214,16 +224,41 @@ export default class I18n {
...
-    if (isSameDate(today, date)) {
-      return this.translate('today');
+    const time = this.formatDate(date, {
+      ...options,
+      style: DateStyle.Time,
+    }).toLocaleLowerCase();
+
+    if (isLessThanOneMinuteAgo(date)) {
+      return this.translate('humanize.now');
+    } else if (isLessThanOneHourAgo(date)) {
```
owain: Each one of these `else if`s is a return so you don't need to make them `else if` they can be standalone early return `if` statements. This makes it slightly easier to parse and means you can move the `time` assignment further down so it's only assigned when needed.

calumptrck:
👋 I forked most of the `DateTime` component from `web` that this PR is based on, it's worth checking out some of the comments there which will be applicable here. https://github.com/Shopify/online-store-ui/pull/590

Also, once this is merged, we should be now be able to replace uses of `DateTime` in web with `i18n.formatDate(... {style: DateStyle.Humanize})`

rexmac:
Updated to use v2.4.1 of `@shopify/javascript-utilities` which includes the bug fixes to prevent false positives for future dates passed to the `isLessThan*` functions.

PR #406, opened 2018-11-13
Add react-effect

lemonmade:
This PR adds a new library, `@shopify/react-effect`. This library will be our way of having a single pass for all things that involve extracting some details from the server. I have a few libraries that can depend on this to make serialization and CSP stuff nicer, and we can also update our react-i18n library to use this same thing in order to reduce additional tree passes.

### DIFF
--- a/packages/react-effect/README.md
+++ b/packages/react-effect/README.md
@@ -3,7 +3,7 @@
 [![Build Status](https://travis-ci.org/Shopify/quilt.svg?branch=master)](https://travis-ci.org/Shopify/quilt)
 [![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](LICENSE.md) [![npm version](https://badge.fury.io/js/%40shopify%2Freact-effect.svg)](https://badge.fury.io/js/%40shopify%2Freact-effect.svg) [![npm bundle size (minified + gzip)](https://img.shields.io/bundlephobia/minzip/@shopify/react-effect.svg)](https://img.shields.io/bundlephobia/minzip/@shopify/react-effect.svg)
 
-A component and set of utilities for performing effects within a universal React app. This package gives you the tools you need to extract details from an application during server rendering using only a single pass.
+This package contains a component and set of utilities for performing multiple effects during one single pass of server rendering in a universal React application.
 
 ## Installation
 
@@ -17,7 +17,7 @@ $ yarn add @shopify/react-effect
 
 This package is largely built around a component, `Effect`. To set up an action to be performed, use the `perform` prop:
 
-```ts
+```tsx
 import {Effect} from '@shopify/react-effect';
 
 export default function MyComponent() {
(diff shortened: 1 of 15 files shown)


packages/react-effect/README.md
```diff
@@ -0,0 +1,121 @@
+# `@shopify/react-effect`
+
+[![Build Status](https://travis-ci.org/Shopify/quilt.svg?branch=master)](https://travis-ci.org/Shopify/quilt)
+[![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](LICENSE.md) [![npm version](https://badge.fury.io/js/%40shopify%2Freact-effect.svg)](https://badge.fury.io/js/%40shopify%2Freact-effect.svg) [![npm bundle size (minified + gzip)](https://img.shields.io/bundlephobia/minzip/@shopify/react-effect.svg)](https://img.shields.io/bundlephobia/minzip/@shopify/react-effect.svg)
+
+A component and set of utilities for performing effects within a universal React app. This package gives you the tools you need to extract details from an application during server rendering using only a single pass.
```
cartogram: I don't think you need give the example of extracting details. Could be combined to form the following one-liner. 

>This package contains a component and set of utilities for performing multiple effects during one single pass of server rendering in a universal React application.

packages/react-effect/README.md
```diff
@@ -0,0 +1,121 @@
...
+import {extract} from '@shopify/react-effect/server';
+
+async function app(ctx) {
+  const app = <App />;
+  await extract(app);
+  ctx.body = renderToString(app);
+}
+```
+
+You may optionally pass a second argument to this function: an array of symbols representing the categories to include in your processing of the tree (matching the `kind` prop on `Extract` components).
```
cartogram: Mine as well just add an example of the second argument.

packages/react-effect/src/Effect.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+interface Props {
+  kind?: symbol;
+  serverOnly?: boolean;
+  clientOnly?: boolean;
+  perform(): void;
+}
+
+export default class Effect extends React.PureComponent<Props>
+  implements Extractable {
```
cartogram: I am so not crazy about this line break that I would suggest moving the `export default Effect` out to the bottom.

packages/react-effect/src/test/Effect.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+  });
+
+  it('does not call perform() when extraction passes false', () => {
+    const spy = jest.fn();
+    const effect = mount(<Effect serverOnly perform={spy} />);
+    simulateExtractCall(effect, false);
+    expect(spy).not.toHaveBeenCalled();
+  });
+
+  it('does not call perform() when extraction passes a an array of symbols to include and there is no kind on the effect', () => {
```
cartogram: `a an` => `an`

packages/react-effect/src/Effect.tsx
lemonmade: Meh, just let prettier do its thing :P

PR #409, opened 2018-11-14
Add react-network

lemonmade:
This PR adds a new library, `react-network`. It uses the `network` and `react-effect` libraries I released today to allow setting details on the actual HTTP response. For right now, I just did a few things I know we need and would like to move in to app code: status code, redirect, and CSP. You can read the README for more details.

I didn't write tests for the components because there were a lot of them and they are all trivial... hope that's OK? :P


packages/react-network/src/components/FontSource.tsx
```diff
@@ -0,0 +1,11 @@
+import * as React from 'react';
+import {CspDirective} from '@shopify/network';
+import SetCspDirective from './SetCspDirective';
+
+interface Props {
+  sources: string | string[];
+}
+
+export default function FontSource({sources}: Props) {
+  return <SetCspDirective directive={CspDirective.FontSrc} value={sources} />;
```
nickjs: I don't see a ton of value in having a separate component for every CspDirective. Can we just do `<SetCspDirective directive={CspDirective.[...]Src}/>` where we need to set it? The `values` prop type appears to be the same for all of them.

packages/react-network/src/components/NetworkEffect.tsx
```diff
@@ -0,0 +1,25 @@
...
+import {Manager} from '../Manager';
+
+interface Props {
+  perform(manager: Manager): void;
+}
+
+export const EFFECT_ID = Symbol('network');
+
+export default function NetworkEffect({perform}: Props) {
+  if (typeof window !== 'undefined') {
```
nickjs: Is there a more explicit server check we can do? I worry about this breaking for example in test environments where the DOM gets mocked.

packages/react-network/src/components/WebrtcSource.tsx
```diff
@@ -0,0 +1,11 @@
+import * as React from 'react';
+import {CspDirective} from '@shopify/network';
+import SetCspDirective from './SetCspDirective';
+
+interface Props {
+  sources: string | string[];
+}
+
+export default function WebrtcSource({sources}: Props) {
```
nickjs: Capitalization should be `WebRtcSource`?

packages/react-network/src/server.ts
```diff
@@ -0,0 +1,74 @@
...
+    this.redirectUrl = url;
+  }
+
+  addStatusCode(statusCode: StatusCode) {
+    this.statusCodes.push(statusCode);
+  }
+
+  addCspDirective(directive: CspDirective, value: string | string[] | boolean) {
+    const normalizedValue = typeof value === 'string' ? [value] : value;
+    const newValue = Array.isArray(value)
```
nickjs: Don't think this check is necessary; between the typing and the line above `value` will always be an Array no?

packages/react-network/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+```bash
+$ yarn add @shopify/react-network
+```
+
+## Usage
+
+This package uses [`@shopify/react-effect`](https://github.com/Shopify/quilt/tree/master/packages/react-effect) to allow your application to communicate various HTTP-related details to the Node server doing React rendering. It also provides a utility function for easily applying these details to a Koa context object.
+
+To start, have your app accept a `Manager` object, and pass in a `ServerManager` object in your server render (on the client rehydration, you can safely omit any manager, and the library will simply no-opt all of the components discussed below):
```
nickjs: `no-opt` -> `no-op`

packages/react-network/README.md
```diff
@@ -0,0 +1,110 @@
...
+  const networkManager = new ServerManager();
+  const app = <App networkManager={networkManager} />;
+  await extract(app);
+
+  applyToContext(ctx, networkManager);
+  ctx.body = renderToString(app);
+}
+```
+
+> Note: You can selectively extract _only_ the translations by using the `EFFECT_ID` exported from `@shopify/react-i18n`, and using this as the second argument to `@shopify/react-effect`’s `extract()` as detailed in its documentation. Most consumers of this package will be fine with just the example above.
```
nickjs: Why is i18n relevant here?

packages/react-network/src/server.ts
```diff
@@ -0,0 +1,74 @@
...
+          : undefined,
+      csp: [...this.csp.entries()].reduce<{
+        [key: string]: string[] | string | boolean;
+      }>((csp, [directive, value]) => ({...csp, [directive]: value}), {}),
+      redirectUrl: this.redirectUrl,
+    };
+  }
+}
+
+export function applyToContext<T extends Context>(
```
nickjs: I'd like to see this broken out into a different file so that a consumer doesn't need to pull in Koa if they're not using Koa/this function.

packages/react-network/src/context.tsx
```diff
@@ -0,0 +1,21 @@
+// https://gist.github.com/tyscorp/24516f1fdb8974a637b075cc9167a690
```
nickjs: Not sure what this is referencing

packages/react-network/src/context.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+import * as React from 'react';
+import {Manager, NoopManager} from './manager';
+
+const {Consumer, Provider} = React.createContext<Manager>(new NoopManager());
```
nickjs: Why do we initialize the default value in the context when it gets overridden when the Provider is created below? Shouldn't line 13 be `{manager = new NoopManager(), children}: Props` or am I missing something?

packages/react-network/package.json
```diff
@@ -0,0 +1,36 @@
...
+    "url": "git+https://github.com/Shopify/quilt.git"
+  },
+  "bugs": {
+    "url": "https://github.com/shopify/quilt/issues"
+  },
+  "homepage": "https://github.com/Shopify/quilt/blob/master/packages/react-network/README.md",
+  "dependencies": {
+    "@shopify/network": "^1.0.0",
+    "@shopify/react-effect": "^1.0.0",
+    "@types/koa": "^2.0.46"
```
nickjs: I think this should be a peer dependency so it can be used by consumers not using Koa (since you went to the trouble of generalizing the API in the first place).

packages/react-network/package.json
```diff
@@ -0,0 +1,36 @@
...
+  "description": "A collection of components that allow you to set common HTTP headers from within your React application.",
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "scripts": {
+    "build": "tsc --p tsconfig.build.json",
+    "prepublishOnly": "yarn run build"
+  },
+  "publishConfig": {
+    "access": "public",
+    "@shopify:registry": "https://registry.npmjs.org"
```
GoodForOneFare: 🔥 the explicit registry (per [the guide](https://development.shopify.io/guides/package_cloud_nodejs/)).

packages/react-network/package.json
```diff
@@ -0,0 +1,36 @@
+{
```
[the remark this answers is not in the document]
GoodForOneFare: `network/package.json` could use this, too.

packages/react-network/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+Specifies a redirect location. `applyToContext` will call `ctx.redirect()` with the passed URL, and set the status code, if you pass the `status` prop.
+
+```tsx
+<Redirect url="/login">
+```
+
+### `<Status />`
+
+Specifies a status code. `applyToContext` will set `ctx.status` with the passed status code. If multiple status codes are set during the navigation of the tree, the highest one will be used. The example below illustrates how you can use this component to always return a 404 status code for a `NotFound` component:
```
GoodForOneFare: I had to look at the tests to figure out if "highest" means "numerically greatest" or "closest to the tree root".

"largest" also reads oddly to me.  So... umm... I'll try to think of a better suggestion 😬

packages/react-network/src/components/FontSource.tsx
lemonmade: There is a component for this, but IMO it is much nicer to provider more tailored to components. Some directives, like the `FooSource` ones, accept arrays of strings. Some, like `UpgradeInsecureRequests`, accept a boolean. A single component means they don't get any tips on what to pass in.

packages/react-network/src/components/NetworkEffect.tsx
lemonmade: Nope, this is the best we have. This is actually kind of a feature for tests, since we generally don't really care that any of these will do anything real in tests (it's basically built-in mocking). Jest does allow per-file setting of the environment though so it is easy enough to verify these as they would appear in Node.

packages/react-network/src/server.ts
lemonmade: No, `value` can still be `boolean` here.

packages/react-network/src/server.ts
lemonmade: This is already only available via the server entry point, and Koa is the only server library we really support. This also only bring in the typings for Koa, it does not bring in the library itself.

packages/react-network/src/context.tsx
lemonmade: lol whoops

packages/react-network/src/context.tsx
lemonmade: It's kind of a matter of preference. This just makes it so that, if anyone uses the consumer directly with no provider in the tree, they don't get an `undefined`, they get a `Manager` (it just doesn't do anything)

PR #413, opened 2018-11-16
Better React HTML/ Serializations

lemonmade:
This PR completely reworks our solution for our HTML/ Serialization stuff around this proposal: https://github.com/Shopify/web-foundation/blob/master/Proposals/02%20-%20Serialization%20in%20application%20code.md. To summarize:

* Merges `@shopify/react-serialize` with @shopify/react-html` and deprecates the serialize library
* Splits up the React HTML library between client and server entry points
* Completely refactors serialization — there is now a manager that can collect strongly-typed serializations from within application code and then apply them to the HTML document

The proposal has more rationale for why these changes are good. It is a big improvement architecturally IMO (less code split all over the place), but even more importantly, it lets us isolate infrequently-used features that need serialization to their respective components/ bundles, freeing up space in the main bundle. Here's an example of using this feature to support Redux within one route, without needing to wrap it around the whole app: https://github.com/lemonmade/perfect-app/blob/master/app/sections/Orphans/Orphans.tsx


lemonmade:
For XSS folks, don't worry too much about the `dangerouslySetInnerHTML` here, it is used on the result of React's `renderToString`, so everything that needs escaping is already escaped.

packages/react-html/README.md
```diff
@@ -12,112 +12,165 @@ $ yarn add @shopify/react-html
...
+  return (
+    <WithSerialized>
+      {translations => {
+        const manager = new Manager({locale}, translations);
+
+        return (
+          <>
+            <Provider manager={manager}>{children}</Provider>
+            <Serialize
+              data={() => ({
```
kaelig: Getting this error when linting the code:
> JSX props should not use arrow functions [react/jsx-no-bind]

packages/react-html/README.md
```diff
@@ -12,112 +12,165 @@ $ yarn add @shopify/react-html
...
-**data**
-Any serializable data that needs to be available from the DOM when the `deferredScripts` are run.
+const {Serialize, WithSerialized} = createSerializer<
+  ReturnType<Manager['extract']>
+>('i18n');
+
+export default function I18n({locale, children}: Props) {
+  return (
+    <WithSerialized>
+      {translations => {
```
kaelig:
```suggestion
      {(translations) => {
```

packages/react-html/README.md
```diff
@@ -12,112 +12,165 @@ $ yarn add @shopify/react-html
...
+
+##### Props
+
+The `<Html />` component accepts the following props:
+
+- `manager`: a `Manager` instance. When provided, the `Html` component will extract all the information from this object and place it in an appropriate place in the document.
+- `children`: the application. It will be rendered to a string and placed inside a `div` with an ID of `app`.
+- `locale`: the language to use for the HTML `lang` attribute.
+- `styles`: descriptors for any style tags you want to include in the HEAD of the document.
+- `scripts`: descriptors for any script tags you want to include in your document. All scripts passed to this property will be deferred by appending them to the end of the document. We encourage this as a default, although you may use `blockingScripts` for any scripts that must be included in the HEAD of the document.
```
TheMallen: I'd end this bullet at `We encourage this as a default`(and maybe link to an article or something on why), since `blockingScripts` is the next item in the list anyway.

packages/react-html/README.md
```diff
@@ -12,112 +12,165 @@ $ yarn add @shopify/react-html
...
+- `blockingScripts`: descriptors for any script tags you want to include in the HEAD of the document. These will block HTML parsing until they are evaluated, so use them carefully.
+- `headMarkup`: additional JSX to be embedded in the head of the document (after styles, but before blocking scripts).
+- `bodyMarkup`: additional JSX to be embedded in the body of the document (before serialization markup and deferred scripts).
```
TheMallen: Is this useful? Seems redundant with children. Not sure why we wouldn't just encourage
```
<Html>
<>
  <App />
  <otherMarkup />
</>
</Html>
```

packages/react-html/src/context.tsx
```diff
@@ -0,0 +1,19 @@
...
+import Manager from './manager';
+
+const {Provider, Consumer} = React.createContext<Manager>(new Manager());
+
+interface Props {
+  manager?: Manager;
+  children: React.ReactNode;
+}
+
+function HtmlManagerProvider({manager, children}: Props) {
```
TheMallen: `HtmlManagerProvider` feels so java.

~Also, isn't `HtmlManager` a misnomer? It doesn't manage HTML, it manages JS serializations, no?~
I still don't like the name though. :/

packages/react-html/src/manager.ts
```diff
@@ -0,0 +1,23 @@
...
+      this.serializations.set(id, data);
+    }
+  }
+
+  get<T>(id: string): T | undefined {
+    return this.serializations.get(id) as T | undefined;
+  }
+
+  extract() {
+    return [...this.serializations.entries()].map(([id, data]) => ({
```
TheMallen: Why do we need to clone the entries array? I think you can just drop the clone and do `.entries().map`

packages/react-html/src/serializer.tsx
```diff
@@ -0,0 +1,35 @@
...
+interface WithSerializedProps<T> {
+  children(data?: T): React.ReactNode;
+}
+
+export function createSerializer<T>(id: string) {
+  function Serialize({data}: SerializeProps<T>) {
+    return (
+      <Consumer>
+        {manager => (
+          <Effect
```
TheMallen: <3

packages/react-html/src/serializer.tsx
```diff
@@ -0,0 +1,35 @@
...
+            serverOnly
+            kind={EXTRACT_ID}
+            perform={() => manager.set(id, data())}
```
TheMallen: :0 did we disable the linting rule against this?!

packages/react-html/README.md
```diff
@@ -2,7 +2,7 @@
 
 [![Build Status](https://travis-ci.org/Shopify/quilt.svg?branch=master)](https://travis-ci.org/Shopify/quilt) [![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](LICENSE.md) [![npm version](https://badge.fury.io/js/%40shopify%2Freact-html.svg)](https://badge.fury.io/js/%40shopify%2Freact-html)
 
-A component to render your React app with no static HTML.
+A collection of utilities for constructing an HTML document that supports serialization, assets, and various `head` tags.
```
TheMallen: I like the taglines to be a bit briefer / punchier personally.
> React components for server rendering HTML documents and making serializing data between the client and server scalable.

Maybe even
> Say goodbye to react-tree-walker and JSON.parse.

kaelig: > React components for server rendering HTML documents and making serializing data between the client and server scalable.

❤️ 

> Say goodbye to react-tree-walker and JSON.parse.

That one would raise the bar quite a bit (requires prior knowledge), but maybe it could be used as additional context for experts? For example:

> React components for server rendering HTML documents and making serializing data between the client and server scalable. Useful as a replacement for [react-tree-walker](https://github.com/ctrlplusb/react-tree-walker) and `JSON.parse`.

packages/react-html/README.md
lemonmade: I think it's mostly about having the guaranteed position relative to some of the other props. It also has a good parallel with `headMarkup`, so I don't personally see a problem with including it.

packages/react-html/src/context.tsx
lemonmade: I know, could just call it `HtmlProvider` or something but w/e. It is exported from the package as just `Provider` so people can pick a name they like for it.

packages/react-html/src/manager.ts
lemonmade: `Map.prototype.entries` returns an iterator, not an array.

screen shot 2018-11-20 at 10 05 13 am

packages/react-html/src/serializer.tsx
lemonmade: I will disable it, there isn't really a better way to do it here.

packages/react-html/README.md
lemonmade: `react-extract` uses `react-tree-walker` under the hood so I don't think it's appropriate to say goodbye to it. I would prefer to leave it as-is since we will soon add title/ meta/ link tag support here, which make focusing on serialization a bit too limited. Happy to punch it up if we have an alternative that makes it clear this is intended for generally managing an HTML document.

packages/react-html/README.md
lemonmade: Prettier wants the existing one because we haven't moved Quilt over to the arrow parens style yet.

packages/react-html/README.md
lemonmade: We'll be turning this rule off some point soon, most developers learning React see (and want to use) arrow functions in JSX all over online examples/ docs.

lemonmade:
@kaelig thanks for the review. Will definitely add a migration guide to the README, but I think that the comprehensive example is already available in the `createSerializer()` section? Maybe more accurately, the `App code` section shows all that you would need to make this work (provider/ manager + serializations somewhere in the tree)

kaelig:
> the comprehensive example is already available in the createSerializer() section? 

This example shows how to use it with react-i18n, which may be very useful to some teams, but as someone who just discovered this package and wants to know how to use it, this is extra cognitive load that makes it complex to learn at a glance.

I'd really love to see a couple of pieces of code at the beginning of the README (or in an example) that show a basic/recommended usage of react-html while showcasing the main benefits to the user without having them read the entire readme (in other words: can it be dumbed down for non-experts like me?).

packages/react-html/README.md
```diff
@@ -166,11 +185,94 @@ The `<Html />` component accepts the following props:
...
 - `bodyMarkup`: additional JSX to be embedded in the body of the document (before serialization markup and deferred scripts).
 
 #### `<Serialize />`
 
 The Serialize component is a component that will take care of rendering a `script` tag with a serialized version of the `data` prop. You should not need this directly; prefer using the `createSerializer()` function within your app code instead.
+
+## Migrating from `@shopify/react-html@4.x`/ `@shopify/react-serialize@1.x` to `@shopify/react-html@5.x`
+
+The changes in this library from version 4.x to 5.x significantly changed the API. Notably, the new version takes over all the previous responsibilities of the `@shopify/react-serialize` library, and changes the API of the `HTML` component. This section summarizes the changes required to migrate.
+
```
kaelig: ❤️

packages/react-html/src/server/components/Html.tsx
```diff
@@ -84,36 +79,28 @@ export default function HTML({
...
   return (
-    <html lang="en" {...htmlAttributes}>
+    <html lang={locale} {...htmlAttributes}>
       <head>
         {helmet.title.toComponent()}
         {helmet.meta.toComponent()}
         {helmet.script.toComponent()}
         {helmet.link.toComponent()}
 
         {stylesMarkup}
```
kaelig: There are times where we'd want markup before styles. Should we simply rely on Helmet for that use-case?

packages/react-html/README.md
```diff
@@ -12,112 +12,267 @@ $ yarn add @shopify/react-html
...
+    ))
+  }
 />
 ```
+
+This is obviously much more work than it was previously. We recommend that, instead of using the method, you move to the patterns described in the [app code](#app-code) section of this document. The new pattern allows you to move your serializations beside the component that actually cares about them, and automatically inserts them into the HTML document.
+
+### `<Serializer />` is now `<Serialize />` and is exported from `@shopify/react-html/server`
+
+The `<Serializer />` component used to be imported from `@shopify/react-serialize`. It has now moved to `@shopify/react-html/server`, and has a slightly nicer name. Note that you should only need this component during migration to the new pattern of serializations living alongside your application code.
```
kaelig: > slightly nicer name

opinion 😋

packages/react-html/README.md
```diff
@@ -12,112 +12,267 @@ $ yarn add @shopify/react-html
...
+```ts
+import {getSerialized} from '@shopify/react-html';
+
+// before
+// was type {data: string}
+const data = getSerialized<string>('my-data');
+
+// after
+// has type string | undefined
+const data = getSerialized<string>('my-data);
```
kaelig: Missing quote:

```suggestion
const data = getSerialized<string>('my-data');
```

packages/react-html/src/server/components/Html.tsx
lemonmade: I think we'd rely on helmet, I am not sure what markup would go above styles other than links/ scripts/ meta tags?

packages/react-html/README.md
```diff
@@ -12,112 +12,267 @@ $ yarn add @shopify/react-html
...
-  integrity?: string;
-}
+```tsx
+createSerializer<string>('MyData');
 
-interface Browser {
-  userAgent: string;
-  supported: boolean;
-}
+// Would create the following components:
```
BPScott:
```suggestion
// Serialize and WithSerialized have the following typings:
```

I'm not sure if typings is the right word there :/

BPScott: Perhaps this should be in two codeblocks to reenforce the idea that these functions are the result of what you just called, instead of being something a user copy+pastes

packages/react-html/README.md
```diff
@@ -12,112 +12,267 @@ $ yarn add @shopify/react-html
...
+export default function I18n({locale, children}: Props) {
+  return (
+    <WithSerialized>
+      {translations => {
+        const manager = new Manager({locale}, translations);
+
+        return (
+          <>
+            <Provider manager={manager}>{children}</Provider>
+            <Serialize
```
BPScott: Should we render something that uses a loaded piece of data in here?

ATM it still feels a little contrived as we load some data but never use it.

packages/react-html/README.md
```diff
@@ -12,112 +12,267 @@ $ yarn add @shopify/react-html
...
+  children: React.ReactNode;
+}
 
-**scripts**
-Descriptors for any script tags you want to include in your document. All scripts passed to this property will be deferred by appending them to the end of the document. We encourage this as a default, although you may use `blockingScripts` for any scripts that must be included in the HEAD of the document.
+const {Serialize, WithSerialized} = createSerializer<
+  ReturnType<Manager['extract']>
+>('i18n');
+
+export default function I18n({locale, children}: Props) {
```
BPScott:
```suggestion
export default function MyComponentThatUsesTranslations({locale, children}: Props) {
```

Might be nice to make it clear that this is a component that uses i18n. On my first pass at reading this I wondered if this name had special meaning based on the imported library and it doesn't

packages/react-html/README.md
```diff
@@ -12,112 +12,267 @@ $ yarn add @shopify/react-html
...
-**blockingScripts**
-Descriptors for any script tags you want to include in the HEAD of the document. These will block HTML parsing until they are evaluated, so use them carefully.
+#### Utility functions
 
-**hideForInitialLoad**
-Sets the body contents to be hidden for the initial render. Use this when injecting stylesheets dynamically in development in order to prevent a flash of unstyled content.
+##### `showPage()`
 
-### Serializers
+This function encapsulates the logic for showing the page in development, where it is hidden on the initial render by default. You must call this function from your client entry point, usually right after hydrating your React app. It returns a promise that resolves after the document is guaranteed to be visible.
```
BPScott: If you *must* call this as part of the client entry point it doesn't really sound like a "Utility" (which makes me think "helper" and "optional").

Placing this example of how to set up your client entry point near the example of how to set up your server entry would be good to make it clear there are multiple steps.

I'd put those preliminary requirements at the start of the usage section, before talking about how to configure a component.

## Usage
### Configuring your Server Entry Point
### Configuring your Client Entry Pont
### Using createSerializer() within a component (but with a better name)

packages/react-html/README.md
```diff
@@ -12,112 +12,267 @@ $ yarn add @shopify/react-html
...
+/>;
+```
+
+### Server
+
+The server entrypoint allows you to easily construct an HTML document for the response. It also re-exports all of the components noted above.
+
+#### `<Html />` and `render()`
+
+The `<Html>` component serves as a top level wrapper for a React application, allowing you to avoid needing any kind of server-side template. The `render()` function creates a stringified version of the HTML document with an appropriate DOCTYPE.
```
BPScott: Death to needing to manually append doctypes 🎉 💜

packages/react-html/README.md
```diff
@@ -12,112 +12,267 @@ $ yarn add @shopify/react-html
...
+
+The component will automatically propagate any usage of the [`react-helmet` library](https://github.com/nfl/react-helmet) in your app’s content to manipulate the title or other top level HTML or HEAD attributes.
+
+##### Props
+
+The `<Html />` component accepts the following props:
+
+- `manager`: a `Manager` instance. When provided, the `Html` component will extract all the information from this object and place it in an appropriate place in the document.
+- `children`: the application. It will be rendered to a string and placed inside a `div` with an ID of `app`.
+- `locale`: the language to use for the HTML `lang` attribute.
```
BPScott: Why not call this prop `lang` so it matches the attribute that gets populated?

packages/react-html/README.md
```diff
@@ -12,112 +12,267 @@ $ yarn add @shopify/react-html
...
+- `styles`: descriptors for any style tags you want to include in the HEAD of the document.
+- `scripts`: descriptors for any script tags you want to include in your document. All scripts passed to this property will be deferred by appending them to the end of the document. We encourage this as a default because it improves the initial rendering performance of your page.
+- `blockingScripts`: descriptors for any script tags you want to include in the HEAD of the document. These will block HTML parsing until they are evaluated, so use them carefully.
+- `headMarkup`: additional JSX to be embedded in the head of the document (after styles, but before blocking scripts).
+- `bodyMarkup`: additional JSX to be embedded in the body of the document (before serialization markup and deferred scripts).
+
+#### `<Serialize />`
```
BPScott: If using this directly isn't needed as people should use createSerializer then why expose it?

packages/react-html/README.md
```diff
@@ -12,112 +12,267 @@ $ yarn add @shopify/react-html
...
+
+The Serialize component is a component that will take care of rendering a `script` tag with a serialized version of the `data` prop. You should not need this directly; prefer using the `createSerializer()` function within your app code instead.
+
+## Migrating from `@shopify/react-html@4.x`/ `@shopify/react-serialize@1.x` to `@shopify/react-html@5.x`
```
BPScott: I love migration notes, but I think they should be in a separate file. sewing-kit recently added a [migration-guide.md](https://github.com/Shopify/sewing-kit/blob/master/migration-guide.md) should we follow that?

packages/react-html/README.md
```diff
@@ -12,112 +12,267 @@ $ yarn add @shopify/react-html
...
+
+The changes in this library from version 4.x to 5.x significantly changed the API. Notably, the new version takes over all the previous responsibilities of the `@shopify/react-serialize` library, and changes the API of the `HTML` component. This section summarizes the changes required to migrate.
+
+### `<HTML />` is now `<Html />`, and exported from `/server` instead of the root package
```
BPScott:
```suggestion
### `<HTML />` is now `<Html />`, and is exported from `@shopify/react-html/server` instead of `@shopify/react-html`
```

packages/react-html/README.md
```diff
@@ -12,112 +12,267 @@ $ yarn add @shopify/react-html
...
+const html = <HTML />;
+
+// after
+import {Html} from '@shopify/react-html/server';
+const html = <Html />;
+```
+
+### `<Html />` `hideForInitialLoad` prop is gone, use `showPage()` to show the page in development
+
+`hideForInitialLoad` was used in development to prevent a flash of unstyled contents. Because almost every project ended up doing `hideForInitialLoad={process.env.NODE_ENV === 'development'}`, we made this the default. As a result, you will need to use a bundler that can transform `process.env.NODE_ENV` for the client, and you will need to make sure to unset the hidden styling on the document. The library exposes a function you can use to do this:
```
BPScott: Can you show a before and after, like in the prior example

packages/react-html/README.md
```diff
@@ -12,112 +12,267 @@ $ yarn add @shopify/react-html
...
-  hideForInitialLoad?: boolean;
-}
+The function returns a pair of components:
 
-interface Asset {
-  path: string;
-  integrity?: string;
-}
+```tsx
+createSerializer<string>('MyData');
```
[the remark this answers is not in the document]
amrocha: Agreed that the code right now is hard to follow, a working example would be appreciated here

packages/react-html/README.md
amrocha: Agreed that this belongs in a separate migration doc

packages/react-html/README.md
tzvipm: How about splitting it up a bit:

> A collection of utilities for constructing an HTML document.

Features:
* Serialization
* Assets
* `head` tags

packages/react-html/README.md
```diff
@@ -12,112 +12,267 @@ $ yarn add @shopify/react-html
 
 ## Usage
 
-The `<HTML>` component serves as a top level wrapper for a react application, allowing you to avoid needing any kind of server side template, in favor of purely using `reactDom.renderToString`.
+This PR exposes two entrypoints:
```
tzvipm: PR -> package 😄

packages/react-html/README.md
```diff
@@ -12,112 +12,267 @@ $ yarn add @shopify/react-html
...
 
-```javascript
-import * as React from 'react';
-import {renderToString} from 'react-dom/server';
+- `@shopify/react-html`: which contains code that can be used on server and client. Code in your `app` and `client` directories should import from this entrypoint.
```
tzvipm: keep in mind that this is for general consumption. Not all users will have an `app` or `client` directory.

tzvipm: Perhaps this:

> `@shopify/react-html`: which contains code that can be used on server and client. Code executed in a browser context (for example, an `app` or `client` directory) should import from this entrypoint.

packages/react-html/README.md
```diff
@@ -12,112 +12,267 @@ $ yarn add @shopify/react-html
...
+- `@shopify/react-html/server`: which contains all of `react-html`, but also includes some features that are not safe to import on the client. Code in your `server` directory should import from this entrypoint.
```
tzvipm: > ... which contains all of the aforementioned `react-html`, but also includes some features that are not safe to run in a browser context. Code in your `server` directory should import from this entrypoint.

packages/react-html/README.md
```diff
@@ -12,112 +12,267 @@ $ yarn add @shopify/react-html
...
+### App code
 
-export default (ctx, next) => {
-  // we have to prepend DOCTYPE to serve valid HTML
-  ctx.body = DOCTYPE + renderToString(
-    <HTML>
-      <MyApp />
-    </HTML>
-  );
+The main entry point of this library is primarily concerned with giving you the ability to [serialize and deserialize information](https://github.com/Shopify/web-foundation/blob/master/Proposals/02%20-%20Serialization%20in%20application%20code.md) from within your application code. There are three parts to this: a `Manager`, which stores and looks up serializations, a `Provider`, that distributes this manager throughout the application, and a function for generating type-safe serialization/ deserialization components.
```
tzvipm: This sounds a bit too technical to me. Have you considered including a brief summary of the proposal in the readme here? Also, perhaps introduce some sort of terminology for the `function for generating type-safe serialization/ deserialization components` or its output?

packages/react-html/src/server/components/Html.tsx
kaelig: You're right, thanks!

packages/react-html/README.md
lemonmade: The data is used to construct the manager? We'd be digging into a completely separate library's usage to show how i18n is used, Kaelig's comment was that this was already too confusing being tied to a different library :/

packages/react-html/README.md
lemonmade: It is not a component that uses i18n, it is a component that provides i18n to the rest of the app with the help of the i18n package.

packages/react-html/README.md
lemonmade: It's an interesting idea, I chose it because the variable you are very likely to have in scope is `locale`, as this is used for the i18n library. IMO the `lang` attribute is more of an implementation detail of the HTML component.

packages/react-html/README.md
lemonmade: For incremental migration.

packages/react-html/README.md
lemonmade: I think it is fine to leave as-is, only Node apps can currently use this. Will leave a note to that effect.

lemonmade:
Thanks all for the comments! I've reworked the docs a bit to focus in on the three areas this is used (server entry, client entry, app code), and it reads much better. If we have additional improvements to the docs, I think they can be done later on.

packages/react-html/README.md
```diff
@@ -12,112 +12,218 @@ $ yarn add @shopify/react-html
...
-  hideForInitialLoad?: boolean;
+function App({htmlManager}: {htmlManager: Manager}) {
+  return <Provider manager={htmlManager}>Hello world!</Provider>;
 }
 
-interface Asset {
-  path: string;
-  integrity?: string;
-}
+// Somewhere in your server
```
BPScott: Split this into two code blocks to make it clearer that they're two files

PR #415, opened 2018-11-19
Don’t serialize fallback translations

lemonmade:
This PR is a more generalized fix to the "don't serialize the fallbacks" that we did in Web. It tells the I18n manager what locale is the fallback locale, so it knows that it can just synchronously resolve the fallback when that locale is requested. It then changes the translations that get extracted by the app to only be those that were resolved async, since that's anything resolve sync will already be part of the bundle on the client.

### DIFF
--- a/packages/react-i18n/src/test/manager.test.ts
+++ b/packages/react-i18n/src/test/manager.test.ts
@@ -602,7 +602,7 @@ describe('Manager', () => {
 
       const translationsByID = manager.extract();
       expect(Object.keys(translationsByID)).toBeArrayOfUniqueItems();
-      // @ts-ignore (Object.values)
+
       const translations = Object.values(translationsByID);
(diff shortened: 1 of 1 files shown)


packages/react-i18n/src/test/manager.test.ts
```diff
@@ -612,6 +629,26 @@ describe('Manager', () => {
...
+      // Connected, but not resolved
+      await manager
+        .connect(
+          connection,
+          noop,
+        )
+        .resolve();
+
+      const translationsByID = manager.extract();
+      // @ts-ignore (Object.values)
```
TheMallen: why this?

lemonmade: It was here before but I could get rid of it without issue /shrug

PR #416, opened 2018-11-20
[koa-shopify-auth] [WIP] ITP 2.1

tylerball:
## :tophat: instructions
1. checkout this branch locally and `yarn link`
1. you can follow the [app_builder instructions](https://gist.github.com/tylerball/5715015e213af21e8eeac8806c1d92d6) to get an app set up quick.
1. in your project `yarn link @shopify/koa-shopify-auth`

## TODO
- [ ] update illustrations
- [ ] clean up names of view methods "create" doesn't seem right.
- [x] tests for template views
- [x] fence off mobile and POS user agents.
- [x] figure out how to get and pass app name through to templates.
- [ ] tests for top level redirect
- [ ] client side tests


packages/koa-shopify-auth/client/enable-cookies.js
```diff
@@ -0,0 +1,45 @@
+(function() {
+  function setCookieAndRedirect() {
+    document.cookie = 'shopify.granted_storage_access=true';
```
francinen: I just thought of this question now, but should have thought of this earlier: Should the cookies we're setting for both the ITP 2.0 and ITP 2.1 workflows be unique to the app being authenticated (e.g., by adding the app ID to the cookie name)? This question also applies to the `shopify_app` gem

cc @ragalie

francinen: Nevermind, I was having a brain fart moment. Talked it through with Mike and this shouldn't be an issue. Sorry!

francinen:
Hey @tylerball! Does this PR only implement the solution for ITP 2.0? I'm only seeing the `Enable cookies` view, which is meant for the version of ITP that was released in September and only works for Safari 12.0.

tylerball:
@TheMallen not entirely, I'm still working on tests but thanks for the feedback. I'm trying to cover as much as I can in `request-storage.js` and the rest of the client JS without having to actually run anything in a browser environment.

MasjZam:
We're looking to send out some [partner comms](https://github.com/Shopify/email-marketing/issues/1444) warning them of the upcoming changes for Safari 12.1+ users with ITP 2.1

Was wondering if we have a timeline on this PR so we can include it in the comms as well

lemonmade:
@tylerball can this either be updated or closed?

ayronshopify:
Closing this in favour of https://github.com/Shopify/quilt/pull/1413/files

PR #424, opened 2018-11-29
Allow keyboard shortcuts for alternate modifier combinations

beefchimi:
In the app I am building, I have a number of keyboard shortcuts I need to map to actions.

**To give a quick example:**
1. `undo` most recent change
2. `redo` most recent change

I need keyboard modifiers to accommodate both Mac and Windows users.

> **Note:** For those curious, `Meta` maps to the "Command" key.

**Undo**
Windows: `Control + z`
Mac: `Meta + z`

**This allows us to do:**

```
interface KeyboardShortcuts {
  held: HeldKey;
  ordered: Key[];
}

const UNDO_KEY_PROPS: KeyboardShortcuts = {
  held: [['Meta'], ['Control']],
  ordered: ['z'],
};

<Shortcut {...UNDO_KEY_PROPS} onMatch={this.handleUndo} />
```

The above code means I can execute the `Shortcut` by pressing either `Control + z` or `Meta + z`.

This proposed change would not break anything, as a single "modifier group" is still supported _(a single array vs a nested array)_.

The code I have committed so far does in fact work! But... I am encountering some TypeScript pains... so, I have not moved forward with writing any tests yet. Would love it if I could get some help figuring out the right way to author this code.

### DIFF
--- a/packages/react-shortcuts/src/ShortcutManager/ShortcutManager.tsx
+++ b/packages/react-shortcuts/src/ShortcutManager/ShortcutManager.tsx
@@ -1,13 +1,13 @@
-import Key, {ModifierKey} from '../keys';
+import Key, {HeldKey, ModifierKey} from '../keys';
 
 const ON_MATCH_DELAY = 500;
 
 export interface Data {
   node: HTMLElement | null | undefined;
   ordered: Key[];
-  held?: ModifierKey[];
+  held?: HeldKey;
   ignoreInput: boolean;
-  onMatch(matched: {ordered: Key[]; held?: ModifierKey[]}): void;
+  onMatch(matched: {ordered: Key[]; held?: HeldKey}): void;
   allowDefault: boolean;
 }
 
@@ -61,6 +61,20 @@ export default class ShortcutManager {
     }
   };
 
+  private allModifiersAreHeld(heldKeys: HeldKey, event: KeyboardEvent) {
+    function hasKeyGroups(groups: HeldKey): groups is HeldKey {
+      return groups.every(key => Array.isArray(key));
+    }
+
+    function keyGroupIsHeld(keyGroup: ModifierKey[]) {
+      return keyGroup.every((key: ModifierKey) => event.getModifierState(key));
+    }
+
+    return hasKeyGroups(heldKeys)
+      ? heldKeys.some(keyGroupIsHeld)
+      : keyGroupIsHeld(heldKeys);
+  }
+
   private updateMatchingShortcuts(event: KeyboardEvent) {
     const shortcuts =
       this.shortcutsMatched.length > 0 ? this.shortcutsMatched : this.shortcuts;
@@ -71,7 +85,7 @@ export default class ShortcutManager {
           return false;
         }
 
-        if (held && !held.every(key => event.getModifierState(key))) {
+        if (held && !this.allModifiersAreHeld(held, event)) {
           return false;
         }
 
@@ -116,7 +130,8 @@ export default class ShortcutManager {
 
 function isFocusedInput() {
   const target = document.activeElement;
-  if (target.tagName == null) {
+
+  if (target == null || target.tagName == null) {
     return false;
   }
 
--- a/packages/react-shortcuts/README.md
+++ b/packages/react-shortcuts/README.md
@@ -45,9 +45,9 @@ export interface Props {
   */
   ordered: Key[];
   /*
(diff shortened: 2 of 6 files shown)


TheMallen:
Seems like you got some type issues going on. This make sense to me as an API, if we can get it working with the other types / add some tests :)
image

packages/react-shortcuts/src/ShortcutManager/ShortcutManager.tsx
```diff
@@ -61,6 +61,22 @@ export default class ShortcutManager {
     }
   };
 
+  private allModifiersAreHeld = (heldKeys: HeldKey, event: KeyboardEvent) => {
+    // eslint-disable-next-line no-warning-comments
+    // TODO: TypeScript complains that `keyGroupIsHeld(heldKeys)`
+    // could still contain `ModifierKey[]`, which is incorrect...
+    // Until this can be resolved, we need to set type to `any` instead of `ModiferKey[]`
+    function keyGroupIsHeld(keyGroup: any) {
```
beefchimi: If I leave this as `ModiferKey[]`, line `77` complains that `heldKeys` could still contain `ModifierKey[]`... which, should not be try based on the `hasKeyGroups` ternary.

Not sure how to solve this...

owain:
Can you currently do the following to achieve the same? Is this just cutting down on the verboseness?

```tsx
<Shortcut
    held={['Control']}
    ordered={['z']}
    onMatch={() => console.log('undo Windows!')}
/>
<Shortcut
    held={['Meta']}
    ordered={['z']}
    onMatch={() => console.log('undo Mac!')}
/>
<Shortcut
    held={['Control']}
    ordered={['y']}
    onMatch={() => console.log('redo Windows!')}
/>
<Shortcut
    held={['Meta', 'Shft']}
    ordered={['z']}
    onMatch={() => console.log('redo Mac!')}
/>
```

packages/react-shortcuts/src/ShortcutManager/ShortcutManager.tsx
TheMallen: You need to do a custom [typeguard](https://basarat.gitbooks.io/typescript/docs/types/typeGuard.html#user-defined-type-guards).

```
   function hasKeyGroups(list: ModifierKey[]): list is WhatevertheNestedTypeIs {
      return heldKeys.every(key => Array.isArray(key))
   }
```

packages/react-shortcuts/src/ShortcutManager/ShortcutManager.tsx
```diff
@@ -61,6 +61,20 @@ export default class ShortcutManager {
...
+    function hasKeyGroups(groups: HeldKey): groups is HeldKey {
+      return groups.every(key => Array.isArray(key));
+    }
+
+    function keyGroupIsHeld(keyGroup: ModifierKey[]) {
+      return keyGroup.every((key: ModifierKey) => event.getModifierState(key));
+    }
+
+    return hasKeyGroups(heldKeys)
+      ? heldKeys.some(keyGroupIsHeld)
```
owain: If this is the case does it make more sense to be strict and error out?

beefchimi: What do you mean "error out"? Both options are valid.

owain: Ignore, I misread this logic 👍

9at8:
This PR means that now two shortcuts exist on both the Mac and Windows. As in, you could Undo on a mac using either `Meta + z` or `Control + z`. Is this intended? Should we detect the OS and depending on it, only subscribe to a single shortcut?

9at8:
I was thinking about using this: https://developer.mozilla.org/en-US/docs/Web/API/NavigatorID/platform haha and make the API something like:

```jsx
<Shortcut
  held={['Control']}
  ordered={['a']}
  alternate={{
    osx: {
      held: ['Meta'],
    }
  }}
/>

<Shortcut
  held={['Control']}
  ordered={['y']}
  alternate={{
    osx: {
      held: ['Meta', 'Shift'],
      ordered: ['z'],
    }
  }}
/>
```

Thoughts?

beefchimi:
@9at8 I don't have opinions on whether or not your change is better than the existing solution. If nothing else, it prevents 2 Shortcuts from executing on the same OS - so that is good.

I have not looked much into the `platform` API you referenced. How reliable is it?

PR #425, opened 2018-11-30
Add public registry

ismail-syed:
Adds back public registries in package.json - [context](https://github.com/Shopify/shipit-engine/issues/846)


GoodForOneFare:
Pretty sure one of my driveby PR comments is the origin of this problem 🤦‍♂️   Thanks for investigating, and fixing.

PR #427, opened 2018-12-03
Add support for title, meta, and link tags to @shopify/react-html

lemonmade:
This PR adds support for a variety of additional special HTML tags to react-html. In doing so, it removes support for `react-helmet`, which is good since we recently identified issues with that library and asynchronously-rendered applications.

The pattern for updating title/ meta/ link tags is a little naive (it just queues all the updates for the next animation frame), but I think it will suffice given that this is not the most common case in our apps.


packages/react-html/src/components/Meta.tsx
```diff
@@ -0,0 +1,39 @@
...
+      this.removeMeta();
+    }
+  }
+
+  render() {
+    return (
+      <Consumer>
+        {manager => (
+          <Effect
+            key={JSON.stringify(this.props)}
```
lemonmade: This is a bit lazy, but it basically just makes it so we don't have to care about the props changing case (since the component will be unmounted and remounted, re-executing `perform` as a result.

packages/react-html/src/components/Link.tsx
```diff
@@ -0,0 +1,37 @@
...
+  render() {
+    return (
+      <Consumer>
+        {manager => (
+          <Effect
+            key={JSON.stringify(this.props)}
+            kind={EFFECT_ID}
+            perform={() => {
+              if (this.removeLink) {
+                this.removeLink();
```
TheMallen: I get that this imperative manager api mostly exists to support the declarative api, but I wonder if it might be nicer to have the manager api also declarative? 

```
<Consumer>
  {(manager) => {
    const {Head} = manager;
    return (
      <Head>
        <link />
      </Head>
    );
 }
</Consumer>
```

This way when / if consumers try to use the manager directly for this sort of thing (I'm sure there will be cases where this has to happen for some reason or another) they still get a nice imperative api without having to use `Effect` or manage callbacks.

lemonmade: Talked IRL, since `Consumer` is not really exposed publicly I think it is OK to keep the API as-is.

tzvipm:
I'm really loving the direction this library is going <3

packages/react-shortcuts/src/Shortcut/Shortcut.tsx
```diff
@@ -17,9 +17,9 @@ export interface Subscription {
 
 export default function Shortcut(props: Props) {
   return (
-    <Consumer>
+    <Context.Consumer>
```
tzvipm: we need to fix the imports here, right?

PR #431, opened 2018-12-05
✨ add a tophat script

TheMallen:
This PR adds a tophat script to quilt, similar to the one we use in some other libraries.

image

image


bin/tophat.js
```diff
@@ -0,0 +1,133 @@
...
+
+process.on('unhandledRejection', error => {
+  process.stderr.write(`${error.stack ? error.stack : error}`, () => {
+    process.exit(1);
+  });
+});
+
+yargs
+  .usage('Usage: $0 $1 <command> [options]')
+  .command({
```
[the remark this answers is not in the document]
TheMallen: Updated it to use `*`, so no positional argument is needed.

bin/watcher.js
```diff
@@ -0,0 +1,104 @@
+/* eslint-disable no-console */
+const fs = require('fs-extra');
+const {join, basename} = require('path');
+
+function isCompiling(dataStr) {
+  return dataStr.match(
+    /File change detected. Starting incremental compilation/,
```
tzvipm: Is there a better way for us to determine this?

TheMallen: not that I know of /shrug

PR #434, opened 2018-12-06
🚨 fix inconsistent component state update

samlanning:
Not sure if you accept these kinds of contributions, but thought i'd try in any case. This addresses a potential inconsistent state update.

React component state updates using `setState` may asynchronously update `this.props` and `this.state`, thus it is not safe to use either of the two when calculating the new state passed to `setState`, and instead the callback-based variant should be used instead. ([details here](https://lgtm.com/rules/1819283066/)).

These were all found on LGTM.com, and there are [a few other alerts](https://lgtm.com/projects/g/Shopify/quilt/alerts) for this project too.

*(Full disclosure: I'm part of the team behind LGTM.com)*

### DIFF
--- a/packages/react-form-state/src/FormState.ts
+++ b/packages/react-form-state/src/FormState.ts
@@ -199,7 +199,7 @@ export default class FormState<
 
   @bind()
   private reset() {
-    this.setState((state, props) => createFormState(props.initialValues));
+    this.setState((_state, props) => createFormState(props.initialValues));
   }
 
   @memoize()
--- a/packages/react-form-state/src/FormState.ts
+++ b/packages/react-form-state/src/FormState.ts
@@ -199,7 +199,7 @@ export default class FormState<
 
   @bind()
   private reset() {
-    this.setState(createFormState(this.props.initialValues));
+    this.setState((state, props) => createFormState(props.initialValues));
   }
 
   @memoize()


packages/react-form-state/src/FormState.ts
```diff
@@ -199,7 +199,7 @@ export default class FormState<
 
   @bind()
   private reset() {
-    this.setState(createFormState(this.props.initialValues));
+    this.setState((state, props) => createFormState(props.initialValues));
```
TheMallen: `state` -> `_state` should satisfy our linting.

TheMallen:
> The variable `set` is created as a result of the assignment, and is not used at all in that function, so that line could effectively be removed. It may be that the comment is confusing though?

There is no variable named `set` after this line though, so there is nothing to be removed. This is [aliased destructuring with a default](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment), which is an understandable false positive to have but is certainly a false positive.

Simple example illustrating what I mean:
image

Regardless, this PR is probably not the right place to discuss that. Feel free to email me or DM me on twitter or something if you want to talk about that more. :)

TheMallen:
Went ahead and merged, I'll release this soon. Thanks so much! :)

samlanning:
> There is no variable named set after this line though, so there is nothing to be removed. This is aliased destructuring with a default, which is an understandable false positive to have but is certainly a false positive.

Ah you're totally right. I was confusing the name of the property from the destructed object with the new variable name. I've opened an issue to address this here: https://github.com/Semmle/ql/issues/664

And thanks for merging! :)

PR #435, opened 2018-12-07
✨ Disable all fields when form is submitting, add `disableOnSubmit`

macdonaldr93:
Part of this Hack Days project: https://hackdays.shopify.io/projects/11630

This PR adds the `disableOnSubmit` prop to `<FormState>`. This prop allows you disable all your fields while submission is happening and prevent users from accidentally editing fields further.

Rational around this... when the form submission is taking time, users can make changes to fields. This leads to two different problems: 

1. there's ambiguity around which value was actually persisted
2. the dirty flag will not get reset after submission

As a best practice, we're proposing form fields should be disabled. As a breaking change this could be default behaviour in the future but for now we're introducing it as an opt-in prop.

Lastly, this maps 1-to-1 with Polaris form components props so developers don't have to keep track of disabling fields themselves.

kapture 2018-12-07 at 15 20 35

🎩 instructions
clone branch
yarn install && yarn build
cd packages/react-form-state
yarn link
go to your own project
yarn link @shopify/react-form-state
try it out :)

cc @atesgoral

### DIFF
--- a/packages/react-form-state/src/FormState.ts
+++ b/packages/react-form-state/src/FormState.ts
@@ -50,6 +50,7 @@ interface Props<Fields> {
   validators?: Partial<ValidatorDictionary<Fields>>;
   onSubmit?: SubmitHandler<Fields>;
   validateOnSubmit?: boolean;
+  disableOnSubmit?: boolean;
   onInitialValuesChange?: 'reset-all' | 'reset-where-changed' | 'ignore';
   externalErrors?: RemoteError[];
 }
@@ -195,7 +196,7 @@ export default class FormState<
   }
 
   private submit = async (event?: Event) => {
-    const {onSubmit, validateOnSubmit} = this.props;
+    const {onSubmit, disableOnSubmit, validateOnSubmit} = this.props;
     const {formData} = this;
 
     if (!this.mounted) {
@@ -212,10 +213,18 @@ export default class FormState<
 
     this.setState({submitting: true});
 
+    if (disableOnSubmit) {
+      this.disableFields();
+    }
+
     if (validateOnSubmit) {
       await this.validateForm();
 
       if (this.hasClientErrors) {
+        if (disableOnSubmit) {
+          this.enableFields();
+        }
+
         this.setState({submitting: false});
         return;
       }
@@ -232,6 +241,10 @@ export default class FormState<
       this.setState({submitting: false});
     } else {
       this.setState({submitting: false, errors});
+    }
+
+    if (disableOnSubmit) {
+      this.enableFields();
     }
   };
 
@@ -390,6 +403,32 @@ export default class FormState<
       errors,
       fields: fieldsWithErrors(fields, [...errors, ...externalErrors]),
     }));
+  }
+
+  private disableFields() {
+    this.setState(({fields}: State<Fields>) => {
+      return {
+        fields: mapObject(fields, field => {
+          return {
+            ...field,
+            disabled: true,
+          };
+        }),
+      };
+    });
+  }
+
+  private enableFields() {
+    this.setState(({fields}: State<Fields>) => {
+      return {
+        fields: mapObject(fields, field => {
+          return {
+            ...field,
+            disabled: undefined,
+          };
+        }),
+      };
+    });
   }
 }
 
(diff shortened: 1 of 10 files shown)


packages/react-form-state/src/FormState.ts
```diff
@@ -371,6 +384,32 @@ export default class FormState<
...
+          return {
+            ...field,
+            disabled: true,
+          };
+        }),
+      };
+    });
+  }
+
+  private unsetDisabledFields() {
```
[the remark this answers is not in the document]
TheMallen: `enableFields`?

michelleyschen:
@macdonaldr93 your explanation makes sense to me, however, the approach I am a un-sure about.

I remember disabling the submit button is actually a bad pattern in a11y cases. And this approach remind me of something very similar.

Before we merged this into form-state and suggest everyone to use this pattern, I am going to post this in #polaris to get some suggestion from the team on what they would suggest.

michelleyschen:
Had a conversation with @dpersing and @solonaarmstrong in polaris channel https://shopify.slack.com/archives/C4Y8N30KD/p1544461278106500

Generally it is ok accessibility wise to disable form fields during submitting, as long as focus management are being consider during submission / after submission. (ie. after submission, the focus should go to the error banner, or the first error field if the banner does not exist.)

An https://github.com/Shopify/polaris-ux issue will be open to explore the form submission pattern in the future. Which will determined if this `disableOnSubmit` behavbious should be default going forward.

Since `disableOnSubmit` is optional for now. I see no problem merging it.

dpersing:
Here's the `polaris-ux` issue @michelleyschen mentioned: https://github.com/Shopify/polaris-ux/issues/221. Please feel free to add and weigh in.

TheMallen:
@macdonaldr93 I'm ok with it for now since it's behind a prop and we aren't at 1.0 yet. If the issue resolves in a way that looks like it warrants this in general we can make it a default behaviour, if not we can leave it or deprecate it if we want.

macdonaldr93:
I need to update Nested and List and add docs! Working on it 👍

packages/react-form-state/src/components/List.tsx
```diff
@@ -53,6 +55,7 @@ export default class List<Fields> extends React.Component<
             initialValue: initialFieldValue,
             dirty: value !== initialFieldValue,
             error: get(error, [index, fieldPath]),
+            disabled,
```
macdonaldr93: This seems a bit weird to me but I don't want to change `disabled` to take an object with the fieldPath... Basically this means that when the disabled prop gets set on the parent, it'll always disable all the children. You can't selectively disable children unless you manage that state yourself...

This is fine for the `disableOnSubmit` feature... but I just want to make sure this isn't going to be a problem later on.

cc @michelleyschen @TheMallen

packages/react-form-state/src/FormState.ts
```diff
@@ -191,10 +192,18 @@ export default class FormState<
...
 
+    if (disableOnSubmit) {
+      this.disableFields();
+    }
+
     if (validateOnSubmit) {
       await this.validateForm();
 
       if (this.hasClientErrors) {
+        if (disableOnSubmit) {
```
michelleyschen: the place where `disableFields` / `enableFields` are called seems to mirror where `submitting` state is changed.
maybe those method can be called in `componentDidUpdate` instead?

packages/react-form-state/src/components/List.tsx
michelleyschen: I think this is fine for now, since we have things like `dirty` that also has value instead of object.
Defin' something we want to change for v2 thou.

TheMallen: Yeah this is only a problem if we update it selectively for some, which we don't do now. Michelle is right that we need to change a lot of how nested stuff works in the future anyway.

TheMallen:
@macdonaldr93 any word on whether Polaris is aligned on this / if we are going forward with this?

macdonaldr93:
I'll follow up with them next week 👍 . Our team has been on this mobile performance sprint so I needed to focus all my energy on that for the past two weeks.

macdonaldr93:
Library is deprecated... Closing.

PR #436, opened 2018-12-10
Add sewing-kit-koa

lemonmade:
This PR adds a package that Koa apps can use to automatically read details from Sewing Kit. It basically extracts the work done in web, plus the ability to have the middleware handle all asset serving as well, which I needed to simplify some of the e2e testing stuff I am working on. This gives us the same level of integration we get from the Rails sewing kit gem in terms of abstracting away the way Sewing Kit stores assets, so should be much easier to implement >1 client bundle, different manifest shapes, consolidating the async asset stuff, and more.

### DIFF
--- a/packages/sewing-kit-koa/src/assets.ts
+++ b/packages/sewing-kit-koa/src/assets.ts
@@ -1,0 +1,88 @@
+import {join} from 'path';
+import {readJson} from 'fs-extra';
+import appRoot from 'app-root-path';
+
+export interface Asset {
+  path: string;
+  integrity?: string;
+}
+
+interface Entrypoint {
+  js: Asset[];
+  css: Asset[];
+}
+
+interface AssetList {
+  entrypoints: {[key: string]: Entrypoint};
+}
+
+interface Options {
+  assetHost: string;
+}
+
+export default class Assets {
+  assetHost: string;
+  private resolvedAssetList?: AssetList;
+
+  constructor({assetHost}: Options) {
+    this.assetHost = assetHost;
+  }
+
+  async scripts({name = 'main'} = {}) {
+    const {js} = getAssetsForEntrypoint(name, await this.getAssetList());
+
+    const scripts =
+      // Sewing Kit does not currently include the vendor DLL in its asset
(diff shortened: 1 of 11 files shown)


packages/sewing-kit-koa/src/middleware.ts
```diff
@@ -0,0 +1,53 @@
...
+  // http://localhost:8080/webpack/assets/. When running in `serveAssets`
+  // mode (the application server also serves the assets), we default to
+  // assuming they have set the asset endpoint to be under /assets. In order
+  // to use this feature, a developer would need to add the following to the
+  // Sewing Kit config that built the assets:
+  //
+  // {
+  //   plugins: [plugins.cdn('/assets/')],
+  // }
+  return serveAssets ? '/assets/' : 'http://localhost:8080/webpack/assets/';
```
TheMallen: I feel like we should have a `peerDependency` on sewing-kit and pull these types of literals from there (expose them if they aren't currently)

packages/sewing-kit-koa/src/assets.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  async scripts({name = 'main'} = {}) {
+    const {js} = getAssetsForEntrypoint(name, await this.getAssetList());
+
+    const scripts =
+      // Sewing Kit does not currently include the vendor DLL in its asset
+      // manifest, so we manually add it here (it only exists in dev).
+      // eslint-disable-next-line no-process-env
+      process.env.NODE_ENV === 'development'
+        ? [{path: `${this.assetHost}dll/vendor.js`}, ...js]
```
TheMallen: I don't like that we are just hardcoding all of these. If sewing-kit ever changes we have nothing that will let us know this is broken other than just knowing.

lemonmade:
@TheMallen I agree that the hardcoded paths suck. IMO, we need to extract the "configuration" part of Sewing Kit from all the other stuff, so that a tool like this could get at it without too much overhead. Peer dependencies on Sewing Kit itself probably don't make much sense since it isn't really a production dependency, whereas this is.

PR #439, opened 2018-12-12
🎉 initialize react-traverse-nodes package

tzvipm:
# Infuse your react tree with data

Part of https://github.com/Shopify/web/issues/5613


packages/react-html/src/server/components/Html.tsx
```diff
@@ -33,7 +33,7 @@ export default function Html({
   headMarkup = null,
   bodyMarkup = null,
 }: Props) {
-  const markup = renderToString(children);
+  const markup = renderToString(<>{children}</>);
```
tzvipm: this was necessary, because `renderToString` expects a single `ReactElement`, but the `Html` component allows any `ReactNode` including multiple or none.

tzvipm:
Closing this as it's stale and I feel it's the wrong approach. The approach taken by the apollo team usign repetitive rendering seems more effective.

PR #441, opened 2018-12-14
Support consolidated manifest

lemonmade:
Thus PR adds support to `sewing-kit-koa` for the new, multi-build client setup that I'm adding in https://github.com/Shopify/sewing-kit/pull/1096. It basically just accepts the user agent, compares it against the new `browsers` field alongside each manifest, and then chooses the right asset from that manifest.

### DIFF
--- a/packages/sewing-kit-koa/src/assets.ts
+++ b/packages/sewing-kit-koa/src/assets.ts
@@ -1,5 +1,6 @@
 import {join} from 'path';
 import {readJson} from 'fs-extra';
+import {matchesUA} from 'browserslist-useragent';
 import appRoot from 'app-root-path';
 
 export interface Asset {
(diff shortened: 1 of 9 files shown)


packages/sewing-kit-koa/src/assets.ts
```diff
@@ -43,21 +55,61 @@ export default class Assets {
...
+    //
+    // 1. If there is no user agent, use the "last" manifest, which is the
+    // least restrictive set of browsers.
+    // 2. If there is only one manifest, use it, regardless of how well it
+    // matches the user agent.
+    // 3. If there is a user agent, find the first manifest where the
+    // browsers it was compiled for matches the user agent, or where there
+    // is no browser restriction on the bundle.
+    // 4. If no matching manifests are found, fall back to the last manifest.
+    if (userAgent == null) {
```
TheMallen: I think early returns would make this less noisy /shrug

packages/sewing-kit-koa/src/assets.ts
```diff
@@ -43,21 +55,61 @@ export default class Assets {
...
+      this.resolvedManifest =
+        consolidatedManifest[consolidatedManifest.length - 1].manifest;
```
TheMallen: can we pull `consolidatedManifest[consolidatedManifest.length -1]` into a semantically named variable?

packages/sewing-kit-koa/src/tests/assets.test.ts
```diff
@@ -125,27 +169,122 @@ describe('Assets', () => {
...
+
+  describe('userAgent', () => {
+    const scriptOne = 'script-one.js';
+    const scriptTwo = 'script-two.js';
+    const chrome71 =
+      'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_0) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/71.0.3578.98 Safari/537.36';
+
+    it('uses the last manifest when no useragent exists', async () => {
+      readJson.mockImplementation(() =>
+        createMockConsolidatedManifest([
```
TheMallen: IMO the word create doesn't really add much except noise for all this mocking

packages/sewing-kit-koa/src/tests/assets.test.ts
```diff
@@ -125,27 +169,122 @@ describe('Assets', () => {
...
-    },
-  };
+  return {js: scripts, css: styles};
+}
+
+function createMockManifest(entrypoints: {[key: string]: Entrypoint} = {}) {
+  return {entrypoints};
+}
+
+function createMockManifestEntry({
```
TheMallen: We might want to expose these for external test usage / put them in their own package later.

packages/sewing-kit-koa/src/assets.ts
lemonmade: Only one thing would become less nested and it's easy to accidentally miss returning in that case, IMO this isn't the best spot for early returns

PR #445, opened 2018-12-19
Content length zero 400 error fix

mserge:
I found issue when testing the /graphql proxy as part of project for shopify-app

Starting 3-4 days Shopify backend always responds with 400 error and response
`
{"errors":{"query":"Required parameter missing or invalid"}}
`
when sending requests to proxy. Hunting for error I found that proxy code sends the headers with:
`
Content-Length: 0
`
as it informed to not parse the body.

GraphQL bodies are quite small so I suppose the parsing couldn't be a problem, as https://github.com/nsimmons/koa-better-http-proxy#parsereqbody advising use it for large payloads.

After change in generated JS code it started to work again.

### DIFF
--- a/packages/koa-shopify-graphql-proxy/src/shopify-graphql-proxy.ts
+++ b/packages/koa-shopify-graphql-proxy/src/shopify-graphql-proxy.ts
@@ -33,7 +33,7 @@ export default function shopifyGraphQLProxy(proxyOptions?: ProxyOptions) {
 
     await proxy(shop, {
       https: true,
-      parseReqBody: false,
+      parseReqBody: true,
       // Setting request header here, not response. That's why we don't use ctx.set()
       // proxy middleware will grab this request header
       headers: {
--- a/packages/koa-shopify-graphql-proxy/src/test/proxy.test.ts
+++ b/packages/koa-shopify-graphql-proxy/src/test/proxy.test.ts
@@ -138,7 +138,7 @@ describe('koa-shopify-graphql-proxy', () => {
         'X-Shopify-Access-Token': accessToken,
       },
       https: true,
-      parseReqBody: false,
+      parseReqBody: true,
     });
   });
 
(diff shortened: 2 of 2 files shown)


mserge:
Please review
@cartogram 
@TheMallen 
@michelleyschen 
@markiiikram

TheMallen:
I'm a little concerned about this from performance and privacy perspectives, and internally we aren't seeing this problem on our proxied graphql requests so I'm inclined to think there are some other factors causing this. Any of the other pinged folks seen this in an app on our side?

mserge:
Mallory (@TheMallen) thank you for feedback. 
I'm using code based on the https://github.com/Shopify/unite-react-node-app-workshop/blob/step5/server/index.js sample. The only large modification I made I've added the additional middleware:

```
var bodyParser = require('koa-bodyparser');
var router = new Router();

app.use(bodyParser());
```
because I have to hook the JSON coming. I can provide yarn.lock if needed. The configuration running under official node docker container with traefik reverse proxy.
 
If you think this fix is not approriate, could we make the bodyparsing configurable from outside of the GraphQL-proxy package?

Like that:
`
app.use(graphQLProxy({bodyParse: true}))
`

TheMallen:
@mserge if you use the `bodyParser` middleware exclusively on routes that need it 

eg.

```
const parser = bodyParser();

router.post('/some-hook', bodyParser, someHandler); 
router.post('/some-form', bodyParser, someHandler);
```

Does the problem go away? Wondering if that is the difference between our setups. Note that doing so is the recommended way to use body parsing middleware, as they can add considerable memory / CPU usage when used everywhere in the app.

mserge:
Yes, this was helpfull comment - it helped a lot, may be we should add this to README as known item?

TheMallen:
@mserge it might make sense to be in an FAQ or a `gotchas` section yes :)

TheMallen:
@mserge  I'm going to close this for now, if you want to open a PR adding this to the README I'd be happy to take a look :)

PR #447, opened 2018-12-20
Format count when handling pluralized translation

mathieulag:
When using react-i18n's pluralization feature, the `count` replacement variable often end up in the translated string. Currently, that count isn't localized, which means that we end up with an unformatted number in the output.

This PR formats the `count` variable as a number before passing it and the rest of the replacement variables to get processed.

### Example

```
// Assuming a dictionary like:
{
  "MyComponent": {
    "searchResult": {
      "one": "{count} widget found",
      "other": "{count} widgets found"
    }
  }
}

// With the locale 'en-US'
i18n.translate('MyComponent.searchResult', {count: 1000});
```

Before: `1000 widgets found`
After: `1,000 widgets found`


packages/react-i18n/src/utilities/translate.tsx
```diff
@@ -118,6 +120,10 @@ function translateWithDictionary(
     if (typeof count === 'number') {
       const group = new Intl.PluralRules(locale).select(count);
       result = result[group];
+
+      additionalReplacements[PLURALIZATION_KEY_NAME] = new Intl.NumberFormat(
```
nimzco: Why not doing `replacements.count = new Intl.NumberFormat(locale).format(count);` here?

packages/react-i18n/src/test/utilities.test.tsx
```diff
@@ -125,14 +125,23 @@ describe('translate()', () => {
...
+        ).toBe('1 foo');
+        expect(
+          translate('foo', {replacements: {count: 2}}, dictionary, locale),
+        ).toBe('2 foos');
+      });
+
+      it('formats the replacement `count`', () => {
+        const dictionary = {foo: {one: '{count} foo', other: '{count} foos'}};
+        expect(
+          translate('foo', {replacements: {count: 1000}}, dictionary, locale),
```
nimzco: It's weird if we pass `fr-fr` as locale, it still returns `1,000` 🤔 . 
It seems to only be in the tests though... 🤔

packages/react-i18n/src/test/utilities.test.tsx
```diff
@@ -125,14 +125,23 @@ describe('translate()', () => {
...
+          translate('foo', {replacements: {count: 1}}, dictionary, locale),
+        ).toBe('1 foo');
+        expect(
+          translate('foo', {replacements: {count: 2}}, dictionary, locale),
+        ).toBe('2 foos');
+      });
+
+      it('formats the replacement `count`', () => {
+        const dictionary = {foo: {one: '{count} foo', other: '{count} foos'}};
+        expect(
```
nimzco: I might also test the `0` case. Just in case it passes `0.00` for some reason one day.

nimzco:
I am personally all for that change, it makes total sense to me. I might forget cases where this could be problematic.
@lemonmade @TheMallen, thoughts on this change?

packages/react-i18n/README.md
```diff
@@ -209,7 +209,7 @@ i18n.translate('MyComponent.searchResult', {count: searchResults});
 ```
 
 As noted above, this functionality depends on the `Intl.PluralRules` global. If this does not exist [for your environment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/PluralRules#Browser_compatibility), we recommend including the [`intl-pluralrules`](https://yarnpkg.com/en/package/intl-pluralrules) polyfill.
-We also recommend to have the `{count}` variable in all of your keys as some languages can use the key `"one"` when the count is `zero` for example. See MDN docs on [Localization and Plurals](https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_and_Plurals).
+We also recommend to have the `{count}` variable in all of your keys as some languages can use the key `"one"` when the count is `zero` for example. See MDN docs on [Localization and Plurals](https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_and_Plurals). By default, `{count}` will be automatically formatted as a number.
```
lemonmade: Maybe give them a pattern for if they want the number differently formatted? I think passing both a `count` (which sets up the plural rule) and a `formattedCount` is probably the right pattern in those cases.

packages/react-i18n/src/utilities/translate.tsx
mathieulag: I started by doing that, but I realized it was modifying the original replacement dictionary passed by the caller.

packages/react-i18n/src/test/utilities.test.tsx
mathieulag: Good point! I will add a test for that 👍

packages/react-i18n/src/utilities/translate.tsx
nimzco: Oh true... 👍

packages/react-i18n/src/test/utilities.test.tsx
mathieulag: After scratching my head on this one I think I found the reason from the [node.js docs](https://nodejs.org/api/intl.html#intl_build_with_a_pre_installed_icu_system_icu):

> Functionalities that only require the ICU library itself, such as String.prototype.normalize() and the WHATWG URL parser, are fully supported under system-icu. Features that require ICU locale data in addition, such as Intl.DateTimeFormat may be fully or partially supported, depending on the completeness of the ICU data installed on the system.

I ran the test they had on that page and our node.js build doesn't have the full ICU data, and to get it we would need to do a custom build. That explains why it always returns `1,000` without regards to the locale.

TL;DR our node.js build doesn't fully support Intl.

packages/react-i18n/README.md
mathieulag: I've added a small note on how to handle this case!

packages/react-i18n/src/test/utilities.test.tsx
nimzco: Oh nice findings. 👍

Issue #433, opened 2018-12-05
A collection of issues while testing react-shortcuts


beefchimi:
Everything here was tested on a Mac - and specifically, my Macbook Pro _(non-touchbar)_ keyboard. It is possible behaviour will be different on a Windows or Linux machine.

- [x] `Fn` key does not register on my Mac _(there is no `event`)_.
- [ ] Since holding `Shift` changes a key to its uppercase variant _(`a` becomes `A`)_, holding `Shift` and using lowercase `ordered` keys will not work.
  - [ ] Should we automatically convert characters to uppercase when `Shift` is used in a `held` array?
  - [ ] Or just cover this in the documentation?
- [ ] Likewise, using the `Alt` key also modifies the original key _(`a` becomes `å`)_, meaning a held `Alt` key won't match on anything available within the `AlphabetKey` type.
- [ ] Interestingly, `CapsLock` as a `held` key only works in one scenario:
  - **Using lowercase `ordered` keys**
  - Begin with `CapsLock` off, hold key, press ordered keys = nothing
  - Begin with `CapsLock` on, hold key, press ordered keys = nothing
  - **Using uppercase `ordered` keys**
  - Begin with `CapsLock` off, hold key, press ordered keys = `onMatch`
  - Begin with `CapsLock` on, hold key, press ordered keys = nothing

cc/ @TheMallen @9at8

9at8:
These are really nice findings!

- `Fn` Key: This does not generate events as defined in the [spec](https://w3c.github.io/uievents-key/#key-Fn). We should remove this type, and others which are not detected 👍 
- `Shift` changing `a` -> `A`: I'm not too opinionated about this, but maybe we should convert lowercase to uppercase when we see a `Shift` in the modifiers. Whatever we do, we should clearly cover this in documentation though!
- Not sure what to do about this 🤔 We can use `event.code` but it has it's own [problems](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code)
- `CapsLock` only works as held if the LED is green, as mentioned [here](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/getModifierState)

thomasletsch:
To give some remarks of a library user:
I just started with implementing some shortcuts and of course used "Shift" and lower case "a". Therefore I would love to be able to use "Shift" + "a" :-)

[...]

PR #448, opened 2018-12-24
[react-shortcuts] 🔥 Remove `Fn` and `FnLock`

9at8:
#433 mentions that the `Fn` key does not generate events. This is expected, according to the spec [here](https://w3c.github.io/uievents-key/#key-Fn). This PR removes those keys from types, so no one uses them.

cc: @TheMallen @beefchimi


beefchimi:
Looks good 👍

PR #454, opened 2019-01-02
[FormState] Adds externalErrors prop

darrenhebner:
### What does this PR do?
This PR adds an `externalErrors` prop to `FormState`. This allows you to initialize the form with external errors. This API also enables the possibility of async errors.

### Example
Here's how you would initialize the form with external errors. Notice that external errors are merged with remote errors and are returned by the `errors` array.

External errors will also be passed down to corresponding fields if a field is specified. 

Form state will not manage your external errors, you will need to clear them yourself if necessary. 

```jsx
class MyComponent extends React.Component {
    state = {
        externalErrors: [{message: 'Example error', field: 'foo'}]
    }

  render() {
    const {externalErrors} = this.state;
    
    return (
      <FormState
        initialValues={{foo: ''}}
        externalErrors={externalErrors}
        onSubmit={async (formDetails) => {
          const response = await submitFormMutation(formDetails);
          
           if (response.errors) {
               // we still need to clear the errors ourselves because formstate 
              // doesn't know what they represent or when to clear them
               this.setState({externalErrors: null});
               return response.errors;
           }
        }}
      >
        {({fields, errors}) => {
           return (
             <Banner>
              {/* the externalErrors get seamlessly added to the errors array */}
               {errors.map(error => <li>{error.message}</li>)}
             </Banner>
               {/* and passed down to matching fields */}
              <TextField {...fields.foo} />
           )
        }}
      </FormState>
    )
  }
}
```

Here's how this API can enable async errors.

```jsx
class MyComponent extends React.Component {
    state = {
        externalErrors: []
    }

  render() {
    const {externalErrors} = this.state;
    
    return (
      <FormState
        initialValues={{foo: ''}}
        externalErrors={externalErrors}
        onSubmit={async (formDetails) => {
          const response = await submitFormMutation(formDetails);
          
           if (response.errors) {
               return response.errors;
           }
        }}
      >
        {({fields, errors}) => {
           return (
             <Banner>
               {errors.map(error => <li>{error.message}</li>)}
             </Banner>
              <TextField {...fields.foo} />
              {/* this would return the same type of error objects */}
               <RemoteValidation 
                   fields={fields} 
                   url={someUrl} 
                  onFail={(externalErrors) => this.setState({externalErrors}}
                />
           )
        }}
      </FormState>
    )
  }
}
```

### Cases that I've tried to consider
- Initializing form with external errors
- Mixing remote errors with external errors
- Updating external errors when remote errors are present

### Proof it works 😄
This screenshot shows a server error and an external error coexisting.
screen shot 2019-01-02 at 4 46 18 pm

### DIFF
--- a/packages/react-form-state/src/FormState.ts
+++ b/packages/react-form-state/src/FormState.ts
@@ -52,6 +52,7 @@ interface Props<Fields> {
   validators?: Partial<ValidatorDictionary<Fields>>;
   onSubmit?: SubmitHandler<Fields>;
   validateOnSubmit?: boolean;
+  externalErrors?: RemoteError[];
   children(form: FormDetails<Fields>): React.ReactNode;
 }
 
@@ -60,6 +61,7 @@ interface State<Fields> {
   fields: FieldStates<Fields>;
   dirtyFields: (keyof Fields)[];
   errors: RemoteError[];
+  externalErrors: RemoteError[];
 }
 
 export default class FormState<
@@ -70,16 +72,31 @@ export default class FormState<
 
   static getDerivedStateFromProps<T>(newProps: Props<T>, oldState?: State<T>) {
     const newInitialValues = newProps.initialValues;
+    const newExternalErrors = newProps.externalErrors || [];
 
     if (oldState == null) {
-      return createFormState(newInitialValues);
+      return createFormState(newInitialValues, newExternalErrors);
     }
 
     const oldInitialValues = initialValuesFromFields(oldState.fields);
     const shouldReinitialize = !isEqual(oldInitialValues, newInitialValues);
+    const oldExternalErrors = oldState.externalErrors;
 
     if (shouldReinitialize) {
-      return createFormState(newInitialValues);
+      return createFormState(newInitialValues, newExternalErrors);
+    }
+
+    if (
+      newExternalErrors !== oldExternalErrors ||
+      newExternalErrors.length !== oldExternalErrors.length
+    ) {
+      return {
+        externalErrors: newExternalErrors,
+        fields: fieldsWithErrors(oldState.fields, [
+          ...newExternalErrors,
+          ...oldState.errors,
+        ]),
+      };
     }
 
     return null;
@@ -110,13 +127,13 @@ export default class FormState<
   }
 
   private get formData() {
-    const {errors} = this.state;
+    const {errors, externalErrors} = this.state;
     const {fields, dirty, valid} = this;
 
     return {
       dirty,
       valid,
-      errors,
+      errors: [...errors, ...externalErrors],
       fields,
     };
   }
@@ -132,9 +149,13 @@ export default class FormState<
   }
 
   private get valid() {
-    const {errors} = this.state;
+    const {errors, externalErrors} = this.state;
 
-    return !this.hasClientErrors && errors.length === 0;
+    return (
+      !this.hasClientErrors &&
+      errors.length === 0 &&
+      externalErrors.length === 0
+    );
   }
 
   private get hasClientErrors() {
@@ -199,7 +220,9 @@ export default class FormState<
 
   @bind()
   private reset() {
-    this.setState((_state, props) => createFormState(props.initialValues));
+    this.setState((_state, props) =>
+      createFormState(props.initialValues, props.externalErrors),
+    );
   }
 
   @memoize()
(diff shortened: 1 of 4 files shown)


packages/react-form-state/src/FormState.ts
```diff
@@ -348,32 +368,42 @@ export default class FormState<
...
-        fields: mapObject(fields, (field, path) => {
-          return {
-            ...field,
-            error: errorDictionary[path],
-          };
-        }),
-      };
+    this.setState({
+      errors,
+      fields: fieldsWithErrors(fields, [...errors, ...externalErrors]),
```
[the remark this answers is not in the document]
darrenhebner: Oh right! I'll update.

TheMallen:
🎉

packages/react-form-state/src/FormState.ts
```diff
@@ -70,16 +72,28 @@ export default class FormState<
...
     const oldInitialValues = initialValuesFromFields(oldState.fields);
     const shouldReinitialize = !isEqual(oldInitialValues, newInitialValues);
+    const oldExternalErrors = oldState.externalErrors;
 
     if (shouldReinitialize) {
-      return createFormState(newInitialValues);
+      return createFormState(newInitialValues, newExternalErrors);
+    }
+
+    if (!isEqual(newExternalErrors, oldExternalErrors)) {
```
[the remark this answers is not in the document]
jim-chou-shopify: EX:
```
newExternalErrors !== oldExternalErrors || 
newExternalErrors.length !== oldExternalErrors.length
```

packages/react-form-state/src/FormState.ts
```diff
@@ -70,16 +72,31 @@ export default class FormState<
...
+      return createFormState(newInitialValues, newExternalErrors);
+    }
+
+    if (
+      newExternalErrors !== oldExternalErrors ||
+      newExternalErrors.length !== oldExternalErrors.length
+    ) {
+      return {
+        externalErrors: newExternalErrors,
+        fields: fieldsWithErrors(oldState.fields, [
```
jim-chou-shopify: We can derive `errors` here by merging external errors  and remote errors.

jim-chou-shopify: I think we can simplify the logic by not associating field with errors using getDerivedStateFromProps.
Instead, we should be compute field with errors in render function with memoization helper.
https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#what-about-memoization

Advantage:
- simpler logic
- createFormState, updateRemoteErrors, getDerivedStateFromProps don't need to associate field with external error

packages/react-form-state/src/FormState.ts
```diff
@@ -110,13 +127,13 @@ export default class FormState<
...
   private get formData() {
-    const {errors} = this.state;
+    const {errors, externalErrors} = this.state;
     const {fields, dirty, valid} = this;
 
     return {
       dirty,
       valid,
-      errors,
+      errors: [...errors, ...externalErrors],
```
jim-chou-shopify: errors can be derived in getDerivedStateFromProps

packages/react-form-state/src/FormState.ts
```diff
@@ -132,9 +149,13 @@ export default class FormState<
...
 
   private get valid() {
-    const {errors} = this.state;
+    const {errors, externalErrors} = this.state;
 
-    return !this.hasClientErrors && errors.length === 0;
+    return (
+      !this.hasClientErrors &&
+      errors.length === 0 &&
+      externalErrors.length === 0
```
jim-chou-shopify: errors can be derived in getDerivedStateFromProps to include externalErrors

packages/react-form-state/src/FormState.ts
```diff
@@ -348,32 +371,40 @@ export default class FormState<
...
-          return {
-            ...field,
-            error: errorDictionary[path],
-          };
-        }),
-      };
-    });
+    this.setState(({fields, externalErrors}) => ({
+      errors,
+      fields: fieldsWithErrors(fields, [...errors, ...externalErrors]),
```
jim-chou-shopify: May be simplified to `this.setState({errors});`.
See https://github.com/Shopify/quilt/pull/454/files#r245368975.

packages/react-form-state/src/FormState.ts
```diff
@@ -385,8 +416,9 @@ function createFormState<Fields>(values: Fields): State<Fields> {
   return {
     dirtyFields: [],
     errors: [],
+    externalErrors,
```
jim-chou-shopify: See https://github.com/Shopify/quilt/pull/454/files#r245368975.
createFormState can probably left unchanged.

packages/react-form-state/src/FormState.ts
```diff
@@ -70,16 +72,31 @@ export default class FormState<
 
   static getDerivedStateFromProps<T>(newProps: Props<T>, oldState?: State<T>) {
     const newInitialValues = newProps.initialValues;
+    const newExternalErrors = newProps.externalErrors || [];
 
     if (oldState == null) {
```
jim-chou-shopify: Does oldState ever equal to undefined or null?
https://github.com/Shopify/quilt/blob/db4eed7982cda2581ea71251f1e49afc2b8b4463/packages/react-form-state/src/FormState.ts#L105

packages/react-form-state/src/FormState.ts
```diff
@@ -70,16 +72,31 @@ export default class FormState<
...
-      return createFormState(newInitialValues);
+      return createFormState(newInitialValues, newExternalErrors);
     }
 
     const oldInitialValues = initialValuesFromFields(oldState.fields);
```
jim-chou-shopify: Not related to the PR but I feel `getDerivedStateFromProps` is not necessary in this component.
- Reinitialization can be achieved in `componentDidUpdate` with access to prevProps

packages/react-form-state/src/FormState.ts
TheMallen: it doesn't, if this gets rebased this line is gone in latest master.

jim-chou-shopify:
Create a branch to include external errors without tracking in the state and `getDerivedStateFromProps` with the latest master for reference.
https://github.com/Shopify/quilt/pull/new/form-state-external-errors

darrenhebner:
I'm closing this PR in favour of https://github.com/Shopify/quilt/pull/461. The new PR accounts for some of Jim's suggestions from this PR. It also rethinks things a bit to work with `onInitialValueChanged`.

PR #456, opened 2019-01-03
Adds `DeepPartial` mapped type

cartogram:
The `Partial<T>` mapped type only applies to the first level of properties. 

This PR adds a `DeepPartial<T>` mapped type that recursively makes optional all properties in a given type.

### DIFF
--- a/packages/useful-types/src/types.ts
+++ b/packages/useful-types/src/types.ts
@@ -4,3 +4,10 @@ export type ThenType<T> = T extends Promise<infer U> ? U : T;
 export type FirstArgument<T> = T extends ((arg: infer U) => any) ? U : never;
 export type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
 export type Props<T> = T extends ComponentType<infer P> ? P : never;
+export type DeepPartial<T> = {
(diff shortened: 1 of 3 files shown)


packages/useful-types/src/types.ts
```diff
@@ -4,3 +4,10 @@ export type ThenType<T> = T extends Promise<infer U> ? U : T;
 export type FirstArgument<T> = T extends ((arg: infer U) => any) ? U : never;
 export type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
 export type Props<T> = T extends ComponentType<infer P> ? P : never;
+export type DeepPartial<T> = {
+  [P in keyof T]?: T[P] extends Array<infer U>
+    ? Array<DeepPartial<U>>
+    : T[P] extends ReadonlyArray<infer U>
+      ? ReadonlyArray<DeepPartial<U>>
```
patsissons: I couldn't get the `ReadonlyArray` to infer properly, have you tried testing that out?

patsissons: i'll try and build a type set that covers this issue

cartogram: Resolved via conversations on slack. Merging as is.

PR #457, opened 2019-01-04
[react-i18n] Removed MissingTimezoneError

vsumner:
react-i18n formatDate required a timezone to be passed. 
Since `Intl` doesn't require a timezone to work the error was removed.

### DIFF
--- a/packages/react-i18n/src/i18n.ts
+++ b/packages/react-i18n/src/i18n.ts
@@ -17,7 +17,6 @@ import {
 } from './constants';
 import {
   MissingCurrencyCodeError,
-  MissingTimezoneError,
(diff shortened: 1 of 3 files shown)


packages/react-i18n/src/i18n.ts
```diff
@@ -15,11 +15,7 @@ import {
   RTL_LANGUAGES,
   Weekdays,
 } from './constants';
-import {
-  MissingCurrencyCodeError,
-  MissingTimezoneError,
```
lemonmade: Can you delete this class?

vsumner: Done. Thanks for catching this.

PR #461, opened 2019-01-07
[FormState] Adds externalErrors prop

darrenhebner:
Note: This PR is a second attempt at adding the `externalErrors` prop. This accounts for the new `onInitialValueChanged` prop.

### What does this PR do?
This PR adds an `externalErrors` prop to `FormState`. This allows you to initialize the form with external errors. This API also enables the possibility of async errors.

### Example
Here's how you would initialize the form with external errors. Notice that external errors are merged with remote errors and are returned by the `errors` array.

External errors will also be passed down to corresponding fields if a field is specified. 

Form state will not manage your external errors, you will need to clear them yourself if necessary. 

```jsx
class MyComponent extends React.Component {
    state = {
        externalErrors: [{message: 'Example error', field: 'foo'}]
    }

  render() {
    const {externalErrors} = this.state;
    
    return (
      <FormState
        initialValues={{foo: ''}}
        externalErrors={externalErrors}
        onSubmit={async (formDetails) => {
          const response = await submitFormMutation(formDetails);
          
           if (response.errors) {
               // we still need to clear the errors ourselves because formstate 
              // doesn't know what they represent or when to clear them
               this.setState({externalErrors: null});
               return response.errors;
           }
        }}
      >
        {({fields, errors}) => {
           return (
             <Banner>
              {/* the externalErrors get seamlessly added to the errors array */}
               {errors.map(error => <li>{error.message}</li>)}
             </Banner>
               {/* and passed down to matching fields */}
              <TextField {...fields.foo} />
           )
        }}
      </FormState>
    )
  }
}
```

Here's how this API can enable async errors.

```jsx
class MyComponent extends React.Component {
    state = {
        externalErrors: []
    }

  render() {
    const {externalErrors} = this.state;
    
    return (
      <FormState
        initialValues={{foo: ''}}
        externalErrors={externalErrors}
        onSubmit={async (formDetails) => {
          const response = await submitFormMutation(formDetails);
          
           if (response.errors) {
               return response.errors;
           }
        }}
      >
        {({fields, errors}) => {
           return (
             <Banner>
               {errors.map(error => <li>{error.message}</li>)}
             </Banner>
              <TextField {...fields.foo} />
              {/* this would return the same type of error objects */}
               <RemoteValidation 
                   fields={fields} 
                   url={someUrl} 
                  onFail={(externalErrors) => this.setState({externalErrors}}
                />
           )
        }}
      </FormState>
    )
  }
}
```

### Cases that I've tried to consider
- Initializing form with external errors
- Mixing remote errors with external errors
- Updating external errors when remote errors are present

### Proof it works 😄
This screenshot shows a server error and an external error coexisting.
screen shot 2019-01-02 at 4 46 18 pm


packages/react-form-state/src/FormState.ts
```diff
@@ -360,31 +383,39 @@ export default class FormState<
...
-            error: errorDictionary[path],
-          };
-        }),
+        fields: mapObject(fields, fieldWithErrorMapper),
       };
     });
   }
 }
 
+function errorDictionaryFactory(errors: RemoteError[]) {
```
TheMallen: I don't really like these names, we should name functions for what they *do* not what they *are*.

packages/react-form-state/src/FormState.ts
```diff
@@ -424,11 +458,17 @@ function createFormState<Fields>(values: Fields): State<Fields> {
     };
   });
 
+  let fieldsWithErrors = fields;
+  if (externalErrors) {
```
TheMallen: A couple things:
- the let, if, reassignment pattern isn't really something this library does anywhere else so this feels a little weird in terms of coding style
- we need to also handle filtering down errors into the fields in `reconcileFormState`
- we could probably simplify this stuff if we only push the errors down into the fields at `formData()`, then `getDerivedStateFromProps` doesn't care about `externalErrors` at all and we don't need to store this stuff in state.

packages/react-form-state/src/FormState.ts
```diff
@@ -91,13 +92,19 @@ export default class FormState<
...
 
-  state = createFormState(this.props.initialValues);
+  state = createFormState(this.props.initialValues, this.props.externalErrors);
   private mounted = false;
 
   componentDidMount() {
     this.mounted = true;
   }
 
+  componentDidUpdate(prevProps) {
```
TheMallen: We shouldn't need this method. At the moment it's redundant with what we're doing in `createFormState`, and if we make the changes I suggested we won't need to alter the state based on `externalErrors` at all.

packages/react-form-state/src/FormState.ts
jim-chou-shopify: IMO, factory pattern is well known pattern.
This function transforms array into object format.

packages/react-form-state/src/FormState.ts
jim-chou-shopify: From what I can tell, errors need to be associated with fields to stay consistent with how RemoteErrors are handled.
Several methods in the component relies on errors being associated with fields:
- `submit`: Prevent submitting of form before errors (remote or client error) are fixed
**How:** If `validateOnSubmit=true` then `hasClientErrors` checks if any field has error.
  - IMO, `hasFieldErrors` makes more sense, since remoteErrors are also associated with fields
- `updateField`: Easy of removing error in field by removing error on field change
**How:** Runs client validation on fields with errors (remote or client error).
In case the error is remote error, the client validation probably return null and error get removed.

In this case, we are associating external errors with fields following the pattern in the component.
We do not want to track externalErrors in state in order to conditionally derive new fields from external errors.

According to the React doc, `getDerivedStateFromProps` is not recommended for most components.
https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops

One of the suggested alternative is to use `componentDidUpdate`.
> If you need to perform a side effect (for example, data fetching or an animation) in response to a change in props, use componentDidUpdate lifecycle instead.

packages/react-form-state/src/FormState.ts
jim-chou-shopify: - Do you suggest two return statement instead?
- Can you elaborate on `we need to also handle filtering down errors into the fields in reconcileFormState`?
- Agree. Maybe my understanding of FormState is unclear.
See https://github.com/Shopify/quilt/pull/461/files#r246853227

packages/react-form-state/src/FormState.ts
darrenhebner: I think something like `getErrorDictionary` is common.

packages/react-form-state/src/FormState.ts
TheMallen: We aren't performing a side-effect. There is no reason to use `componentDidUpdate` over `getDerivedStateFromProps` for mutating state.

packages/react-form-state/src/FormState.ts
TheMallen: * Two return statements or a ternary assignment is more in line with the rest of the code (and most of our react code in general)

* If is set to `reset-where-changed` when initial values change it's state derivation logic will be using`reconcileFormState`. As such we need to facilitate that by having the same error pass-down logic there too

*Spoke to @darrenhebner about this on slack, we definitely need this in state due to it being independently clearable on blur of a field. Responded to your other comment as well, but we definitely are a component that needs derived state (one could argue we should maybe be fully uncontrolled instead, but that would be a massive breaking api change, and probably an opt-in mode) and `componentDidUpdate` has no benefits for us over `getDerivedStateFromProps`.

TheMallen:
@darrenhebner We might want to rename `onInitialValuesChange` since this PR also secretly makes that also effect the behaviour of `externalErrors`. We might need to go with the more verbose naming I originally had `stateDerivationStrategy` or `reconciliationStrategy` :/ (this rename would be a breaking change)

packages/react-form-state/src/FormState.ts
TheMallen: Factory is a well known pattern but naming functions with a factory suffix is not. Half of the functions in a given JS codebase are technically factories (they output custom objects based on input), but we don't name everything `translationComponentFactory`, `apolloConnectedComponentFactory`, `fixtureFactory`. We tend to name them based on what they do instead.

TheMallen: For example: if it transforms arrays into an object format we should call it something like `mapErrorsToDictionary`.

darrenhebner:
Ok I _think_ everything should be in a good place 😅. @TheMallen I'm not really sure what to do about the naming of the `onInitialValuesChanged` prop. If we want to make that change in the PR I can do it.

packages/react-form-state/src/FormState.ts
```diff
@@ -71,27 +73,52 @@ export default class FormState<
...
 
   static getDerivedStateFromProps<T>(newProps: Props<T>, oldState: State<T>) {
-    const {initialValues, onInitialValuesChange} = newProps;
+    const {
+      initialValues,
+      onInitialValuesChange,
+      externalErrors = [],
+    } = newProps;
+
+    let nextState;
```
jim-chou-shopify: I thought using let is discouraged instead you should have separate return statements.

darrenhebner: Good call. I've cleaned this up.

tzvipm:
Could this PR have broken react-form-state's validation? See https://github.com/Shopify/quilt/issues/500. Note that `0.7.4` is working in the linked codesandbox.

darrenhebner:
@tzvipm - I think it's likely that this broke validation, but I'm not yet positive how. Do we usually revert or try to fix forward in Quilt?

Issue #466, opened 2019-01-11
jest-koa-mocks: custom state not included in createMockContext method


colinrcummings:
## Overview

This does not work:

```
import { createMockContext } from "@shopify/jest-koa-mocks";

const ctx = createMockContext({
    state: {
      foo: "bar"
    }
  });

console.log(ctx.state); // {}
```

## Consuming repo

@shopify/jest-koa-mocks

PR #467, opened 2019-01-11
🐛fix custom state in jest-koa-mocks createMockContext method

colinrcummings:
Addresses #466.


colinrcummings:
Thanks for shipping this so quickly!

PR #468, opened 2019-01-11
Adds User Timing Mock

cartogram:
This PR adds a `user-timing` mock to `jest-dom-mocks`.

I started this PR thinking I would mock the entire `window.performance` API, but after realizing that the global exists in the most recent versions of `jest`, opted for just providing a way to mock user timing metrics instead.

### DIFF
--- a/packages/jest-dom-mocks/src/user-timing.ts
+++ b/packages/jest-dom-mocks/src/user-timing.ts
@@ -1,0 +1,59 @@
+import set from 'lodash/set';
+
+type UserTimingMock = typeof window.performance.timing;
+
+export default class Performance {
+  private isUsingMockUserTiming = false;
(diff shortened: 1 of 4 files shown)


packages/jest-dom-mocks/src/user-timing.ts
```diff
@@ -0,0 +1,59 @@
...
+    };
+
+    set(window.performance, 'timing', mockTiming);
+    this.isUsingMockUserTiming = true;
+  }
+
+  restore() {
+    if (!this.isUsingMockUserTiming) {
+      throw new Error(
+        'You tried to restore window.performance when it was already restored.',
```
cartogram: `window.performance` => `window.performance.timing`

PR #469, opened 2019-01-16
Make the enable_cookies route redirect to ${prefix}/auth/inline

beppu:
Previously, the **/auth/enable_cookies** route redirected to:

```
https://${window.shopOrigin}/admin/apps/${window.apiKey} 
```

which seems useless, because all it ever does is 404.  This results in the first installation attempt failing.

```
GET /shopify/auth?hmac=b7d85bd1976f2f4e3d2badab0473ed47d456855f199cc15917470f1da04ed213&shop=dbap-dev.myshopify.com&timestamp=1547106686 200 2.777 ms - 802
GET /shopify/auth/enable_cookies?shop=dbap-dev.myshopify.com 200 1.664 ms - 10350
```

However, the **/auth/enable_cookies** route does set the `shopifyTestCookie` so the second time around and ever subsequent time afterward, the installation works.  Also, notice that subsequent installation attempts go to **/auth/inline** instead. 

```
GET /shopify/auth?hmac=cc729de08ca82c5bb06e28872e4fc23b0a396678c2af9ec9e95b90e5b0619e37&shop=dbap-dev.myshopify.com&timestamp=1547106888 200 1.463 ms - 786
GET /shopify/auth/inline?shop=dbap-dev.myshopify.com 302 1.916 ms - 619
```

## Solution

What this pull request implements is a change to the **/auth/enable_cookies** route that redirects to **/auth/inline**.  This makes installation ALWAYS work on the first try and every subsequent try.  Let me know if this seems like a reasonable change, @ragalie ?

### DIFF
--- a/packages/koa-shopify-auth/src/auth/create-enable-cookies.ts
+++ b/packages/koa-shopify-auth/src/auth/create-enable-cookies.ts
@@ -1,3 +1,4 @@
+import querystring from 'querystring';
 import {Context} from 'koa';
 
 import Error from './errors';
@@ -10,10 +11,13 @@ const FOOTER = `Cookies let the app authenticate you by temporarily storing your
 information. They expire after 30 days.`;
 const ACTION = 'Enable cookies';
 
-export default function createEnableCookies({apiKey}: OAuthStartOptions) {
+export default function createEnableCookies({apiKey, prefix}: OAuthStartOptions) {
   return function enableCookies(ctx: Context) {
-    const {query} = ctx;
+    const {host, query} = ctx;
     const {shop} = query;
+    const path = `${prefix}/auth/inline`;
+    const params = {shop};
+    const queryString = querystring.stringify(params);
 
     if (shop == null) {
       ctx.throw(400, Error.ShopParamMissing);
@@ -370,7 +374,7 @@ export default function createEnableCookies({apiKey}: OAuthStartOptions) {
     (function() {
       function setCookieAndRedirect() {
         document.cookie = "shopify.cookies_persist=true";
-        window.location.href = window.shopOrigin + "/admin/apps/" + window.apiKey;
+        window.location.href = "https://${host}${path}?${queryString}"
       }
 
       function shouldDisplayPrompt() {


ragalie:
Thanks! I'm not sure why it wasn't something that we caught earlier...

I _think_ this would work, but does more than it needs to. The existing code was redirecting to the application root URL in the embedded context. I think we'll get the same end result (but always working) by redirecting to the application root URL _outside_ the embedded context.

For this to work, we'd either need to redefine `prefix` to mean the application root (and then the auth stuff will always be mounted at `/auth` from the application root) or add a new option `applicationPath`. Then we can redirect to that value.

I haven't tried that, but I think that would be a better way of going about this, since it avoid an extra OAuth flow in some cases.

beppu:
Apologies for the late reply.

> The existing code was redirecting to the application root URL in the embedded context. I think we'll get the same end result (but always working) by redirecting to the application root URL outside the embedded context.

As far as I can tell, this doesn't seem to be true.  The application root URL is already being used OUTSIDE of the embedded context on the first try (before the `shopifyTest` cookie exists), and it fails.  I presume it's because the application root URL is not yet valid for an uninstalled app.  

We start here.  Make sure all the cookies for the app's hostname are cleared first.

partners

The first try uses the application root URL outside of the embedded context and it fails presumably because the app isn't installed yet, and it's not yet meaningful.

fail

If we try the same thing again from the partners admin, it works as intended, because the `shopifyTest` cookie exists, and it takes a different code path that does the right thing.

second try

---

I'll still see if I can skip going to `/auth/inline` and go straight to `/admin/oauth/request_grant`, but I wanted to try to show how the application root URL was being used and causing a failure.

lemonmade:
@ragalie do you know what the next steps should be here and on https://github.com/Shopify/quilt/pull/523?

ragalie:
I'm sorry I've lost context on this; @tylerball might know better. The best I can say is that this implementation should probably follow the same logic we use in [shopify_app](https://github.com/Shopify/shopify_app).

mkevinosullivan:

Thank you!

PR #473, opened 2019-01-18
[react-i18n] Add currency code types

ismail-syed:
This exposes the `CurrencyCode` enum from `types/graphql/core-types`. This is almost always used in conjunction with `@shopify/react-i18n` so it should probably be provided and consumed from here. 

I also added a `CHANGELOG.md` for this repo. @michelleyschen, was this package left our when we added changelogs or was not having a changelog for this package intentional?


michelleyschen:
the changlot thing was to fix any package that has a different version number than what's on npm.
Adding it now it perfect thou!

PR #474, opened 2019-01-18
Allow to specify multiple env vars in withEnv

robin-drexler:
Hey 👋,

this is just an idea that crossed my mind when I had a look at the repository. Feel free to close this PR if you don't deem it useful.

## What does it do?
This PR allows to use `withEnv` not only with `NODE_ENV`, but also lets consumers pass in an object to set multiple environment variables. Those variables will also be reset to their previous values after the callback has run.

Before:
```ts
import withEnv from '@shopify/with-env';

it('does one thing in development', () => {
  withEnv('development', () => {
    // your code here
  });
});
```

Now also supports:

```ts
import withEnv from '@shopify/with-env';

it('does one thing', () => {
  withEnv({MY_ENV_ONE: 'test', ANOTHER_ENV: 'test-2'}, () => {
    // your code here
  });
});
```

### DIFF
--- a/packages/with-env/src/index.ts
+++ b/packages/with-env/src/index.ts
@@ -1,10 +1,28 @@
-export default function withEnv<T = any>(env: string, callback: () => T): T {
-  const previousEnv = process.env.NODE_ENV;
-  function resetEnv() {
-    process.env.NODE_ENV = previousEnv;
-  }
+import {Env} from './types';
 
-  process.env.NODE_ENV = env;
+export default function withEnv<T = any>(env: Env, callback: () => T): T {
+  const envToSet = typeof env === 'string' ? {NODE_ENV: env} : env;
+
+  const previousEnv = Object.keys(envToSet).reduce((previousEnv, key) => {
+    // only care for keys that already have an entry in `process.env`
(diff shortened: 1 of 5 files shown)


packages/with-env/src/index.ts
```diff
@@ -1,11 +1,19 @@
...
+  const previousEnv = Object.keys(envToSet).reduce((previousEnv, key) => {
+    previousEnv[key] = process.env[key];
+    return previousEnv;
+  }, {});
+
+  Object.assign(process.env, envToSet);
+
   function resetEnv() {
-    process.env.NODE_ENV = previousEnv;
+    Object.assign(process.env, previousEnv);
```
TheMallen: one thing to note is that this won't un-assign anything that was previously undefined, so we may want to instead clear off all the keys from `process.env` before doing this assign.

robin-drexler: Great catch, thank you!
I've fixed it, added a test for that case and force-pushed the commit (as I only rebased against master :))

TheMallen:
Looking good with the latest changes, thanks!

robin-drexler:
@TheMallen thanks for the review!

Should I just go ahead and merge the PR? :)

Issue #472, opened 2019-01-17
Remove lodash from react-form-state


tzvipm:
## Overview

We currently use `lodash` in `@shopify/react-form-state`. This ends up causing lodash to be included in whichever vendor bundle `react-form-state` ends up in. Stranger yet, somehow this happens even if `lodash` is also in a different vendors bundle, so we end up getting it and `lodash-decorators` twice (~20kb).

## Consuming repo

`Shopify/web`

## Checklist

- [x] Please delete the labels section before submitting your issue
- [x] I have described this issue in a way that is actionable (if possible)

TheMallen:
Yeah, we can make simpler versions of set / deepEquals pretty easily. I had a branch that did this actually. We could even drop the decorators probably, though memoize would be a pain to handroll

PR #475, opened 2019-01-21
🔥 Remove lodash in react-form-state

tzvipm:
Closes https://github.com/Shopify/quilt/issues/472

### DIFF
--- a/packages/react-form-state/src/FormState.ts
+++ b/packages/react-form-state/src/FormState.ts
@@ -89,7 +89,7 @@ export default class FormState<
 
   state = createFormState(this.props.initialValues);
   private mounted = false;
-  private fieldsWithHandlers = new Map<string, any>();
+  private fieldsWithHandlers = new WeakMap();
 
   componentDidMount() {
     this.mounted = true;
@@ -211,26 +211,21 @@ export default class FormState<
 
   private fieldWithHandlers = <Key extends keyof Fields>(
     field: FieldStates<Fields>[Key],
-    fieldPath: string & Key,
+    fieldPath: Key,
   ) => {
-    const hashKey = `${fieldPath}:${JSON.stringify(field)}`;
-    let ret: FieldState<Fields[Key]> & {name: Key} & {
-      onChange(newValue: any): void;
-      onBlur(): void;
-    };
-    if (this.fieldsWithHandlers.has(hashKey)) {
+    if (this.fieldsWithHandlers.has(field)) {
       // eslint-disable-next-line typescript/no-non-null-assertion
-      ret = this.fieldsWithHandlers.get(hashKey)!;
-    } else {
-      ret = {
-        ...(field as FieldState<Fields[Key]>),
-        name: fieldPath,
-        onChange: this.updateField.bind(this, fieldPath),
-        onBlur: this.blurField.bind(this, fieldPath),
-      };
-      this.fieldsWithHandlers.set(hashKey, ret);
+      return this.fieldsWithHandlers.get(field)!;
     }
-    return ret;
+
+    const result = {
+      ...(field as FieldState<Fields[Key]>),
+      name: String(fieldPath),
+      onChange: this.updateField.bind(this, fieldPath),
+      onBlur: this.blurField.bind(this, fieldPath),
+    };
+    this.fieldsWithHandlers.set(field, result);
+    return result;
   };
 
   private updateField<Key extends keyof Fields>(
--- a/packages/react-form-state/src/FormState.ts
+++ b/packages/react-form-state/src/FormState.ts
@@ -1,11 +1,7 @@
 /* eslint-disable no-case-declarations */
 import * as React from 'react';
-import isEqual from 'lodash/isEqual';
-import isArray from 'lodash/isArray';
-import set from 'lodash/set';
-import {memoize, bind} from 'lodash-decorators';
 
-import {mapObject} from './utilities';
+import {mapObject, set, isEqual} from './utilities';
 import {
   FieldDescriptors,
   FieldState,
@@ -93,6 +89,7 @@ export default class FormState<
 
   state = createFormState(this.props.initialValues);
   private mounted = false;
+  private fieldsWithHandlers = new Map<string, any>();
 
   componentDidMount() {
     this.mounted = true;
@@ -133,15 +130,14 @@ export default class FormState<
     });
   }
 
-  @bind()
-  public reset() {
+  public reset = () => {
     return new Promise(resolve => {
       this.setState(
         (_state, props) => createFormState(props.initialValues),
         () => resolve(),
       );
     });
-  }
+  };
 
   private get dirty() {
     return this.state.dirtyFields.length > 0;
@@ -172,8 +168,7 @@ export default class FormState<
     return fieldDescriptors;
   }
 
-  @bind()
-  private async submit(event?: Event) {
+  private submit = async (event?: Event) => {
     const {onSubmit, validateOnSubmit} = this.props;
     const {formData} = this;
 
@@ -212,21 +207,31 @@ export default class FormState<
     } else {
       this.setState({submitting: false, errors});
     }
-  }
+  };
 
-  @memoize()
-  @bind()
-  private fieldWithHandlers<Key extends keyof Fields>(
+  private fieldWithHandlers = <Key extends keyof Fields>(
     field: FieldStates<Fields>[Key],
-    fieldPath: Key,
-  ) {
-    return {
-      ...(field as FieldState<Fields[Key]>),
-      name: fieldPath,
-      onChange: this.updateField.bind(this, fieldPath),
-      onBlur: this.blurField.bind(this, fieldPath),
+    fieldPath: string & Key,
+  ) => {
+    const hashKey = `${fieldPath}:${JSON.stringify(field)}`;
+    let ret: FieldState<Fields[Key]> & {name: Key} & {
+      onChange(newValue: any): void;
+      onBlur(): void;
     };
-  }
+    if (this.fieldsWithHandlers.has(hashKey)) {
+      // eslint-disable-next-line typescript/no-non-null-assertion
+      ret = this.fieldsWithHandlers.get(hashKey)!;
+    } else {
+      ret = {
+        ...(field as FieldState<Fields[Key]>),
+        name: fieldPath,
+        onChange: this.updateField.bind(this, fieldPath),
+        onBlur: this.blurField.bind(this, fieldPath),
+      };
+      this.fieldsWithHandlers.set(hashKey, ret);
+    }
+    return ret;
+  };
 
   private updateField<Key extends keyof Fields>(
     fieldPath: Key,
@@ -446,7 +451,7 @@ function runValidator<T, F>(
     return validate(value, fields);
   }
 
-  if (!isArray(validate)) {
+  if (!Array.isArray(validate)) {
     // eslint-disable-next-line consistent-return
     return;
   }
--- a/packages/react-form-state/src/components/List.tsx
+++ b/packages/react-form-state/src/components/List.tsx
@@ -1,6 +1,4 @@
 import * as React from 'react';
-import get from 'lodash/get';
-import {memoize, bind} from 'lodash-decorators';
 
 import {FieldDescriptor, FieldDescriptors, ValueMapper} from '../types';
 import {mapObject, replace} from '../utilities';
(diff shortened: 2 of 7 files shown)


packages/react-form-state/src/components/Nested.tsx
```diff
@@ -49,7 +54,7 @@ export default class Nested<Fields> extends React.Component<
           name: `${name}.${fieldPath}`,
           initialValue: initialFieldValue,
           dirty: value !== initialFieldValue,
-          error: get(error, fieldPath),
+          error: error[fieldPath],
```
tzvipm: I don't understand why we were using `get` here. doesn't mapObject just give us the root keys? cc @TheMallen

packages/react-form-state/src/FormState.ts
```diff
@@ -133,15 +130,14 @@ export default class FormState<
     });
   }
 
-  @bind()
-  public reset() {
+  public reset = () => {
```
tzvipm: bind can be replaced with class members and arrow functions.

packages/react-form-state/src/FormState.ts
```diff
@@ -212,21 +207,31 @@ export default class FormState<
...
-    fieldPath: Key,
-  ) {
-    return {
-      ...(field as FieldState<Fields[Key]>),
-      name: fieldPath,
-      onChange: this.updateField.bind(this, fieldPath),
-      onBlur: this.blurField.bind(this, fieldPath),
+    fieldPath: string & Key,
+  ) => {
+    const hashKey = `${fieldPath}:${JSON.stringify(field)}`;
```
tzvipm: memoize can be replaced by storing a private cache for return values, in the form of a map, and caching based on a computed hash key.

tzvipm: Is the stringify here a bit too much? Is there another way we can compute this hash key?

packages/react-form-state/src/FormState.ts
```diff
@@ -446,7 +451,7 @@ function runValidator<T, F>(
     return validate(value, fields);
   }
 
-  if (!isArray(validate)) {
+  if (!Array.isArray(validate)) {
```
tzvipm: not sure why we needed lodash for this in the past.

packages/react-form-state/src/FormState.ts
```diff
@@ -1,11 +1,7 @@
 /* eslint-disable no-case-declarations */
 import * as React from 'react';
-import isEqual from 'lodash/isEqual';
-import isArray from 'lodash/isArray';
-import set from 'lodash/set';
-import {memoize, bind} from 'lodash-decorators';
 
-import {mapObject} from './utilities';
+import {mapObject, set, isEqual} from './utilities';
```
tzvipm: these function the same as the lodash equivalents, for the purposes of our app.

packages/react-form-state/src/components/List.tsx
```diff
@@ -45,14 +45,14 @@ export default class List<Fields> extends React.Component<
...
-          const initialFieldValue = get(initialValue, [index, fieldPath]);
+          const initialFieldValue = initialValue[index][fieldPath];
           return {
             value,
             onBlur,
             name: `${name}.${index}.${fieldPath}`,
             initialValue: initialFieldValue,
             dirty: value !== initialFieldValue,
-            error: get(error, [index, fieldPath]),
+            error: error && error[index] && error[index][fieldPath],
```
tzvipm: `get` used to handle undefined scenarios.

packages/react-form-state/src/components/tests/List.test.tsx
```diff
@@ -8,7 +8,7 @@ import {lastCallArgs} from '../../tests/utilities';
 import FormState from '../..';
 
 describe('<FormState.List />', () => {
-  it('passes form state into child function for each index of the given array', () => {
+  it.only('passes form state into child function for each index of the given array', () => {
```
tzvipm: Oops. Need to :fire: this.

packages/react-form-state/src/utilities.ts
```diff
@@ -27,3 +25,64 @@ export function replace<T>(array: T[], targetIndex: number, newValue: T) {
...
+    if (baseline === null) {
+      return false;
+    }
+    const keysInValue = Object.keys(value);
+    const keysInBaseline = Object.keys(baseline);
+    if (keysInValue.length !== keysInBaseline.length) {
+      return false;
+    }
+    for (const key of keysInValue) {
+      if (!isEqual(value[key], baseline[key])) {
```
tzvipm: could this cause problems if `value` has a key that is explicitly `undefined`? Do I need to check all of the keys in `baseline` too?

tzvipm: Or else I suppose I could skip a bunch of this by simply calling `isEqual` on the keys 🤔.

packages/react-form-state/src/utilities.ts
```diff
@@ -27,3 +25,64 @@ export function replace<T>(array: T[], targetIndex: number, newValue: T) {
...
+        return false;
+      }
+    }
+    return true;
+  }
+  return value === baseline;
```
TheMallen: small thing, but we might want to put this check first to speed up prop comparisons where the object is reference equal across renders

packages/react-form-state/src/utilities.ts
```diff
@@ -27,3 +25,64 @@ export function replace<T>(array: T[], targetIndex: number, newValue: T) {
...
+  } else {
+    const [current, ...rest] = path;
+    return {
+      ...(lhs as any),
+      [current]: set(lhs[current], rest, value),
+    } as InputType;
+  }
+}
+
+export function isEqual(value, baseline) {
```
TheMallen: Aren't these implicit anys? Shouldn't we type it `any`, `any`?

packages/react-form-state/src/FormState.ts
lemonmade: Code before likely used a `WeakMap`, now this is almost certainly leaky.

packages/react-form-state/src/FormState.ts
```diff
@@ -212,21 +207,31 @@ export default class FormState<
...
+  private fieldWithHandlers = <Key extends keyof Fields>(
     field: FieldStates<Fields>[Key],
-    fieldPath: Key,
-  ) {
-    return {
-      ...(field as FieldState<Fields[Key]>),
-      name: fieldPath,
-      onChange: this.updateField.bind(this, fieldPath),
-      onBlur: this.blurField.bind(this, fieldPath),
+    fieldPath: string & Key,
```
lemonmade: why?

packages/react-form-state/src/FormState.ts
```diff
@@ -212,21 +207,31 @@ export default class FormState<
...
+    let ret: FieldState<Fields[Key]> & {name: Key} & {
+      onChange(newValue: any): void;
+      onBlur(): void;
     };
-  }
+    if (this.fieldsWithHandlers.has(hashKey)) {
+      // eslint-disable-next-line typescript/no-non-null-assertion
+      ret = this.fieldsWithHandlers.get(hashKey)!;
+    } else {
+      ret = {
```
lemonmade: `ret` is not a word

packages/react-form-state/src/FormState.ts
lemonmade: Hmm actually maybe not if this is an instance member on the class...

tzvipm: Ooh, interesting...

packages/react-form-state/src/FormState.ts
```diff
@@ -93,6 +89,7 @@ export default class FormState<
 
   state = createFormState(this.props.initialValues);
   private mounted = false;
+  private fieldsWithHandlers = new Map<string, any>();
```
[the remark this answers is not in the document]
tzvipm: I don't think so. I need a second type for the `Map` values and not all the values will have the same type. suggestions?

packages/react-form-state/src/FormState.ts
```diff
@@ -212,21 +207,31 @@ export default class FormState<
...
+    fieldPath: string & Key,
+  ) => {
+    const hashKey = `${fieldPath}:${JSON.stringify(field)}`;
+    let ret: FieldState<Fields[Key]> & {name: Key} & {
+      onChange(newValue: any): void;
+      onBlur(): void;
     };
-  }
+    if (this.fieldsWithHandlers.has(hashKey)) {
+      // eslint-disable-next-line typescript/no-non-null-assertion
```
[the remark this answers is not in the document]
tzvipm: Agreed.

packages/react-form-state/src/FormState.ts
```diff
@@ -212,21 +207,31 @@ export default class FormState<
...
+      ret = this.fieldsWithHandlers.get(hashKey)!;
```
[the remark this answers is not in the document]
tzvipm: I could do that :+1:

packages/react-form-state/src/utilities.ts
```diff
@@ -27,3 +25,64 @@ export function replace<T>(array: T[], targetIndex: number, newValue: T) {
     return newValue;
   });
 }
+
+export function set<InputType extends Object>(
+  lhs: InputType,
```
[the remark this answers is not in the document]
tzvipm: suggestions? Essentially this is calculating the object in which to set the associated key. Maybe `parentObject`?

packages/react-form-state/src/utilities.ts
```diff
@@ -27,3 +25,64 @@ export function replace<T>(array: T[], targetIndex: number, newValue: T) {
...
+    return lhs;
+  } else if (path.length === 1) {
+    return {
+      ...(lhs as any),
+      [path[0]]: value,
+    };
+  } else {
+    const [current, ...rest] = path;
+    return {
+      ...(lhs as any),
```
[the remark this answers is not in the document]
tzvipm: I see. I made this immutable on purpose, because we're using this is `setState` and returning the value.

packages/react-form-state/src/utilities.ts
tzvipm: interesting. So we return `true` early if they are reference-equal and objects?

packages/react-form-state/src/FormState.ts
tzvipm: Typescript didn't like it without this. I forgot to investigate after getting the build to work; thanks for catching.

packages/react-form-state/src/FormState.ts
tzvipm: I don't think `WeakMap` would work here, TBH. We need to cache handlers based only on the field path, but still ensure that pure components receive the exact same object reference unless the field data changes. I'll think about this a bit more and try to find a more efficient solution. @TheMallen, what should we do in a case where the field state is `deepEqual`, but the reference is different? Should we just pass through the object? OR should we return a cached reference as per the test suite (wrt. re-rendering PureComponents)?

packages/react-form-state/src/FormState.ts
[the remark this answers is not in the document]
TheMallen: They should have a generic type no?

packages/react-form-state/src/utilities.ts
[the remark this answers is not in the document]
TheMallen: makes sense

packages/react-form-state/src/utilities.ts
TheMallen: well, reference equal at all I should think

packages/react-form-state/src/FormState.ts
TheMallen: `result` or `resultField`?

packages/react-form-state/src/utilities.ts
tzvipm: Ah, yes. But if they're objects and not reference equal, they may still be deep equal, whereas if they're not objects, then strict equality is the same as deep equality. I'll change this shortly.

packages/react-form-state/src/FormState.ts
TheMallen: - we don't need to cache them based on field path (the old memoize was based on the `field` object itself as it was the first param)
- don't worry about deep equality, we control the way this stuff is updated and it's all done immutably

packages/react-form-state/src/components/List.tsx
```diff
@@ -45,14 +45,14 @@ export default class List<Fields> extends React.Component<
       const innerFields: FieldDescriptors<Fields> = mapObject(
         fieldValues,
         (value, fieldPath) => {
-          const initialFieldValue = get(initialValue, [index, fieldPath]);
+          const initialFieldValue = initialValue[index][fieldPath];
```
axelerator: This is not a transparent replacement.
`get(initialValue, [index, fieldPath])` this returns `undefined`
whereas 
`initialValue[index][fieldPath];` throws an exception 
if `initialValue[index]` yields undefined

PR #477, opened 2019-01-21
Unified effect/ extract

lemonmade:
This PR updates `react-effect` and its consuming libraries to a new pattern for extracting stuff on the server, heavily inspired by [Apollo](https://github.com/apollographql/react-apollo/blob/master/src/getDataFromTree.ts): it just keeps rendering the app to string, collecting promises as it goes, and resolving + doing it again when there are still promises queued for a given pass.

This is a much dumber approach than the tree walk we had before, but means that we don't have to wait for `react-tree-walker` to update and support modern React features like the context API/ hooks. It also means that we can piggy back on top of Apollo and use the same tree traversals for it, which is exactly what I've done with `react-effect-apollo`, a bridging library that provides the necessary context (and renders the necessary Effect) to collect and resolve Apollo queries in a single pass.

While not strictly necessary, I took the change of what I consider to be a breaking change to update `react-effect` to no longer run effects on `componentDidMount`. I immediately regretted that I made it do both, and this will be more confusing when hooks come in and have a better solution to the client-side part of this. I updated the one real user of the client-side effects, `react-html`, to just use a nicer custom component for itself. I also changed the `kind` you can provide to the Effect component to be an object with optional logic to run between/ after each tree traversal; this is useful for libraries like `react-html` where we want to ditch the computed title and such from earlier tree traversals, and only keep the "last" one.

You can see this stuff in action on this branch of web-proving-ground: https://github.com/Shopify/web-proving-ground/compare/unified-extract


packages/react-effect/README.md
```diff
@@ -60,44 +52,49 @@ async function app(ctx) {
 }
 ```
 
-You may optionally pass a second argument to this function: an array of symbols representing the categories to include in your processing of the tree (matching the `kind` prop on `Extract` components).
+You may optionally pass an options object that contains the following keys )all of which are optional):
```
GoodForOneFare:
```suggestion
You may optionally pass an options object that contains the following keys (all of which are optional):
```

packages/react-effect/src/server.tsx
```diff
@@ -0,0 +1,60 @@
+import * as React from 'react';
+import {renderToStaticMarkup} from 'react-dom/server';
```
TheMallen: don't we want to use `renderToString` so that we can hydrate the result?

tzvipm: This is just for extraction, right? So we don't actually render the result. That said, using `renderToString` is a neat idea, since it does eliminate a single tree traversal. But is it as performant per traversal?

packages/react-effect/documentation/migrating-version-1-to-2.md
```diff
@@ -0,0 +1,53 @@
+# Migrating from `@shopify/react-effect@1.x` to `@shopify/react-effect@1.x`
```
tzvipm:
```suggestion
# Migrating from `@shopify/react-effect@1.x` to `@shopify/react-effect@2.x`
```

packages/react-effect/package.json
```diff
@@ -23,10 +23,11 @@
   },
   "homepage": "https://github.com/Shopify/quilt/blob/master/packages/react-effect/README.md",
   "dependencies": {
-    "react-tree-walker": "^4.3.0",
+    "@shopify/useful-types": "^1.1.0",
```
tzvipm: `react-dom` needs to be here, right? Or is it a peer dependency?

packages/react-effect/README.md
```diff
@@ -60,44 +52,49 @@ async function app(ctx) {
...
-    // When implementing your own version of this, you should
-    // implement your own check for the effect "kind". The
-    // Effect component does this automatically.
-    if (
-      include === true ||
-      (Array.isArray(include) && include.includes(EFFECT_ID))
-    ) {
-      this.setState({extracting: true});
-    }
+    await extract(app, {
```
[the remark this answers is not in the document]
TheMallen: I wonder if we should call it something else, I don't find `render` very obvious.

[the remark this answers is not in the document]
lemonmade: Any suggestions? Maybe `decorate`?

packages/react-effect/src/server.tsx
lemonmade: Yeah, I think this is a reasonable API since it matches Apollo, and gives the flexibility to the user to choose how to handle it.

packages/react-effect/package.json
lemonmade: Good catch, need to add it to peer dependencies

PR #479, opened 2019-01-23
Deprecate react-preconnect in favour of react-html

lemonmade:
This PR adds a `<Preconnect />` component to `react-html`, which obviates the need for `react-preconnect`. This package was still referencing Helmet which we know to be a dangerous, non-thread-safe package, so I think it's fine to remove it, and one less import is always nice.

### DIFF
--- a/packages/react-html/README.md
+++ b/packages/react-html/README.md
@@ -228,7 +228,7 @@ Renders a `<link />` tag with the necessary props to specify a favicon. Accepts 
 
 ### `<Preconnect />`
 
-Renders a `<link />` tag with the necessary props to fully preconnect to the host specified by the `source` prop.
+Renders a `<link />` tag that preconnects the browser to the host specified by the `source` prop. You can read more about preconnecting on [Google’s guide to resource prioritization](https://developers.google.com/web/fundamentals/performance/resource-prioritization#preconnect).
 
 ### `<Serialize />`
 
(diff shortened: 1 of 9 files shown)


packages/react-html/src/components/tests/Preconnect.test.tsx
```diff
@@ -0,0 +1,20 @@
+import * as React from 'react';
+import {mount} from 'enzyme';
+
+import Link from '../Link';
+import Preconnect from '../Preconnect';
+
+describe('<Preconnect />', () => {
+  it('renders a <Link /> with DNS prefetch and preconnect props', () => {
```
cartogram: `renders a <Link /> with a rel prop set to DNS prefetch and preconnect` is more accurate IMO.

packages/react-html/README.md
```diff
@@ -226,6 +226,10 @@ This component accepts a string child, which will be used to set the title of th
 
 Renders a `<link />` tag with the necessary props to specify a favicon. Accepts a `source` property that should be the image source for the favicon.
 
+### `<Preconnect />`
+
+Renders a `<link />` tag with the necessary props to fully preconnect to the host specified by the `source` prop.
```
cartogram: Just a thought: I'm on the fence if the word `necessary` is too vague and we should just say what it does. I understand you are trying to shield the implementation details to the consumer, which is why I'm on the fence.

PR #480, opened 2019-01-24
🤖 enable clabot

TheMallen:
Apparently this is a thing we should do, check slack for context or ping me.

### DIFF
--- a/.github/probots.yml
+++ b/.github/probots.yml
@@ -1,0 +1,2 @@
+enabled:
+  - cla


.github/probots.yml
```diff
@@ -0,0 +1,2 @@
+enabled:
+  - cla
```
tzvipm: what are the implications of this? Do we need to enable or change any checks in our github repo settings?

PR #481, opened 2019-01-25
Adds metadata components

cartogram:
This PR adds a few new components for rendering metadata to the `react-html` package.

The new components are:
- `<AppleHomeScreen />`
- `<Responsive />`


README.md
```diff
@@ -34,6 +34,7 @@ Each package has its own `README` and documentation describing usage.
 | koa-metrics | [directory](packages/koa-metrics) | [![npm version](https://badge.fury.io/js/%40shopify%2Fkoa-metrics.svg)](https://badge.fury.io/js/%40shopify%2Fkoa-metrics) |
 | koa-shopify-auth | [directory](packages/koa-shopify-auth) | [![npm version](https://badge.fury.io/js/%40shopify%2Fkoa-shopify-auth.svg)](https://badge.fury.io/js/%40shopify%2Fkoa-shopify-auth) |
 | koa-shopify-graphql-proxy | [directory](packages/koa-shopify-graphql-proxy) | [![npm version](https://badge.fury.io/js/%40shopify%2Fkoa-shopify-graphql-proxy.svg)](https://badge.fury.io/js/%40shopify%2Fkoa-shopify-graphql-proxy) |
+| koa-shopify-webhooks | [directory](packages/koa-shopify-webhooks) | [![npm version](https://badge.fury.io/js/%40shopify%2Fkoa-shopify-webhooks.svg)](https://badge.fury.io/js/%40shopify%2Fkoa-shopify-webhooks) |
```
cartogram: Unrelated change.

packages/react-html/README.md
```diff
@@ -230,6 +230,18 @@ Renders a `<link />` tag with the necessary props to specify a favicon. Accepts
 
 Renders a `<link />` tag that preconnects the browser to the host specified by the `source` prop. You can read more about preconnecting on [Google’s guide to resource prioritization](https://developers.google.com/web/fundamentals/performance/resource-prioritization#preconnect).
 
+### `<Baseline />`
```
TheMallen: I don't really like this name, especially since we have the concept of `baseline` bundles in sewing-kit. Actually I'm not sure we want this to be a component. I sort of feel like the basic meta tags should just always be rendered by `HTML` and then `title` and `favicon` can either be used directly or passed as props to `HTML` itself.

packages/react-html/README.md
```diff
@@ -230,6 +230,18 @@ Renders a `<link />` tag with the necessary props to specify a favicon. Accepts
...
+
+Renders a minimal set of `<Meta />` tags to specify basic app information. Accepts a `title` property that is passed as the children of a `<Title />` component, and a `favicon` source that is passed to a `<Favicon />` component.
+
+### `<Responsive />`
```
TheMallen: I'm also not sure we ever don't want these

packages/react-html/README.md
lemonmade: I like Mal's idea here 👍

packages/react-html/README.md
lemonmade: This one I think is good, the incantation to get the right meta tags is always a pain in the but IMO

packages/react-html/src/components/tests/AppleHomeScreen.test.tsx
```diff
@@ -0,0 +1,73 @@
...
+        content: 'yes',
+      }),
+    ).toHaveLength(1);
+  });
+
+  it('renders a <Meta /> with `name="apple-mobile-web-app-status-bar-style" content="black"` by default', () => {
+    const appleHomeScreen = mount(<AppleHomeScreen />);
+
+    expect(
+      findMeta(appleHomeScreen.find(Meta), {
```
lemonmade: Feels a bit awkward to have `findMeta(thing.find(Meta)`, maybe just `findMeta(thing, `?

cartogram:
@TheMallen Updated:
1. removed `<Baseline />` altogether and merged with `<HTML />` using the title prop as a fallback if one was not extracted from the page.
2. fixed up the `findMeta` utility and moved it into the `AppleHomeScreen` component file since that is the only place it is used (since I remove `<Baseline />`)

packages/react-html/README.md
TheMallen: Yeah, I just mean do we ever want to not have this? Why not just have the component always include them

PR #483, opened 2019-01-25
🐛 [react-form-state] Fix isEqual / dirty state for dates

tzvipm:


packages/react-form-state/src/tests/utilities.test.ts
```diff
@@ -0,0 +1,46 @@
+import {isEqual} from '../utilities';
```
tzvipm: I could get rid of these, but I wrote them before I found `fast-deep-equal`, and it makes it easier if we later switch to our own implementation.

TheMallen: I don't think it's useful to test the other library, we can just get these back from git if we ever switch. If you want to add a test to make sure we never break dates again I'd rather see it at the level of testing that the dirty state behaviour is working for them.

PR #486, opened 2019-01-29
Introduce unformatCurrency utility

ismail-syed:
This PR introduces a utility to return the unformatted normalized version of formatted money input. Eg. "$1,234.50" normalizes to "1234.50". This is quite trivial for locales where `,` are used for integer groupings and `.` are used for decimals but gets a bit difficult when integer group symbols or decimal symbols can be `,` or  `.` or  `'` or ` `.  - [Use case](https://github.com/Shopify/web/pull/10077)


packages/react-i18n/src/i18n.ts
```diff
@@ -227,6 +262,19 @@ export default class I18n {
       });
     }
   }
+
+  private decimalSymbol(currencyCode: string) {
+    const {symbol} = this.getCurrencySymbolLocalized(this.locale, currencyCode);
+    const templatedInput = Number('1000.00');
```
ismail-syed: A bit of a clever-ish trick to get the decimal symbol for the respective locale. Ideally, we would want to use [`Intl.Numberformat.formatToParts()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NumberFormat/formatToParts) but it's not supported on all majore browsers and the [polyfill](https://github.com/tc39/proposal-intl-formatToParts/blob/master/polyfill.js) seems to only provide support for `Intl.DateTimeFormat.formatToParts`

packages/react-i18n/src/test/i18n.test.ts
```diff
@@ -346,6 +346,42 @@ describe('I18n', () => {
     });
   });
 
+  describe('#unformatCurrency()', () => {
```
ismail-syed: TODO: Assuming reviewers are okay with the approach in the utilities, add tests for more unique locales.

packages/react-i18n/src/i18n.ts
```diff
@@ -150,6 +155,36 @@ export default class I18n {
...
   }
 
+  unformatCurrency(input: string, currencyCode: string): string {
+    const nonDigits = /\D/g;
+    const decimal = this.decimalSymbol(currencyCode);
+    const lastDecimalIndex = input.lastIndexOf(decimal);
+
+    if (decimal === DECIMAL_NOT_SUPPORTED) {
+      const amount = input.replace(nonDigits, '');
+      return `${amount}.${DECIMAL_VALUE_FOR_CURRENCIES_WITHOUT_DECIMALS}`;
```
nimzco: Should we even return a decimal if they aren't supported by the currency?

ismail-syed: IMO, yes. The intention of `unformatCurrency` is to provide a standard representation of the formatted currency so it can then be used to send off a back end, or even back to`i18n.formatCurrency`. This is what we do in rails today. If a UI component formats a price to `12` in `JPY`, we store it as `12.00` in the backend.

packages/react-i18n/src/i18n.ts
```diff
@@ -150,6 +155,36 @@ export default class I18n {
...
+    const decimalValue = input
+      .substring(lastDecimalIndex + 1)
+      .replace(nonDigits, '');
+
+    const normalizedValue = `${integerValue}${normalizedDecimal}${decimalValue}`;
+    const invalidValue = normalizedValue === '' || normalizedValue === '.';
+
+    return invalidValue
+      ? DEFAULT_VALUE
+      : parseFloat(normalizedValue).toFixed(2);
```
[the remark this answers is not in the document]
nimzco: We could store that data on the constant file maybe? I've created a JSON file with `Currency => number of decimals`:
https://gist.github.com/nimzco/37178863c096bc863243b5d92b3e1fab

packages/react-i18n/src/i18n.ts
nimzco: Oh I see. 
I would argue, why not returning a `number` instead of a `string` then? We would avoid some other problems, like the `toFixed(2)`?

packages/react-i18n/src/i18n.ts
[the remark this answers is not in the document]
ismail-syed: Remove `toFixed` then?

packages/react-i18n/src/i18n.ts
ismail-syed: My understanding is that we store all our money inputs as strings. I believe `toFixed(2)` conversation has less to do with a string/number issue and more to do with knowing how many decimals places the consumer wants.

packages/react-i18n/src/i18n.ts
```diff
@@ -227,6 +262,19 @@ export default class I18n {
...
     }
   }
+
+  private decimalSymbol(currencyCode: string) {
+    const {symbol} = this.getCurrencySymbolLocalized(this.locale, currencyCode);
+    const templatedInput = Number('1000.00');
+    const formattedInput = this.formatCurrency(templatedInput, {
+      currency: currencyCode,
+    }).replace(symbol, '');
+    const decimal = formattedInput.charAt(formattedInput.length - 3);
```
Korri: As Nima mentioned about toFixed(2), this won't work either for some currencies for example:
`new Intl.NumberFormat('ar', { style: 'currency', currency: 'JOD'}).format('1000.00');` => `د.أ.‏ 1,000.000`

packages/react-i18n/src/i18n.ts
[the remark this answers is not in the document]
ismail-syed: Ouu, that `decimals.json` is super neat. But we also need to understand how many decimal places based on the locale. Looking at the [`mdn examples`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NumberFormat) `en-IN` is uses 3 decimal places.

packages/react-i18n/src/i18n.ts
nimzco: This also won't work for JPY because there is no decimal delimiter for that currency.
What about something like this:
- Use `templatedInput = 1`, not 1000 to avoid edge cases and
```js
new Intl.NumberFormat('en', { style: 'currency', currency: currencyCode})
  .format('1')
  .replace('1', '')
  .replace(currencyCode, '')
  .replace(/\s/, '')
```

And then get the first character?

- For JPY it should be undefined
- For JOD, it should be `.` since you'll have `.000` left.

Does it make sense?

We might want to try with arabic as well because the currency code will be `right-to-left` which will place the `.` at the right of the decimals instead of at the left... 🤦‍♂️

packages/react-i18n/src/i18n.ts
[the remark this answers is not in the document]
clauderic: At minimum you need the currency's precision (number of decimal) and the decimal mark in order to parse a string into a number. You can create a json file with that data for all currencies as suggested by @nimzco, though ideally we would have a GraphQL endpoint to only fetch the data for the display currency https://github.com/Shopify/intl-localization/issues/8.

You can take a look at the work I had started on this a while ago, https://github.com/Shopify/money-js/blob/master/src/parse.ts, IIRC it worked for all currencies we support (3 decimal, no decimals, 2 decimals, dot as decimal mark, etc), though I wasn't 100% happy with the code.

We have a fairly exhaustive test suite in core, you should make sure whatever solution you come up with passes all of the test cases https://github.com/Shopify/shopify/blob/master/test/javascripts/admin/unit/money_test.coffee

packages/react-i18n/src/i18n.ts
clauderic: I would recommend avoiding trying to guess the decimal mark, I think it's a fairly error-prone process (at least in my experience) and I'd feel much more confident in a solution that used the metadata we already have on currencies we support

nimzco: True, we went a bit far when we can just use the JSON file with the data since we have the currency code... Thanks for bringing us back to the light Clauderic 🙏

ismail-syed: 😞 I had a feeling. I would really like to use `Intl.NumberFormat.formatToParts` instead but can't seem to find a polyfill for it.

packages/react-i18n/src/i18n.ts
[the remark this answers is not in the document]
nimzco: We can probably start by embedding the `json` file with only the data we want and revisit later when the endpoint is created. Also, the downside of the endpoint is that the method will have to be `async` :/ 

TLDR; let's use a mimum json config file with just the data we need. :)

packages/react-i18n/src/i18n.ts
clauderic: IIRC, I think the only danger with relying on the currency metadata is that for some currencies, depending on the locale, the decimal separator is not the same. For instance, EUR can either be `,` or `.`:

screen shot 2019-01-29 at 11 39 12 am

clauderic: So you may need to guess the decimal mark after all :/

ismail-syed: > So you may need to guess the decimal mark after all :/

"guessing" the decimal was really my least preferred choice.

nimzco: Oh true...
But looking at the code, it seems that you use that method only to see if the currency supports decimal, so why not have a function that just does that?
Be like: 

```js
function supportsDecimal(currencyCode) {
  returns !CURRENCIES_WITH_NO_DECIMALS.includes(currencyCode)
}
```

ismail-syed:
> TLDR; let's use a mimum json config file with just the data we need. :)

I tested this out in my branch in /web using the manifest approach but this get's increasingly more difficult with different locales. 

We can convert all the locales for all the currencies in a manifest so we'll need sensible defaults for specific edgecases. 

```
const number = 123456.789;

console.log('en-US: ', new Intl.NumberFormat('en-US', { style: 'currency', currency: 'USD' }).format(number));
console.log('fr-FR: ', new Intl.NumberFormat('fr-FR', { style: 'currency', currency: 'USD' }).format(number));
console.log('de-DE: ', new Intl.NumberFormat('de-DE', { style: 'currency', currency: 'USD' }).format(number));
console.log('de-CH: ', new Intl.NumberFormat('de-CH', { style: 'currency', currency: 'USD' }).format(number));
```

```
> "en-US: " "$123,456.79"
> "fr-FR: " "123 456,79 $US"
> "de-DE: " "123.456,79 $"
> "de-CH: " "$ 123'456.79"
```

Korri:
Looks like we have to use the solution @nimzco suggested: 
> This also won't work for JPY because there is no decimal delimiter for that currency.
> What about something like this:
> 
> * Use `templatedInput = 1`, not 1000 to avoid edge cases and
> 
> ```js
> new Intl.NumberFormat('en', { style: 'currency', currency: currencyCode})
>   .format('1')
>   .replace('1', '')
>   .replace(currencyCode, '')
>   .replace(/\s/, '')
> ```
>
> [...]

We could probably do something similar for the thousand separator.

packages/react-i18n/src/utilities/decimals.json
```diff
@@ -0,0 +1,174 @@
+{
```
nimzco: Can we make it a Map to follow what we have in the [`constants` file](https://github.com/Shopify/quilt/blob/master/packages/react-i18n/src/constants.ts#L42)?
And could you move that file into a `constant` folders actually? 🙏

packages/react-i18n/src/i18n.ts
```diff
@@ -179,10 +180,11 @@ export default class I18n {
 
     const normalizedValue = `${integerValue}${normalizedDecimal}${decimalValue}`;
     const invalidValue = normalizedValue === '' || normalizedValue === '.';
+    const decimalPlaces = decimals[currencyCode.toUpperCase()];
```
nimzco: Maybe have a fallback here, just in case?

packages/react-i18n/src/test/i18n.test.ts
```diff
@@ -346,6 +346,52 @@ describe('I18n', () => {
...
+
+      it('handles prefix without a space', () => {
+        getCurrencySymbol.mockReturnValue(mockSymbolResult);
+
+        const i18n = new I18n(defaultTranslations, defaultDetails);
+        expect(i18n.unformatCurrency('$1,234.50', 'USD')).toBe('1234.50');
+      });
+    });
+
+    it('unformats formatted input to 2 digits', () => {
```
nimzco: Can you add tests with currencies that do not have digits (JPY) and some that have 3 digits (JOD)? 🙏

packages/react-i18n/src/i18n.ts
```diff
@@ -31,6 +33,11 @@ export interface TranslateOptions {
   scope: RootTranslateOptions<any>['scope'];
 }
 
+// Used for currecies that don't use fractional units (eg. JPY)
+const DECIMAL_NOT_SUPPORTED = 'N/A';
+const DECIMAL_VALUE_FOR_CURRENCIES_WITHOUT_DECIMALS = '00';
+const DEFAULT_VALUE = '';
```
nimzco: I realize the name `DEFAULT_VALUE` is too generic here since it's defined in the `i18n.ts` file. Maybe you don't need that specific constant and can just return `''` in your method.

packages/react-i18n/src/i18n.ts
```diff
@@ -150,6 +156,39 @@ export default class I18n {
     return this.formatNumber(amount, {as: 'currency', ...options});
   }
 
+  unformatCurrency(input: string, currencyCode: string): string {
+    const nonDigits = /\D/g;
```
nimzco: There is one problem I see with this: Japanese full-width characters/numbers. See: https://stackoverflow.com/questions/46290595/japanese-regex-for-numbers

But that's another issue, I definitely won't block this PR because of this. It's more of an FYI/TIL 👍

ismail-syed:
Thanks @nimzco! I'm going to do one last smoke test in /web for my own sanity. Will merge and bump a release once I've verified this.

ismail-syed:
Testing in /web worked. Merging on  ✅

PR #491, opened 2019-01-31
Initial async package

lemonmade:
This is the first step in a three-part process of introducing our own, super-smart async component management. This PR introduces the build-ey parts; a custom babel plugin, and a Webpack plugin. These are largely copied from `react-loadable` and the work we added on top in [sewing-kit](https://github.com/Shopify/sewing-kit/tree/master/packages/webpack-react-loadable-plugin).

I've included this package in Quilt because it also exposes a few types I will use in the react parts of this. I also skipped tests for now because these are basically copied and because we have no great way of testing these things :P

Here's the steps going forward:

* Ship `@shopify/async`
* Integrate the Babel and Webpack parts of this package into sewing-kit
* Ship `@shopify/react-async` and `@shopify/react-graphql`, which will rely on this package to properly load scripts on server and client


packages/async/webpack.d.ts
```diff
@@ -0,0 +1 @@
+export * from './dist/webpack-plugin';
```
tzvipm: dist is right here?

tzvipm:
https://github.com/babel-utils/babel-plugin-tester wouldn’t work for this?

packages/async/src/babel-plugin.ts
```diff
@@ -0,0 +1,143 @@
...
+
+        const source = path.node.source.value;
+
+        const processImports = processPackages.get(source) || [];
+
+        if (processImports.length === 0) {
+          return;
+        }
+
+        const factoryImports = path.get('specifiers').filter(specifier => {
```
tzvipm: Sorry, maybe I'm blind, but where does `path` come from here?

packages/async/src/babel-plugin.ts
```diff
@@ -0,0 +1,143 @@
+import * as Types from '@babel/types';
+import {NodePath, Binding} from '@babel/traverse';
+
+const DEFAULT_PROCESS_PACKAGES = {
```
tzvipm: are these packages to process or packages *that* process? Maybe call them `processorPackageNames`?

packages/async/src/babel-plugin.ts
tzvipm: what is a factory import?

packages/async/src/babel-plugin.ts
```diff
@@ -0,0 +1,143 @@
...
+          if (binding != null) {
+            processBinding(binding, t);
+          }
+        }
+      },
+    },
+  };
+}
+
+function processBinding(binding: Binding, t: typeof Types) {
```
tzvipm: what is a binding? How are we processing it?

packages/async/package.json
```diff
@@ -0,0 +1,37 @@
...
+  "homepage": "https://github.com/Shopify/quilt/blob/master/packages/async/README.md",
+  "dependencies": {},
+  "devDependencies": {
+    "@babel/types": "^7.3.0",
+    "@types/babel__traverse": "^7.0.5",
+    "@types/webpack": "^4.4.24",
+    "typescript": "~3.0.1"
+  },
+  "files": [
+    "dist/*"
```
ismail-syed: We'll need to include the `webpack.js`, and `babel.js` in here as well.

packages/async/src/babel-plugin.ts
```diff
@@ -0,0 +1,143 @@
...
+  '@shopify/react-async': ['createAsyncComponent'],
+  '@shopify/react-graphql': ['createAsyncQueryComponent'],
+};
+
+interface State {
+  processPackages?: Map<string, string[]>;
+  opts?: {packages?: {[key: string]: string[]}};
+}
+
+export default function asyncBabelPlugin({types: t}: {types: typeof Types}) {
```
ismail-syed: I had some [tests](https://github.com/Shopify/quilt/blob/678536dd2ef0ab7817687ed5026f2e19473dc3ed/packages/async-chunks/src/lib/babel/tests/index.test.tsx) from when I worked on the babel plugin. Might be worth pulling some of those in.

packages/async/webpack.d.ts
lemonmade: Yes

packages/async/src/babel-plugin.ts
lemonmade: Path is an extremely common term to use for this in babel plugins, it is what [their handbook](https://github.com/jamiebuilds/babel-handbook/blob/master/translations/en/plugin-handbook.md#toc-unit-testing) uses

packages/async/src/babel-plugin.ts
lemonmade: They are packages that get processed, will rename for clarity

packages/async/src/babel-plugin.ts
lemonmade: called it `addIdOption`, but `Binding` is also a common term and has a type definition right next to it.

packages/async/package.json
```diff
@@ -0,0 +1,42 @@
...
+  "devDependencies": {
+    "@babel/core": "^7.2.2",
+    "@babel/plugin-syntax-dynamic-import": "^7.2.0",
+    "@babel/types": "^7.3.0",
+    "@types/babel__core": "^7.0.4",
+    "@types/babel__traverse": "^7.0.5",
+    "@types/webpack": "^4.4.24",
+    "typescript": "~3.0.1"
+  },
+  "files": [
```
ismail-syed: Where did the webpack plugin go?
Is it being added to the [webpack-asset-metadata-plugin](https://github.com/Shopify/sewing-kit/pull/1140/files) instead?

lemonmade: Correct, moved it to this PR: https://github.com/Shopify/sewing-kit/pull/1140

PR #492, opened 2019-01-31
Async React and GraphQL packages

lemonmade:
Sorry for the beast of a PR, but a lot of it is tests and docs. This PR adds two new packages, both based on the new [async package](https://github.com/Shopify/quilt/pull/491):

- `react-async`, which replaces `react-loadable` as our way of creating async components, and also offers much more, like easy preloading/ prefetching based on URL
- `react-graphql`, which sits on top of `react-apollo` and creates more strongly-typed query components, and exposes a way to create query components that asynchronously load their query, making it suitable for use with the prefetching stuff in `react-async`.

All of this is done while fully supporting server side rendering.

The docs do a reasonable job of documenting these packages, so I will leave the description there. My big questions on this PR are just about the naming. I use `Preload` to mean "download just scripts", and `Prefetch` to mean "download scripts and data", but I feel there must be something better I could use to distinguish between those two roles. I also use `Prepare` to mean "register this component to prefetch when the user intends to go to a URL", but I hate that name too.

I'll just quickly show a comprehensive example that illustrates why these are such important components:

```tsx
import {Props} from '@shopify/useful-types';
import {createAsyncComponent, Prepare} from '@shopify/react-async';
import {createAsyncQueryComponent} from '@shopify/react-graphql';

// An "async" Query component, with super strict type checking that is inferred from
// the types we embed in our GraphQL documents!
const ProductDetailsQuery = createAsyncQueryComponent({
  load: () => import('./graphql/ProductDetailsQuery.graphql',
});

// The actual ProductDetails component, but async-ed, and with static components
// to preload, prefetch, and keep it fresh. Of course, all the props are preserved based
// on the inference of our load() method
const ProductDetails = createAsyncComponent({
  load: () => import('./ProductDetails'),
  renderPrefetch: (props: Props<typeof ProductDetailsQuery.Prefetch>) => <ProductDetailsQuery.Prefetch {...props} />,
  renderPreload: () => <ProductDetailsQuery.Preload />,
  renderKeepFresh: (props: Props<typeof ProductDetailsQuery.KeepFresh>) => <ProductDetailsQuery.KeepFresh {...props} />,
});

// The "regular" way
<ProductDetails id="1" />

// On ProductIndex, we might want to just preload all the scripts for that page:
<ProductDetails.Preload />

// On ProductVariantDetails, we might want to keep that product fresh because we are soon to return:
<ProductDetails.KeepFresh id="1" />

// Across the whole admin, we might want to preload the data for a particular product
// when we are hovering over, focused in, or clicking on a link that goes to /products/:id
<Prepare
  component={ProductDetails}
  url={/^\/products\/\d+\/?$}
  mapUrlToProps={(url) => ({id: url.split('/').pop()})}
/>
```


packages/react-async/src/manager.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export const noopManager: Manager = {
+  registered: new Set(),
+  register() {
+    return () => {};
+  },
+  markAsUsed() {},
+};
+
+export class AsyncManager implements Manager {
```
ismail-syed: I personally think this is too vague of a name, maybe `AsyncComponentManager `.

packages/react-async/src/Provider.tsx
```diff
@@ -0,0 +1,24 @@
...
+import {AsyncManager, Manager} from './manager';
+import {Prefetcher} from './Prefetcher';
+import {AsyncContext} from './context';
+
+interface Props {
+  children?: React.ReactNode;
+  manager?: Manager;
+}
+
+export class AsyncProvider extends React.PureComponent<Props> {
```
ismail-syed: Same here, maybe use `AsyncComponentProvider` to let readers know it's related to all `createAsyncComponent` that will be introduced app-wide.

packages/react-async/README.md
```diff
@@ -0,0 +1,147 @@
...
+renderToString(
+  <AsyncProvider manager={asyncManager}>
+    <App />
+  </AsyncProvider>,
+);
+
+const moduleIds = [...asyncManager.used];
+```
+
+These module IDs can be looked up in the manifest created by `@shopify/async`’s Webpack plugin. If you are using [`sewing-kit-koa`](https://github.com/Shopify/quilt/tree/master/packages/sewing-kit-koa), you can follow the instructions from that package to automatically collect the required JavaScript and CSS bundles.
```
ismail-syed: > These module IDs can be looked up in the manifest created by `@shopify/async`’s Webpack plugin.

Might be worth including the default manifest path `build/client/async-manifest.json`.

packages/react-async/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+`createAsyncComponent` is a function for creating components that are loaded asynchronously on initial mount. However, the resulting component does more than just help you split up your application along component lines; it also supports customized rendering for loading, and creates additional components for smartly preloading or preloading the component’s bundle. Best of all, in conjunction with the Babel and Webpack plugins provided by [`@shopify/async`](https://github.com/Shopify/quilt/tree/master/packages/async), you can easily extract the bundles needed to render your application during server side rendering.
+
+To start, import the `createAsyncComponent` function. The simplest use of this function requires just a `load` function, which returns a promise for a component:
+
+```tsx
+import {createAsyncComponent} from '@shopify/react-async';
+
+const MyComponent = createAsyncComponent({
+  load: () => import('./MyComponent'),
```
ismail-syed: `renderLoading` is meant for the skeleton fallback components, right? Might be worth including that in this examples

packages/react-async/README.md
```diff
@@ -0,0 +1,147 @@
...
+});
+
+// All of these are available:
+<MyComponent />
+<MyComponent.Prefetch />
+<MyComponent.Preload />
+<MyComponent.KeepFresh />
+```
+
+By default, `Preload`, `Prefetch`, and `KeepFresh` all simply prefetch the bundle for the component in the background. However, you can provide additional markup to render during these phases with the `renderPreload`, `renderPrefetch`, and `renderKeepFresh` options to `createAsyncComponent`:
```
tzvipm: How do these phases work? Is there an example async lifecycle we could provide?

packages/react-async/README.md
```diff
@@ -0,0 +1,147 @@
...
+- `Prefetch`: loading resources that will be used by the component
+- `Preload`: loading resources **and** data that will be used by the component
+- `KeepFresh`: loading resources and data that will be used by the component, and keeping data up to date
+
+If you want props for your `Preload`, `Prefetch`, or `KeepFresh` components, simply provide them in the `render` option for that component. The resulting components will have those prop types baked in.
+
+```tsx
+const MyComponent = createAsyncComponent({
+  load: () => import('./MyComponent'),
+  renderPreload: ({priority}: {priority: 'high' | 'low'}) => (
```
tzvipm: what does `low` priority mean in this case? Is that more of a `defer` than a `preload`?

packages/react-async/README.md
```diff
@@ -0,0 +1,147 @@
...
+});
+
+// This is a type error, we need a `priority` prop!
+<MyComponent.Preload />;
+
+// Ah, much better!
+<MyComponent.Preload priority="high" />;
+```
+
+This system is designed to work well with our [`@shopify/react-graphql` package](https://github.com/Shopify/quilt/tree/master/packages/react-graphql). Simply create an async GraphQL query using that library, and then `Prefetch`, `Preload`, and `KeepFresh` that component alongside the React component itself:
```
tzvipm: relative link here?

packages/react-async/src/Prefetcher.tsx
```diff
@@ -0,0 +1,176 @@
...
+
+interface Props {
+  manager: Manager;
+}
+
+interface State {
+  url?: string;
+}
+
+export const HOVER_DELAY = 100;
```
tzvipm: what are the units? maybe call this `HOVER_DELAY_MS`?

packages/react-async/src/Provider.tsx
lemonmade: Same answer as above, it is not just for async components.

packages/react-async/README.md
lemonmade: The manifest path depends on Rails versus Node, and is not something we otherwise mention since `sewing-kit-koa` handles it internally

packages/react-async/README.md
lemonmade: not sure what you mean by phases, but the docs below show what we recommend using each one for.

packages/react-async/README.md
lemonmade: this is just an example to show that you can include arbitrary props in `renderPreload` and that it will thread those typings through

packages/react-async/README.md
tzvipm: Neither am I. You mention the word phases  in the readme.

lemonmade: Gotcha, thanks. Removed that part because it wasn't needed 👍

lemonmade:
I am feeling very confident with this and want to get moving on integrating a demo of it in the admin, so I am going to ship this. We can discuss API decisions once we see it in the wild.

PR #494, opened 2019-02-01
Have ExtractedLocation contain null for locales that are not available (instead of undefined)

Korri:
This will allow use to serialize and keep the information as our serializer removed undefined properties.

It will remove the necessity of a hack around it on [web#10270](https://github.com/Shopify/web/pull/10270)

I decided to change only the `ExtractedTranslations` interface as it will minimise the impact on other systems, makes for a much simpler implementation (as `undefined` is the return value in case of failure of a lot of methods like `Map.get`), and its the only place we need `null`.

### DIFF
--- a/packages/react-i18n/src/manager.ts
+++ b/packages/react-i18n/src/manager.ts
@@ -17,7 +17,7 @@ export interface ConnectionResult {
 }
 
 export interface ExtractedTranslations {
-  [key: string]: TranslationDictionary | undefined;
+  [key: string]: TranslationDictionary | null;
 }
 
 export default class Manager {
@@ -26,7 +26,7 @@ export default class Manager {
   }
 
   private subscriptions = new Map<Subscriber, Connection>();
-  private translations = new Map<string, TranslationDictionary | undefined>();
+  private translations = new Map<string, TranslationDictionary | null>();
   private asyncTranslationIds: string[] = [];
   private translationPromises = new Map<string, Promise<void>>();
 
@@ -48,7 +48,7 @@ export default class Manager {
     return this.asyncTranslationIds.reduce<ExtractedTranslations>(
       (extracted, id) => ({
         ...extracted,
-        [id]: this.translations.get(id),
+        [id]: this.translations.get(id) || null,
       }),
       {},
     );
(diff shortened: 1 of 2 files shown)


packages/react-i18n/src/manager.ts
```diff
@@ -32,15 +32,15 @@ export default class Manager {
     initialTranslations: ExtractedTranslations = {},
   ) {
     for (const [id, translation] of Object.entries(initialTranslations)) {
-      this.translations.set(id, translation);
+      this.translations.set(id, translation || undefined);
```
nimzco: Is this really needed?

Korri: Yes, `this.translations` is `TranslationDictonary | undefined` so `null` is not a valid value.

Do you think using
```javascript
this.translations.set(id, translation === null ? undefined : translation);
```
instead be more readable?

nimzco: Oh right I missed that `initialTranslations` was an `ExtractedTranslations`. `|| undefined` is fine with me 👍

lemonmade: I think I would rather change this so that `this.translations` has the same "no translation found" type as we give when you serialize it (`null`)

Korri:
> Just one change I think would make things a bit simpler, then LGTM 👍

Hey @lemonmade, as I mentioned here:

> I decided to change only the ExtractedTranslations interface as it will minimise the impact on other systems, makes for a much simpler implementation (as undefined is the return value in case of failure of a lot of methods like Map.get), and its the only place we need null.

I did do it that way originally, and it makes the code much more complicated (need to add a lot of `|| null` everywhere, or some non-trivial things like changing the [TranslationGetter](https://github.com/Shopify/quilt/blob/dc185ab3626c442c8e7d503a7e860cb9aaf43339/packages/react-i18n/src/connection.ts#L4-L6) interface to return `null`), so I decided against it, but if you still thinks its best I'm ok to do it.

lemonmade:
I don't think any public API needs to change, just how we store them internally. You might have to update where we store them to do `|| null` but I think it makes it conceptually simpler.

PR #502, opened 2019-02-05
Remove context destructuring

cartogram:
This PR removes a few cases where we immediately destructure the `Consumer` and `Provider` from `React.CreateContext`. These to lead to weird bugs in React ^16.6.0.

### DIFF
--- a/packages/react-network/src/context.tsx
+++ b/packages/react-network/src/context.tsx
@@ -1,7 +1,10 @@
 import * as React from 'react';
 import {Manager, NoopManager} from './manager';
 
-const {Consumer, Provider} = React.createContext<Manager>(new NoopManager());
+const Context = React.createContext<Manager>(new NoopManager());
+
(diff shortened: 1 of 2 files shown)


packages/react-network/src/context.tsx
```diff
@@ -1,7 +1,10 @@
 import * as React from 'react';
 import {Manager, NoopManager} from './manager';
 
-const {Consumer, Provider} = React.createContext<Manager>(new NoopManager());
+const Context = React.createContext<Manager>(new NoopManager());
+
+const Consumer = Context.Consumer;
+const Provider = Context.Provider;
```
lemonmade: I think this has the same issue as destructuring, no? I think we need to not export the consumer separately.

cartogram: With further testing, I don't think we need to do anything anymore. I'm going to close this for now.

Issue #500, opened 2019-02-05
ValidateList and ValidateNested functionality is broken (0.7.0 version is OK)


dcugalj:
## Overview

Played with https://codesandbox.io/s/x94lqqkxxq. Validation on options in Variants section does nothing

## Consuming repo

> What repo were you working in when this issue occurred?

...

## Labels

- [ ] Add the `Type: Bug` label to this issue.

### Area

- [ ] Add any relevant `Area: <area>` labels to this issue

### Scope

- Is this issue related to a specific package?

  - [ ] Tag it with the `Package: <package_name>` label.

---

## Checklist

- [ ] Please delete the labels section before submitting your issue
- [ ] I have described this issue in a way that is actionable (if possible)

tzvipm:
Note: 0.7.4 is also working in that codesandbox.

PR #504, opened 2019-02-06
[FormState] Fixes validation when external errors are not provided

darrenhebner:
Fixes: https://github.com/Shopify/quilt/issues/500

## What does this PR do
This PR fixes a bug where inline validation was not working if `externalErrors` are not provided.

## How does it fix it
Updates the equality check that is used to determine if the value of `externalErrors` has changed. If a value was not provided, we use an empty array as a fallback value. Since this is a new array, the equality check was failing when we checked if the new value was the same as the old value.

When this failed, we would return updated fields with no validation errors.

## Proof it works
Here's validation working when external errors are not provided
validation without external error

Here's validation working when external errors _are_ provided
validation with external error

### DIFF
--- a/packages/react-form-state/src/tests/FormState.test.tsx
+++ b/packages/react-form-state/src/tests/FormState.test.tsx
@@ -450,6 +450,32 @@ describe('<FormState />', () => {
       const {errors} = lastCallArgs(renderPropSpy);
       expect(errors).toEqual([...submitErrors, ...externalErrors]);
     });
+
+    it('does not reset fields if externalErrors are not provided', () => {
+      const renderPropSpy = jest.fn(() => null);
+      const errorMessage = faker.lorem.sentence();
+
+      mount(
+        <FormState
+          initialValues={{
+            product: faker.commerce.productName,
+          }}
+          validators={{
+            product: () => errorMessage,
+          }}
+        >
+          {renderPropSpy}
+        </FormState>,
+      );
+
+      const {fields} = lastCallArgs(renderPropSpy);
+      fields.product.onChange(faker.commerce.productName);
+      fields.product.onBlur();
+
+      const {fields: updatedFields} = lastCallArgs(renderPropSpy);
+
+      expect(updatedFields.product.error).toEqual(errorMessage);
+    });
   });
 
   describe('reset()', () => {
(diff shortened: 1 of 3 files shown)


packages/react-form-state/src/tests/FormState.test.tsx
```diff
@@ -450,6 +450,32 @@ describe('<FormState />', () => {
       const {errors} = lastCallArgs(renderPropSpy);
       expect(errors).toEqual([...submitErrors, ...externalErrors]);
     });
+
+    it('does not reset fields if externalErrors are not provided', () => {
```
darrenhebner: I'm a bit stuck with this test. It actually passes with the old equality check and I'm not sure why.

tzvipm:
@darrenhebner I'm also baffled as to why this is passing with the old equality check. When I try in the codesandbox, the errors for all the fields end up being `undefined`. This makes me doubt our test setup. @TheMallen, do you have any thoughts as to why this might be happening?

jamesism:
:wave: was also looking at this issue yesterday (should have checked here first!), but also could not figure out this testing thing. It almost seems as though `getDerivedStateFromProps` isn't being hit on state updates when we just manually invoke `onChange`/`onBlur` from the children as function spy. IMO it may make sense (for simplicity's sake) to drop `getDerivedStateFromProps` and use `componentDidUpdate` for the `initialValues`/`externalErrors` reconciliation.

Anyway, I think there is another thing to be addressed here in the `fieldsWithErrors` function that may be confounding things. As it stands all `error` keys on the existing `Fields` collection will be overwritten, even if `externalErrors` hasn't specified an error for that key. I think we can fix with a fallback in the `error` key assignment:
```
return mapObject(fields, (field, path) => {
    return {
      ...field,
      error: errorDictionary[path] || field.error,
    };
  });

// or an early return
return mapObject(fields, (field, path) => {
  if (!errorDictionary[path]) {
    return field;
  }

  return {
    ...field,
    error: errorDictionary[path],
  };
});
```
I have this locally on this branch but wasn't sure what the etiquette is around pushing commits to other people's branches. Please let me know if I'm missing something about this function or if you'd like me to push that change up to this branch.

darrenhebner:
@jamesism - Hey James, I like your solution for updating `fieldsWithErrors`. That makes sense to me. The was some discussion around using `componentDidUpdate` here: https://github.com/Shopify/quilt/pull/461#discussion_r247595226. We ended up deciding to stick with `getDerivedStateFromProps` for this scenario.

I can push up that change for `fieldsWithErrors` shortly. 👍

darrenhebner:
Pushed up the change. I'll merge if everything else looks good.

PR #505, opened 2019-02-07
✨ Add babel-plugin-react-i18n

tzvipm:
This PR adds a plugin for Babel that auto-fills `withI18n`'s arguments from an adjacent translations folder. The Babel plugin is exported from the `@shopify/react-i18n/babel` entrypoint:
 ```js
// babel.config.js
{
  plugins: [
    ['@shopify/react-i18n/babel'],
  ],
}
```
 This plugin will look for an adjacent translations folder containing, at minimum, an `en.json` file (the default locale). It will then iterate over each reference to the `withI18n` decorator, if the reference is a call expression with no arguments, and inject the appropriate arguments.
 ```js
// Within MyComponent.tsx:
 withI18n();
 // Becomes:
 import _en from './translations/en.json';
 withI18n({
  id: 'MyComponent_<hash>',
  fallback: _en,
  async translations(locale) {
    const dictionary = await import(/* webpackChunkName: "MyComponent_<hash>-i18n" */ `./translations/${locale}.json`);
    return dictionary;
  },
});
```


packages/babel-plugin-react-i18n/src/BabelPluginReactI18n.ts
```diff
@@ -0,0 +1,97 @@
...
+  template,
+}: {
+  types: typeof Types;
+  template: TemplateBuilder<Types.ImportDeclaration | Types.ObjectExpression>;
+}) {
+  const fallbackTranslationsImport = template(
+    `import enTranslations from './translations/en.json';`,
+    {sourceType: 'module'},
+  );
+  function i18nParams({ID}) {
```
tzvipm: Note: I needed to use a function here, because the default replacement logic doesn't seem to handle replacements within comments.

packages/babel-plugin-react-i18n/src/BabelPluginReactI18n.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  return {
+    visitor: {
+      Program(_path: NodePath<Types.Program>, state: State) {
+        const lastImport = _path
+          .get('body')
+          .filter(subPath => subPath.isImportDeclaration())
+          .pop();
+        if (lastImport) {
+          state.importer = importDecl => lastImport.insertAfter(importDecl);
```
cartogram: It is very convenient that it is a local import and you can get away with just this and still put it in the right place. 👍Cool.

packages/babel-plugin-react-i18n/src/BabelPluginReactI18n.ts
```diff
@@ -0,0 +1,97 @@
...
+          expr.arguments.length === 0
+        ) {
+          const {filename, filenameRelative} = this.file.opts;
+          const enJSONPath = path.resolve(
+            path.dirname(filename),
+            'translations',
+            'en.json',
+          );
+          if (!fs.existsSync(enJSONPath)) {
+            return;
```
cartogram: Do we just want to bail here? Not sure if you considered writing the file or warning that they have a decorator with no matching translation file.

packages/babel-plugin-react-i18n/src/test/BabelPluginReactI18n.test.ts
```diff
@@ -0,0 +1,87 @@
...
+          // eslint-disable-next-line typescript/no-non-null-assertion
+          resolve(result!.code!);
+        }
+      },
+    );
+  });
+}
+
+describe('babel-pluin-react-i18n', () => {
+  it('injects arguments when adjacent translations exist', async () => {
```
cartogram: I'd consider adding a test to make sure it doesn't do anything to components without the decorator.

packages/babel-plugin-react-i18n/src/test/BabelPluginReactI18n.test.ts
```diff
@@ -0,0 +1,87 @@
...
+    const fixturePath = path.resolve(
+      __dirname,
+      'fixtures',
+      'adjacentTranslations',
+      'MyComponent.js',
+    );
+
+    const code = await runFixture(fixturePath);
+    expect(code).toMatchInlineSnapshot(`
+"import React from 'react';
```
cartogram: This is a nitpick: I did snapshots in webgen to start but have since written a custom matcher that calls a few helpers for comparing file contents (stips their indents or formats, etc..). Historically we don't like snapshots and there isn't any reason this should be an exception IMO. I'd suggest using the matcher I wrote or writing a simple helper function. Then re-write these tests to be more specific (that it adds the import, generated an id, transformed the arguments, etc...). This might be overkill for such a small package/transformed code, but just thought I'd point it out.

packages/babel-plugin-react-i18n/src/BabelPluginReactI18n.ts
tzvipm: Having a decorator with no translations file is perfectly valid here; we just want to leave it as is in that case, so I think bailing is the right approach.

packages/babel-plugin-react-i18n/src/BabelPluginReactI18n.ts
```diff
@@ -0,0 +1,97 @@
...
+        sourceType: 'module',
+        plugins: ['dynamicImport'],
+        preserveComments: true,
+      },
+    )();
+  }
+
+  return {
+    visitor: {
+      Program(_path: NodePath<Types.Program>, state: State) {
```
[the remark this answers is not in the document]
tzvipm: it shadows the `oath` import otherwise. That said, I think the underscore is weird, so I'll rename these to `nodePath` 👍

packages/babel-plugin-react-i18n/README.md
```diff
@@ -0,0 +1,61 @@
...
+$ yarn add @shopify/babel-plugin-react-i18n
+```
+
+## Usage
+
+Simply add `@shopify/react-i18n` to your list of babel plugins. For example, via `.babelrc`:
+
+```json
+{
+  "plugins": ["@shopify/react-i18n"]
```
lemonmade: Does Babel not get confused with this and the actual package? Did you consider just throwing this in to the package itself?

packages/babel-plugin-react-i18n/src/BabelPluginReactI18n.ts
```diff
@@ -0,0 +1,97 @@
...
+
+export default function injectWithI18nArguments({
+  types: t,
+  template,
+}: {
+  types: typeof Types;
+  template: TemplateBuilder<Types.ImportDeclaration | Types.ObjectExpression>;
+}) {
+  const fallbackTranslationsImport = template(
+    `import enTranslations from './translations/en.json';`,
```
lemonmade: I think there is an API on path to get a unique binding name that we probably want to use instead

packages/babel-plugin-react-i18n/src/BabelPluginReactI18n.ts
```diff
@@ -0,0 +1,97 @@
...
+          state.importer = importDecl =>
+            _path.get('body')[0].insertBefore(importDecl);
+        }
+      },
+      CallExpression(_path: NodePath<Types.CallExpression>, state: State) {
+        const expr = _path.node;
+        const {callee} = expr;
+        if (
+          t.isIdentifier(callee) &&
+          (callee as Types.Identifier).name === WITH_I18N_DECORATOR_NAME &&
```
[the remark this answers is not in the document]
lemonmade: We need to be a bit more guarded on what we transform. In general, it is better to go from the import of withI18n, then find all the bindings, then process each of those, IMO. You can see an example of this with the async transform I added.

packages/babel-plugin-react-i18n/src/BabelPluginReactI18n.ts
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// based on postcss-modules implementation
+// see https://github.com/css-modules/postcss-modules/blob/60920a97b165885683c41655e4ca594d15ec2aa0/src/generateScopedName.js
+function generateID(filename: string) {
+  const hash = stringHash(filename)
+    .toString(36)
+    .substr(0, 5);
+  const extension = path.extname(filename);
+  const legible = path.basename(filename).replace(extension, '');
```
lemonmade: I think basename takes a second argument for removing the extension

packages/babel-plugin-react-i18n/src/test/BabelPluginReactI18n.test.ts
tzvipm: I do think this makes sense overall, but I agree it may be overkill for such a small package.

packages/babel-plugin-react-i18n/src/BabelPluginReactI18n.ts
```diff
@@ -0,0 +1,97 @@
...
+          const enJSONPath = path.resolve(
+            path.dirname(filename),
+            'translations',
+            'en.json',
+          );
+          if (!fs.existsSync(enJSONPath)) {
+            return;
+          }
+          state.importer(
+            fallbackTranslationsImport() as Types.ImportDeclaration,
```
lemonmade: To me this logic feels backwards. I’d rather do the following:

1. Check if the file actually includes an import for React-i18n
2. Check if that binding is ever used without arguments
3. Check if there is a translation file
4. Process each no-argument i18n call if all of the above are true

packages/babel-plugin-react-i18n/src/BabelPluginReactI18n.ts
```diff
@@ -0,0 +1,97 @@
...
+          }) as Types.ObjectExpression);
+        }
+      },
+    },
+  };
+}
+
+// based on postcss-modules implementation
+// see https://github.com/css-modules/postcss-modules/blob/60920a97b165885683c41655e4ca594d15ec2aa0/src/generateScopedName.js
+function generateID(filename: string) {
```
lemonmade: 👌

packages/babel-plugin-react-i18n/README.md
tzvipm: I thought babel plugins are always supposed to begin with `babel-plugin-*`, but I guess that's not the case?

packages/babel-plugin-react-i18n/src/BabelPluginReactI18n.ts
tzvipm: that is technically more correct, I suppose. I considered that approach, but I was worried about the second step. I suppose we could use `state` to do this entire thing in a single traversal. 🤔

packages/babel-plugin-react-i18n/README.md
tzvipm: How would you throw this into the package? I guess using some sort of `@shopify/react-i18n/babel` suffix? Would babel know how to find that?

lemonmade: You’d just use the full path to the module, yeah

packages/react-i18n/src/test/babel-plugin.test.ts
```diff
@@ -0,0 +1,107 @@
...
+import _en from './translations/en.json';
+
+function MyComponent({
+  i18n
+}) {
+  return i18n.translate('key');
+}
+
+export default withI18n({
+  id: 'MyComponent_1qre7',
```
tzvipm: The normalization script we use in `async` doesn't have a clear way to handle a hash like this. thoughts?

tzvipm: Ooh, I guess I could mock out `string-hash` 🤔

packages/babel-plugin-react-i18n/src/BabelPluginReactI18n.ts
cartogram: Ah right, sound good 👍

packages/react-i18n/README.md
```diff
@@ -274,6 +274,42 @@ export default function App({
...
+import _en from './translations/en.json';
+
+withI18n({
+  id: 'MyComponent_<hash>',
+  fallback: _en,
+  async translations(locale) {
+    try {
+      const dictionary = await import(/* webpackChunkName: "MyComponent_<hash>-i18n" */ `./translations/${locale}.json`);
+      return dictionary;
+    } catch (err) {}
```
lemonmade: Can we make a change so that this wrapping isn't needed? I thought I had it working at one point where the `translations` could throw and it would just do the right thing

packages/react-i18n/src/babel-plugin.ts
```diff
@@ -0,0 +1,145 @@
...
+}: {
+  types: typeof Types;
+  template: TemplateBuilder<Types.ImportDeclaration | Types.ObjectExpression>;
+}) {
+  function fallbackTranslationsImport({id}) {
+    return template(`import ${id} from './translations/en.json';`, {
+      sourceType: 'module',
+    })();
+  }
+  function i18nParams({id, fallbackID, decoratorName}) {
```
lemonmade: Newlines :D

packages/react-i18n/src/babel-plugin.ts
```diff
@@ -0,0 +1,145 @@
...
+    bindingName,
+    filename,
+    filenameRelative,
+    lastImport,
+  }) {
+    if (!hasTranslations(filename)) {
+      return;
+    }
+
+    let fallbackID;
```
lemonmade: Is there any reason not just to generate the fallback ID here?

packages/react-i18n/src/babel-plugin.ts
```diff
@@ -0,0 +1,145 @@
...
+        continue;
+      }
+
+      const args = callExpression.get('arguments');
+      if (args.length !== 0) {
+        continue;
+      }
+
+      if (fallbackID == null) {
+        fallbackID = refPath.scope.generateUidIdentifier('en').name;
```
lemonmade: This logic seems a bit incorrect to me. It generates the ID in the scope of the reference, but only on the first reference, and then reuses that for everything.

packages/react-i18n/src/babel-plugin.ts
```diff
@@ -0,0 +1,145 @@
...
+        lastImport.insertAfter(fallbackTranslationsImport({id: fallbackID}));
```
lemonmade: This also probably fails if there are multiple `i18n` calls?

lemonmade: I think it's OK to say it only supports one, we should just make the code reflect that

packages/react-i18n/src/babel-plugin.ts
```diff
@@ -0,0 +1,145 @@
...
+      }
+
+      callExpression.replaceWith(
+        i18nParams({
```
lemonmade: `Params` is a bit of an odd name given that this is actually a call expression

packages/react-i18n/src/babel-plugin.ts
```diff
@@ -0,0 +1,145 @@
...
+    )();
+  }
+  function addI18nArguments({
+    binding,
+    bindingName,
+    filename,
+    filenameRelative,
+    lastImport,
+  }) {
+    if (!hasTranslations(filename)) {
```
lemonmade: Maybe put the `hasTranslations` check as late as you can, since it's the most expensive?

packages/react-i18n/src/test/babel-plugin.test.ts
```diff
@@ -0,0 +1,111 @@
...
+      }
+
+      export default MyComponent;
+      `,
+    );
+
+    expect(await transform(code, optionsForFile('MyComponent.tsx'))).toBe(code);
+  });
+});
+async function normalize(code: string) {
```
lemonmade: Newline above

packages/react-i18n/src/test/babel-plugin.test.ts
```diff
@@ -0,0 +1,111 @@
...
+        return 'some string';
+      }
+
+      export default MyComponent;
+      `,
+    );
+
+    expect(await transform(code, optionsForFile('MyComponent.tsx'))).toBe(code);
+  });
+});
```
lemonmade: There are a few other tests that might be useful:

* Doesn't transform different imports from react-i18n
* Doesn't transform `withI18n` from other libraries
* Transforms `withI18n` when it was renamed during import

packages/react-i18n/babel.js
```diff
@@ -0,0 +1 @@
+module.exports = require('./dist/babel-plugin');
```
lemonmade: You'd need to add this + the .d.ts file to the `files` array in `package.sjon`

tzvipm: this package doesn't have a files array. should I add one?

packages/react-i18n/README.md
```diff
@@ -90,7 +90,7 @@ The most commonly-used feature of the `@shopify/react-i18n` library is looking u
 Translations are provided using two keys in the `withI18n` decorator:
 
 - `fallback`: a translation file to use when translation keys are not found in the locale-specific translation files. These will usually be your English translations, as they are typically the most complete.
-- `translations`: a function which takes the locale and returns one of: nothing (no translations for the locale), a dictionary of key-value translation pairs, or a promise of one of the above. Alternatively, you can pass an object where the keys are locales, and the values are either translation dictionaries, or promises for translation dictionaries.
+- `translations`: a function which takes the locale and returns one of: nothing (no translations for the locale), a dictionary of key-value translation pairs, or a promise of one of the above. The `translations` function can also throw and `react-i18n` will handle the situation gracefully. Alternatively, you can pass an object where the keys are locales, and the values are either translation dictionaries, or promises for translation dictionaries.
```
lemonmade: Can you make sure we have in here somewhere the recommendation that translations are inside a collocated `translations` directory, and that this pattern gets preferential treatment with the Babel plugin? I think we need to reinforce it a bit since it is on the magical side

packages/react-i18n/src/babel-plugin.ts
```diff
@@ -0,0 +1,155 @@
...
+
+  function addI18nArguments({
+    binding,
+    bindingName,
+    filename,
+    filenameRelative,
+    fallbackID,
+    insertImport,
+  }) {
+    let _hasTranslations;
```
lemonmade: Why underscore, and no initializer (so has `any` type)?

packages/react-i18n/src/babel-plugin.ts
```diff
@@ -0,0 +1,155 @@
...
+        continue;
+      }
+
+      const args = callExpression.get('arguments');
+      if (args.length !== 0) {
+        continue;
+      }
+
+      if (_hasTranslations == null) {
+        _hasTranslations = hasTranslations(filename);
```
lemonmade: Might be worth commenting on why you are pushing this to be later, since it makes the code read a little awkwardly (for good reason, though!)

packages/react-i18n/src/babel-plugin.ts
```diff
@@ -0,0 +1,155 @@
...
+      ImportDeclaration(
+        nodePath: NodePath<Types.ImportDeclaration>,
+        state: State,
+      ) {
+        if (nodePath.node.source.value !== '@shopify/react-i18n') {
+          return;
+        }
+        const {specifiers} = nodePath.node;
+        for (const specifier of specifiers) {
+          if (
```
lemonmade: Might be better to flip the condition and `continue` if false? There is a lot of nesting that goes on here

packages/react-i18n/src/decorator.tsx
```diff
@@ -162,6 +162,13 @@ function normalizeTranslationGetter(
   translations: TranslationMap | TranslationGetter,
 ) {
   return typeof translations === 'function'
-    ? translations
+    ? // eslint-disable-next-line consistent-return
+      (locale: string) => {
```
lemonmade: Excellent 👍

PR #510, opened 2019-02-11
✨ Upgrade to React 16.8.1

tzvipm:
This PR prepares our repo for React Hooks by upgrading all strict dependencies and dev dependencies to react `16.8.1`.

### DIFF
--- a/packages/react-i18n/src/test/i18n.test.ts
+++ b/packages/react-i18n/src/test/i18n.test.ts
@@ -607,7 +607,7 @@ describe('I18n', () => {
     });
 
     it('uses UTC when defaultTimezone is Etc/GMT+12', () => {
-      const date = new Date();
+      const date = new Date('2018-01-01T12:34:56-12:00');
       const defaultTimezone = 'Etc/GMT+12';
 
       const i18n = new I18n(defaultTranslations, {
@@ -616,7 +616,7 @@ describe('I18n', () => {
       });
       const expected = new Intl.DateTimeFormat(defaultDetails.locale, {
         timeZone: 'UTC',
-      }).format(date);
+      }).format(new Date('2018-01-01'));
 
       expect(
         i18n.formatDate(date, {
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "@types/enzyme": "^3.1.10",
     "@types/enzyme-adapter-react-16": "^1.0.3",
     "@types/jest": "^23.3.5",
-    "@types/react": "16.7.22",
+    "@types/react": "16.8.2",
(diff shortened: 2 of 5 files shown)


packages/react-compose/package.json
```diff
@@ -27,7 +27,7 @@
     "tslib": "^1.9.3"
   },
   "peerDependencies": {
-    "react": "^16.3.2"
+    "react": "^16.8.1"
```
lemonmade: Does it actually need such a high version constraint?

packages/react-i18n/src/test/i18n.test.ts
```diff
@@ -616,7 +616,7 @@ describe('I18n', () => {
       });
       const expected = new Intl.DateTimeFormat(defaultDetails.locale, {
         timeZone: 'UTC',
-      }).format(date);
+      }).format(new Date('2018-01-01'));
```
lemonmade: Why this

tzvipm: This was an existing bug that only seemed to come up when I was coding in the wee hours of the morning. I modified this code to reflect the test above it.

packages/react-compose/package.json
tzvipm: Fair point. I thought we were intending to use hooks everywhere, but I see now that this is not one of the places on our hooks roadmap, so no need to limit consumers to react `16.8.x` 👍 1+

packages/react-form-state/package.json
```diff
@@ -27,7 +27,7 @@
     "tslib": "^1.9.3"
   },
   "peerDependencies": {
-    "react": "^16.4.1"
+    "react": "^16.8.1"
```
tzvipm: Here we will need the higher peer dependency for a hooks rewrite.

lemonmade: Yes but we should wait until we actually do the hook version to bump the peer dependency

packages/useful-types/package.json
```diff
@@ -23,7 +23,7 @@
   },
   "homepage": "https://github.com/Shopify/quilt/blob/master/packages/useful-types/README.md",
   "dependencies": {
-    "@types/react": ">=16.4.0",
+    "@types/react": "^16.8.1",
```
tzvipm: I wasn't sure about this one. Should I leave as is as well? It seems like we want this one to be a peer dependency too, no?

lemonmade: IMO types must stay as dependencies, so that installing a package can never result in broken typings. Having the widest version range is very beneficial here, though, to reduce the crazy `@types` duplication that can happen.

Issue #513, opened 2019-02-11
[Type: Bug] [Package: react-i18n] Timezone related test failing only at certain times of the day


jamesism:
## Overview

While running the quilt test suite in the hours around midnight (+/- 2) I noticed a failing test from `react-i18n`. I don't have timestamps in my prompt unfortunately, so I don't have an exact time. 

![image](https://user-images.githubusercontent.com/887305/52588288-16e27000-2e0a-11e9-9fe6-5fd33a12c45f.png)

Please excuse my weird terminal background.

## Consuming repo

> What repo were you working in when this issue occurred?

`react-form-state`, but hunch is definitely that it is about the time at which I was running the tests and specific to `react-i18n`.

jamesism:
cc @shopify/insights-web-developers : was mentioned in slack that you added this test

joshuajay:
Yeah, this is likely some kind of DST related bug. If the late night test failures are problematic we can ship a quick PR that skips the test while we figure out the underlying issue..

joshuajay:
looks like this was fixed in https://github.com/Shopify/quilt/pull/510

PR #514, opened 2019-02-11
[i18n] Fix default timezone test

joshuajay:
Fixes https://github.com/Shopify/quilt/issues/513

The issue is only reproducible at a certain time of day, the test passes as of right now, but to be absolutely certain we should try testing it at a X hours close to midnight.


joshuajay:
looks like it was fixed in https://github.com/Shopify/quilt/pull/510

PR #515, opened 2019-02-12
Changing display none to visibility hidden of on body to allow measuring on render

dleroux:
Currently the `Sticky` component of `Polaris` needs to measure the height of the topbar in order to know where to start making `Sticky` elements position: fixed. Since the body element is set to `display: none` until styles are loaded, this prevents the measuring to occur. Changing to visibility: hidden fixes this issue.


dleroux:
@lemonmade how can I test this in web?

PR #521, opened 2019-02-20
Fix null length property

tzvipm:
The current plugin was causing many errors of the form:
```
./app/components/<name>/<name>.tsx
TypeError: Cannot read property 'length' of undefined
```

I anticipate something is weird with the AST we're getting from babel, and I'm hoping this PR will fix that.


tzvipm:
After some more inspection, with chrome devtools for node, I am closing this as unnecessary.

PR #525, opened 2019-02-21
Change webpackMode to lazy-once for async i18n bundles

tzvipm:
See https://webpack.js.org/api/module-methods/#magic-comments. Thanks to @GoodForOneFare for finding this.

### DIFF
--- a/packages/react-i18n/src/babel-plugin.ts
+++ b/packages/react-i18n/src/babel-plugin.ts
@@ -28,7 +28,7 @@ export default function injectWithI18nArguments({
       id: '${id}',
       fallback: ${fallbackID},
       async translations(locale) {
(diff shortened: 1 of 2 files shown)


GoodForOneFare:
When I tried a proof of concept for this, I recall it going poorly.  Details are dim, but I _think_ production builds ended up downloading files for already-serialized translations.  That said, a few recent i18n perf improvements may have fixed that problem.

So, are we sure this will have a positive impact?

tzvipm:
@GoodForOneFare what do you mean by already serialized translations? Do you mean that it downloads a bundle containing the fallback when we already have it loaded synchronously? I think we can avoid that using `/* webpackExclude: /en\.json$/ */`.

GoodForOneFare:
Yeah 👍  Downloading stuff that SSR has already embedded.

Issue #519, opened 2019-02-19
FormState.List is broken when initialValue is empty array


marina101:
## Overview
*Urgent*
We noticed that FormState.List no longer works for us on a form where initialValue is an empty array. It used to work and we haven't changed our code, so we started digging in the List source. I think the regression got introduced here: `https://github.com/Shopify/quilt/commit/300389252b8d367b66b93690c4cda502cb9e7025#diff-68157ac38fbcaa003b546774b1857293R48`  on line 48 in `packages/react-form-state/src/components/List.tsx`.  When `initialValue[index]` is called on an empty array with index 0, it returns undefined, so `const initialFieldValue = initialValue[index][fieldPath];` throws an error.

This is blocking us on a form that is payments compliance related with a hard deadline coming up soon, so we'd really appreciate it if it would be possible to fix this as soon as possible. Relates to Shopify/shopify#188581

## Consuming repo

Shopify/web

### Scope

- Is this issue related to a specific package?

  - [x] Tag it with the `Package: <package_name>` label.

---

## Checklist

- [x] Please delete the labels section before submitting your issue
- [x] I have described this issue in a way that is actionable (if possible)

axelerator:
Introduced here:
https://github.com/Shopify/quilt/pull/475/files#diff-68157ac38fbcaa003b546774b1857293R48

PR #531, opened 2019-02-26
Fix react-form-state List component

marina101:
Fixes: #519 

This is a fix for a regression that was introduced in this commit: https://github.com/Shopify/quilt/pull/475/files#diff-68157ac38fbcaa003b546774b1857293R48 on line 48. The bug is that initialValue[index][fieldPath] throws an exception instead of returning undefined the way it previously did when initialValue[index] is undefined. I added an extra check to see if initialValue.index is defined and return undefined if its not.

🎩  it with our form (which has a polymorphic list and uses the List from react-form-state) and now it working instead of throwing an error (relates to #11012).

### DIFF
--- a/packages/react-form-state/src/components/List.tsx
+++ b/packages/react-form-state/src/components/List.tsx
@@ -45,7 +45,8 @@ export default class List<Fields> extends React.Component<
       const innerFields: FieldDescriptors<Fields> = mapObject(
         fieldValues,
         (value, fieldPath) => {
-          const initialFieldValue = initialValue[index][fieldPath];
+          const initialFieldValue =
+            initialValue[index] && initialValue[index][fieldPath];
           return {
             value,
             onBlur,


packages/react-form-state/src/components/List.tsx
```diff
@@ -45,7 +45,8 @@ export default class List<Fields> extends React.Component<
       const innerFields: FieldDescriptors<Fields> = mapObject(
         fieldValues,
         (value, fieldPath) => {
-          const initialFieldValue = initialValue[index][fieldPath];
+          const initialFieldValue = initialValue[index] && initialValue[index][fieldPath] ?
```
tzvipm: hmm... do we need the ternary here? Does `initialValue[index] && initialValue[index][fieldPath]` not work on its own?

marina101: On its own it would return a boolean rather than the value (if the value exists)

axelerator: I'm not sure this is relevant, but the original implementation of `_.get` works with a path, where our implementation does not:
```
> a = { b: { c: 'foo'}}
> _.get(a, 'b.c')
"foo"
> a['b.c']
undefined
```

tzvipm: I don't think it will be relevant in our particular case. The main reaosn we were using `get`, AFAICT, is related to undefined behaviour. In terms of the `&&` operator, it will actually return the value without the ternary.

marina101: 👍  today I learned that && don't always return a bool in JS!

PR #539, opened 2019-02-27
🎨 use graphql for webhook registration

markiiikram:
This PR updates `koa-shopify-webhooks` to use GraphQL. I don't believe anything needs to be updated in the README. Thanks for helping me with this PR @TheMallen.


markiiikram:
@TheMallen please correct me if I'm wrong, but I don't believe receiving webhooks needs updating as Shopify is handling the request, and we're just doing a safe compare to auth it.
https://help.shopify.com/en/api/getting-started/webhooks#receiving-a-webhook

PR #542, opened 2019-02-28
✨ `performance` now tracks first-input-delay

GoodForOneFare:
Basically, this measures the milliseconds between the user's first event, and the browser passing the event on to handlers.  See https://github.com/GoogleChromeLabs/first-input-delay for more information.

Note: this is only useful when the first-input-delay polyfill is present in the document's head.

### DIFF
--- a/packages/performance/src/first-input-delay.d.ts
+++ b/packages/performance/src/first-input-delay.d.ts
@@ -1,0 +1,11 @@
+// window.perfMetrics is created by a polyfill that *must* be injected into
+// the document head
+// (see https://github.com/GoogleChromeLabs/first-input-delay).
+
+interface PerfMetrics {
+  onFirstInputDelay(handler: (duration: number, event: Event) => void): void;
+}
+
+interface Window {
+  perfMetrics?: PerfMetrics;
+}
--- a/packages/performance/src/performance.ts
+++ b/packages/performance/src/performance.ts
@@ -159,6 +159,16 @@ export class Performance {
           | TimeToFirstContentfulPaintEvent);
       });
     }
+
+    if (typeof window !== undefined && window.perfMetrics !== undefined) {
+      window.perfMetrics.onFirstInputDelay(delay => {
+        this.lifecycleEvent({
+          type: EventType.FirstInputDelay,
+          start: now() - delay,
+          duration: delay,
+        });
+      });
+    }
   }
 
   on<T extends keyof EventMap>(event: T, handler: EventMap[T]) {
--- a/packages/performance/src/test/performance.test.ts
+++ b/packages/performance/src/test/performance.test.ts
@@ -1,5 +1,9 @@
+import {FirstArgument} from '@shopify/useful-types';
 import {Performance} from '../performance';
 import {Navigation} from '../navigation';
+import {EventType} from '../types';
+
+type FirstInputDelayCallback = FirstArgument<PerfMetrics['onFirstInputDelay']>;
 
 describe('Performance', () => {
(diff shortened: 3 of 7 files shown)


GoodForOneFare:
Trying to figure out tests for this, but I'd appreciate comments now :)

packages/performance/src/performance.ts
```diff
@@ -159,6 +159,16 @@ export class Performance {
           | TimeToFirstContentfulPaintEvent);
       });
     }
+
+    if (typeof perfMetrics !== 'undefined') {
+      perfMetrics.onFirstInputDelay(delay => {
+        this.lifecycleEvent({
```
GoodForOneFare: Is this really a lifecycle event?  It's a thing that happens once, so I think so.  But it's also the only one that concerns `duration` instead of `start`.  Should we expose a way for consumers to know which value is significant?

packages/performance/src/performance.ts
```diff
@@ -159,6 +159,16 @@ export class Performance {
...
+          type: EventType.FirstInputDelay,
+          start: 0,
```
lemonmade: Can we make start `now` so we know when it was triggered? It will look weird here in the performance charts

packages/performance/src/first-input-delay.d.ts
```diff
@@ -0,0 +1,6 @@
+// window.perfMetrics is created by a polyfill that *must* be injected into
+// the document head
+// (see https://github.com/GoogleChromeLabs/first-input-delay).
+declare const perfMetrics: {
```
lemonmade: Hmm might be better to augment `Window` with this as an optional property?

packages/performance/src/performance.ts
lemonmade: Or, maybe, if this is triggered after the delay end, start is `now() - delay`?

packages/performance/src/performance.ts
lemonmade: This definitely does feel a bit weird but I think we are kind of stuck putting it here. Consumers at least have a way of figuring out whether it was some sort of duration event or a point in time by looking at `duration`, but I agree it's not the cleanest abstraction. IMO, this is fine for the time being.

packages/performance/src/test/performance.test.ts
```diff
@@ -12,4 +16,47 @@ describe('Performance', () => {
     performance.finish();
     expect(spy).toHaveBeenCalledWith(expect.any(Navigation));
   });
+
+  describe.skip('first-input-delay', () => {
```
GoodForOneFare: @lemonmade tests aren't runnable right now, but I think they'd pass once a `performance` mock is available.  Can you sanity check, please?

packages/performance/src/test/performance.test.ts
```diff
@@ -12,4 +16,47 @@ describe('Performance', () => {
...
+      const perfMetrics = new MockPerfMetrics();
+      window.perfMetrics = perfMetrics;
+
+      const lifecycleSpy = jest.fn();
+      // eslint-disable-next-line no-new
+      const performance = new Performance();
+      performance.on('lifecycleEvent', lifecycleSpy);
+
+      // eslint-disable-next-line typescript/no-non-null-assertion
+      perfMetrics.callback!(1000, new Event('click'));
```
lemonmade: Oh interesting, I wonder if we should capture the event type as metadata for the `fid` event?

GoodForOneFare: Let's get some results in first (aka, _I'd like to ship this now_ 😛 )

PR #544, opened 2019-02-28
improving the DATE_TIME_PARTS_REGEX to support fractional seconds

patsissons:
### What problem is this PR solving?

If a timestamp ever includes fractional seconds the regex will outright fail. This patch allows those seconds to be captured and inserted into the parsed date. This can be particularly useful if a date formatter emits empty data for the fractional seconds (i.e., the formatter emits `.000`), which adds no new information but causes the parse the fail.

### Reviewers’ hat-rack :tophat:

The added new tests should cover all new inputs for this updated regex pattern.

Because this is an optional field all previous usages should not be affected.

### DIFF
--- a/packages/dates/src/parse-date-string.ts
+++ b/packages/dates/src/parse-date-string.ts
@@ -1,36 +1,26 @@
 import {applyTimeZoneOffset} from './apply-time-zone-offset';
-
-/**
- * Allowed date string formats
- * yyyy-mm-dd
- * yyyy-mm-ddThh:mm:ss
- * yyyy-mm-ddThh:mm:ss+hh:mm
- * yyyy-mm-ddThh:mm:ss-hh:mm
- * yyyy-mm-ddThh:mm:ssZ
- */
-const DATE_TIME_PARTS_REGEX = /^(\d{4})-(\d{2})-(\d{2})(?:T(\d{2}):(\d{2}):(\d{2})(?:(Z|(?:(\+|-)(\d{2}):(\d{2}))))?)?$/;
+import {parseDateStringParts} from './parse-date-string-parts';
 
 export function parseDateString(dateString: string, timeZone?: string) {
-  const dateTimeParts = new RegExp(DATE_TIME_PARTS_REGEX).exec(dateString);
+  const dateTimeParts = parseDateStringParts(dateString);
 
   if (dateTimeParts == null) {
     return null;
   }
 
-  const [
-    // @ts-ignore
-    _,
-    rawYear,
-    rawMonth,
-    rawDay,
-    rawHour,
-    rawMinute,
-    rawSecond,
+  const {
+    year: rawYear,
+    month: rawMonth,
+    day: rawDay,
+    hour: rawHour,
+    minute: rawMinute,
+    second: rawSecond,
+    millisecond: rawMillisecond,
     timeZoneOffset,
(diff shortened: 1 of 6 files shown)


joshuajay:
> However, this is a breaking change for `parseDateStringParts` as it adds a new element to the array of fields (not at the end).

Looks like `parseDateStringParts` doesn't exist before this change, so it should still be non-breaking

patsissons:
> Looks like `parseDateStringParts` doesn't exist before this change, so it should still be non-breaking

Haha right, future me is getting too intertwined with present me 😁

packages/dates/src/parse-date-string.ts
```diff
@@ -3,49 +3,60 @@ import {applyTimeZoneOffset} from './apply-time-zone-offset';
...
-export function parseDateString(dateString: string, timeZone?: string) {
+export function parseDateStringParts(dateString: string) {
   const dateTimeParts = new RegExp(DATE_TIME_PARTS_REGEX).exec(dateString);
 
   if (dateTimeParts == null) {
     return null;
   }
 
+  // slice the first regex part (the full match) off
+  return Array.from(dateTimeParts).slice(1);
```
joshuajay: I always feel a bit uncomfortable when a function returns an array, it might be better to move the destructuring and `parseInt`'ing into this function, and return an object with all of the 'parts'.

That also allows us to surface a `DateStringParts` interface in the API

packages/dates/src/parse-date-string.ts
```diff
@@ -3,49 +3,60 @@ import {applyTimeZoneOffset} from './apply-time-zone-offset';
...
+ * yyyy-mm-ddThh:mm:ss.fff
+ * yyyy-mm-ddThh:mm:ss.fff+hh:mm
+ * yyyy-mm-ddThh:mm:ss.fff-hh:mm
+ * yyyy-mm-ddThh:mm:ss.fffZ
  */
-const DATE_TIME_PARTS_REGEX = /^(\d{4})-(\d{2})-(\d{2})(?:T(\d{2}):(\d{2}):(\d{2})(?:(Z|(?:(\+|-)(\d{2}):(\d{2}))))?)?$/;
+const DATE_TIME_PARTS_REGEX = /^(\d{4})-(\d{2})-(\d{2})(?:T(\d{2}):(\d{2}):(\d{2})(?:\.(\d{1,3}))?(?:(Z|(?:(\+|-)(\d{2}):(\d{2}))))?)?$/;
 
-export function parseDateString(dateString: string, timeZone?: string) {
+export function parseDateStringParts(dateString: string) {
```
joshuajay: are we planning to export this from the package, or is it only being exported for testing purposes?

patsissons: yea it is intended to be a package export.

packages/dates/src/parse-date-string.ts
patsissons: I would be fine with that. i'll make the change.

packages/dates/src/parse-date-string.ts
joshuajay: Ah, then we should add this to the README as well
Perhaps give it it's own file / test file

patsissons: yea sounds good!

PR #545, opened 2019-02-28
adding isFutureDate function

patsissons:
adds a new `isFutureDate` function to `@shopify/dates`. relatively simple, does what the label says. The included test should cover its functionality.

### DIFF
--- a/packages/dates/README.md
+++ b/packages/dates/README.md
@@ -65,6 +65,19 @@ const timeZone2 = 'Australia/Perth';
 const timeZoneOffset = getTimeZoneOffset(date, timeZone1, timeZone2);
 ```
 
+### `isFutureDate`
+
+Takes in a date object and an optional now date object to compare against (defaulting to a new date object). Returns a boolean indicating whether or not the first date is in the future.
+
+```ts
+import {isFutureDate} from '@shopify/dates';
+
+const now = '2018-01-01Z00:00';
+const date = '2018-01-02Z23:59';
+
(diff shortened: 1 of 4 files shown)


packages/dates/src/index.ts
```diff
@@ -10,4 +11,3 @@ export * from './is-tomorrow';
 export * from './parse-date-string';
 export * from './sanitise-date-string';
 export * from './unapply-time-zone-offset';
-export * from './map-deprecated-timezones';
```
joshuajay: why'd this get pulled out?

patsissons: good question, it definitely should not be pulled out.

patsissons:
I will update the function signature to include `now = new Date()` so that if need be, you can provide your own timezone specific _now_.

packages/dates/README.md
```diff
@@ -65,6 +65,19 @@ const timeZone2 = 'Australia/Perth';
...
 
+### `isFutureDate`
+
+Takes in a date object and an optional now date object to compare against (defaulting to a new date object). Returns a boolean indicating whether or not the first date is in the future.
+
+```ts
+import {isFutureDate} from '@shopify/dates';
+
+const now = '2018-01-01Z00:00';
+const date = '2018-01-02Z23:59';
```
joshuajay: these should be Date objects

patsissons: ack, apologies. will correct right away.

Issue #136, opened 2018-05-28
Move @shopify/javascript-utilities to quilt


tzvipm:
See https://github.com/Shopify/javascript-utilities/issues/33.

Currently, `javascript-utilities` is set up as a large package with multiple exports. This pattern doesn't work well with our monorepo approach, and doesn't really feel like a proper package boundary.

1. Packages need to be created for each of the javascript-utilities functions:
  - decorators can be moved into @shopify/javascript-decorators
  - files like feature-detect can be moved into their own packages (ex. @shopify/feature-detection)

2. Create an issue in the `javascript-utilites` repo for deprecating the library, add an upgrade path document for where to find the new utilities, and ensure that no key libraries (ex. `polaris`) are depending on the old packages.

michelleyschen:
Propose packages to be created from `javascript-utilities`

- [ ] [decorators](https://github.com/Shopify/javascript-utilities/tree/master/src/decorators) => create `@shopify/javascript-decorators` 
- [ ] [dates](https://github.com/Shopify/javascript-utilities/blob/master/src/dates.ts) => merge into [quilt/dates](https://github.com/Shopify/quilt/tree/master/packages/dates)
- [ ] [doms](https://github.com/Shopify/javascript-utilities/blob/master/src/doms.ts)  +  [fastdom](https://github.com/Shopify/javascript-utilities/blob/master/src/fastdom.ts) + [focus](https://github.com/Shopify/javascript-utilities/blob/master/src/focus.ts) => create `@shopify/dom`
- [ ] [events](https://github.com/Shopify/javascript-utilities/blob/master/src/events.ts) => create `@shopify/events`
- [ ] [feature-detect](https://github.com/Shopify/javascript-utilities/blob/master/src/feature-detect.ts) => create `@shopify/feature-detect`
- [ ] [math](https://github.com/Shopify/javascript-utilities/blob/master/src/math.ts) + [geometry](https://github.com/Shopify/javascript-utilities/blob/master/src/geometry.ts) => create `@shopify/math`
- [ ] [url](https://github.com/Shopify/javascript-utilities/blob/master/src/url.ts) => create `@shopify/url`
- [ ] [others/noop](https://github.com/Shopify/javascript-utilities/blob/master/src/other.ts#L1-L2) => ???
- [ ] [others/createUniqueIDFactory](https://github.com/Shopify/javascript-utilities/blob/master/src/other.ts#L4-L7) => ???

michelleyschen:
[...]

PR #548, opened 2019-03-01
✨ add a small memoize decorator

michelleyschen:
Part of https://github.com/Shopify/web/issues/10480 and https://github.com/Shopify/quilt/issues/136

The memoize utility should do the following:
* When the method run in server => return identity function
* When the method run in client =>
  
  * Use `WeakMap` when there is only 1 argument and it is an object (ideally not immutable, other wise this is useless)
  * Use `Map` with 50-100 limit on everything else including having resolver.

It will also replace `@shopify/javascript-utilities/decorators/memoize`

### DIFF
--- a/packages/decorators/CHANGELOG.md
+++ b/packages/decorators/CHANGELOG.md
@@ -1,0 +1,12 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
+and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+### Added
+
+- `@shopify/decorators` package
--- a/packages/decorators/README.md
+++ b/packages/decorators/README.md
@@ -1,0 +1,107 @@
+# `@shopify/decorators`
+
+[![Build Status](https://travis-ci.org/Shopify/quilt.svg?branch=master)](https://travis-ci.org/Shopify/quilt)
+[![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](LICENSE.md) [![npm version](https://badge.fury.io/js/%40shopify%2Fdecorators.svg)](https://badge.fury.io/js/%40shopify%2Fdecorators.svg) [![npm bundle size (minified + gzip)](https://img.shields.io/bundlephobia/minzip/@shopify/decorators.svg)](https://img.shields.io/bundlephobia/minzip/@shopify/decorators.svg)
+
+A set of decorators to aid your javascript journey.
+
+## Installation
+
+```bash
+$ yarn add @shopify/decorators
+```
+
+## Usage
+
+### `memoize`
+
+The memoize decorator creates a function that memoizes the results of the function it is decorating.
+The cache key for storing the results are based on the first argument provided to the memoized function.
+If the memoization key cannot be inferred from the first argument alone, a `resolver` should be passed in to ensure a unique key. (ex: the unique key is in the second argument, or the unique key is a combination of a few arguments)
+
+Know that memoization will be skipped on server process and the cached results have a maximum limit of 50 entries on a first in first out basis.
+
+#### Memoizing a simple function
+
+```ts
+import {memoize} from '@shopify/decorators';
+
+class MyClass {
+  @memoize()
+  addOne(number: number) {
+    return number + 1;
+  }
+}
+
+const myClass = new MyClass();
+
+myClass.addOne(1); // -> 2, addOne is executed
+myClass.addOne(1); // -> 2, result is from cache
+```
+
+#### Memoizing a function with object as argument
+
+When memoizing a function with object as first argument, make sure the object is immutable.
+
+```ts
+import {memoize} from '@shopify/decorators';
+
+class MyClass {
+  @memoize()
+  getValues(someObject: {one: string; two: string}) {
+    return;
+  }
+}
+
+const myClass = new MyClass();
+
+const testObject1 = {one: 1, two: 2};
+myClass.getValues(testObject1); // -> [1, 2], getValues is executed
+myClass.getValues(testObject1); // -> [1, 2], result is from cache
+
+testObject1.two = 3;
+myClass.getValues(testObject1); // -> [1, 2], result is from cache, BAD
+```
+
+#### Memoizing a function while providing a resolver
+
+The resolver takes in the same arguments as the function it is decorating.
+Be sure that the resolver returns a unique identifer.
+
+```ts
+import {memoize} from '@shopify/decorators';
+
+class MyClass {
+  @memoize((command: string, value: string) => `${command}${value}`)
+  getByCommand(command: string, value: string) {
+    // implementation for getByCommand
+  }
+}
+
+const myClass = new MyClass();
+
+myClass.getByCommand('command name 1', 'command value 1'); // runCommand is executed
+myClass.getByCommand('command name 1', 'command value 2'); // runCommand is executed
+```
+
+Next let's fix the example from [above](#memoizing-a-function-with-object-as-argument) so the results will always be correct.
+
+```ts
+import {memoize} from '@shopify/decorators';
+
+class MyClass {
+  @memoize((someObject: {id: string; value: string}) => `${id}${value}`)
+  getValues(someObject: {id: string; value: string}) {
+    return Object.values(someObject);
+  }
+}
+
+const myClass = new MyClass();
+
+const testObject1 = {id: 1, value: 2};
+myClass.getValues(testObject1); // -> [1, 2], getValues is executed
+myClass.getValues(testObject1); // -> [1, 2], result is from cache
+
+testObject1.value = 3;
+myClass.getValues(testObject1); // -> [1, 3], getValues is executed
+```
--- a/packages/decorators/package.json
+++ b/packages/decorators/package.json
@@ -1,0 +1,31 @@
+{
+  "name": "@shopify/decorators",
+  "version": "0.0.0",
+  "license": "MIT",
+  "description": "A set of decorators to aid your JavaScript journey.",
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "sideEffects": false,
+  "scripts": {
+    "build": "tsc --p tsconfig.build.json",
+    "prepublishOnly": "yarn run build"
+  },
+  "publishConfig": {
+    "access": "public",
+    "@shopify:registry": "https://registry.npmjs.org"
+  },
+  "author": "Shopify Inc.",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/Shopify/quilt.git"
+  },
+  "bugs": {
+    "url": "https://github.com/shopify/quilt/issues"
+  },
+  "homepage": "https://github.com/Shopify/quilt/blob/master/packages/decorators/README.md",
+  "dependencies": {},
+  "devDependencies": {
+    "typescript": "~3.0.1"
+  },
+  "files": ["dist/*"]
+}
--- a/packages/decorators/src/index.ts
+++ b/packages/decorators/src/index.ts
@@ -1,0 +1,1 @@
+export {default as memoize} from './memoize';
--- a/packages/decorators/src/memoize.ts
+++ b/packages/decorators/src/memoize.ts
@@ -1,0 +1,74 @@
+interface MemoizeMap<T, U> {
+  get(key: T): U;
+  has(key: T): boolean;
+  set(key: T, value: U): MemoizeMap<T, U>;
+}
+
+type Resolver<T extends Function> = T;
+
+export const MAX_MAP_ENTRIES = 50;
+
+export default function memoize<Method extends Function>(
+  resolver?: Resolver<Method>,
+): MethodDecorator {
+  return function<T>(
+    _target: Object,
+    _propertyKey: string | symbol,
+    descriptor: TypedPropertyDescriptor<T>,
+  ) {
+    const method = descriptor.value;
+
+    if (!method || !(method instanceof Function)) {
+      return descriptor;
+    }
+
+    descriptor.value = memoized(method as any, resolver);
+    return descriptor;
+  };
+}
+
+function memoized<F extends Function>(method: F, resolver?: Resolver<F>): F {
+  const weakMapCache = new WeakMap();
+  const mapCache = new Map();
+  const mapKeys: any[] = [];
+
+  return (function memoized(...args) {
+    if (typeof window === 'undefined') {
+      return method.apply(this, args);
+    }
+
+    const useWeakMap =
+      args.length === 1 && typeof args[0] === 'object' && !resolver;
+
+    let key;
+    if (useWeakMap) {
+      key = args[0];
+    } else if (resolver && resolver instanceof Function) {
+      key = resolver(...args);
+    } else {
+      key = args[0];
+    }
+
+    const cache: MemoizeMap<any, any> = useWeakMap ? weakMapCache : mapCache;
+    if (cache.has(key)) {
+      return cache.get(key);
+    }
+
+    const result = method.apply(this, args);
+
+    if (useWeakMap) {
+      weakMapCache.set(key, result);
+    } else {
+      mapCache.set(key, result);
+      mapKeys.push(key);
+
+      if (mapCache.size > MAX_MAP_ENTRIES) {
+        const oldestKey = mapKeys[0];
+        mapCache.delete(oldestKey);
+        mapKeys.shift();
+      }
+    }
+
+    return result;
+  } as unknown) as F;
+}
--- a/packages/decorators/src/tests/memoize-server.test.ts
+++ b/packages/decorators/src/tests/memoize-server.test.ts
@@ -1,0 +1,43 @@
+/**
+ * @jest-environment node
+ */
+
+import memoize from '../memoize';
+
+describe('memoize()', () => {
+  describe('server', () => {
+    it('recalculates the result when first argument changed', () => {
+      const spy = jest.fn();
+      class MyClass {
+        @memoize()
+        addOne(number: number) {
+          spy();
+          return number + 1;
+        }
+      }
+
+      const myClass = new MyClass();
+
+      expect(myClass.addOne(1)).toEqual(2);
+      expect(myClass.addOne(2)).toEqual(3);
+      expect(spy).toBeCalledTimes(2);
+    });
+
+    it('recalculates the result when the first argument stay the same', () => {
+      const spy = jest.fn();
+      class MyClass {
+        @memoize()
+        addOne(number: number) {
+          spy();
+          return number + 1;
+        }
+      }
+
+      const myClass = new MyClass();
+
+      expect(myClass.addOne(1)).toEqual(2);
+      expect(myClass.addOne(1)).toEqual(2);
+      expect(spy).toBeCalledTimes(2);
+    });
+  });
+});
--- a/packages/decorators/src/tests/memoize.test.ts
+++ b/packages/decorators/src/tests/memoize.test.ts
@@ -1,0 +1,166 @@
+/**
+ * @jest-environment jsdom
+ */
+
+import memoize, {MAX_MAP_ENTRIES} from '../memoize';
+
+describe('memoize()', () => {
+  describe('client', () => {
+    it('recalculates the result when first argument changed', () => {
+      const spy = jest.fn();
+      class MyClass {
+        @memoize()
+        addOne(number: number) {
+          spy();
+          return number + 1;
+        }
+      }
+
+      const myClass = new MyClass();
+
+      expect(myClass.addOne(1)).toEqual(2);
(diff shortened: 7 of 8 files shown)


packages/javascript-decorators/README.md
```diff
@@ -0,0 +1,76 @@
...
+getValues(testObject1); // -> [1, 2], getValues is executed
+getValues(testObject1); // -> [1, 2], result is from cache
+
+testObject1.value = 3;
+getValues(testObject1); // -> [1, 2], result is from cache
+
+getValues({id: 1, value: 2}); // -> [1, 2], getValues is executed because the object is new
+```
+
+#### memorizing a function with resolver
```
lemonmade: `Memoizing`, also sentence case

packages/javascript-decorators/README.md
```diff
@@ -0,0 +1,76 @@
...
+    return number + 1
+}
+
+addOne(1); // -> 2, addOne is executed
+addOne(1); // -> 2, result is from cache
+```
+
+#### memorizing a function with object as argument
+
+When memorizing a function with object as first argument, make sure the object is immutable.
```
lemonmade: memoizing

packages/javascript-decorators/README.md
```diff
@@ -0,0 +1,76 @@
...
+}
+
+const testObject1 = {id: 1, value: 2};
+getValues(testObject1); // -> [1, 2], getValues is executed
+getValues(testObject1); // -> [1, 2], result is from cache
+
+testObject1.value = 3;
+getValues(testObject1); // -> [1, 2], result is from cache
+
+getValues({id: 1, value: 2}); // -> [1, 2], getValues is executed because the object is new
```
lemonmade: Weird that we say to use a resolver in the comment above, but don't show it until the example below

packages/javascript-decorators/README.md
```diff
@@ -0,0 +1,76 @@
...
+```
+
+#### memorizing a function with resolver
+
+The resolver takes in the same arguments as the funciton it is decorating.
+Make sure the resolver return a unique identifer to ensure
+
+```js
+const resolver = (someObject: Object) => `${someObject.id}-${someObject.value}`;
```
lemonmade: We shouldn't show examples we wouldn't want people to copy. We should have a better type for `Object` here

packages/javascript-decorators/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+getValues({id: 1, value: 2}); // -> [1, 2], getValues is executed because the object is new
+```
+
+#### memorizing a function with resolver
+
+The resolver takes in the same arguments as the funciton it is decorating.
+Make sure the resolver return a unique identifer to ensure
+
+```js
```
lemonmade: `ts`

packages/javascript-decorators/README.md
```diff
@@ -0,0 +1,76 @@
...
+const resolver = (someObject: Object) => `${someObject.id}-${someObject.value}`;
+
+@memoize(resolver)
```
lemonmade: Can you definitely do decorators on function declarations? I thought it was only on methods

packages/javascript-decorators/src/memoize.ts
```diff
@@ -0,0 +1,51 @@
+export interface MemoizeMap<T, U> {
+  get(key: T): U;
+  has(key: T): boolean;
+  set(key: T, value: U): MemoizeMap<T, U>;
+}
+
+export const MAX_MAP_ENTIRES = 50;
+
+export default function Memoize(resolver?: Function) {
```
lemonmade: Why uppercase letter?

packages/javascript-decorators/src/tests/memoize-server.test.ts
```diff
@@ -0,0 +1,28 @@
...
+    it('recalculate the result when first argument changed', () => {
+      const spy = jest.fn(addOne);
+      const memoized = memoize()(spy);
+
+      expect(memoized(1)).toEqual(2);
+      expect(memoized(2)).toEqual(3);
+      expect(spy).toBeCalledTimes(2);
+    });
+
+    it('recalculate the result when the first argument stay the same', () => {
```
lemonmade: It doesn't recalculate the result though?

packages/javascript-decorators/src/tests/memoize-server.test.ts
```diff
@@ -0,0 +1,28 @@
...
+      const spy = jest.fn(addOne);
+      const memoized = memoize()(spy);
+
+      expect(memoized(1)).toEqual(2);
+      expect(memoized(1)).toEqual(2);
+      expect(spy).toBeCalledTimes(2);
```
lemonmade: Shouldn't it only be one here?

packages/javascript-decorators/src/tests/memoize-server.test.ts
lemonmade: Oh this is for the server, never mind

packages/javascript-decorators/src/tests/memoize-server.test.ts
```diff
@@ -0,0 +1,28 @@
+/**
+ * @jest-environment node
+ */
+
+import memoize from '../memoize';
+
+describe('memoize()', () => {
+  describe('server', () => {
+    const addOne = (number: number) => number + 1;
```
lemonmade: I would not share these between tests personally, it is very confusing in the context of testing number of calls

packages/javascript-decorators/src/memoize.ts
```diff
@@ -0,0 +1,51 @@
...
+  const weakMapCache = new WeakMap();
+  const mapCache = new Map();
+
+  return function(...args: any[]) {
+    if (typeof window === 'undefined') {
+      return func.apply(this, args);
+    }
+
+    const useWeakMap =
+      args.length === 0 && typeof args[0] === 'object' && !resolver;
```
lemonmade: Shouldn't this be `1`?

packages/javascript-decorators/src/memoize.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export const MAX_MAP_ENTIRES = 50;
+
+export default function Memoize(resolver?: Function) {
+  return function(...args: any[]) {
+    return Memoized(args[0], resolver);
+  };
+}
+
+function Memoized(func: Function, resolver?: Function) {
```
lemonmade: These types are much too weak, the returned function should be the same type as the input one, and the resolver function type can be inferred from the arguments of the memoized function

packages/javascript-decorators/src/memoize.ts
lemonmade: Also, why is this separated from the `memoized` thing below?

packages/javascript-decorators/src/memoize.ts
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    const useWeakMap =
+      args.length === 0 && typeof args[0] === 'object' && !resolver;
+
+    let key;
+    if (useWeakMap) {
+      key = args[0];
+    } else if (resolver) {
+      key = resolver.apply(this, args);
```
lemonmade: What is `this` here?

packages/javascript-decorators/src/memoize.ts
```diff
@@ -0,0 +1,51 @@
...
+    const cache: MemoizeMap<any, any> = useWeakMap ? weakMapCache : mapCache;
+    if (cache.has(key)) {
+      return cache.get(key);
+    }
+
+    const result = func.apply(this, args);
+
+    if (useWeakMap) {
+      weakMapCache.set(key, result);
+    } else if (mapCache.size < MAX_MAP_ENTIRES) {
```
lemonmade: This means that once the "spots" for moization are filled, no other values are memoized, which is not what we want (we want the most recent `MAX_MAP_ENTRIES` (typo) to be memoized, and to drop the "old" ones)

michelleyschen:
@lemonmade this is meant to replace `@shopify/javascript-utilities/decorators/memoize` which is why it is a decorator package. I would like to move all the things from `@shopify/javascript-utilities/decorators/` over here.

Can you point out what is missing to use `memoize` as a decorator?

packages/javascript-decorators/src/memoize.ts
michelleyschen: I think this should be consistent with how the method is run.

packages/javascript-decorators/src/memoize.ts
```diff
@@ -0,0 +1,74 @@
...
+    descriptor: TypedPropertyDescriptor<F>,
+  ) {
+    const method: F | undefined = descriptor.value;
+
+    if (!method || !(method instanceof Function)) {
+      return descriptor;
+    }
+
+    // @ts-ignore
+    descriptor.value = memoized<F>(method, resolver);
```
michelleyschen: The error here without `ts-ignore`
```
Type 'F' does not satisfy the constraint 'Function'.
```

if I remove `<F>`, I get an error on `resolver`
```
Argument of type 'Resolver<F> | undefined' is not assignable to parameter of type 'Resolver<F & Function> | undefined'.
  Type 'Resolver<F>' is not assignable to type 'Resolver<F & Function> | undefined'.
    Type 'Function | unknown[]' is not assignable to type 'Resolver<F & Function> | undefined'.
      Type 'Function' is not assignable to type 'Resolver<F & Function>'.
        Type 'Function' is not assignable to type 'Resolver<F & Function>'.
          Type 'Resolver<F>' is not assignable to type 'Resolver<F & Function>'.
```

michelleyschen: nvm I figure it out

packages/javascript-decorators/src/memoize.ts
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    const useWeakMap =
+      args.length === 1 && typeof args[0] === 'object' && !resolver;
+
+    let key;
+    if (useWeakMap) {
+      key = args[0];
+    } else if (resolver) {
+      key = resolver.apply(this, args);
```
GoodForOneFare: If `resolver` always uses a strong map, should it be typed to enforce a primitive result?

packages/javascript-decorators/src/memoize.ts
GoodForOneFare: 🐟 I can't think why we'd want the resolver to have access to the class instance.  That feels like a foot-gun.

Poking at this a bit: a resolver that tries to access `this.addOne` passes type checks, but fails at runtime because `this === undefined`.

GoodForOneFare: So this should probably be `key = resolver(...args)`?

packages/javascript-decorators/src/memoize.ts
```diff
@@ -0,0 +1,73 @@
+export interface MemoizeMap<T, U> {
+  get(key: T): U;
+  has(key: T): boolean;
+  set(key: T, value: U): MemoizeMap<T, U>;
+}
+
+type Resolver<T extends Function> = T extends (...args: infer T) => any
```
GoodForOneFare: This PR is inspiring me to read up on advanced types 😬 😆

packages/javascript-decorators/src/memoize.ts
```diff
@@ -0,0 +1,73 @@
+export interface MemoizeMap<T, U> {
```
GoodForOneFare: This seems to be an internal thing, so does it need to be exported?

packages/javascript-decorators/src/tests/memoize.test.ts
```diff
@@ -0,0 +1,128 @@
...
+
+      testObject1.one = 2;
+      expect(testClass.getValues(testObject1)).toEqual([1, 2]);
+
+      expect(spy).toBeCalledTimes(1);
+    });
+  });
+
+  describe('customized resolver', () => {
+    it('recalculate the result when the resolver result was changed', () => {
```
GoodForOneFare: This is the only block that uses `getName`, so I'd prefer to see a specific `class ResolverTester` defined in this block.

This fits with our convention of keeping test helpers isolated, and not overloading them with many methods.

GoodForOneFare: Looking at the other `describe` blocks, it looks like we could have an isolated tester class for those, too.

packages/javascript-decorators/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+[![Build Status](https://travis-ci.org/Shopify/quilt.svg?branch=master)](https://travis-ci.org/Shopify/quilt)
+[![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](LICENSE.md) [![npm version](https://badge.fury.io/js/%40shopify%2Fjavascript-decorators.svg)](https://badge.fury.io/js/%40shopify%2Fjavascript-decorators.svg) [![npm bundle size (minified + gzip)](https://img.shields.io/bundlephobia/minzip/@shopify/javascript-decorators.svg)](https://img.shields.io/bundlephobia/minzip/@shopify/javascript-decorators.svg)
+
+A set of decorators to aid your javascript journey.
+
+## Installation
+
+```bash
+$ yarn add @shopify/javascript-decorators
```
GoodForOneFare: Nit: these can be used with TypeScript!   Should it be `@shopify/decorators`?

packages/javascript-decorators/src/tests/memoize.test.ts
```diff
@@ -0,0 +1,128 @@
...
+    it('gets the result from cache when the first argument stay the same', () => {
+      const spy = jest.fn();
+      const testClass = createTestClass({addOneSpy: spy});
+
+      expect(testClass.addOne(1)).toEqual(2);
+      expect(testClass.addOne(1)).toEqual(2);
+      expect(spy).toHaveBeenCalledTimes(1);
+    });
+
+    it('does not put the result into map cache, if map cache is over its max limit', () => {
```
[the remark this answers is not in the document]
GoodForOneFare: Oh, it's just first-in-first-out.  K, so `deletes oldest cache item when the cache limit is reached`.  D'you think there's any value in making this an LRU cache?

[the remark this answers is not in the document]
michelleyschen: > D'you think there's any value in making this an LRU cache

mm...maybe, but I really just hoped that we don't go over 50. 
I dont think we would have a lot of callback in the UI that would be over 50 items.

packages/javascript-decorators/src/memoize.ts
michelleyschen: It could return an object as well.

And if it return object, its likely the object is being created in the resolver and won't benefit from caching using WeakMap.

packages/javascript-decorators/src/memoize.ts
michelleyschen: thank you for explaining! will change this.

packages/javascript-decorators/src/memoize.ts
michelleyschen: I found this in a github comment!
This is how you enforce the parameters type using the type of a function.
The post also mentioned something similar to enforce the return type as well.

Thou I read in Chris's PR yesterday, `ReturnType` is actually a thing people use so much that it is included in the base library!

packages/javascript-decorators/README.md
michelleyschen: was trying to be more specific, my assumption being that npm != javascript?

packages/javascript-decorators/src/tests/memoize.test.ts
michelleyschen: I isolated them to a create class method that gets call.
which is a pattern I have seem before as well (for things like `graphqlClient`)
if there more worries for class that should not be create like this?

GoodForOneFare: Having a test class with _every_ type of memoize usage is convenient for the writer of the test, but inconvenient for new readers.  Before I can understand any test, I have to jump down to the helper class, and understand how every `@memoize` method works :/

If each `describe` block defines a test helper, I can focus on that, and not jump around so much.

packages/javascript-decorators/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+const testObject1 = {id: 1, value: 2};
+myClass.getValues(testObject1); // -> [1, 2], getValues is executed
+myClass.getValues(testObject1); // -> [1, 2], result is from cache
+
+testObject1.value = 3;
+myClass.getValues(testObject1); // -> [1, 2], result is from cache, BAD
+```
+
+#### Memoizing a function with resolver
```
lemonmade: Nit: you can only memoize a method with what we currently have

packages/javascript-decorators/README.md
```diff
@@ -0,0 +1,117 @@
...
+
+The resolver takes in the same arguments as the function it is decorating.
+Be sure that the resolver returns a unique identifer.
+
+This is how you can use a resolver to make sure the results of the function will always be correct.
```
lemonmade: This feels like all one paragraph. I am not really sure this last sentence makes sense though, it isn't really about them being "correct", it is about getting a memoized value for an object that changes from call to call

packages/javascript-decorators/src/memoize.ts
lemonmade: Don't use `infer T` though, since `T` is already the generic name of the function type (generally, this ends up being called `U`, but you can call it `Arguments` or something more meaningful

packages/javascript-decorators/src/memoize.ts
```diff
@@ -0,0 +1,73 @@
...
+    _propertyKey: string | symbol,
+    descriptor: TypedPropertyDescriptor<F>,
+  ) {
+    const method: F | undefined = descriptor.value;
+
+    if (!method || !(method instanceof Function)) {
+      return descriptor;
+    }
+
+    descriptor.value = memoized(method, resolver);
```
lemonmade: @tzvipm do you recall the issues with decorators? I seem to remember there being some special incantations required to make sure this stuff didn't break in some browsers

packages/javascript-decorators/src/memoize.ts
```diff
@@ -0,0 +1,73 @@
...
+    return descriptor;
+  };
+}
+
+function memoized<F extends Function>(method: F, resolver?: Resolver<F>): F {
```
lemonmade: I would export this as sell (maybe as `memoizeFunction`), since it is useful on its own for functions that aren't methods

packages/javascript-decorators/src/memoize.ts
```diff
@@ -0,0 +1,73 @@
...
+  const weakMapCache = new WeakMap();
+  const mapCache = new Map();
+
+  return (function(...args) {
```
lemonmade: Would be nice to name this, at least something like `memoized`

packages/javascript-decorators/src/memoize.ts
```diff
@@ -0,0 +1,73 @@
...
+
+    const result = method.apply(this, args);
+
+    if (useWeakMap) {
+      weakMapCache.set(key, result);
+    } else {
+      mapCache.set(key, result);
+
+      if (mapCache.size > MAX_MAP_ENTRIES) {
+        const oldestKey = mapCache.keys().next().value;
```
lemonmade: This feels really expensive. Can we maybe keep an array around that we append to/ can pop from when it is > 50 items?

packages/javascript-decorators/src/tests/memoize-server.test.ts
lemonmade: Super nit: `recalculates` for it to read like a sentence

packages/javascript-decorators/src/tests/memoize-server.test.ts
lemonmade: nvm comment above, this is for server tests 👍

packages/javascript-decorators/src/tests/memoize.test.ts
lemonmade: I strongly agree, repetition feels like it would be very useful here

packages/javascript-decorators/README.md
michelleyschen: what we currently have?

packages/javascript-decorators/src/memoize.ts
michelleyschen: I may need an example, not sure I know exactly what you mean.

lemonmade:
```ts
type Resolver<T extends Function> = T extends (...args: infer Arguments) => any
  ? Arguments
  : never;
```

packages/javascript-decorators/src/memoize.ts
michelleyschen: In that case I am going to do Gord's suggestion of LRU cache

lemonmade: I think an LRU cache is going to be a fair bit trickier to implement. Pushing keys on to an array and unshifting when length > 50 will be much easier and get most of the benefit IMO

GoodForOneFare: Agreed.  Otherwise this small decorator starts to become big 😄

PR #550, opened 2019-03-05
Allow custom error handling for i18n errors

lemonmade:
Right now, when a translation is missing (or in a few other cases), we throw an error from the i18n library and are done with it. IMO this is the proper default; React provides ways to catch this error in the tree and recover gracefully. However, Web has been bitten a few times by this where incomplete test coverage/ being a little careless with the translations + no additional error handling means that the whole page gets wiped.

This PR adds the necessary hook for us to catch the error. With this PR, if the error is not re-thrown by the `onError` handler, it will return an empty string instead. IMO this is not really better than the default (since the content will not make sense), and it means that in some cases we still have errors thrown (where we can't return any sensible "default" result), but I seem to be in the minority opinion on this.

### DIFF
--- a/packages/react-i18n/src/i18n.ts
+++ b/packages/react-i18n/src/i18n.ts
@@ -17,7 +17,11 @@ import {
   currencyDecimalPlaces,
   DEFAULT_DECIMAL_PLACES,
 } from './constants';
-import {MissingCurrencyCodeError, MissingCountryError} from './errors';
+import {
+  MissingCurrencyCodeError,
+  MissingCountryError,
+  I18nError,
+} from './errors';
 import {
   getCurrencySymbol,
   translate,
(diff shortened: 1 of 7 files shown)


packages/react-i18n/src/i18n.ts
```diff
@@ -129,7 +142,12 @@ export default class I18n {
       };
     }
 
-    return translate(id, normalizedOptions, this.translations, this.locale);
+    try {
+      return translate(id, normalizedOptions, this.translations, this.locale);
+    } catch (error) {
+      this.onError(error);
+      return '';
```
tzvipm: Should we make the consumer supply a component for this? Maybe some sort of empty state is better than the empty string? Like a line that looks like it never loads?

lemonmade: I think this gets really tricky because then you probably want a bunch of context about the type of error and other info. These errors should be rare and dealt with immediately, so I think this is enough to get that done.

PR #551, opened 2019-03-05
Fix polyfills in browser

tzvipm:
As it turns out, there was a problem with the standard import method that lead to a ReferenceError for `exports` and there was a problem with the `exports = {}` syntax, because `module.exports` is read-only, but the `require()` approach works well for the browser environment and the previous approach (with imports) is sufficient for node environments.

### DIFF
--- a/packages/polyfills/src/baseline.node.ts
+++ b/packages/polyfills/src/baseline.node.ts
@@ -2,5 +2,3 @@ import '@babel/polyfill';
 
 // eslint-disable-next-line import/extensions
 import '@shopify/polyfills/fetch.node';
-
-export = {};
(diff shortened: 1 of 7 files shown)


tzvipm:
This thread provides context on the issue, for our future selves: https://github.com/Microsoft/TypeScript/issues/14351

tzvipm:
In particular, the typescript compiler was including code like this in our library's generated js:
```
Object.defineProperty(exports, "__esModule", { value: true });
```

But, it seems that `exports` is undefined in `the browser / webpack`, although `module.exports` is defined. Confusingly, when running this code through `babel-loader`, it seems to have kept the `exports` piece intact, instead of integrating it with webpack's ecosystem.

One fix for this is to put `export = {}` in the input typescript code and the compiler will transpile that to `module.exports = {}` and get rid of the previous code referencing `exports`.

This worked well for building the client and server bundles, but it ended up throwing an error at runtime since `module.exports` is read only in webpack's runtime configuration.

As a solution, using `require` in the `browser` bundles instead of es6-style imports, insures that TypeScript doesn't inject the `exports` code, but also protects us from trying to assign to `module.exports` at runtime.

PR #552, opened 2019-03-05
Add Provider and decorator back

vsumner:
The provider uses the new context api and the decorator uses the new hook.


packages/react-i18n/src/decorator.ts
```diff
@@ -0,0 +1,20 @@
...
+export interface WithI18nProps {
+  i18n: I18n;
+}
+
+export function withI18n(i18nOptions: RegisterOptions) {
+  return <P extends WithI18nProps>(
+    WrappedComponent: React.ComponentType<P>,
+  ): React.ComponentType<P> => props => {
+    const [i18n] = useI18n(i18nOptions);
+    return React.createElement(WrappedComponent, {
```
lemonmade: Can we use JSX here?

packages/react-i18n/src/hooks.tsx
```diff
@@ -35,7 +35,7 @@ export function useI18n({
 
   React.useEffect(
     () => {
-      return manager.subscribe(ids, (translations, details) => {
+      manager.subscribe(ids, (translations, details) => {
```
lemonmade: I think this needs to be returned, we might just need to rework the return value a bit. If we don't return this, it will never unsubscribe from the manager.

packages/react-i18n/src/decorator.ts
```diff
@@ -0,0 +1,20 @@
...
+import {RegisterOptions} from './manager';
+import {useI18n} from './hooks';
+import I18n from './i18n';
+
+export interface WithI18nProps {
+  i18n: I18n;
+}
+
+export function withI18n(i18nOptions: RegisterOptions) {
+  return <P extends WithI18nProps>(
```
lemonmade: This no longer does the `hoistNonReactStatics` stuff, which I think was part of the old one?

packages/react-i18n/README.md
```diff
@@ -232,6 +225,32 @@ i18n.translate('MyComponent.searchResult', {
 });
 ```
 
+#### `withI18n` decorator
```
lemonmade: I feel like we need a slightly better story here. Right now, this is great for legacy purposes, which you've noted in the doc. However, this would also be the only option for class-based components. How would you feel about adding a `<WithI18n>` component that provides the i18n instance in a render prop (this brings it in to a class-based component in a similar way to how it will end up looking in a function component, though obviously with more nesting)? Then the docs could be structured basically like `The right way` => `If you need classes` => `If you have legacy code`

packages/react-i18n/src/decorator.ts
```diff
@@ -0,0 +1,20 @@
...
+    WrappedComponent: React.ComponentType<P>,
+  ): React.ComponentType<P> => props => {
+    const [i18n] = useI18n(i18nOptions);
```
lemonmade: I think to truly replicate what existed before, this will need to also grab the `ShareTranslations` bit of the return value and render it around the wrapped component

packages/react-i18n/src/hooks.tsx
vsumner: 👍 It was failing to compile because the return type was boolean.

packages/react-i18n/src/decorator.ts
vsumner: 👍

packages/react-i18n/src/hooks.tsx
vsumner: Fixed.

packages/react-i18n/src/decorator.ts
vsumner: Added `< ShareTranslations></ShareTranslations>` wrapper.

packages/react-i18n/src/decorator.ts
vsumner: I added `hoistStatics` which was what the old one used. https://github.com/Shopify/quilt/blob/master/packages/react-i18n/src/decorator.tsx#L153

packages/react-html/src/tests/context.test.tsx
```diff
@@ -92,9 +92,9 @@ describe('<Provider />', () => {
     const manager = new Manager();
 
     mount(
-      <Provider manager={manager}>
+      <HtmlProvider manager={manager}>
         <div />
-      </Provider>,
+      </HtmlProvider>,
```
lemonmade: These are I assume just problems with the base branch eh?

packages/react-i18n/README.md
```diff
@@ -42,17 +42,15 @@ export default function App() {
 
 ### Internationalized components
 
-Components must connect to the i18n context in order to get access to the many internationalization utilities this library provides. You can use the `withI18n` decorator to add an `i18n` prop to your component:
+Components must connect to the i18n context in order to get access to the many internationalization utilities this library provides. You can use the `useI18n` hook to access `i18n` in your component:
```
lemonmade: Probably want to document `ShareTranslations` too but we can wait on that for the "combined" final PR

packages/react-i18n/src/Provider.tsx
```diff
@@ -0,0 +1,14 @@
+import * as React from 'react';
+import {I18nContext} from './context';
+import Manager from './manager';
+
+interface Props {
+  manager?: Manager;
+  children: React.ReactNode;
+}
+
+export function Provider({manager, children}: Props) {
```
lemonmade: What do you think, do we still need this component? I feel like maybe it's fine to just expose the raw context object directly but I am not sure :S

packages/react-html/src/tests/context.test.tsx
vsumner: Yes. I was working on tests.

packages/react-i18n/src/Provider.tsx
vsumner: I think the advantage of exposing the raw context instead is that this is a documented pattern for the React Context API. So if you are using context anywhere else in react, this is the same thing.

PR #555, opened 2019-03-06
[polyfills] Fix conflict with isomorphic fetch

tzvipm:
The version of `whatwg-fetch` used by `isomorphic-fetch` is version `2.0.4`. This line in our package.json caused a conflict in node environments that use `isomorphic-fetch`, as they end up running code from `whatwg-fetch@3.0.0` that only works in a browser environment.

### DIFF
--- a/packages/polyfills/src/fetch.node.ts
+++ b/packages/polyfills/src/fetch.node.ts
@@ -1,1 +1,16 @@
-import 'isomorphic-fetch';
+// general logic and approach taken from
+// https://github.com/matthew-andrews/isomorphic-fetch/blob/master/fetch-npm-node.js
+
+const nodeFetch = require('node-fetch');
+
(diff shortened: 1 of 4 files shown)


packages/polyfills/src/fetch.node.ts
```diff
@@ -1 +1,16 @@
...
+// https://github.com/matthew-andrews/isomorphic-fetch/blob/master/fetch-npm-node.js
+
+const nodeFetch = require('node-fetch');
+
+function wrappedFetch(url: string, options) {
+  const finalURL = /^\/\//.test(url) ? `https:${url}` : url;
+  return nodeFetch.call(this, finalURL, options);
+}
+
+if (!(global as any).fetch) {
```
lemonmade: Would probably be nicer to augment the global

Issue #557, opened 2019-03-07
FormState.Nested does not re-render children if nested fields don't change


patsissons:
## Overview

In cases where nested fields are bound to components that also react to parent component state, when the parent component state changes the nested component will not re-render because only the parent state has changed, not the bound `FormState` fields.

A simple example would be a `<TextField />` that is bound to a form field, but perhaps has its `readOnly` prop driven by parent component state. When the parent component calls `setState` the `FormState.Nested` will see a re-render, but because the bound field doesn't change the `shouldComponentUpdate` returns `false` and the render props are never invoked.

This behaviour was originally observed using a Modal whose `open` prop was driven by parent component state. The modal would never open because the state change would never result in the nested component re-rendering.

## Consuming repo

> What repo were you working in when this issue occurred?

https://github.com/Shopify/web

amireynoso:
To add on that, the same problem is observed in `FormState.List`.

Another example (and it's actually my use case): a dropdown select which options are loaded from GraphQL. So, initially, it’s an empty array that then gets populated. Since the actual value of the field never changes, the dropdown doesn’t get re-rendered.

amireynoso:
Looks like this was introduced in #312, with good reasoning behind it.

I'm thinking maybe adding a new prop to these components (maybe something like `update`, defaults to `false`) that if set, will ignore the checks in `shouldComponentUpdate`. This way, we keep the performance boost for most cases, and it's possible to opt-out of it. I don't, however, really like the idea of changing / adding to the API. 🤔 

@jfangrad @TheMallen What do you think?

patsissons:
Yea my initial thought was to add a `state?: {}` prop and then include a deep compare in the `shouldComponentUpdate`

amireynoso:
I just opened https://github.com/Shopify/quilt/pull/558 to address it, would love some eyes on it  👀

amireynoso:
Resolved in `@shopify/react-form-state@0.9.0`. 🙂

PR #558, opened 2019-03-07
Add optional prop to <FormState.Nested> and <FormState.List> to re-render

amireynoso:
## What's the problem
As described in #557, `<FormState.Nested>` and `<FormState.List>` are not updating their children, unless the passed `field` changes.

While this is the expected behaviour (introduced in #312 to improve the performance of these components), there are cases in which it's necessary for the children to be updated. For example, if any of the child form elements also depend on a state change or a prop other than `field` changing.

## How it's solved
I considered changing the logic within `shouldComponentUpdate`, but ended up going with a non-intrusive solution:

- ✨ Added a new prop `update` to both `<FormState.Nested>` and `<FormState.List>`
   - `update` is optional, therefore falsy by default.
   - Setting this to `true` will by-pass the logic in `shouldComponentUpdate`
- ✅ Added new tests to cover the use of this prop
- 📝 Added info to the docs on why the children don't update (which could be unexpected) and instructions on how to use the prop

## Tophatting 🎩

I tophatted the changes in my app using `yarn tophat`. I'd recommend testing this in any app that's using either `<FormState.Nested>` or `<FormState.List>` and making sure that by using `update`, the child form elements get re-rendered. 

The tests should cover this case and can also be used as a playground to 🎩 it!

### DIFF
--- a/packages/react-form-state/docs/building-forms.md
+++ b/packages/react-form-state/docs/building-forms.md
@@ -508,6 +508,21 @@ export function ProductPage() {
 }
 ```
 
+#### Performance
+
+`<FormState.Nested />` will only re-render its children if `value`, `initialValue` or `error` on `field` change for performance reasons. While this is highly recommended, there are cased in where you might want to by-pass this check. To force it to always re-render its children, you can pass `update`:
+
+```tsx
+<FormState.Nested field={fields.firstVariant} update>
+  {fields => (
+    <>
+      <TextField label="Option" {...fields.option} />
+      <TextField label="Value" {...fields.value} />
+    </>
+  )}
+</FormState.Nested>
+```
+
 ### `<FormState.List />`
 
 Sometimes your data might be best represented using an array of objects.
@@ -592,6 +607,21 @@ For our example above, it makes sense to assume that each combination of `option
   field={fields.variants}
   getChildKey={(variant) => `${variant.option}-${variant.value}`}
 >
+```
+
+#### Performance
+
+`<FormState.List />` will only re-render its children if `value`, `initialValue` or `error` on `field` change for performance reasons. While this is highly recommended, there are cased in where you might want to by-pass this check. To force it to always re-render its children, you can pass `update`:
+
+```tsx
+<FormState.List field={fields.variants} update>
+  {fields => (
+    <>
+      <TextField label="Option" {...fields.option} />
+      <TextField label="Value" {...fields.value} />
+    </>
+  )}
+</FormState.List>
 ```
 
 ## Putting it all together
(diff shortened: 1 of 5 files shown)


michelleyschen:
I am having a hard time seeing how this will work - for example Pat's use case of the children should re-render when the modalVisible change.

Can you provided an example code?

amireynoso:
Hey @michelleyschen! You'd use it like this:

```
<FormState.Nested field={fields.variants} update>
  {fields => (
    <>
      <TextField label="Option" {...fields.option} />
      <TextField label="Value" {...fields.value} />
    </>
  )}
</FormState.List>
```

By setting `update` (on either `<FormState.Nested>` or `<FormState.List>`), you're opting out completely of the performance checks that prevent these components to render if the passed `field` has not changed.

Considering that there could be many cases in which you'd want children to update, I thought it'd be better to just allow it to be turned off.

On the case I described in #557, my component should re-render on 2 different prop changes, and in Pat's case, it's a state change. 🤔I'm not sure how to make it smart enough to allow re-render for specific prop/state changes.

michelleyschen:
ok I see, ya this make sense for the time being.

lemonmade:
Hmm, I am not sure this really does what we need. Most likely, I think we should just revert back to using regular `PureComponent`s for these, because it is very unexpected that they would stop an update from happening that has nothing to do with them. If we don't want that, then `update` should at least be a prop we could pass an arbitrary string to; right now, it would be very hard to tell the component to `update`, then `update` again, because you will need to flip it to false/ undefined in between.

patsissons:
the `update` flag should just be opting out of the standard _should update_ logic. maybe the flag name could be changed to be more representative, like `skipUpdateCheck` (ugh, naming things is hard).

lemonmade:
But this is just not a thing you should need. I don't think we should keep performance optimizations that simply break React updates.

patsissons:
ah i see, you're suggesting to just remove the attempt at optimizing perf altogether. Seems reasonable.

amireynoso:
@lemonmade Thanks for your input! I was wondering about this too at the beginning. When it was first introduced in #312, it seemed to be a legitimate concern for some implementations, so, as to not introduce breaking changes, I thought we should at least provide a way to opt-out of it.

I merged but haven't released yet as I was pulled into a meeting. Should I just revert this PR, and open a new one removing `shouldComponentUpdate` for both components? Would that then be a major release?

lemonmade:
I would prefer to revert and remove the `shouldComponentUpdate`, if we want to it's easy enough to go back to this, but adding this and removing it is a breaking change (removing `shouldComponentUpdate` is fine, since it's just an internal optimization and no one should be depending on it(

amireynoso:
That makes sense! Thanks, @lemonmade. 

I opened https://github.com/Shopify/quilt/pull/564 reverting this PR (do I need approvals for that, or I can just go ahead and merge?). I will open a new PR removing the logic in `shouldComponentUpdate` early next week.

PR #295, opened 2018-09-19
React shortcuts context refactor

hughxie:
This pr refactors react-shortcuts to use the new React16.3+ context API. Resolves #196. 

Should be good to go after #347 is merged👍

PR #559, opened 2019-03-07
react-shortcuts: Remove prop-types as a dependency

BPScott:
PR #295 refactored react-shortcuts to use the modern context API which
means prop-types is no longer needed

Pinging @kyledurand for review as they reviewed the original PR


kyledurand:
👍 Looks good to me as long as it's 💚

PR #566, opened 2019-03-08
upgrading to GraphQL v14

patsissons:
Upgrading `graphql-testing`, `jest-mock-apollo`, `react-effect-apollo`, and `react-graphql` to support GraphQL v14


package.json
```diff
@@ -55,6 +58,7 @@
     "plop": "^2.0.0",
     "prettier": "^1.14.0",
     "react": "^16.8.1",
+    "react-apollo": "^2.5.2",
```
patsissons: these new packages added to `devDependencies` are just to squelch some peer dependency warnings.

packages/react-graphql/package.json
```diff
@@ -27,10 +27,12 @@
     "@shopify/async": "^1.1.0",
     "@shopify/react-async": "^1.2.2",
     "@shopify/useful-types": "^1.1.1",
-    "graphql-typed": "^0.3.0",
+    "apollo-client": "^2.5.1",
```
patsissons: I am adding `apollo-client@^2.5.1` to the dependencies list because it is a peer dependency for `react-apollo`. We could alternatively shuffle this dep to `peerDependencies`, but i think it makes more sense as a direct dependency?

packages/jest-mock-apollo/package.json
```diff
@@ -23,21 +23,21 @@
...
-    "apollo-link": "^1.2.3",
-    "graphql-tool-utilities": "^0.9.0",
+    "apollo-link": "^1.2.9",
+    "graphql-tool-utilities": "^0.10.0",
     "tslib": "^1.9.3"
   },
   "devDependencies": {
-    "apollo-cache-inmemory": "^1.2.4",
-    "apollo-client": "^2.3.4",
+    "apollo-cache-inmemory": "^1.5.1",
```
lemonmade: Do we strictly need versions this high? Do lower versions have an incorrect peer dependency? I'd like us to be lenient on react/apollo/typescript dependencies where we can be

packages/react-graphql/package.json
lemonmade: Hmm yeah I guess this is fine since react-apollo is a regular dependency, but looser version restriction would be 💯 if possible

packages/jest-mock-apollo/package.json
patsissons: i didn't check specifically, but i think all needed some sort of a bump. I can definitely peer into their deps and see what is the min requirement.

patsissons: managed to lower the following packages a bit:

* `apollo-cache-inmemory@^1.4.0`
* `apollo-client@^2.4.9`
* `graphql-tag@^2.10.0`
* `react-apollo@^2.2.3`

These are the bare minimum support levels for GraphQL 14

PR #570, opened 2019-03-11
Remove updating logic in FormState.Nested and FormState.List

amireynoso:
## What's the problem
As described in #557, `<FormState.Nested>` and `<FormState.List>` are not updating their children, unless the passed `field` changes.

While this is the expected behaviour (introduced in #312 to improve the performance of these components), there are cases in which it's necessary for the children to be updated. For example, if any of the child form elements also depend on a state change or a prop other than `field` changing.

I first attempted to solve this problem in #558 by adding a new prop to these components which would enable the consumer to "by-pass" the check. However, [it was brought up](https://github.com/Shopify/quilt/pull/558#issuecomment-471034178) by @lemonmade that the checks might not need to be there at all since it breaks the expected behaviour of how a React component should work.

☝️ _The changes introduced in #558 were later reverted in #564._

## How it's solved
This PR is basically reverting the changes introduced in #312:

- :zap: Use `React.PureComponent` instead of `React.Component`
- :fire: Removed `shouldComponentUpdate()` from `<FormState.Nested>` and `<FormState.List>` altogether
- :fire: Removed tests for the updating logic
- :memo: Updated CHANGELOG with proposed release (minor version)

## Tophatting 🎩

I tophatted the changes in my app using `yarn tophat`. I'd recommend testing this in any app that's using either `<FormState.Nested>` or `<FormState.List>` and making sure the child form elements get re-rendered.

### DIFF
--- a/packages/react-form-state/CHANGELOG.md
+++ b/packages/react-form-state/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
+## [0.9.0]
+
+### Removed
+
+- `<Nested />` and `<List />`: removed logic in `shouldComponentUpdate()` limiting updates
+
 ## [0.8.0]
 
 ### Added
--- a/packages/react-form-state/src/components/List.tsx
+++ b/packages/react-form-state/src/components/List.tsx
@@ -9,30 +9,11 @@ interface Props<Fields> {
   getChildKey?(item: Fields): string;
 }
 
-export default class List<Fields> extends React.Component<
+export default class List<Fields> extends React.PureComponent<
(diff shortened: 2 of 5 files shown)


packages/react-form-state/CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
+## [0.9.0]
+
+### Removed
```
lemonmade: I think this is more of an `Update`

amireynoso: You're correct, I just re-read the changelog guidelines and this is definitely `Changed`. I'll update it after merging and before releasing, so as not to lose the ✅.

PR #572, opened 2019-03-12
reducing min apollo version requirements

patsissons:
* `apollo-cache-inmemory`: `^1.4.0` → `^1.3.6`
* `apollo-client`: `^2.4.9` → `^2.3.8` 

## `apollo-cache-inmemory`

The min requirements in [`1.3.6`](https://github.com/apollographql/apollo-client/blob/apollo-cache-inmemory%401.3.6/packages/apollo-cache-inmemory/package.json#L50) align with GraphQL v14, [`1.3.5`](https://github.com/apollographql/apollo-client/blob/apollo-cache-inmemory%401.3.5/packages/apollo-cache-inmemory/package.json#L48) does not have support.

## `apollo-client`

The min requirements in [`2.3.8`](https://github.com/apollographql/apollo-client/blob/apollo-client%402.3.8/packages/apollo-client/package.json#L58) align with GraphQL v14, [`2.3.7`](https://github.com/apollographql/apollo-client/blob/apollo-client%402.3.7/packages/apollo-client/package.json#L58) technically supports GraphQL v14 but has no support for v13.

We may want to consider setting the min requirement to [`2.4.3`](https://github.com/apollographql/apollo-client/blob/apollo-client%402.4.3/packages/apollo-client/package.json#L59) which aligns with the `apollo-cache-inmemory@1.3.6` release. It is not completely clear if maintaining release alignment between these two libraries is necessary. I do not believe it is necessary, so we'll start with the actual min requirements first.


patsissons:
yea i just added a lot of links because i was looking at the wrong versions when i upgraded previously. now we have hard linked evidence 😄

PR #573, opened 2019-03-13
Pass props to renderLoading prop in react-async

cartogram:
Minor change to add a `props` argument in `renderLoading`.

Allows us to do things like this:

```
createAsyncComponent({
  load: () => import(/* webpackChunkName: 'someComponent' */ './SomeComponent'),
  renderLoading: ({someProp}) => (
    <SomeComponent
      someProp={someProp}
    />
  ),
});
```


packages/react-async/src/component.tsx
```diff
@@ -64,7 +64,7 @@ export function createAsyncComponent<
         load={load}
         id={id}
         defer={defer}
-        renderLoading={renderLoading}
+        renderLoading={() => renderLoading && renderLoading(componentProps)}
```
lemonmade: I don't think we really want to return undefined here if `renderLoading` is not provided

packages/react-async/src/tests/component.test.tsx
```diff
@@ -27,7 +27,7 @@ describe('createAsyncComponent()', () => {
     expect(asyncComponent.find(Async).props()).toMatchObject({
       load,
       id,
-      renderLoading,
+      renderLoading: expect.any(Function),
```
lemonmade: We need a test to make sure this is called with the props

packages/react-async/src/component.tsx
lemonmade: We also probably want to make it so that the `async.renderLoading` prop is called with the props, too.

PR #574, opened 2019-03-13
Add pass information to react-effect callbacks

lemonmade:
This PR introduces two features to `@shopify/react-effect`:

1. A `maxPasses` to limit possible spinning forever. This defaults to `5` max iterations
2. `afterEachPass`/ `betweenEachPass` now provide information about the pass, including which iteration we are on, and how long the render/ resolution parts took.

I want to use these to try to remove some possibilities of render-related weirdness in the Web.

### DIFF
--- a/packages/react-effect/src/server.tsx
+++ b/packages/react-effect/src/server.tsx
@@ -1,19 +1,24 @@
 import * as React from 'react';
 import {renderToStaticMarkup} from 'react-dom/server';
 import {EffectContext, EffectManager} from './context';
+import {Pass} from './types';
 
 interface Options {
   include?: symbol[] | boolean;
+  maxPasses?: number;
   decorate?(element: React.ReactElement<any>): React.ReactElement<any>;
   renderFunction?(element: React.ReactElement<{}>): string;
-  betweenEachPass?(): any;
-  afterEachPass?(): any;
+  betweenEachPass?(pass: Pass): any;
+  afterEachPass?(pass: Pass): any;
 }
+
+const DEFAULT_MAX_PASSES = 5;
 
 export function extract(
   app: React.ReactElement<any>,
   {
     include,
+    maxPasses = DEFAULT_MAX_PASSES,
     decorate = identity,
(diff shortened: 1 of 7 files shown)


packages/react-effect/src/server.tsx
```diff
@@ -1,19 +1,24 @@
...
+  maxPasses?: number;
   decorate?(element: React.ReactElement<any>): React.ReactElement<any>;
   renderFunction?(element: React.ReactElement<{}>): string;
-  betweenEachPass?(): any;
-  afterEachPass?(): any;
+  betweenEachPass?(pass: Pass): any;
+  afterEachPass?(pass: Pass): any;
 }
 
+const DEFAULT_MAX_PASSES = 5;
```
GoodForOneFare: 5 feels like a silent footgun for consumers, but you're way more aware of realistic numbers for this so, 🔋

If someone wants to be proactive about this, they could add `afterEachPass` handler that logs/alerts when their app approaches the limit, right?

packages/react-effect/src/server.tsx
```diff
@@ -27,30 +32,75 @@ export function extract(
...
+        await afterEachPass({
+          index,
+          finished: false,
+          renderDuration,
+          resolveDuration,
+        });
+      }
+
+      if (index + 1 >= maxPasses) {
+        return result;
```
GoodForOneFare: So at this point, SSR will just render whatever it has, and unresolved promises will appear client-side with their loading state?

packages/react-effect/src/server.tsx
lemonmade: Yes, I will add thorough logging for this on web's side

packages/react-effect/src/server.tsx
lemonmade: Correct. They will get a "mostly server rendered page ™️", and the client will rehydrate with the data it has, and otherwise simply start those queries/ async translation loading/ whatever itself

PR #576, opened 2019-03-18
Add IntersectionObserver users

lemonmade:
This PR adds a couple consumers for the new `IntersectionObserver` component. I added `DeferTiming.InViewport`, which meant that it was natural to add "only load in viewport" to `react-async` and `react-import-remote`. While adding tests, I noticed that we didn't actually test the "load on idle" bits, so I added a mock for `requestIdleCallback` as well.

### DIFF
--- a/packages/react-async/src/tests/Async.test.tsx
+++ b/packages/react-async/src/tests/Async.test.tsx
@@ -2,6 +2,9 @@ import * as React from 'react';
 import {mount} from 'enzyme';
 import {Effect} from '@shopify/react-effect';
 import {trigger} from '@shopify/enzyme-utilities';
+import {DeferTiming} from '@shopify/async';
+import {IntersectionObserver} from '@shopify/react-intersection-observer';
+import {requestIdleCallback} from '@shopify/jest-dom-mocks';
 
 import {Async} from '../Async';
 import {AsyncAssetContext} from '../context/assets';
@@ -14,11 +17,26 @@ jest.mock('@shopify/react-effect', () => ({
   },
 }));
 
+jest.mock('@shopify/react-intersection-observer', () => ({
+  ...require.requireActual('@shopify/react-intersection-observer'),
+  IntersectionObserver() {
+    return null;
+  },
+}));
+
 function MockComponent() {
   return null;
 }
 
 describe('<Async />', () => {
+  beforeEach(() => {
+    requestIdleCallback.mock();
+  });
+
+  afterEach(() => {
+    requestIdleCallback.restore();
+  });
+
   it('does not call render() before the resolved module is available', () => {
     const promise = createResolvablePromise(MockComponent);
     const render = jest.fn(() => null);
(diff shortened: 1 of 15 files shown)


packages/react-import-remote/src/ImportRemote.tsx
```diff
@@ -50,24 +62,45 @@ export default class ImportRemote extends React.PureComponent<Props, never> {
...
-    try {
-      const imported = await load(source, getImport, nonce);
-      onImported(imported);
-    } catch (error) {
-      onError(error);
-    }
-  }
+        try {
+          const imported = await load(source, getImport, nonce);
+          onImported(imported);
```
GoodForOneFare: Is `onImported` going to be called every time `<ImportRemote defer={DeferTiming.InViewport}>` scrolls into view?    It feels like the kind of thing that should only be called once.

packages/react-import-remote/src/ImportRemote.tsx
```diff
@@ -5,6 +5,10 @@ import {
   WindowWithRequestIdleCallback,
   DeferTiming,
 } from '@shopify/async';
+import {
```
GoodForOneFare: Isn't this prettymuch unshakeable? :/  Would it make more sense for `react-intersection-observer` to provide an intersectable extension to import remote?

packages/react-import-remote/README.md
```diff
@@ -80,4 +80,8 @@ Callback that gets called with the imported global
 
 **defer**
 
-A member of the `DeferTiming` enum (from `@shopify/async`) allowing the import request to occur either on mount (`DeferTiming.Mount`) or until the browser is idle (`DeferTiming.Idle`; requires a polyfill for `window.requestIdleCallback`).
+A member of the `DeferTiming` enum (from `@shopify/async`) allowing the import request to wait until:
+
+- Component mount (`DeferTiming.Mount`; this is the default)
```
GoodForOneFare: Bullets > run-on sentences, yay! 👏

packages/react-async/src/tests/Async.test.tsx
```diff
@@ -94,11 +112,43 @@ describe('<Async />', () => {
...
+      const load = jest.fn(() => promise.promise);
+
+      mount(<Async defer={DeferTiming.Idle} load={load} render={() => null} />);
+      expect(load).not.toHaveBeenCalled();
+
+      requestIdleCallback.runIdleCallbacks();
+      expect(load).toHaveBeenCalled();
+    });
+
+    it('loads the module in on viewport intersection when defer is DeferTiming.InViewport', () => {
```
cartogram: `loads the module in on viewport intersection` => `loads the module on viewport intersection `

packages/react-import-remote/src/ImportRemote.tsx
lemonmade: No, we had logic to remove this once the module had loaded. To make it more bulletproof, I've added a `loading` piece to state that also hides it during the process of loading the module.

packages/react-import-remote/src/ImportRemote.tsx
lemonmade: It is not tree shakeable, but the module is very small. I don't think we can cleanly expose a way to inject stuff in that doesn't rely on shaky type checking and module-scoped variables. I know this isn't ideal for consumers that don't need everything, but web is probably the only real consumer of this right now, and we'll deal with better tree shaking for all of these later.

packages/react-import-remote/src/ImportRemote.tsx
```diff
@@ -50,24 +62,45 @@ export default class ImportRemote extends React.PureComponent<Props, never> {
...
+      return (
+        <>
+          <Preconnect source={url.origin} />
+          {intersectionObserver}
+        </>
+      );
     }
 
-    return null;
+    return intersectionObserver;
```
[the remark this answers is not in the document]
lemonmade: Damn :( I will remove the `display: contents` bits for now.

PR #577, opened 2019-03-18
Support unexpected usage of decimal symbols

ismail-syed:
This PR improves `i18n.unformatCurrency` to handle the usage of `.` as the decimal symbol for locales where the `.` is not actually the expected decimal symbol. This is because users may input numbers with a `.` as the decimal symbol and expect to to work even though the formatted version of the input yields `,` as the decimal symbol. A good example of this is the `fr` locale. 

As per a conversation with @Korri, we should support this workflow for users. 
 
> Although some poeple might still use a `.` in France instead of a `,`
> I know some bank interface that allow `.`, some others that don’t (in Québec)

Related to https://github.com/Shopify/web/issues/10999
cc/ @loic-d

### DIFF
--- a/packages/react-i18n/src/i18n.ts
+++ b/packages/react-i18n/src/i18n.ts
@@ -191,8 +191,8 @@ export default class I18n {
     // and expect it to be treated as the decimal symbol for their locale.
     const hasExpectedDecimalSymbol = input.lastIndexOf(expectedDecimal) !== -1;
     const hasPeriodAsDecimal = input.lastIndexOf(PERIOD) !== -1;
-    const usePeriodDecimal = !hasExpectedDecimalSymbol && hasPeriodAsDecimal;
-    const decimalSymbolToUse = usePeriodDecimal ? PERIOD : expectedDecimal;
+    const usesPeriodDecimal = !hasExpectedDecimalSymbol && hasPeriodAsDecimal;
+    const decimalSymbolToUse = usesPeriodDecimal ? PERIOD : expectedDecimal;
     const lastDecimalIndex = input.lastIndexOf(decimalSymbolToUse);
 
     const integerValue = input
--- a/packages/react-i18n/src/i18n.ts
+++ b/packages/react-i18n/src/i18n.ts
@@ -184,9 +184,16 @@ export default class I18n {
     // This decimal symbol will always be '.' regardless of the locale
     // since it's our internal representation of the string
     const symbol = this.decimalSymbol(currencyCode);
-    const decimal = symbol === DECIMAL_NOT_SUPPORTED ? '.' : symbol;
+    const expectedDecimal = symbol === DECIMAL_NOT_SUPPORTED ? '.' : symbol;
 
-    const lastDecimalIndex = input.lastIndexOf(decimal);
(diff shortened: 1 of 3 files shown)


packages/react-i18n/src/test/i18n.test.ts
```diff
@@ -568,6 +568,36 @@ describe('I18n', () => {
         });
         expect(i18n.unformatCurrency('1,233.45', 'EUR')).toBe('1233.45');
       });
+
+      describe('fr locale', () => {
```
ismail-syed: Any other edge cases we'd like to have covered?

packages/react-i18n/src/i18n.ts
```diff
@@ -184,19 +185,26 @@ export default class I18n {
...
     // since it's our internal representation of the string
     const symbol = this.decimalSymbol(currencyCode);
-    const decimal = symbol === DECIMAL_NOT_SUPPORTED ? '.' : symbol;
+    const expectedDecimal = symbol === DECIMAL_NOT_SUPPORTED ? PERIOD : symbol;
+
+    // For locales that use non-period symbols as the decimal symbol, users may still input a period
+    // and expect it to be treated as the decimal symbol for their locale.
+    const hasExpectedDecimalSymbol = input.lastIndexOf(expectedDecimal) !== -1;
+    const hasPeriodAsDecimal = input.lastIndexOf(PERIOD) !== -1;
+    const usePeriodDecimal = !hasExpectedDecimalSymbol && hasPeriodAsDecimal;
```
loic-d: `usesPeriodDecimal`?

packages/react-i18n/src/test/i18n.test.ts
loic-d: Do we have to be more specific on the locale? `fr-FR` and `fr-CA` will probably have different behaviours?

Korri:
Not sure we should accept `555.555.55` as a valid number, it does not make sense in any country.

ismail-syed:
> Not sure we should accept 555.555.55 as a valid number, it does not make sense in any country.

As per the `Intl.NumberFormat` API, `it` locale uses `.` as a thousandths separator

```
const number = 55555555.55;
console.log(new Intl.NumberFormat('it', { style: 'currency', currency: 'USD' }).format(number));
> "55.555.555,55 USD"
```

`it` could be a good candidate for another test?

Korri:
My concern is more that we should not accept numbers where decimal and thousands separator are the same character as it is clearly a typo, and we might be making a wrong assumption (did they forget a number of did they typo the thousand separator?)

ismail-syed:
> we should not accept numbers where decimal and thousands separator are the same character 

What do you mean by "not accept". In production rails today, in standard `en-US` locale, if the merchant inputs `555.555.555.55` into the price field in product show, it formats to `555,555,555.55`.

Korri:
> 555.555.555.55



> In production rails today, in standard `en-US` locale, if the merchant inputs `555.555.555.55` into the price field in product show, it formats to `555,555,555.55`.

TIL, then this is definitely out of scope for this PR, sorry about that!

PR #594, opened 2019-03-21
♻️ refactor to use the enhancer

michelleyschen:
Refactor the decorator to use the enhancer lib and only keep the part where we make it a class decorator.


packages/decorators/src/tests/memoize.test.ts
```diff
@@ -39,35 +39,6 @@ describe('memoize()', () => {
       expect(myClass.addOne(1)).toEqual(2);
       expect(spy).toHaveBeenCalledTimes(1);
     });
-
```
GoodForOneFare: Can we 🔥 more of these tests?  e.g., `'recalculates the result when first argument changed'` feels like it's just testing what [`function-enhancers` already covers](https://github.com/Shopify/quilt/blob/master/packages/function-enhancers/src/test/memoize.test.ts#L9-L20).

packages/decorators/src/tests/memoize.test.ts
```diff
@@ -39,35 +39,6 @@ describe('memoize()', () => {
...
-    it('deletes oldest cache item when the cache limit is reached', () => {
```
michelleyschen: this test is deleted because 1. MAX_MAP_ENTRIES is not being exported 2. this is no longer the functionality of the decorator

PR #595, opened 2019-03-22
Fix export path of createMockCookies

nuc:
Apparently `createMockContext` was exported instead.

Thanks


nuc:
CLA signed.

lemonmade:
Thanks for the fix @nuc, sorry it took us so long to merge :(

PR #596, opened 2019-03-22
[Performance] Memoize Intl.DateTimeFormat & Int.NumberFormat

macdonaldr93:
Closes https://github.com/Shopify/marketing-technology/issues/4130

We recently made some updates to the StartEndDates component to make it render much faster. PR for that https://github.com/Shopify/web/pull/11956. One of the big performance hits was creating a new `Int.DateTimeFormat` instance every time we were going to format instead of just referencing the same instance and calling the `.format` method.

I ran some tests here: https://jsperf.com/intl-datetimeformat-instantiation-vs-reference-same-one and it's significantly faster to reference the same instance.

I feel like a lot of parts in Shopify will use the same locale and same options for formatting when using the i18n library. Memoizing this is a good idea so that we can reference the same instance.

### DIFF
--- a/packages/react-i18n/src/i18n.ts
+++ b/packages/react-i18n/src/i18n.ts
@@ -1,4 +1,5 @@
 import {languageFromLocale, regionFromLocale} from '@shopify/i18n';
+import memoizeFn from 'lodash/memoize';
 import {memoize} from '@shopify/javascript-utilities/decorators';
 import {
   I18nDetails,
@@ -41,6 +42,16 @@ export interface TranslateOptions {
 const DECIMAL_NOT_SUPPORTED = 'N/A';
 const PERIOD = '.';
 const DECIMAL_VALUE_FOR_CURRENCIES_WITHOUT_DECIMALS = '00';
+const memoizedDateTimeFormatter = memoizeFn(
+  dateTimeFormatter,
+  (locale: string, options: Intl.DateTimeFormatOptions = {}) =>
+    `${locale}${JSON.stringify(options)}`,
+);
+const memoizedNumberFormatter = memoizeFn(
+  numberFormatter,
+  (locale: string, options: Intl.NumberFormatOptions = {}) =>
+    `${locale}${JSON.stringify(options)}`,
+);
 
 export default class I18n {
   readonly locale: string;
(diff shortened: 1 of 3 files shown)


packages/react-i18n/src/i18n.ts
```diff
@@ -306,6 +306,19 @@ export default class I18n {
 
     return decimal.length === 0 ? DECIMAL_NOT_SUPPORTED : decimal;
   }
+
+  @memoize(
+    (
+      locale: string,
+      options: Intl.DateTimeFormatOptions & {style?: DateStyle} = {},
```
macdonaldr93: I think the `DateStyle` helps a lot here... Maybe we can add more of these to accommodate more cases so that less devs use the `options` object.

packages/react-i18n/src/i18n.ts
```diff
@@ -306,6 +306,19 @@ export default class I18n {
...
+    ) => `${locale}${JSON.stringify(options)}`,
```
macdonaldr93: not convinced this is the best key

packages/react-i18n/src/i18n.ts
lemonmade: I don't think you need the `{style}` typing in here, it is removed before this is ever called

packages/react-i18n/src/i18n.ts
```diff
@@ -1,4 +1,5 @@
 import {languageFromLocale, regionFromLocale} from '@shopify/i18n';
+import memoizeFn from 'lodash/memoize';
```
macdonaldr93: @michelleyschen can I use that new memoize function here? Not sure if it's ready yet

michelleyschen: I am going to published it toady :)
however, you may not want to use it since it doesn't run in server.

macdonaldr93:
This is ready for review again! This time it has tests and everything :)

packages/react-i18n/src/test/i18n.test.ts
```diff
@@ -305,6 +305,36 @@ describe('I18n', () => {
       expect(i18n.formatNumber(0.12345, options)).toBe(expected);
     });
 
+    it('updates format on multiple calls', () => {
```
jim-chou-shopify: Do we need to test that it does not create new object if same arguments are called twice?

macdonaldr93: I don't this is a particularly testable case. The instance of `Intl` is outside the scope of this class and would be considered private.

PR #601, opened 2019-03-22
[FormState] Pass lastSubmitErrors to render prop

b-ryu:
### What is this PR doing?

This PR adds client-side submission errors into the `errors` array passed to the render prop.

When users try to submit a form with `FormState` and `validateOnSubmit` is set to `true`, the validators will run on the form and block a call to `onSubmit` if any errors arise.

These errors are then surfaced in `errors`, in the same way remote and external errors are.

This is helpful in cases where specific UI (such as `ErrorBanner`) may need to be updated on submit attempts, but not on field changes. (e.g. [here](https://github.com/Shopify/web/issues/11901) and [here](https://github.com/Shopify/web/issues/12060))

### DIFF
--- a/packages/react-form-state/src/FormState.ts
+++ b/packages/react-form-state/src/FormState.ts
@@ -1,7 +1,7 @@
 /* eslint-disable no-case-declarations */
 import * as React from 'react';
 
-import {mapObject, set, isEqual} from './utilities';
+import {mapObject, set, isEqual, flatMap} from './utilities';
 import {
   FieldDescriptors,
   FieldState,
@@ -185,6 +185,12 @@ export default class FormState<
     });
   }
 
+  private get clientErrors() {
+    const {fields} = this.state;
+
+    return flatMap(Object.values(fields), ({error}) => collectErrors(error));
+  }
+
   private get fields() {
     const {fields} = this.state;
     const fieldDescriptors: FieldDescriptors<Fields> = mapObject(
@@ -216,8 +222,10 @@ export default class FormState<
     if (validateOnSubmit) {
       await this.validateForm();
 
-      if (this.hasClientErrors) {
-        this.setState({submitting: false});
+      const clientErrors = this.clientErrors;
+
+      if (clientErrors.length > 0) {
+        this.setState({submitting: false, errors: clientErrors});
         return;
       }
     }
@@ -527,4 +535,24 @@ function runAllValidators<FieldMap>(
     ...state,
     fields: updatedFields,
   } as State<FieldMap>;
+}
+
+function collectErrors(
+  message: string | any[] | {} | undefined,
+): RemoteError[] {
+  if (!message) {
+    return [];
+  }
+
+  if (typeof message === 'string') {
+    return [{message}];
+  }
+
+  if (Array.isArray(message)) {
+    return flatMap(message, itemError => collectErrors(itemError));
+  }
+
+  return flatMap(Object.values(message), nestedError =>
+    collectErrors(nestedError),
+  );
 }
--- a/packages/react-form-state/src/utilities.ts
+++ b/packages/react-form-state/src/utilities.ts
@@ -50,3 +50,13 @@ export function set<InputType extends Object>(
     } as InputType;
   }
 }
+
+export function flatMap<T>(
+  array: any[],
+  mapper: (item: any, index?: number) => T | T[],
+): T[] {
+  return array.reduce(
+    (acc, item, index) => acc.concat(mapper(item, index)),
+    [],
+  );
+}
--- a/packages/react-form-state/docs/building-forms.md
+++ b/packages/react-form-state/docs/building-forms.md
@@ -396,6 +396,8 @@ function MyComponent() {
 }
 ```
 
+In addition to being propagated down to their respective `fields`, these client-side validation errors can be accessed together at the top level via the `errors` array, in the same way as your remote errors (returned from `onSubmit`) and external errors (see below).
+
 To learn more about building validators, and the built in functions exposed by this package, check out the [validators guide](validators.md).
 
 ## External errors
--- a/packages/react-form-state/src/tests/FormState.test.tsx
+++ b/packages/react-form-state/src/tests/FormState.test.tsx
@@ -2,9 +2,10 @@ import * as React from 'react';
 import faker from 'faker';
 import {mount} from 'enzyme';
 
-import FormState from '..';
+import FormState, {validate, validateNested} from '..';
 import {lastCallArgs} from './utilities';
 import {Input, InputField} from './components';
+import {validateList} from '../validators';
 
 describe('<FormState />', () => {
   it('passes form state into child function', () => {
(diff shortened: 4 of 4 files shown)


packages/react-form-state/src/FormState.ts
```diff
@@ -185,6 +189,24 @@ export default class FormState<
     });
   }
 
+  private get clientErrors() {
+    const {fields} = this.state;
+    const clientErrors: RemoteError[] = [];
+
+    Object.keys(fields).forEach(fieldPath => {
```
patsissons: I think this will encounter problems with nested/list fields. I have been testing out function that will recursively extract the errors and field paths 👍 

```tsx
function collectErrors(
  message: string | any[] | {} | undefined,
  field?: string,
): RemoteError[] {
  if (!message) {
    return [];
  }

  if (typeof message === 'string') {
    return [{message, field: field ? [field] : undefined}];
  }

  if (Array.isArray(message)) {
    return message.flatMap(itemError => collectErrors(itemError, field));
  }

  return Object.entries(message).flatMap(([nestedField, nestedError]) =>
    collectErrors(nestedError, field ? `${field}.${nestedField}` : nestedField),
  );
}
```

and then i was building the client errors with this getter

```tsx
  private get clientErrors() {
    const {fields} = this.state;

    return Object.entries(fields)
      .flatMap(([field, {error}]) => collectErrors(error, field))
      .map(({message, field}) => ({
        field,
        message,
      }));
  }
```

b-ryu: That's a good point, I was wondering about the nested/list field cases 🤔
I guess my reasoning for writing this method like I did is:
- the existing method `hasClientErrors` contains similar logic for detecting validation errors
- the type of `this.state.fields` is `FieldStates`, which I don't _think_ is shaped like a nested object or a list, but I may be mistaken

I'll try tophatting/testing this with nested/list examples and see what I find 👍

patsissons: i was testing on the consumer side with the `formData.fields` which should be similar (it looks like it's derived from `this.state.fields`). i'll give it a test in this repo to see how it works.

patsissons: 🎩 in my playground is working 👍 

this is what i tested.

`FormState.tsx`
```tsx
export interface FormData<Fields> {
  fields: FieldDescriptors<Fields>;
  dirty: boolean;
  valid: boolean;
  clientErrors: RemoteError[]; // <---- added
  errors: RemoteError[];
}

  private get formData() {
    const {errors} = this.state;
    const {externalErrors = []} = this.props; // <-- added
    const {clientErrors, fields, dirty, valid} = this;

    return {
      dirty,
      valid,
      clientErrors, // <-- added
      errors: [...errors, ...externalErrors],
      fields,
    };
  }

  private get clientErrors(): RemoteError[] {
    const {fields} = this.state;

    return Object.entries(fields)
      .flatMap(([field, {error}]) => collectErrors(error, field))
      .map(({message, field}) => ({
        field,
        message,
      }));
  }

function collectErrors(
  message: string | any[] | {} | undefined,
  field?: string,
): RemoteError[] {
  if (!message) {
    return [];
  }

  if (typeof message === 'string') {
    return [{message, field: field ? [field] : undefined}];
  }

  if (Array.isArray(message)) {
    return message.flatMap(itemError => collectErrors(itemError, field));
  }

  return Object.entries(message).flatMap(([nestedField, nestedError]) =>
    collectErrors(nestedError, field ? `${field}.${nestedField}` : nestedField),
  );
}
```

`Playground` (in `web`)
```tsx
/* eslint-disable */

import PlaygroundProvider from 'foundation/components/PlaygroundProvider';

import {Button, TextField} from '@shopify/polaris';
import FormState, {
  validate,
  validateList,
  validateNested,
} from '@shopify/react-form-state';
import {ErrorBanner} from 'components';

export default function Playground() {
  return (
    <PlaygroundProvider>
      <FormState
        initialValues={{single: '', nested: {text: ''}, list: [{text: ''}]}}
        validators={{
          single: [
            validate(() => false, 'single error 1'),
            validate(() => false, 'single error 2'),
          ],
          nested: validateNested({text: validate(() => false, 'nested error')}),
          list: validateList({text: validate(() => false, 'list error')}),
        }}
        validateOnSubmit
      >
        {({
          clientErrors,
          dirty,
          fields: {single, nested, list},
          reset,
          submit,
        }) => (
          <div>
            {clientErrors.length > 0 && <ErrorBanner errors={clientErrors} />}
            <TextField label="Single" {...single} />
            <FormState.Nested field={nested}>
              {({text}) => <TextField label="Nested" {...text} />}
            </FormState.Nested>
            <FormState.List field={list}>
              {({text}) => <TextField label="List" {...text} />}
            </FormState.List>
            <Button onClick={reset} disabled={!dirty}>
              Reset
            </Button>
            <Button onClick={submit} disabled={!dirty}>
              Submit
            </Button>
          </div>
        )}
      </FormState>
    </PlaygroundProvider>
  );
}
```

image

`formData.clientErrors`
image

**NOTE**: i needed to add `"esnext.array"` to the `tsconfig.base.json`'s `lib`

b-ryu: Great! I'll add this in and write some tests 👍 Hopefully we can get this merged soon so we can deal with the dependant issues.

packages/react-form-state/src/utilities.ts
```diff
@@ -50,3 +50,13 @@ export function set<InputType extends Object>(
     } as InputType;
   }
 }
+
+export function flatMap<T>(
```
patsissons: i suspect you're probably safe to pull in the `esnext.array` lib, unless your goal is to specifically avoid it

b-ryu: I was running into some issues with getting the testing environment to recognize `flatMap`. I'll give it another shot.

patsissons: let me know if you want me to have a peek

b-ryu: Took a closer look at this, I think the issue lies with our `ts-jest` dependencies/version. Haven't found a conclusive fix, but open to other ideas in the meantime.

Meanwhile I found [this](https://github.com/Shopify/quilt/pull/475), which gives me a little more confidence in handrolling our own `flatMap`, seeing that we wrote our own implementations of other utilities. But having the real thing is probably ideal 👍

patsissons: Ohhhh we need a polyfill. I think we use `core-js` in `graphql-tools-web`.

packages/react-form-state/docs/validators.md
```diff
@@ -229,3 +229,11 @@ import {validateList} from '@shopify/react-form-state';
     });
   }}
 ```
+
+### `lastSubmitErrors`
+
+When the user attempts to submit an invalid form with `validateOnSubmit` set to `true`, the validators will run on the form, preventing a call to the `onSubmit` callback if an error arises.
```
TheMallen: If this only ever contains client errors, should we maybe add the word `client` errors the name?

packages/react-form-state/src/FormState.ts
```diff
@@ -230,9 +244,9 @@ export default class FormState<
 
     if (errors.length > 0) {
       this.updateRemoteErrors(errors);
-      this.setState({submitting: false});
+      this.setState({submitting: false, lastSubmitErrors: []});
     } else {
-      this.setState({submitting: false, errors});
+      this.setState({submitting: false, lastSubmitErrors: [], errors});
```
TheMallen: Alternatively, we could just also pass `errors` to `lastSubmitErrors` here, but that might end up being confusing / redundant.

b-ryu:
> I think a slightly better way to handle this would be to store what we currently call `errors` as `remoteErrors`, do propagation down to fields using those, name the new state field we are adding here `clientSubmitErrors`, and then when it comes time to actually push errors down into the render prop combine them into the `errors` prop.

Makes sense 👍 I guess something we would have to watch out for is if existing consumers are doing something with `errors` with the expectation that they're only ever going to be remote errors, then they might have to make some adjustments.

patsissons:
Something that could work is to inject the validation errors into `errors` on submit. since we shouldn't be injecting any remote/submit errors anyways (since the validation failure should prevent the submit), it should be safe to just dump validation errors in there. That way there is a single pattern for surfacing form errors, always use `errors`. This should be either a no-op change for existing usage, or a nice improvement in error visualization if the consumer is using form validation.

TheMallen:
@patsissons  the only issue I see with dumping it into errors is that those get propagated down to fields if they have a `fieldPath`

patsissons:
could we omit the `field` when collecting validation errors? i don't believe it's used for the error banner

b-ryu:
Just waiting on a final ✅ from @TheMallen (or someone else on Foundations) and I'll rebase and merge 👍 🙏

Issue #489, opened 2019-01-30
react-compose context typings breaks Typescript 3.2.x


BPScott:
## Overview

When attempting to use typescript 3.2 we got lots of context warnings errors as components seem to expect explicit specification of props that provided by context (and thus you don't actually need to state them explicitly)

From https://shopify.slack.com/archives/CCNRSKLTH/p1545081241166000 (via @henrytao-me) and potentially related to https://shopify.slack.com/archives/CCNRSKLTH/p1546468712025100 (via @patsissons):

```ts
const Component = compose<Props>(
  fooEnhancer,
  barEnhancer,
)(StatelessComponent);

const result = mount(<Component baz="baz" />);
```

>The output `Component` from compose should not require `bar` and `foo` though.
The problem comes from all of the `C` generic here https://github.com/Shopify/quilt/blob/master/packages/react-compose/src/index.ts#L14. Removing all `C` works as expected.

I believe in this example `StatelessComponent` accepts the Prop `baz`, while `fooEnhancer` and `barEnhancer` provide the `foo` and `bar` props via context.

My typescript-fu isn't strong enough adequately explain why removing the `& C` fixes this or what knock on effects it may have.

## Consuming repo

> What repo were you working in when this issue occurred?

Within polaris-react this exposes itself on the [typescript 3.2](https://github.com/Shopify/polaris-react/tree/typescript-3.2) branch. Checkout that branch then run `yarn run sk type-check` and see several errors, that mostly relate to this issue - where it complains about a lack of `polaris` and `context` props that are provided via context.

BPScott:
This also means that consumers of polaris can't update to TS 3.2: https://github.com/Shopify/polaris-react/issues/711

TheMallen:
I think the `& C` is to capture any static methods / properties the component has outside of those that are on every `ReactComponent`.  Without it if your component has any subcomponents on it as static properties (some Polaris components do this iirc) TS would not know about them.

patsissons:
It might be possible to shuffle `& C` to `hoistStatics` typing and eliminate the issue. That is how `recompose` does it (I think). It depends on where the type is getting inferred I suppose.

https://github.com/acdlite/recompose/blob/master/docs/API.md#hoiststatics

BPScott:
[Continuing disclaimers that this is beyond my TS ability)

[...]

PR #602, opened 2019-03-22
Upgrade to typescript 3.2.1 and fix react-compose static fields

henrytao-me:
This PR fix react-compose static fields in typescript > 3.2.1.

Reference https://github.com/Shopify/quilt/issues/489

### DIFF
--- a/packages/react-compose/src/index.ts
+++ b/packages/react-compose/src/index.ts
@@ -4,6 +4,11 @@ import hoistStatics from 'hoist-non-react-statics';
 export type ReactComponent<P> = React.ComponentType<P>;
 export type ComponentClass = React.ComponentClass<any>;
 
+export type StaticFields<Object> = {
+  [Key in keyof Object]: Key extends 'prototype' ? never : Key
+}[keyof Object];
+export type Statics<Object> = Pick<Object, StaticFields<Object>>;
+
 export type WrappingFunction = (
   Component: ReactComponent<any>,
 ) => ReactComponent<any>;
@@ -13,7 +18,7 @@ export default function compose<Props>(
 ) {
   return function wrapComponent<ComposedProps, C>(
     OriginalComponent: ReactComponent<ComposedProps> & C,
-  ): ReactComponent<Props> & C {
+  ): ReactComponent<Props> & Statics<typeof OriginalComponent> {
     const result: ReactComponent<ComposedProps> = wrappingFunctions.reduceRight(
       (component: ReactComponent<any>, wrappingFunction: WrappingFunction) =>
         wrappingFunction(component),
@@ -23,6 +28,6 @@ export default function compose<Props>(
     return hoistStatics(
       result as ComponentClass,
       OriginalComponent as ComponentClass,
-    ) as ReactComponent<Props> & C;
+    ) as ReactComponent<Props> & Statics<typeof OriginalComponent>;
   };
 }
--- a/packages/react-compose/src/index.ts
+++ b/packages/react-compose/src/index.ts
@@ -4,7 +4,7 @@ import hoistStatics from 'hoist-non-react-statics';
 export type ReactComponent<P> = React.ComponentType<P>;
 export type ComponentClass = React.ComponentClass<any>;
 
-export type StaticFields<Object> = {
+type StaticFields<Object> = {
   [Key in keyof Object]: Key extends 'prototype' ? never : Key
 }[keyof Object];
 export type Statics<Object> = Pick<Object, StaticFields<Object>>;
--- a/packages/react-compose/src/index.ts
+++ b/packages/react-compose/src/index.ts
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import hoistStatics from 'hoist-non-react-statics';
-import {NonReactStatics} from 'packages/useful-types/src';
+import {NonReactStatics} from '@shopify/useful-types';
 
 export type ReactComponent<P> = React.ComponentType<P>;
 export type ComponentClass = React.ComponentClass<any>;
--- a/packages/react-compose/src/index.ts
+++ b/packages/react-compose/src/index.ts
@@ -1,13 +1,9 @@
 import * as React from 'react';
 import hoistStatics from 'hoist-non-react-statics';
+import {NonReactStatics} from 'packages/useful-types/src';
 
 export type ReactComponent<P> = React.ComponentType<P>;
 export type ComponentClass = React.ComponentClass<any>;
-
-type StaticFields<Object> = {
-  [Key in keyof Object]: Key extends 'prototype' ? never : Key
-}[keyof Object];
-export type Statics<Object> = Pick<Object, StaticFields<Object>>;
 
 export type WrappingFunction = (
   Component: ReactComponent<any>,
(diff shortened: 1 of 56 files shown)


henrytao-me:
There are few places that I think it must be a better way to do it but I am not really sure.

packages/jest-mock-router/src/index.ts
```diff
@@ -78,5 +78,5 @@ export function createLocation(stubs?: Partial<FullLocation>): FullLocation {
       return JSON.stringify(this);
     },
     ...stubs,
-  };
+  } as FullLocation;
```
lemonmade: Why was this cast needed?

packages/react-compose/src/index.ts
```diff
@@ -4,6 +4,11 @@ import hoistStatics from 'hoist-non-react-statics';
 export type ReactComponent<P> = React.ComponentType<P>;
 export type ComponentClass = React.ComponentClass<any>;
 
+type StaticFields<Object> = {
+  [Key in keyof Object]: Key extends 'prototype' ? never : Key
```
lemonmade: I think you will want to omit a few more fields here:

```tsx
// See https://github.com/mridgway/hoist-non-react-statics/blob/master/src/index.js#L6
type ReactStatics = 'displayName' | 'getDerivedStateFromProps' | 'getDerivedStateFromErrors' | 'childContextTypes` | 'contextType' | 'contextTypes' | 'getDefaultProps' | 'propTypes';

type NonReactStatics<Object> = Pick<Object, Exclude<keyof Object, ReactStatics>>;
```

packages/react-compose/src/index.ts
```diff
@@ -13,7 +18,7 @@ export default function compose<Props>(
 ) {
   return function wrapComponent<ComposedProps, C>(
     OriginalComponent: ReactComponent<ComposedProps> & C,
```
lemonmade: I think we can maybe do `C extends ReactComponent<ComposedProps>` for the generic, then ditch the `& C` bits here

packages/react-i18n/src/decorator.tsx
```diff
@@ -34,6 +34,11 @@ type TranslationMap = {
   [key: string]: MaybePromise<TranslationDictionary | undefined>;
 };
 
+type StaticFields<Object> = {
+  [Key in keyof Object]: Key extends 'prototype' ? never : Key
+}[keyof Object];
+export type Statics<Object> = Pick<Object, StaticFields<Object>>;
```
lemonmade: Shoot sucks that we have this in multiple places, we should maybe put these types in `@shopify/useful-types` instead.

BPScott:
Hi @henrytao-me thanks so much for looking into this.

I've tested this in polaris-react's typescript 3.2 branch and it seems to remove 80% of our compilation errors. We've still got a few remaining compilation errors but they all seem unrelated to this compose issue.

henrytao-me:
@BPScott Cool. Thanks for tophat. I will tryin to update this branch and ship it soon.

henrytao-me:
FYI, for some reasons, the static fields stop working after rebasing. I am investigating...

packages/react-compose/src/index.ts
henrytao-me: I tried and looks like it doesn't work unfortunately.

henrytao-me:
This PR is good for another look. 🙇 
cc @lemonmade @BPScott

packages/react-compose/src/index.ts
```diff
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import hoistStatics from 'hoist-non-react-statics';
+import {NonReactStatics} from 'packages/useful-types/src';
```
lemonmade: I know this is the automatic path VSCode adds, but it doesn't actually work once deployed. You'll need to change this to `@shopify/useful-types`

packages/react-i18n/src/decorator.tsx
```diff
@@ -11,13 +12,14 @@ export interface WithI18nProps {
...
-  ): React.ComponentType<OwnProps> & C => {
+  ): React.ComponentType<OwnProps> &
+    NonReactStatics<typeof WrappedComponent> => {
     function WithTranslations(props: OwnProps) {
       const [i18n, ShareTranslations] = useI18n(i18nOptions);
 
       return (
         <ShareTranslations>
-          <WrappedComponent {...props} i18n={i18n} />
+          <WrappedComponent {...props as any} i18n={i18n} />
```
lemonmade: What was the error here?

packages/useful-types/src/types.ts
```diff
@@ -53,3 +53,14 @@ export type NonOptionalKeys<T> = {
...
+
+type ReactStatics =
+  | 'displayName'
+  | 'getDerivedStateFromProps'
+  | 'getDerivedStateFromErrors'
+  | 'childContextTypes'
+  | 'contextType'
+  | 'contextTypes'
+  | 'getDefaultProps'
+  | 'propTypes';
```
lemonmade: Did you have a particular list you were referencing here?

packages/react-compose/src/index.ts
henrytao-me: Oh. I missed that. Fixing it now.

packages/react-i18n/src/decorator.tsx
henrytao-me: It complains about this without casting to any. I don't have a proper solution for it yet.

```
(parameter) WrappedComponent: (React.ComponentClass<OwnProps & WithI18nProps, any> & C) | (React.FunctionComponent<OwnProps & WithI18nProps> & C)
Type 'OwnProps & { i18n: I18n; }' is not assignable to type '(IntrinsicAttributes & LibraryManagedAttributes<ComponentClass<OwnProps & WithI18nProps, any> & C, PropsWithChildren<OwnProps & WithI18nProps>>) | (IntrinsicAttributes & LibraryManagedAttributes<...>)'.
  Type 'OwnProps & { i18n: I18n; }' is not assignable to type 'IntrinsicAttributes & LibraryManagedAttributes<FunctionComponent<OwnProps & WithI18nProps> & C, PropsWithChildren<OwnProps & WithI18nProps>>'.
    Type 'OwnProps & { i18n: I18n; }' is not assignable to type 'LibraryManagedAttributes<FunctionComponent<OwnProps & WithI18nProps> & C, PropsWithChildren<OwnProps & WithI18nProps>>'.ts(2322)
```

packages/useful-types/src/types.ts
henrytao-me: Add your reference `// Reference https://github.com/mridgway/hoist-non-react-statics/blob/master/src/index.js#L6`

packages/react-intersection-observer/src/IntersectionObserver.tsx
```diff
@@ -30,5 +30,5 @@ export const IntersectionObserver = React.memo(function IntersectionObserver({
 
   useValueTracking(intersection, newValue => onIntersectionChange(newValue));
 
-  return <Wrapper ref={ref}>{children}</Wrapper>;
+  return React.createElement(Wrapper, {ref}, children);
```
henrytao-me: Discussed with Chris offline, this is an alternative way to create dynamic component.

henrytao-me:
This PR is ready to ship if we don't have any other comments. 🙇

BPScott:
Noting extra from me, aside from one more thank you for taking ownership of this :shipit:

patsissons:
Thanks so much for making this effort, this will have a very positive impact on a lot of dependent projects. 🎉

henrytao-me:
Thanks Chris for your magic typescript skill.

PR #603, opened 2019-03-25
@shopify/react-testing

lemonmade:
This PR introduces an alternative testing library to Enzyme. We've had numerous troubles with Enzyme in the past that made me decide to dive into this:

* It has not always been very quick to pick up new React features, meaning we either have to do hacks to test them, avoid testing them, or wait to upgrade React
* It has a lot of features we don't like and try to actively dissuade people from, like `state`/ `setState`/ other things that deal with parts that we consider to be implementation details within a React component
* It does not support some features we need/ we have had trouble submitting some features upstream. Two notable examples are `trigger` and automatic mount tracking/ global "unmount all"
* Its API is not well structured to promote type safety

The main alternative to Enzyme is https://github.com/kentcdodds/react-testing-library. This library is better for keeping up with React's many changes, but the tests it promotes are completely different from our own style (very much driven by the DOM and not really caring about component boundaries).

The [README](https://github.com/Shopify/quilt/compare/react-testing?expand=1#diff-f0d1b8ad002396fa1b604452005d6f67) in this PR details the library fairly well. It is roughly similar to Enzyme, but has built-in support for mount tracking/ `trigger`, better support for strict typing, and a minimal API that only exposes the things we actively use a lot of. I rewrote [one set of tests](https://github.com/Shopify/quilt/compare/react-testing?expand=1#diff-b64453815182825f823a20c413143c95) to show off some parts of the API.

In reviewing, aside from the obvious "do we need this" feedback (that I'd love to hear!), there's one big API decision I'd like some thoughts on: returning arrays from the "findAll" style methods, and returning `null` from the "findOne" style methods. Doing this simplifies the library a great deal, but does mean that non-null assertions are more frequently needed. Personally, I think that's appropriate, since there is no typing guarantee we can give that any "find" operation will return anything.

### DIFF
--- a/packages/react-testing/README.md
+++ b/packages/react-testing/README.md
@@ -1,0 +1,285 @@
+# `@shopify/react-testing`
+
+[![Build Status](https://travis-ci.org/Shopify/quilt.svg?branch=master)](https://travis-ci.org/Shopify/quilt)
+[![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](LICENSE.md) [![npm version](https://badge.fury.io/js/%40shopify%2Freact-testing.svg)](https://badge.fury.io/js/%40shopify%2Freact-testing.svg)
+
+A library for testing React components according to [Shopify conventions](https://github.com/Shopify/web-foundation/blob/master/Best%20practices/React/Testing.md).
+
+## Table of contents
+
+1. [Installation](#installation)
+1. [Usage](#usage)
+1. [FAQ](#faq)
+
+## Installation
+
+```bash
+$ yarn add @shopify/react-testing
+```
+
+## Usage
+
+This library allows you to test React components with a focus on type safety and testing based on a component’s external API. In order to keep the API small and easy-to-use, it will generally track to only the latest minor release of React.
+
+### <a name="mount"></a> `mount(element: React.ReactElement<any>)`
+
+Mounts a component to the DOM and returns a [`Root`](#root) instance. Note that in order for this to work, you must have a simulated browser environment, such as the `jsdom` environment that Jest uses.
+
+### <a name="destroyAll"></a> `destroyAll()`
+
+All mounted components are tracked in-memory. `destroyAll()` forcibly unmounts all mounted components and removes the DOM node used to house them. You should run this after each test that mounts a component (this is often done in a global `afterEach` hook).
+
+### <a name="root"></a> `Root<Props>`
+
+A `Root` object represents a mounted React tree. Most of the properties and methods it exposes are simply forwarded to the [`Element`](#element) instance representing the top-level component you rendered:
+
+- [#children](#children)
+- [#descendants](#descendants)
+- [#props](#props)
+- [#isDOM](#isDOM)
+- [#instance](#instance)
+- [#domNode](#domNode)
+- [#domNodes](#domNodes)
+- [#html()](#html)
+- [#text()](#text)
+- [#is()](#is)
+- [#prop()](#prop)
+- [#find()](#find)
+- [#findAll()](#findAll)
+- [#findWhere()](#findWhere)
+- [#findAllWhere()](#findAllWhere)
+- [#trigger()](#trigger)
+- [#triggerKeypath()](#triggerKeypath)
+
+This object also has a number of methods that only apply to the root of a component tree:
+
+#### <a name="mount"></a> `mount()`
+
+Re-mounts the component to the DOM. If the component is already mounted, this method will throw an error.
+
+#### <a name="unmount"></a> `unmount()`
+
+Unmounts the component from the DOM. If the component is not already mounted, this method will throw an error. This method can be useful for testing side effects that occur in `componentWillUnmount` or `useEffect` hooks.
+
+#### <a name="setProps"></a> `setProps(props: Partial<Props>)`
+
+Allows you to change a subset of the props specified when the component was originally mounted. This can be useful to test behaviour that is only caused by a change in props, such as `getDerivedStateFromProps` or its equivalent `useRef`/ `useState` hook version.
+
+#### <a name="act"></a> `act<T>(action: () => T): T`
+
+Performs an action in the context of a React [`act() block`](https://reactjs.org/docs/test-utils.html#act), then updates the internal representation of the React tree. You **must** use this whenever performing an action that will cause the React tree to set state and re-render, such as simulating event listeners being called. Failing to do so will print a warning, and the React tree will not be updated for subsequent calls to methods such as `find()`.
+
+```tsx
+function MyComponent() {
+  const [clicked, setClicked] = React.useState(false);
+
+  React.useEffect(
+    () => {
+      const handler = () => setClicked(true);
+      document.body.addEventListener('click', handler);
+      return () => document.body.removeEventListener('click', handler);
+    },
+    [setClicked],
+  );
+
+  return clicked ? <div>I’ve been clicked!</div> : <div>Nothing yet</div>;
+}
+
+const myComponent = mount(<MyComponent />);
+
+// If you don’t do this, you’ll see a warning and the subsequent assertion
+// will fail
+myComponent.act(() => simulateClickOnBody());
+
+expect(myComponent.text()).toContain('I’ve been clicked!');
+```
+
+#### <a name="destroy"></a> `destroy()`
+
+Unmounts the component and removes its associated DOM node. This method ensures that nothing leaks between tests. It is called on all un-destroyed `Root` objects when you call [`destroyAll()`](#destroyAll)
+
+#### <a name="forceUpdate"></a> `forceUpdate()`
+
+Forces the root component to re-render. This can be necessary in some cases where globals change in a way that does not already cause a "natural" React update, but in general, this method should not be necessary.
+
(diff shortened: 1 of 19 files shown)


lemonmade:
If people are happy with this, I also have a few matchers I'd like to add, then I consider this to be mostly feature-complete

packages/react-testing/README.md
```diff
@@ -0,0 +1,267 @@
...
+
+## Usage
+
+This library is broadly similar to [Enzyme](https://github.com/airbnb/enzyme), but with a focus on type safety and testing based on a component’s external API. In order to keep the API small and easy-to-use, it will generally track to only the latest minor release of React.
+
+### <a name="mount"></a> `mount(element: React.ReactElement<any>)`
+
+Mounts a component to the DOM and returns a [`Root`](#root) instance. Note that in order for this to work, you must have a simulated browser environment, such as the `jsdom` environment that Jest uses.
+
+Unlike Enzyme, there are no options to `render` or `shallow` render a component. These patterns do not work well under many circumstances, so there is no API in this library to use them.
```
cartogram: Don't bother talking about or comparing to Enzyme at all IMO.

I could see a separate FAQ part in this readme that is like "How does this library compare to Enzyme?" or "Why use this library over Enzyme?" where the differences are listed. But in the main part of this Readme, just focus quickly on the things this library does and consider that some new devs might not even know what Enzyme is.

packages/react-testing/src/errors.ts
```diff
@@ -0,0 +1,63 @@
...
+      return;
+    }
+
+    originalConsoleError.call(console, ...args);
+  };
+
+  function handleErrorEvent(event: ErrorEvent) {
+    const {error} = event;
+
+    // I want to silence all errors and know what I'm doing
```
TheMallen: lol

packages/react-testing/src/root.tsx
```diff
@@ -0,0 +1,249 @@
...
+  MaybeFunctionReturnType as ReturnType,
+} from '@shopify/useful-types';
+
+import {TestWrapper} from './TestWrapper';
+import {Element, Predicate} from './element';
+import {Tag, Fiber, ReactInstance, FunctionKeys} from './types';
+import {withIgnoredReactLogs} from './errors';
+
+// eslint-disable-next-line typescript/no-var-requires
+const {findCurrentFiberUsingSlowPath} = require('react-reconciler/reflection');
```
TheMallen: o.o

packages/react-testing/src/root.tsx
```diff
@@ -0,0 +1,249 @@
...
+
+  prop<K extends keyof Props>(key: K) {
+    return this.withRoot(root => root.prop(key));
+  }
+
+  find<Type extends React.ComponentType<any> | string>(type: Type) {
+    return this.withRoot(root => root.find(type));
+  }
+
+  findAll<Type extends React.ComponentType<any> | string>(type: Type) {
```
TheMallen: Can we deviate from the enzyme APIs and closer to common functional APIs for this stuff?  Personally I think it would be nice to have like `find`, `filter` instead of `find`, `findAll`

packages/react-testing/src/root.tsx
```diff
@@ -0,0 +1,249 @@
...
+
+  private get mounted() {
+    return this.wrapper != null;
+  }
+
+  constructor(private tree: React.ReactElement<Props>) {
+    this.mount();
+  }
+
+  perform<T>(action: () => T, {update = true} = {}): T {
```
TheMallen: Any reason not to call this `act` for symmetry with react's API?

packages/react-testing/src/root.tsx
```diff
@@ -0,0 +1,249 @@
...
+
+  findWhere(predicate: Predicate) {
+    return this.withRoot(root => root.findWhere(predicate));
+  }
+
+  findAllWhere(predicate: Predicate) {
+    return this.withRoot(root => root.findAllWhere(predicate));
+  }
+
+  trigger<K extends string & FunctionKeys<Props>>(
```
TheMallen: I love `trigger` living here.

packages/react-testing/src/root.tsx
```diff
@@ -0,0 +1,249 @@
...
+    prop: K,
+    ...args: Arguments<Props[K]>
+  ): ReturnType<NonNullable<Props[K]>> {
+    return this.withRoot(root => root.trigger(prop, ...args));
+  }
+
+  triggerKeypath<T = unknown>(keypath: string, ...args: unknown[]) {
```
TheMallen: Why isn't this just an overload of `trigger`?

packages/react-testing/src/tests/root.test.tsx
```diff
@@ -0,0 +1,247 @@
...
+      for (const queued of queue) {
+        queued();
+      }
+
+      expect(perform).toHaveBeenCalled();
+    });
+  });
+
+  describe('errors', () => {
+    const Thrower: React.ComponentType = () => {
```
TheMallen: Should we test suspense style promise throwing? Or leave it till it's more mature

packages/react-testing/README.md
```diff
@@ -0,0 +1,267 @@
...
+- [#props](#props)
+- [#isDOM](#isDOM)
+- [#instance](#instance)
+- [#domNode](#domNode)
+- [#domNodes](#domNodes)
+- [#html()](#html)
+- [#text()](#text)
+- [#is()](#is)
+- [#prop()](#prop)
+- [#find()](#find)
```
TheMallen: I mentioned below I'd prefer `find`, `filter` to `find`, `findAll`, but that didn't cover the difference between the regular and `where` prefixed versions.

I think here's roughly how I'd like to see this look

```
find(p: predicate<Element>): Element
filter(p: predicate<Element>): Element[]
querySelector(selector: string | ReactComponentType): Element
querySelectorAll(selector: string | ReactComponentType): Element[]
```

This way instead of basing our APIs on Enzyme they hew close to the standard JS APIs. It means a slightly more annoying time changing over to this library, but I think it is more grokkable to total newcomers and helps give our testing library it's own identity as a sensible solution.

TheMallen:
I also wonder if we're going to be bit by not having an equivalent to `simulate`, I know we almost always use `trigger` but we might want to run this by the Polaris folks to see if they would be hurt by it.

packages/react-testing/README.md
TheMallen: If `querySelector` is too verbose I would also accept 
- `query` / `queryAll`
- the library exporting some helper predicates `find(matchSelector('query'))`, `filter(matchComponent(MyComponent))`
- `find` and `filter` also accepting query selectors as an override
- `$` (JK)

packages/react-testing/src/root.tsx
lemonmade: This is a tough one. Personally, I find `filter` to be confusing, mostly because some methods return arrays, and arrays have their own filter. `filter` also sounds like the object is a collection (which is not the case). Technically `find` should have all the same problems but in my brain it doesn't get associated with arrays in the same way. Maybe I am crazy?

packages/react-testing/src/root.tsx
lemonmade: I don't like the act name :P but you're right, better to match them

packages/react-testing/src/root.tsx
lemonmade: The override causes us to lose the strong typing of the return value/ arguments/ available keys, as it just flattens down to be `(key: string, ...args: unknown[]): unknown` :(

packages/react-testing/src/tests/root.test.tsx
lemonmade: We don't really use it since it doesn't support SSR so I think we should leave it out for now.

packages/react-testing/README.md
lemonmade: My rule of thumb here was to make the common thing the easiest thing, which is why I prefer having `.shortName(Component)` be available since we do this by far the most often. I am not totally opposed to query/ querySelector but it does feel odd because we do not allow you to pass in arbitrary DOM-ey selectors, even though you would probably expect that given the alignment of naming.

TheMallen: Ah, I didn't catch that difference, that is a good reason not to use `querySelector` as a name, but also probably a good reason not to use the same name as enzyme's ([wrapper#find](https://airbnb.io/enzyme/docs/api/ReactWrapper/find.html) does allow dom style selectors).

I think the best option then is `find` and `filter` accepting `predicate | ComponentType`

packages/react-testing/src/root.tsx
TheMallen: fair fair :(

packages/react-testing/src/root.tsx
TheMallen: I'm having trouble responding to this because I like that some methods return arrays that have their own filter :P 

It's not an array but semantically `filter` and `find` make just as much sense as in an array IMO.

Not something I feel *that* strongly about, I just always found enzyme's choice here a bit weird.

lemonmade:
We are going to merge this for now, try it out for a bit, and see if we then want to make any adjustments.

PR #605, opened 2019-03-25
⚡️ Deferred scripts are now in `<head>`

GoodForOneFare:
For pages with a large body payload, this may allow the DOM parser to start loading external content slightly sooner.

When testing in web, I could never get this to make a measurable difference, but it's possible that a page with many translations + Apollo data could be improved by this change.

### DIFF
--- a/packages/react-html/src/server/components/tests/Html.test.tsx
+++ b/packages/react-html/src/server/components/tests/Html.test.tsx
@@ -87,14 +87,14 @@ describe('<Html />', () => {
   });
 
   describe('deferedScripts', () => {
-    it('generates a script tag in the body with the `defer` attribute', () => {
+    it('generates a script tag in the head with the `defer` attribute', () => {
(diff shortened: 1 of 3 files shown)


packages/react-html/src/server/components/tests/Html.test.tsx
```diff
@@ -172,15 +196,7 @@ describe('<Html />', () => {
...
-      const serializerIndex = findIndex(bodyContent, element =>
-        element.is(Serialize),
-      );
-
-      const scriptsIndex = findIndex(bodyContent, element =>
-        element.is(Script),
-      );
-
-      expect(serializerIndex).toBeLessThan(scriptsIndex);
+      expect(bodyContent.find(Serialize)).toHaveLength(1);
```
lemonmade: I think you can just do `html.find('body').find(Serialize)` here

PR #606, opened 2019-03-26
[DO NOT MERGE] Proof of concept - remove root tsconfig.json

GoodForOneFare:
Still some scrappiness to work through, but here's how using [Jest's multi-project runner](https://jestjs.io/blog/2017/05/06/jest-20-delightful-testing-multi-project-runner) could remove the root tsconfig.

The interesting bit of the changeset is https://github.com/Shopify/quilt/pull/606/commits/3120e896eef7ab88b915b2c6e8c094cecb54a0af.

### y tho?
* Should improve type checking for tests (once we remove `excludes` from nested tsconfigs)
* Easier for devs to understand tests vs build TypeScript config
* 🔥 **Emoji test ids** 🔥 
q2 (zsh) 2019-03-26 08-43-05

### TODO
* Coverage will still require a global config; haven't really dug into how to get this going, but it looks possible
* Figure out how to use `...` in Jest mocks without `tslib` being flagged as a hoisting violation
* Figure out why multiple versions of `react-is` broke module resolution
* Is the build broken?  It took _minutes_ to build this, then failed on address / address-mock compilation :/

### How 🎩?
`yarn jest` should run all tests without any failures


packages/react-async/src/tests/Async.test.tsx
```diff
@@ -17,12 +17,17 @@ jest.mock('@shopify/react-effect', () => ({
   },
 }));
 
-jest.mock('@shopify/react-intersection-observer', () => ({
-  ...require.requireActual('@shopify/react-intersection-observer'),
-  IntersectionObserver() {
-    return null;
-  },
-}));
+jest.mock('@shopify/react-intersection-observer', () =>
```
GoodForOneFare: Had to rewrite these to prevent `...` transpiling to `tslib_1.assign`.  tslib refs are considered hoisting / memory leak bugs.  We could either:
- Add tslib* to Jest's blessed variable list
- Tweak tsconfig's `target` / `module` / `importHelpers` settings (probably not feasible here; it worked for sewing-kit)

Issue #585, opened 2019-03-21
Re-write all components from all packages using hooks


lemonmade:

lemonmade:
Remaining:

- [ ] Stuff in `react-async`
- [ ] Stuff in `react-google-analytics`
- [ ] `ImportRemote` (Matt's on it)
- [ ] `ShopifyAppRoutePropagator` (Michelle's on it)
- [ ] `ShotcutProvider`
- [ ] `TrackingPixel`

lemonmade:
Going to call this finished for now 👍

PR #608, opened 2019-03-26
✨ Adds `useImportRemote()` hook 

cartogram:
Part of #585

This PR adds a `useImportRemote` hook and refactors the ImportRemote component to use this hook. 

DEMO -> https://monosnap.com/file/SuTFwyrW2OPt6IEYL80m5DQ36q6zcw

### Usage

I am suggesting the following API for this hook.

```tsx
function useImportRemote<Imported = any>(
  source: string,
  options: Options<Imported>,
): {
  state: State;
  imported: Imported | null;
  intersectionRef: any;
}

interface Options<Imported> {
  nonce?: string;
  defer?: DeferTiming;
  onError?(error: Error): void;
  getImport(window: Window): Imported;
  onImported?(imported: Imported): void;
}

type State = 'loading' | 'error' | 'loaded' | null;

```

### DIFF
--- a/packages/jest-dom-mocks/src/intersection-observer.ts
+++ b/packages/jest-dom-mocks/src/intersection-observer.ts
@@ -10,6 +10,8 @@ export default class IntersectionObserverMock {
 
   private isUsingMockIntersectionObserver = false;
   private originalIntersectionObserver = (global as any).IntersectionObserver;
+  private originalIntersectionObserverEntry = (global as any)
+    .IntersectionObserverEntry;
 
   simulate(
     entry:
@@ -43,6 +45,9 @@ export default class IntersectionObserverMock {
 
     const setObservers = (setter: (observers: Observer[]) => Observer[]) =>
       (this.observers = setter(this.observers));
+
+    (global as any).IntersectionObserverEntry = () => {};
+    (global as any).IntersectionObserverEntry.prototype.intersectionRatio = 0;
 
     (global as any).IntersectionObserver = class FakeIntersectionObserver {
       constructor(
@@ -87,6 +92,8 @@ export default class IntersectionObserverMock {
     }
 
     (global as any).IntersectionObserver = this.originalIntersectionObserver;
+    (global as any).IntersectionObserverEntry = this.originalIntersectionObserverEntry;
+
     this.isUsingMockIntersectionObserver = false;
     this.observers.length = 0;
   }
--- a/packages/jest-dom-mocks/src/intersection-observer.ts
+++ b/packages/jest-dom-mocks/src/intersection-observer.ts
@@ -47,7 +47,15 @@ export default class IntersectionObserverMock {
       (this.observers = setter(this.observers));
 
     (global as any).IntersectionObserverEntry = () => {};
-    (global as any).IntersectionObserverEntry.prototype.intersectionRatio = 0;
+    Object.defineProperty(
+      IntersectionObserverEntry.prototype,
+      'intersectionRatio',
+      {
+        get() {
+          return 0;
+        },
+      },
+    );
 
     (global as any).IntersectionObserver = class FakeIntersectionObserver {
       constructor(
--- a/packages/react-import-remote/CHANGELOG.md
+++ b/packages/react-import-remote/CHANGELOG.md
@@ -8,6 +8,13 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 <!-- ## [Unreleased] - -->
 
+## 2.0.0 - 2019-05-03
+
+### Changed
+
+- Changes to the props of `<ImportRemote />` mid-render will cause the import to be canceled. ([608](https://github.com/Shopify/quilt/pull/608))
+- Added `useImportRemote()` hook ([608](https://github.com/Shopify/quilt/pull/608))
+
 ### Added
 
 - This CHANGELOG [(#575)](https://github.com/Shopify/quilt/pull/575)
--- a/packages/react-import-remote/CHANGELOG.md
+++ b/packages/react-import-remote/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 ### Changed
 
 - Changes to the props of `<ImportRemote />` mid-render will cause the import to be canceled. ([608](https://github.com/Shopify/quilt/pull/608))
+- Removed `onError` callback prop. `onImported` now receives an error in the case that an error occurs during the import. ([608](https://github.com/Shopify/quilt/pull/608))
 - Added `useImportRemote()` hook ([608](https://github.com/Shopify/quilt/pull/608))
 
 ### Added
--- a/packages/react-import-remote/README.md
+++ b/packages/react-import-remote/README.md
@@ -15,6 +15,8 @@ $ yarn add @shopify/react-import-remote
 
 The provided utilities are intended only for external scripts that load globals. Other JavaScript should use the native `import()` operator for asynchronously loading code. These utilities cache results by source, so only a single `script` tag is ever added for a particular source.
 
+### <ImportRemote />
+
 ```tsx
 import * as React from 'react';
 import ImportRemote from '@shopify/react-import-remote';
@@ -25,23 +27,22 @@ interface WindowWithGlobal extends Window {
   remoteGlobal: RemoteGlobal;
 }
 
-class MyComponent extends React.Component {
-  ...
+function MyComponent() {
+  return (
+    <ImportRemote
+      preconnect
+      source="https://some-external-service.com/global.js"
+      getImport={}
+      onImported={(result: RemoteGlobal | Error) => {
+        if (result instanceof Error) {
+          // do something with error result
+        }
 
-  render() {
-    ...
-
-    return (
-      <ImportRemote
-        preconnect
-        source="https://some-external-service.com/global.js"
-        getImport={(window: WindowWithGlobal) => window.remoteGlobal}
-        onError={(error: Error) => this.setState({error})}
-        onImported={(remoteGlobal: RemoteGlobal) => doSomethingWithGlobal(remoteGlobal)}
-        defer={DeferTiming.Mount}
-      />
-    );
-  }
+        // do something with successful result
+      }}
+      defer={DeferTiming.Mount}
+    />
+  );
 }
 ```
 
@@ -51,9 +52,8 @@ class MyComponent extends React.Component {
 interface Props<Imported = any> {
   source: string;
   preconnect?: boolean;
-  onError(error: Error): void;
   getImport(window: Window): Imported;
-  onImported(imported: Imported): void;
+  onImported(imported: Imported | Error): void;
   defer?: DeferTiming;
 }
 ```
@@ -66,17 +66,13 @@ Source of the script to load the global from
 
 Generates a preconnect link tag for the source’s domain using [`@shopify/react-preconnect`](https://github.com/Shopify/quilt/tree/master/packages/react-preconnect)
 
-**onError**
-
-Callback that takes in `error` is called if an error occurs
-
 **getImport**
 
 Callback that takes in `window` with the added global and returns the global added to the `window` by the new script
 
 **onImported**
 
-Callback that gets called with the imported global
+Callback that gets called with the imported global or an `error` if one occurs
 
 **defer**
 
@@ -85,3 +81,62 @@ A member of the `DeferTiming` enum (from `@shopify/async`) allowing the import r
 - Component mount (`DeferTiming.Mount`; this is the default)
 - Browser idle (`DeferTiming.Idle`; if `window.requestIdleCallback` is not available, it will load on mount), or
 - Component is in the viewport (`DeferTiming.InViewport`; if `IntersectionObserver` is not available, it will load on mount)
+
+Note, changing any of these values while rendering will cancel the import.
+
+### useImportRemote()
+
+The above can also be accomplished using the hook, `useImportRemote()`, that is available in this package.
+
+```tsx
+import * as React from 'react';
+import {useImportRemote, Status} from '@shopify/react-import-remote';
+import {DeferTiming} from '@shopify/async';
+
+function MyComponent() {
+  const {result} = useImportRemote(
+    'https://some-external-service.com/global.js',
+  );
+
+  if (result.status === Status.Failed) {
+    // do something with error result
+  }
+
+  if (result.status === Status.Complete) {
+    // do something with successful result
+  }
+
+  return null;
+}
+```
+
+## Interface
+
+```tsx
+useImportRemote<Imported = unknown>(
+  source: string, // source of the script to load the global from
+  options: Options<Imported>, // see Options interface below
+): {
+  result: Result<Imported>; // see Result interface below
+  intersectionRef: React.Ref<HTMLElement | null>; // with defer: DeferTiming.InViewport, a ref to identify element to observe
+}
+
+interface Options<Imported> {
+  nonce?: string; // a unique cryptographic nonce to add to the generated script tag
+  defer?: DeferTiming; // refer to the defer prop above
+  getImport(window: Window): Imported; // refer to the getImport prop above
+}
+
+export enum Status {
+  Initial = 'Initial',
+  Failed = 'Failed',
+  Complete = 'Complete',
+  Loading = 'Loading',
+}
+
+type Result<Imported = any> =
+  | {status: Status.Initial}
+  | {status: Status.Loading}
+  | {status: Status.Failed; error: Error}
+  | {status: Status.Complete; imported: Imported};
+```
--- a/packages/react-import-remote/README.md
+++ b/packages/react-import-remote/README.md
@@ -13,7 +13,31 @@ $ yarn add @shopify/react-import-remote
 
 ## Usage
 
-The provided utilities are intended only for external scripts that load globals. Other JavaScript should use the native `import()` operator for asynchronously loading code. These utilities cache results by source, so only a single `script` tag is ever added for a particular source.
+The package provides a hook and component that are intended for loading external scripts. These utilities cache results by source, so only a single `script` tag is ever added for a particular source.
+
+### useImportRemote()
+
+```tsx
+import * as React from 'react';
+import {useImportRemote, Status} from '@shopify/react-import-remote';
+import {DeferTiming} from '@shopify/async';
+
+function MyComponent() {
+  const {result} = useImportRemote(
+    'https://some-external-service.com/global.js',
+  );
+
+  if (result.status === Status.Failed) {
+    // do something with error result
+  }
+
+  if (result.status === Status.Complete) {
+    // do something with successful result
+  }
+
+  return null;
+}
+```
 
 ### <ImportRemote />
 
@@ -46,18 +70,6 @@ function MyComponent() {
 }
 ```
 
-## Interface
-
-```ts
-interface Props<Imported = any> {
-  source: string;
-  preconnect?: boolean;
-  getImport(window: Window): Imported;
-  onImported(imported: Imported | Error): void;
-  defer?: DeferTiming;
-}
-```
-
 **source**
 
 Source of the script to load the global from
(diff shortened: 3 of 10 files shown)


packages/react-import-remote/src/ImportRemote.tsx
```diff
@@ -1,73 +1,38 @@
...
-    const {source, preconnect} = this.props;
-
-    if (preconnect) {
-      const url = new URL(source);
-      return <Preconnect source={url.origin} />;
-    }
-
-    return null;
+  if (global && onImported) {
+    onImported(global);
```
TheMallen: We shouldn't be calling these callbacks in render itself, I think these want to be more like:
```
useEffect(() => {
    if (global) {
      onImported(global)
    }
}, [global])
```
```
useEffect(() => {
   if (error) {
      onError(error)
    }
}, [error])
```

packages/react-import-remote/src/test/ImportRemote.test.tsx
```diff
@@ -163,16 +166,12 @@ describe('<ImportRemote />', () => {
...
+      expect(importRemote.find(IntersectionObserver)).toHaveReactProps({
+        threshold: 0,
+      });
 
-      await trigger(
-        importRemote.find(IntersectionObserver),
-        'onIntersectionChange',
-        {isIntersecting: true},
-      );
+      // perform something that gets us into view
```
cartogram: Not sure how to do the "perform" part of this test now.

packages/react-import-remote/src/hooks.ts
```diff
@@ -0,0 +1,100 @@
...
+type State = 'loading' | 'error' | 'loaded' | null;
+
+export function useImportRemote<Imported = any>(
+  source: string,
+  options: Options<Imported>,
+): {
+  state: State;
+  imported: Imported | null;
+  intersectionRef: any;
+} {
```
lemonmade: I am not a big fan of this typing, it technically makes `{state: 'error', imported: Imported}` possible. I'd rather see `{error: Error} | {imported: Imported} | {loading: true}` (you'll probably want to flesh those out so that `error` is `undefined` in all types when no error, `loading` is `false` in all types when not loading, and `imported` is `null` in all types when not loaded)

packages/react-import-remote/src/hooks.ts
```diff
@@ -0,0 +1,100 @@
...
+    defer = DeferTiming.Mount,
+    nonce = '',
+    getImport,
+    onError = () => {},
+    onImported = () => {},
+  } = options;
+
+  const loadRemote = React.useCallback(
+    () => {
+      return new Promise(async resolve => {
```
lemonmade: I usually do `(async () => {})()` since you just resolve at the end

packages/react-import-remote/src/hooks.ts
```diff
@@ -0,0 +1,100 @@
...
+        try {
+          setState('loading');
+          const importResult = await load(source, getImport, nonce);
+          setImported(importResult);
+          onImported(importResult);
```
lemonmade: I don't think this callback really makes sense in a hook, personally (it kind of makes sense in the component version, but you can implement it using `useOnValueChange` there instead

packages/react-import-remote/src/hooks.ts
```diff
@@ -0,0 +1,100 @@
...
+        } catch (error) {
+          setState('error');
+          onError(error);
```
lemonmade: Same as above

packages/react-import-remote/src/hooks.ts
```diff
@@ -0,0 +1,100 @@
...
+
+export function useImportRemote<Imported = any>(
+  source: string,
+  options: Options<Imported>,
+): {
+  state: State;
+  imported: Imported | null;
+  intersectionRef: any;
+} {
+  const [intersection, intersectionRef] = useIntersection();
```
lemonmade: I forget if this one is expensive, but it might be one where it's better to only do it conditionally, but guard against them flipping between `defer` options (to prevent you from being able to actually violate the rule of hooks)

packages/react-import-remote/src/hooks.ts
```diff
@@ -0,0 +1,97 @@
...
+interface Options<Imported> {
+  nonce?: string;
+  defer?: DeferTiming;
+  onError?(error: Error): void;
+  getImport(window: Window): Imported;
+  onImported?(imported: Imported): void;
+}
+
+export enum Status {
+  Initial,
```
TheMallen: Can we use string enums here? They're just a bit nicer to debug.

packages/react-import-remote/src/test/ImportRemote.test.tsx
TheMallen: Not sure you can other than just mocking out the inner hook, or mocking whatever it uses to decide to fire an event.

cartogram:
@mallen/ @lemonmade this is ready for another look.

packages/jest-dom-mocks/src/intersection-observer.ts
```diff
@@ -44,6 +48,9 @@ export default class IntersectionObserverMock {
     const setObservers = (setter: (observers: Observer[]) => Observer[]) =>
       (this.observers = setter(this.observers));
 
+    (global as any).IntersectionObserverEntry = () => {};
+    (global as any).IntersectionObserverEntry.prototype.intersectionRatio = () => {};
```
lemonmade: Shouldn't this return a number?

packages/jest-dom-mocks/src/intersection-observer.ts
```diff
@@ -87,6 +94,9 @@ export default class IntersectionObserverMock {
     }
 
     (global as any).IntersectionObserver = this.originalIntersectionObserver;
+    (global as any).IntersectionObserverEntry = this.originalIntersectionObserverEntry;
+    (global as any).intersectionRatio = this.originalIntersectionRatio;
```
lemonmade: Do we ever mock this originally?

packages/react-import-remote/src/ImportRemote.tsx
```diff
@@ -21,91 +14,48 @@ export interface Props<Imported = any> {
...
+  });
+
+  const intersectionObserver =
+    defer === DeferTiming.InViewport && intersectionRef ? (
+      <div ref={intersectionRef} />
+    ) : null;
+
+  React.useEffect(
+    () => {
+      switch (result.status) {
```
lemonmade: Hmm I feels weird to me that result is a nested object. I would have expected something more like:

```ts
{status: Status.Complete, imported: T, intersectionRef: Ref} | {status: Status.Error, error: Error, intersectionRef: Ref}
```

packages/react-import-remote/src/ImportRemote.tsx
```diff
@@ -21,91 +14,48 @@ export interface Props<Imported = any> {
...
+          onError(result.error);
+          return;
+        case Status.Complete:
+          onImported(result.imported);
       }
-    } else if (defer === DeferTiming.Mount) {
-      await this.loadRemote();
-    }
+    },
+    [result, onError, onImported],
```
lemonmade: It is technically slightly more correct to split this across two effects, because this will cause `onImported` to run again even if only `onError` changed

lemonmade: Although I wonder if it's better to just take a breaking change and change the callback to be a single callback that takes `Error | T`?

packages/react-import-remote/src/hooks.ts
```diff
@@ -0,0 +1,123 @@
...
+  );
+
+  const deferOption = React.useRef(defer);
+
+  if (deferOption.current !== defer) {
+    throw new Error(
+      [
+        'You’ve changed the defer strategy on an <ImportRemote />',
+        'component after it mounted. This is an expensive operation',
+        'that must be avoided in order to dismiss this error.',
```
lemonmade: It's not just that it must be avoided, we literally don't allow it.

packages/react-import-remote/src/hooks.ts
```diff
@@ -0,0 +1,123 @@
...
+  Loading = 'Loading',
+}
+
+type Result<Imported = any> =
+  | {status: Status.Initial}
+  | {status: Status.Loading}
+  | {status: Status.Failed; error: Error}
+  | {status: Status.Complete; imported: Imported};
+
+export function useImportRemote<Imported = any>(
```
lemonmade: Might be better to do `unknown`

packages/react-import-remote/src/ImportRemote.tsx
```diff
@@ -21,91 +14,48 @@ export interface Props<Imported = any> {
...
+export default function ImportRemote({
+  source,
+  nonce,
+  preconnect,
+  onError,
+  getImport,
+  onImported,
+  defer,
+}: Props) {
+  const {result, intersectionRef} = useImportRemote(source, {
```
lemonmade: We should make `ImportRemote` generic, and pass that generic through to the props and here.

packages/react-import-remote/src/hooks.ts
```diff
@@ -0,0 +1,123 @@
...
+      [
+        'You’ve changed the defer strategy on an <ImportRemote />',
+        'component after it mounted. This is an expensive operation',
+        'that must be avoided in order to dismiss this error.',
+      ].join(' '),
+    );
+  }
+
+  let intersection;
+  let intersectionRef;
```
lemonmade: Give these types

packages/react-import-remote/src/hooks.ts
```diff
@@ -0,0 +1,123 @@
...
+        if (
+          idleCallbackHandle.current != null &&
+          typeof (window as any).cancelIdleCallback === undefined
+        ) {
+          (window as any).cancelIdleCallback(idleCallbackHandle.current);
+          idleCallbackHandle.current = null;
+        }
+      };
+    },
+    [defer, loadRemote, intersection, nonce, getImport, source],
```
lemonmade: We should make a note in the README that you really can't let any of these values change part-way through rendering, because any change will cancel the load (unlike before)

packages/react-import-remote/src/test/ImportRemote.test.tsx
```diff
@@ -102,9 +105,13 @@ describe('<ImportRemote />', () => {
       const spy = jest.fn();
       load.mockImplementation(() => promise);
 
-      mount(<ImportRemote {...mockProps} onImported={spy} />);
+      const importRemote = await mount(
+        <ImportRemote {...mockProps} onImported={spy} />,
+      );
 
-      await promise;
+      importRemote.act(async () => {
```
lemonmade: Hmm async `act` should not work yet...

packages/react-import-remote/src/ImportRemote.tsx
cartogram: The way I have it groups things that change together.

packages/jest-dom-mocks/src/intersection-observer.ts
```diff
@@ -44,6 +46,9 @@ export default class IntersectionObserverMock {
     const setObservers = (setter: (observers: Observer[]) => Observer[]) =>
       (this.observers = setter(this.observers));
 
+    (global as any).IntersectionObserverEntry = () => {};
+    (global as any).IntersectionObserverEntry.prototype.intersectionRatio = 0;
```
lemonmade: This is usually a getter right? Might be better to match the property descriptor a little better.

packages/react-import-remote/CHANGELOG.md
```diff
@@ -8,6 +8,13 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 <!-- ## [Unreleased] - -->
 
+## 2.0.0 - 2019-05-03
+
+### Changed
+
+- Changes to the props of `<ImportRemote />` mid-render will cause the import to be canceled. ([608](https://github.com/Shopify/quilt/pull/608))
```
lemonmade: Also removed `onError`?

packages/react-import-remote/README.md
```diff
@@ -66,17 +66,13 @@ Source of the script to load the global from
...
-Callback that takes in `error` is called if an error occurs
-
 **getImport**
 
 Callback that takes in `window` with the added global and returns the global added to the `window` by the new script
 
 **onImported**
 
-Callback that gets called with the imported global
+Callback that gets called with the imported global or an `error` if one occurs
```
lemonmade: How do you feel about this change? I am still kind of waffling back and forth.

packages/react-import-remote/README.md
```diff
@@ -85,3 +81,62 @@ A member of the `DeferTiming` enum (from `@shopify/async`) allowing the import r
...
+}
+
+export enum Status {
+  Initial = 'Initial',
+  Failed = 'Failed',
+  Complete = 'Complete',
+  Loading = 'Loading',
+}
+
+type Result<Imported = any> =
```
lemonmade: Use `unknown` here

packages/react-import-remote/src/ImportRemote.tsx
```diff
@@ -1,111 +1,59 @@
...
+  const intersectionObserver =
+    defer === DeferTiming.InViewport && intersectionRef ? (
+      <div ref={intersectionRef as React.RefObject<HTMLDivElement>} />
+    ) : null;
+
+  React.useEffect(
+    () => {
+      switch (result.status) {
+        case Status.Failed:
+          onImported(new Error(result.error.message));
```
lemonmade: Why not just pass the error directly?

packages/react-import-remote/src/hooks.ts
```diff
@@ -0,0 +1,121 @@
...
+        'component after it has mounted. This is not supported.',
+      ].join(' '),
+    );
+  }
+
+  let intersection: IntersectionObserverEntry | null = null;
+  let intersectionRef: React.Ref<HTMLElement | null> = null;
+
+  // Normally this would be dangerous but because we are
+  // guranteed to have thrown if the defer option changes
```
lemonmade: `guaranteed`

packages/react-import-remote/src/hooks.ts
```diff
@@ -0,0 +1,121 @@
...
+    [intersection, intersectionRef] = useIntersection();
+  }
+  /* eslint-enable react-hooks/rules-of-hooks */
+
+  const loadRemote = React.useCallback(
+    async () => {
+      try {
+        setResult({status: Status.Loading});
+        const importResult = await load(source, getImport, nonce);
+        setResult({status: Status.Complete, imported: importResult});
```
lemonmade: Do you need to check if you are still mounted here?

packages/react-import-remote/src/hooks.ts
```diff
@@ -0,0 +1,121 @@
...
+          loadRemote();
+        }
+      } else if (defer === DeferTiming.Mount) {
+        loadRemote();
+      }
+
+      return () => {
+        if (
+          idleCallbackHandle.current != null &&
+          typeof (window as any).cancelIdleCallback === undefined
```
lemonmade: This is wrong, `typeof (window as any).cancelIdleCallback === 'function'`

packages/react-import-remote/src/test/ImportRemote.test.tsx
lemonmade: whatever this is doing it's not doing what you think so we should probably just skip it until async act is actually real.

packages/react-import-remote/src/test/ImportRemote.test.tsx
```diff
@@ -137,45 +124,85 @@ describe('<ImportRemote />', () => {
     it('does not render any preconnect link by default', () => {
       const importRemote = mount(<ImportRemote {...mockProps} />);
 
-      expect(importRemote.find(Preconnect)).toHaveLength(0);
+      expect(importRemote.findAll(Preconnect)).toHaveLength(0);
```
lemonmade: `.not.toContainReactComponent`

packages/react-import-remote/src/test/ImportRemote.test.tsx
```diff
@@ -137,45 +124,85 @@ describe('<ImportRemote />', () => {
...
     });
 
     it('creates a preconnect link with the source’s origin when preconnecting is requested', () => {
       const importRemote = mount(<ImportRemote {...mockProps} preconnect />);
 
-      expect(importRemote.find(Preconnect).prop('source')).toBe(
+      expect(importRemote.find(Preconnect)!.prop('source')).toBe(
```
lemonmade: `toContainReactComponent(Preconnect, {source: }`

packages/react-import-remote/src/test/ImportRemote.test.tsx
```diff
@@ -137,45 +124,85 @@ describe('<ImportRemote />', () => {
...
+            <ImportRemote {...mockProps} defer={defer} />
+          </>
+        );
+      }
+
+      const mockComponent = mount(<MockComponent />);
+
+      expect(() =>
+        mockComponent.act(() => {
+          mockComponent.find('button')!.trigger('onClick');
```
lemonmade: You shouldn't need to wrap this in an act, trigger is already wrapped in an act

packages/react-import-remote/src/ImportRemote.tsx
cartogram: It's not an instance of `Error`, it is an `Event`.

packages/react-import-remote/README.md
cartogram: I think it's fine. I always felt like there were too many props on this component. 

As long as we have an example, which I added in the Readme.

cartogram:
@lemonmade thanks for all the feedback, I've addressed all of the last comments.

packages/react-import-remote/README.md
```diff
@@ -25,23 +27,22 @@ interface WindowWithGlobal extends Window {
...
-        defer={DeferTiming.Mount}
-      />
-    );
-  }
+function MyComponent() {
+  return (
+    <ImportRemote
+      preconnect
+      source="https://some-external-service.com/global.js"
+      getImport={}
```
lemonmade: Provide a real value here

packages/react-import-remote/README.md
```diff
@@ -85,3 +81,62 @@ A member of the `DeferTiming` enum (from `@shopify/async`) allowing the import r
 - Component mount (`DeferTiming.Mount`; this is the default)
 - Browser idle (`DeferTiming.Idle`; if `window.requestIdleCallback` is not available, it will load on mount), or
 - Component is in the viewport (`DeferTiming.InViewport`; if `IntersectionObserver` is not available, it will load on mount)
+
+Note, changing any of these values while rendering will cancel the import.
+
+### useImportRemote()
```
lemonmade: I'd probably flip the docs to do the hook first, up to you though.

packages/react-import-remote/README.md
```diff
@@ -85,3 +81,62 @@ A member of the `DeferTiming` enum (from `@shopify/async`) allowing the import r
...
+  Complete = 'Complete',
+  Loading = 'Loading',
+}
+
+type Result<Imported = unknown> =
+  | {status: Status.Initial}
+  | {status: Status.Loading}
+  | {status: Status.Failed; error: Error}
+  | {status: Status.Complete; imported: Imported};
+```
```
lemonmade: Personally, I am mostly anti-types in readmes, they are likely to get out of date.

packages/react-import-remote/src/ImportRemote.tsx
lemonmade: Why is it not a real error?

lemonmade: Your types seem to indicate it is a real error?

packages/react-import-remote/src/test/ImportRemote.test.tsx
lemonmade: Please delete it or make it not be async.

packages/react-import-remote/src/test/ImportRemote.test.tsx
```diff
@@ -81,53 +83,39 @@ describe('<ImportRemote />', () => {
...
 
-  describe('onError()', () => {
-    it('is not called when the script loads successfully', () => {
-      const spy = jest.fn();
-      mount(<ImportRemote {...mockProps} onError={spy} />);
-      expect(spy).not.toHaveBeenCalled();
-    });
-
-    it('is called with a script loading error', async () => {
+    // eslint-disable-next-line jest/no-disabled-tests
```
lemonmade: Can you comment on why it is disabled?

Issue #589, opened 2019-03-21
Add a package for any common "utility" hooks


lemonmade:
I know I have ended up writing a `useToggleState` hook several times, I'm sure there are others.

PR #609, opened 2019-03-27
Adds basic `react-hooks` package

cartogram:
Closes: https://github.com/Shopify/quilt/issues/589

Adding an initial react-hooks package for primitive/utility hooks and added a few to start: `useTimeout` and `useValueTracking`.

Edit decided to call it `useOnValueChange`


TheMallen:
No  `useInterval`? :P

packages/react-hooks/src/hooks/timeout.ts
```diff
@@ -0,0 +1,29 @@
+import * as React from 'react';
+
+export default function useTimeout(callback, delay: number) {
+  const callbackRef = React.useRef<() => void>();
+
+  React.useEffect(
```
lemonmade: I don’t understand why this ref/ separate effect is needed

lemonmade: Hmm I guess so you can change the callback and the original timeout will be respected? Imo that API aid not great, it’s better to just say they need to pass the same function, otherwise we cancel the timeout

packages/react-hooks/src/hooks/timeout.ts
```diff
@@ -0,0 +1,29 @@
...
+  );
+  React.useEffect(
+    // eslint-disable-next-line consistent-return
+    () => {
+      function tick() {
+        if (callbackRef.current) {
+          callbackRef.current();
+        }
+      }
+      if (delay !== null) {
```
lemonmade: Not a possible case

packages/react-hooks/src/hooks/timeout.ts
```diff
@@ -0,0 +1,29 @@
...
+        // eslint-disable-next-line prefer-const
+        let id = setTimeout(tick, delay);
+
+        return () => clearTimeout(id);
+      }
+    },
+    [delay],
+  );
+}
```
lemonmade: No tests?

cartogram:
@lemonmade changes addressed and added tests 💅

packages/react-hooks/src/hooks/tests/on-value-change.test.tsx
```diff
@@ -0,0 +1,31 @@
...
+  it('calls the onChange handler when the initial value and the changed value changes', () => {
+    const spy = jest.fn();
+    const initialValue = 'initial';
+    const value = 'changed';
+    const wrapper = mount(<MockComponent value={initialValue} spy={spy} />);
+
+    wrapper.setProps({value});
+
+    expect(spy).toHaveBeenCalledWith(value, initialValue);
+  });
```
lemonmade: Adding a test that it doesn't get called when the component re-renders but the value doesn't change would be good too 👍

packages/react-hooks/src/hooks/tests/timeout.test.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+  it('calls the handler after the delay', () => {
+    const spy = jest.fn();
+
+    mount(<MockComponent spy={spy} />);
+
+    timer.runAllTimers();
+
+    expect(spy).toHaveBeenCalled();
+  });
```
lemonmade: A test that changing the callback/ duration unsubscribes the first timeout would be good

PR #611, opened 2019-03-27
✨ working debounce utility

michenly:
No docs just yet.
`debounce` is much more complex `memoize` and I have this irie feeling that I must have miss some test cases...

### DIFF
--- a/packages/function-enhancers/src/debounce.ts
+++ b/packages/function-enhancers/src/debounce.ts
@@ -1,0 +1,62 @@
+export interface DebounceOptions {
+  leading?: boolean;
+  trailing?: boolean;
+}
+
+interface Timer {
+  ref(): void;
+  unref(): void;
+}
+
+export default function debounce<
+  FuncToDebounce extends (...args: any[]) => any
+>(
+  funcToDebounce: FuncToDebounce,
+  // the number to delay in milliseconds
+  wait = 0,
+  options?: DebounceOptions,
+): FuncToDebounce {
+  const {leading = false, trailing = true} = options || {};
+
+  let result: ReturnType<FuncToDebounce>;
+  let lastArgs: Parameters<FuncToDebounce> | undefined;
+  let lastThis: FuncToDebounce | undefined;
+  let timer: Timer | undefined;
+
+  function expireTimer() {
+    timer = undefined;
+  }
+
+  function expireAndExecute() {
+    result = funcToDebounce.apply(lastThis, lastArgs);
+    lastThis = undefined;
+    lastArgs = undefined;
+    expireTimer();
+
+    return result;
+  }
+
+  return function debounced(...args: Parameters<FuncToDebounce>) {
+    // eslint-disable-next-line consistent-this
+    lastThis = this;
+    lastArgs = args;
+
+    if (!timer) {
+      if (leading) {
+        expireAndExecute();
+      }
+
+      if (trailing) {
+        timer = setTimeout(expireAndExecute, wait);
+      }
+    }
+
+    if (leading) {
+      timer = setTimeout(expireTimer, wait);
+    } else if (trailing && leading) {
+      return expireAndExecute();
+    }
+
+    return result;
+  } as FuncToDebounce;
+}
--- a/packages/function-enhancers/src/test/debounce.test.ts
+++ b/packages/function-enhancers/src/test/debounce.test.ts
@@ -1,0 +1,183 @@
+/**
+ * @jest-environment jsdom
+ */
+
+import debounce from '../debounce';
+
+describe('debounce()', () => {
+  beforeEach(() => {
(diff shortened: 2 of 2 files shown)


packages/function-enhancers/src/debounce.ts
```diff
@@ -0,0 +1,61 @@
...
+  trailing?: boolean;
+}
+
+export default function debounce<F extends Function>(
+  method: F,
+  // the number to delay in milliseconds
+  wait: number = 0,
+  options?: DebounceOptions,
+): F {
+  const leading =
```
lemonmade: Doing this via destructuring will feel a lot nicer

packages/function-enhancers/src/debounce.ts
```diff
@@ -0,0 +1,61 @@
...
+    options && options.leading !== undefined ? options.leading : false;
+  const trailing =
+    options && options.trailing !== undefined ? options.trailing : true;
+
+  let result;
+  let lastArgs;
+  let lastThis;
+  let timer;
```
lemonmade: Types for these would be 💯

packages/function-enhancers/src/debounce.ts
```diff
@@ -0,0 +1,61 @@
...
+  function expireTimer() {
+    timer = undefined;
+  }
+
+  function expireAndExecute() {
+    expireTimer();
+    executeMethod();
+  }
+
+  return (function debounced(...args) {
```
lemonmade: Again, we should be able to at least type the arguments here

packages/function-enhancers/src/debounce.ts
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    if (leading) {
+      timer = setTimeout(expireTimer, wait);
+    } else if (trailing && leading) {
+      return executeMethod();
+    }
+
+    return result;
+  } as unknown) as F;
```
lemonmade: This isn't strictly true because it does not have the same return value as the original function

packages/function-enhancers/src/debounce.ts
```diff
@@ -0,0 +1,61 @@
...
+        executeMethod();
+      }
+
+      if (trailing) {
+        timer = setTimeout(expireAndExecute, wait);
+      }
+    }
+
+    if (leading) {
+      timer = setTimeout(expireTimer, wait);
```
lemonmade: What does this bit do?

packages/function-enhancers/src/debounce.ts
```diff
@@ -0,0 +1,61 @@
...
+    options && options.leading !== undefined ? options.leading : false;
+  const trailing =
+    options && options.trailing !== undefined ? options.trailing : true;
+
+  let result;
+  let lastArgs;
+  let lastThis;
+  let timer;
+
+  function executeMethod() {
```
lemonmade: Should this not always expire the timer too?

packages/function-enhancers/src/test/debounce.test.ts
```diff
@@ -0,0 +1,183 @@
...
+      const spy = jest.fn((value: string) => value);
+      const wait = 100;
+      const debounced = debounce(spy, wait, {leading: true, trailing: false});
+
+      expect(debounced('a')).toEqual('a');
+      expect(spy).toHaveBeenCalledTimes(1);
+      expect(spy).toHaveBeenLastCalledWith('a');
+    });
+
+    it('should only execute the first of the debounced function calls', () => {
```
lemonmade: Try not to prefix with `should` (all tests *should* do the right thing :D)

packages/function-enhancers/src/test/debounce.test.ts
```diff
@@ -0,0 +1,183 @@
...
+
+      jest.advanceTimersByTime(wait);
+
+      expect(spy).toHaveBeenCalledTimes(2);
+      expect(spy).toHaveBeenLastCalledWith('c');
+    });
+  });
+
+  describe('leading=true & trailing=false', () => {
+    it('execute the debounced function immediately', () => {
```
lemonmade: I don't blame you for this, but one common thing I see in your test names is that they don't quite read as a full sentence. In most cases (like this one), all you are missing is an `s` on the verb (`it executes`)

packages/function-enhancers/src/debounce.ts
michelleyschen: can you elaborate on this? I don't really see why

lemonmade: Because if the function returned anything, and the debounce causes the function not to run synchronously, there is no return value at all for the debounced version, but there is for the non-debounced version.

michenly:
Close and use https://github.com/Shopify/quilt/pull/829 instead

PR #613, opened 2019-03-27
use React.memo on TrackingPixel

cartogram:
This PR uses React.memo to 🔥 some lodash.


michelleyschen:
I thought `React.memo` is only shallow comparison while `lodash/isEqual` is a deep comparison.
Don't think we can replace it here.

PR #614, opened 2019-03-27
Add a promise mock

lemonmade:
This PR adds a mocked version of the global `Promise`. When mocked, it uses the `promise` module, which uses timers instead of all the node micro queue stuff. This means that, by mocking and then running timers, we can have promises synchronously resolve. This will be necessary in order to stuff all the promise-resolve-causing-React-update stuff we do (most notably, `resolveAllGraphQL`) to be able to run in an `act` block. This approach is detailed here: https://github.com/facebook/react/issues/14769#issuecomment-462528230 (it's what Facebook folks say they do).

Hopefully the need for this in React land will be temporary, but it's probably generally useful anyways.

### DIFF
--- a/packages/jest-dom-mocks/src/promise.ts
+++ b/packages/jest-dom-mocks/src/promise.ts
@@ -1,0 +1,47 @@
+import FakePromise from 'promise';
+
+export default class Promise {
+  private OriginalPromise = global.Promise;
+  private isUsingFakePromise = false;
+
+  mock() {
+    if (this.isUsingFakePromise) {
+      throw new Error(
+        'Promise is already mocked, but you tried to mock it again.',
+      );
+    }
+
+    jest.useFakeTimers();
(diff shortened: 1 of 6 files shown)


packages/jest-dom-mocks/src/promise.ts
```diff
@@ -0,0 +1,47 @@
+import FakePromise from 'promise';
+
+export default class Promise {
+  private OriginalPromise = global.Promise;
```
cartogram: No capital

threepointone:
We just released an alpha 16.9.0-alpha.0 with an async `act()` implementation, that removes the need for this the polyfill/transpilation/fake timers. Feel free to give it a spin and if it resolves your issues without all the other stuff. Longer docs and explainer coming next week, but I'll be happy to pitch in if it doesn't make sense.

PR #615, opened 2019-03-27
Adds `useShortcut` hook

cartogram:
This PR adds a `useShortcut()` hook to `react-shortcuts`. https://github.com/Shopify/quilt/issues/585

### Usage

```
useShortcut(
    ['F', 'O', 'O'],
    ({ordered, held}) => {
      console.table(ordered);
      console.table(held);
    },
    {
      held: [['Shift']],
     //... other options passed to shortcutManager
    },
  );
```

### DIFF
--- a/packages/react-shortcuts/src/Shortcut/hooks.ts
+++ b/packages/react-shortcuts/src/Shortcut/hooks.ts
@@ -1,0 +1,55 @@
+import * as React from 'react';
+import {ShortcutContext} from '../ShortcutProvider';
+import Key, {HeldKey} from '../keys';
+
+export interface Subscription {
+  unsubscribe(): void;
+}
+
+export interface Options {
+  held?: HeldKey;
+  node?: HTMLElement | null;
+  ignoreInput?: boolean;
+  allowDefault?: boolean;
+}
+
+export default function useShortcut(
+  ordered: Key[],
+  onMatch: (matched: {ordered: Key[]; held?: HeldKey}) => void,
+  options: Options = {},
+) {
+  const shortcutManager = React.useContext(ShortcutContext);
+  const subscription = React.useRef<Subscription | null>(null);
+  const {node, held, ignoreInput, allowDefault} = options;
+
+  React.useEffect(
+    () => {
+      if (node != null) {
+        return;
+      }
+
+      if (shortcutManager == null) {
+        return;
+      }
+
+      subscription.current = shortcutManager.subscribe({
+        onMatch,
+        ordered,
+        node,
+        held,
+        ignoreInput: ignoreInput || false,
+        allowDefault: allowDefault || false,
+      });
+
+      // eslint-disable-next-line consistent-return
+      return () => {
+        if (subscription.current == null) {
+          return;
+        }
+
+        subscription.current.unsubscribe();
+      };
+    },
+    [node, ordered, onMatch, held, ignoreInput, allowDefault],
+  );
+}
--- a/packages/react-shortcuts/src/ShortcutProvider/ShortcutProvider.tsx
+++ b/packages/react-shortcuts/src/ShortcutProvider/ShortcutProvider.tsx
@@ -1,16 +1,15 @@
 import * as React from 'react';
 import ShortcutManager from '../ShortcutManager';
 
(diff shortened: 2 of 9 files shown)


packages/react-shortcuts/src/Shortcut/hooks.ts
```diff
@@ -0,0 +1,55 @@
...
+      subscription.current = context.shortcutManager.subscribe({
+        onMatch,
+        ordered,
+        node,
+        held: options.held,
+        ignoreInput: options.ignoreInput || false,
+        allowDefault: options.allowDefault || false,
+      });
+
+      // eslint-disable-next-line consistent-return
```
cartogram: May want to turn this off as part of https://github.com/Shopify/eslint-plugin-shopify/issues/231

packages/react-shortcuts/src/ShortcutProvider/ShortcutProvider.tsx
```diff
@@ -9,8 +9,9 @@ export interface Props {
   children?: React.ReactNode;
 }
 
-export const {Provider, Consumer} = React.createContext<Context>({});
+export const ShortcutContext = React.createContext<Context>({});
 
+export const {Consumer, Provider} = ShortcutContext;
```
cartogram: Figured we should still export these individually even though they are included in the above line.

packages/react-shortcuts/src/Shortcut/hooks.ts
```diff
@@ -0,0 +1,55 @@
...
+      return () => {
+        if (subscription.current == null) {
+          return;
+        }
+
+        subscription.current.unsubscribe();
+      };
+    },
+    [ordered],
```
lemonmade: I think this needs to include everything you use (`onMatch`, `ordered`, `node`, `held`, etc)

packages/react-shortcuts/src/Shortcut/hooks.ts
```diff
@@ -0,0 +1,55 @@
...
+  const subscription = React.useRef<Subscription | null>(null);
+
+  React.useEffect(
+    () => {
+      const {node} = options;
+      if (node != null) {
+        return;
+      }
+
+      if (context.shortcutManager == null) {
```
lemonmade: So the context this gives is an object that has a `shortcutManager` property? Seems super weird...

packages/react-shortcuts/src/Shortcut/hooks.ts
```diff
@@ -0,0 +1,55 @@
...
+        if (subscription.current == null) {
+          return;
+        }
+
+        subscription.current.unsubscribe();
+      };
+    },
+    [ordered],
+  );
+}
```
lemonmade: Can we re-write the component in terms of the hook?

packages/react-shortcuts/src/ShortcutProvider/ShortcutProvider.tsx
lemonmade: We'll want to cut a new major version (if we rewrite the component with a hook), so I think we should stop exporting these separately (also, if we could make it so context isn't a big old object, that would be 💯)

packages/react-shortcuts/src/Shortcut/hooks.ts
```diff
@@ -0,0 +1,55 @@
...
+      // eslint-disable-next-line consistent-return
+      return () => {
+        if (subscription.current == null) {
+          return;
+        }
+
+        subscription.current.unsubscribe();
+      };
+    },
+    [ordered, onMatch, options],
```
lemonmade: I am not sure but I think this might be your problem in tests. If this object changes all the time, which it probably does, you may be losing your subscription too often. I would pass all options individually instead.

cartogram:
@TheMallen I am not sure how much value is in testing the hook. The `<Shortcut />` component is almost like a mocked component and all the tests would be the same. Have a look at these and let me know if you agree/disagree. https://github.com/Shopify/quilt/pull/615/files#diff-313aebedf4a8d8b08991e49638c60a96 Personally feel like they are just repeats.

PR #616, opened 2019-03-28
Hook route propagator

michelleyschen:
Part of #585
Update current lib to use react hook


packages/react-shopify-app-route-propagator/src/test/react-shopify-app-route-propagator.test.tsx
```diff
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import {mount} from 'enzyme';
+import {mount} from '@shopify/react-testing';
```
michelleyschen: can we do a beta release of `@shopify/react-testing` and iterate on it?

packages/react-shopify-app-route-propagator/src/test/react-shopify-app-route-propagator.test.tsx
```diff
@@ -57,19 +57,37 @@ describe('@shopify/react-shopify-app-route-propagator', () => {
...
-    expect(appBridgeHistoryMock.dispatch).toHaveBeenCalledWith(
+    expect(appBridgeHistoryMock.dispatch).toHaveBeenLastCalledWith(
       AppBridgeHistory.Action.REPLACE,
-      firstPath,
+      secondPath,
     );
-    expect(appBridgeHistoryMock.dispatch).toHaveBeenCalledWith(
+  });
+
+  it('does not dispatch a replace action when the location updates but the value stay the same', () => {
```
michelleyschen: was able to make an improvement easily using hook!

packages/react-shopify-app-route-propagator/package.json
```diff
@@ -24,7 +24,7 @@
   "homepage": "https://github.com/Shopify/quilt/blob/master/packages/react-shopify-app-route-propagator/README.md",
   "peerDependencies": {
     "@shopify/app-bridge": "^0.7.3",
```
michelleyschen: since this will be a breaking change, should I take this opportunity to update app-bridge to at least 1.0.0?

packages/react-shopify-app-route-propagator/CHANGELOG.md
```diff
@@ -4,9 +4,11 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
-<!-- Unreleased changes should go to UNRELEASED.md -->
+## [Unreleased]
 
----
+### Enhancements
+
+- Update to use react hook ([#616](https://github.com/Shopify/quilt/pull/616))
```
cartogram: Is this needed if it doesn't change the Usage?

packages/react-shopify-app-route-propagator/src/index.ts
```diff
@@ -1,4 +1,4 @@
-import * as React from 'react';
+import {useState, useEffect} from 'react';
```
cartogram: I've not been destructuring the imports and just doing `React.useState`. Do we have a preference?

packages/react-shopify-app-route-propagator/src/index.ts
```diff
@@ -15,58 +15,38 @@ export interface Props {
 
 export const MODAL_IFRAME_NAME = 'app-modal-iframe';
 
-export default class ReactShopifyAppRoutePropagator extends React.PureComponent<
-  Props,
-  never
-> {
-  private historyInstance: History.History | undefined;
+export default function ReactShopifyAppRoutePropagator({app, location}: Props) {
+  const [history] = useState(History.create(app));
```
lemonmade: I think `useRef` is more appropriate here (never changes, old version was an instance property, which is roughly equivalent to a ref)

packages/react-shopify-app-route-propagator/src/index.ts
```diff
@@ -15,58 +15,38 @@ export interface Props {
 
 export const MODAL_IFRAME_NAME = 'app-modal-iframe';
 
-export default class ReactShopifyAppRoutePropagator extends React.PureComponent<
-  Props,
-  never
-> {
-  private historyInstance: History.History | undefined;
+export default function ReactShopifyAppRoutePropagator({app, location}: Props) {
```
lemonmade: Can this just be a hook instead? You can keep a component (that just calls the hook and returns `null`), but a custom hook here would probably be nice for folks

lemonmade: API-wise, it is weird to me that you need to manually pass in `location`, I am surprised it does not just come in through a hook

packages/react-shopify-app-route-propagator/src/index.ts
lemonmade: Do we not need to ever respond to a change in `app`? I assume not since the old one doesn't.

packages/react-shopify-app-route-propagator/src/test/react-shopify-app-route-propagator.test.tsx
lemonmade: I'll get a release out today 👍

packages/react-shopify-app-route-propagator/src/index.ts
michelleyschen: the old one doesn't, should we? I don't think its typical for app to change after initialization.

packages/react-shopify-app-route-propagator/src/index.ts
TheMallen: you can also do `React, {useState, useEffect}`

packages/react-shopify-app-route-propagator/src/index.ts
TheMallen: Now might be the right time to change this since it is already a breaking release.

michelleyschen: do we want to have dependency on react router? I am scare because the v4 -> v5 fiasco

packages/react-shopify-app-route-propagator/README.md
```diff
@@ -65,74 +30,135 @@ export interface Props {
...
 
-Allowing you to use it with a string
+### `useRoutePropagator`
 
-```javascript
-<ShopifyRoutePropagator location="/foo/bar?thing=true" app={app} />
-```
-
-Or a location object
+This example uses [app bridge](https://help.shopify.com/en/api/embedded-apps/app-bridge#set-up-your-app) to create an app instance and `withRouter` HoC form `react-router-router`.
```
TheMallen: > instance and `withRouter` HoC form `react-router-router`

should be

> instance and the `withRouter` enhancer from `react-router`

(enhancer or higher-order-component are preffered over acronyms)

packages/react-shopify-app-route-propagator/README.md
```diff
@@ -3,58 +3,23 @@
...
-      <ShopifyRoutePropagator location={location} app={app} />
-    ) : null;
-  }
-}
-```
-
 ## Usage
 
-The component has the following props signature
+In both `useRoutePropagator` and `<RoutePropagator />` you need two to pass in two variables with the signature:
```
cartogram: You’ve got an extra two here

TheMallen: I think it might be clearer to write this as something like

```
Both the hook and component versions of this library take the same two parameters:
```

packages/react-shopify-app-route-propagator/README.md
```diff
@@ -65,74 +30,135 @@ export interface Props {
...
-    </BrowserRouter>
-  );
-})
+    );
+  }
+}
 ```
 
-### With React Router (`<Route>`)
+### Additional ways of getting location object
```
TheMallen: > getting *a* location object

packages/react-shopify-app-route-propagator/README.md
cartogram: form should be from

packages/react-shopify-app-route-propagator/README.md
```diff
@@ -65,74 +30,135 @@ export interface Props {
...
+        </React.Fragment>
+      }}
+    </Route>
   );
 }
 ```
+
+#### With a string
+
+You can simply set location string value (but take care of the re-rending yourself)
```
TheMallen: typo on rerendering

packages/react-shopify-app-route-propagator/README.md
```diff
@@ -65,74 +30,135 @@ export interface Props {
...
+  );
+})
 ```
 
-### With React Router (withRouter)
+### `<RoutePropagator />`
 
-An easy way to mount the component with minimal configuration in a [react-router](https://github.com/ReactTraining/react-router) application is to first import the `withRouter` Higher-Order-Component and create a wrapped version of the propagator.
+This example assume that you have [polaris-react](https://github.com/Shopify/polaris-react) v3 or higher.
+Plus `withRouter` HoC form `react-router-router`.
```
cartogram: from

cartogram: react-router-dom

packages/react-shopify-app-route-propagator/README.md
```diff
@@ -3,58 +3,23 @@
...
 
-`<ShopifyRoutePropagator />` is a simple component to synchronize a Shopify embedded app's client side routing with the outer iframe host. It assumes the embedded app is either using [Shopify's App Bridge Library](https://help.shopify.com/en/api/embedded-apps/app-bridge) or [Polaris v3+ with Shopify App Bridge](https://polaris.shopify.com/components/structure/app-provider#section-initializing-the-shopify-app-bridge)
+This package contains `useRoutePropagator` hook and `<RoutePropagator />` component to synchronize a Shopify embedded app's client side routing with the outer iframe host. It assumes the embedded app is either using [Shopify's App Bridge Library](https://help.shopify.com/en/api/embedded-apps/app-bridge) or [Polaris v3+ with Shopify App Bridge](https://polaris.shopify.com/components/structure/app-provider#section-initializing-the-shopify-app-bridge)
```
TheMallen: > This package contains both a hook, `useRoutePropagator`, and component, `<RoutePropagator />`, API. Both of these allow you to...

packages/react-shopify-app-route-propagator/src/hook.ts
```diff
@@ -0,0 +1,57 @@
...
+import {LocationOrHref} from './types';
+
+import {
+  getSelfWindow,
+  getTopWindow,
+  getOrigin,
+  MODAL_IFRAME_NAME,
+} from './globals';
+
+export default function useRoutePropagator(
```
TheMallen: nit, might want to call this `useRoutePropagation`

packages/react-shopify-app-route-propagator/README.md
michelleyschen: I correct the typo but `withRouter` is actually from `react-router` thou?

packages/react-shopify-app-route-propagator/README.md
```diff
@@ -65,74 +30,135 @@ export interface Props {
...
+import React from 'react';
+import {Switch, Route, withRouter} from 'react-router'
+import {BrowserRouter} from 'react-router-dom'
+import createApp, {getShopOrigin} from '@shopify/app-bridge';
+import {useRoutePropagation} from '@shopify/react-shopify-app-route-propagator';
 
-Or even with `window.location` (though then you will need to make sure it rerenders yourself)
+export default withRouter(function(props) {
+  useRoutePropagation(
+    createApp({
```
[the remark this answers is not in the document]
michelleyschen: maybe, maybe not? but since both polaris & app bridge does not have context API or hook support, I really don't know where the app could be coming from.

I think the safest thing is to assume you create it here, and that the consumer will read the app bridge documentation and figure out they need to find a way to share this app instance throughout out the entire app.

packages/react-shopify-app-route-propagator/README.md
```diff
@@ -65,74 +30,135 @@ export interface Props {
...
+class Routes extends React.Component {
+  // This line is very important! It tells React to attach the `polaris`
+  // object to `this.context` within your component.
+  static contextTypes = {
+    polaris: PropTypes.object,
+  };
+
+  render() {
+    ...
+    const app = this.context.polaris.appBridge;
```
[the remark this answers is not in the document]
michelleyschen: yes, its really sad... I contemplated about offering a wrapper to get this and export a Polaris Context so you can use the Provider/Consumer pattern.

but that's not part of this library for sure.

packages/react-shopify-app-route-propagator/README.md
```diff
@@ -65,74 +30,135 @@ export interface Props {
...
 If you prefer things more explicit you can just get the `location` value to pass in explicitly by using `<Route>`'s children as a render prop.
 
 ```javascript
 import { Switch, Route, withRouter } from 'react-router'
 import { BrowserRouter } from 'react-router-dom'
-import ShopifyRoutePropagator from '@shopify/react-shopify-app-route-propagator';
+import {RoutePropagator} from '@shopify/react-shopify-app-route-propagator';
 
 export default function() {
```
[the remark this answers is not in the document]
michelleyschen: the location is coming from `Route` component's render prop, if I use a hook here, I will have to use the `withRouter` enhancer then it just a repeat of the first example.

packages/react-shopify-app-route-propagator/README.md
```diff
@@ -65,74 +30,135 @@ export interface Props {
...
-})
+    );
+  }
+}
 ```
 
-### With React Router (`<Route>`)
+### Additional ways of getting a location object
+
+#### With React Router (`<Route>`)
```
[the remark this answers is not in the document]
michelleyschen: it differ by how you get the location object.

packages/react-shopify-app-route-propagator/src/globals.ts
```diff
@@ -9,3 +9,5 @@ export function getTopWindow() {
 export function getOrigin() {
   return location.origin;
 }
+
+export const MODAL_IFRAME_NAME = 'app-modal-iframe';
```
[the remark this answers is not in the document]
michelleyschen: I thought it would be good have a variable for this since it can change it core ever change it.

packages/react-shopify-app-route-propagator/README.md
```diff
@@ -65,74 +30,135 @@ export interface Props {
...
+const location = '/foo/bar?thing=true';
+```
+
+#### With `window.location`
+
+Also need to take care of the re-rendering yourself
+
+```javascript
+const location = window.location;
+```
```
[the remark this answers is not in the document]
michelleyschen: that go back to discussion that didn't finished earlier.
should we list `react-router` as a peer-dependency for this library or not?

packages/react-shopify-app-route-propagator/README.md
```diff
@@ -65,74 +30,135 @@ export interface Props {
...
+      apiKey: 'API key from Shopify Partner Dashboard',
+      shopOrigin: getShopOrigin(),
+    }),
+    props.location
+  );
 
-```javascript
-<ShopifyRoutePropagator location={window.location} app={app} />
+  return (
+    <React.Fragment>
```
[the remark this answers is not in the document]
lemonmade: Bump

packages/react-shopify-app-route-propagator/README.md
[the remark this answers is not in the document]
lemonmade: Still feels weird to me, I would probably make a note of it (I think they could certainly write their own decorator to get the app bridge app out of context).

packages/react-shopify-app-route-propagator/README.md
```diff
@@ -65,74 +30,135 @@ export interface Props {
...
 ```
 
-Or a location object
+### `<RoutePropagator />`
 
-```javascript
-<ShopifyRoutePropagator location={this.props.location} app={app} />
-```
+This example assume that you have [polaris-react](https://github.com/Shopify/polaris-react) v3 or higher.
+Plus the `withRouter` enhancer from `react-router-router`.
```
lemonmade: Not a full sentence

packages/react-shopify-app-route-propagator/README.md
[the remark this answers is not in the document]
lemonmade: I don't think it's different enough to bother with the example.

packages/react-shopify-app-route-propagator/README.md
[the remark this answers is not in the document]
lemonmade: I think no, it's not a peer dependency, but don't bother showing them examples that would not be relevant in a single-page app.

packages/react-shopify-app-route-propagator/src/hook.ts
```diff
@@ -0,0 +1,57 @@
...
+  getTopWindow,
+  getOrigin,
+  MODAL_IFRAME_NAME,
+} from './globals';
+
+export default function useRoutePropagation(
+  app: ClientApplication<any>,
+  location: LocationOrHref,
+) {
+  const history = React.useRef(History.create(app));
```
lemonmade: It's not technically safe to do this because `app` changing break this, but maybe that's OK?

packages/react-shopify-app-route-propagator/src/test/hook.test.tsx
```diff
@@ -1,12 +1,18 @@
 import * as React from 'react';
-import {mount} from 'enzyme';
 import {ClientApplication} from '@shopify/app-bridge';
 import {History as AppBridgeHistory} from '@shopify/app-bridge/actions';
+import {mount} from '@shopify/react-testing';
+import {MODAL_IFRAME_NAME} from '../globals';
+import useRoutePropagation from '../hook';
 
-import RoutePropagator, {MODAL_IFRAME_NAME} from '..';
+function MockRoutePropagator({app, location}) {
```
lemonmade: Hmm I thought we were still testing the "component version" of the hook when there already was one, rather than creating a mock component for tests?

packages/react-shopify-app-route-propagator/README.md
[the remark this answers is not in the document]
michelleyschen: deleting the examples :)

packages/react-shopify-app-route-propagator/src/test/hook.test.tsx
michelleyschen: I remember the complete opposite for some reason 😖

PR #618, opened 2019-03-28
🐛 Make the webpack plugins return regular objects intead of Modules

amrocha:
Currently the webpack plugin in `react-i18n` gives consumers a `Module` object to work with. This isn't what's expected, and this PR changes it to return a simple Object instead.

Dynamic `import` statements return object-like data primitives called `Modules`. These modules behave roughly like an object in terms of access, but don't have any of the standard `Object` functions since they're not created with an `Object` as a prototype.

I believe the expectation here is that react-i18n return a standard object for people to manipulate, not a Module, hence this change. For example, this caused a regression in `web` when `@shopify/polaris` was updated to use a custom simplified `merge` function, instead of `lodash`'s merge function. `lodash` could deal with `Modules`, but the new simplified function expects simple `Objects`.

### DIFF
--- a/packages/react-i18n/src/test/babel-plugin.test.ts
+++ b/packages/react-i18n/src/test/babel-plugin.test.ts
@@ -37,7 +37,8 @@ describe('babel-pluin-react-i18n', () => {
           id: 'MyComponent_${defaultHash}',
           fallback: _en,
           async translations(locale) {
-            return await import(/* webpackChunkName: "MyComponent_${defaultHash}-i18n", webpackMode: "lazy-once" */ \`./translations/$\{locale}.json\`);
+            const dictionary = await import(/* webpackChunkName: "MyComponent_${defaultHash}-i18n", webpackMode: "lazy-once" */ \`./translations/$\{locale}.json\`);
+            return {...dictionary};
           },
         })(MyComponent);
         `,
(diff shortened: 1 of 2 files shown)


amrocha:
I considered doing a one liner like:
```js
    return {
      ...(await import(/* webpackChunkName: "Locales-i18n-[request]" */ `locales/${locale}.json`)),
    };
```

But the two-step approach seemed clearer to me. Opinions?

amrocha:
Oh you're right, I didn't realize we could do that as well. Updated the PR :D

packages/react-i18n/src/test/babel-plugin.test.ts
```diff
@@ -37,7 +37,8 @@ describe('babel-pluin-react-i18n', () => {
           id: 'MyComponent_${defaultHash}',
           fallback: _en,
           async translations(locale) {
-            return await import(/* webpackChunkName: "MyComponent_${defaultHash}-i18n", webpackMode: "lazy-once" */ \`./translations/$\{locale}.json\`);
+            const dictionary = await import(/* webpackChunkName: "MyComponent_${defaultHash}-i18n", webpackMode: "lazy-once" */ \`./translations/$\{locale}.json\`);
+            return dictionary.default;
```
lemonmade: I think it may be safer to do `dictionary && dictionary.default` since there is not a 100% guarantee a module will be resolved totally correctly (and undefined is a valid value to return)

amrocha: That makes sense, especially since this is added to the code post-compile. Done!

PR #619, opened 2019-03-29
Adds `react-toast` Package

cartogram:
wip


packages/react-toast/src/hooks.tsx
```diff
@@ -0,0 +1,57 @@
+import * as React from 'react';
+import {Toast} from '@shopify/polaris';
```
TheMallen: I'm not sure I like this depending on polaris. It would be cool if we just had the provider take a `renderToast` function.

packages/react-toast/src/hooks.tsx
```diff
@@ -0,0 +1,57 @@
...
+export interface ToastActions {
+  show(options: ToastDescription): void;
+  hide(): void;
+}
+
+export interface Props {
+  children: any;
+}
+
+export function ToastProvider({children}: Props) {
```
TheMallen: Nit: Can we split the context stuff into `/context`? Seems weird to me to have a file called `hooks` containing jsx.

packages/react-toast/src/hooks.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const [toasts, setToasts] = React.useState<ToastDescription[]>([]);
+
+  const show = (toast: ToastDescription) => {
```
TheMallen: We had talked about supporting the declarative API as well, should the context maybe be like 

```
{
  show(toast: ToastDescription): void;
  hide(): void;
  Toast: ReactComponent<ToastDescription, never>;
}
```

so you can do

```
const {Toast} = useToast()

return (
  /* rest of render*/
 {error && <Toast /*some props*/ /> 
);

```

PR #622, opened 2019-03-30
Add hook-ey `react-intersection-observer` API

lemonmade:
This PR adds a new `useIntersection` hook to `react-intersection-observer`, and slightly reworks the API of the component version to feel more closely related. The README has all of the goods, but basically, the hook version just gives you the state + a ref to throw on a DOM node, so it's quite simple to use the hook version:

```tsx
function MyComponent() {
  const [intersection, intersectionRef] = useIntersection();

  return (
    <div ref={intersectionRef}>Intersection: {intersection.isIntersecting}</div>
  );
}
```

### DIFF
--- a/packages/react-intersection-observer/README.md
+++ b/packages/react-intersection-observer/README.md
@@ -13,13 +13,37 @@ $ yarn add @shopify/react-intersection-observer
 
 ## Usage
 
-This package exports an `IntersectionObserver` component. This component will create its own DOM node that will be observed, with optional `onIntersecting` and `onNotIntersecting` props being called as appropriate. `onIntersecting` and `onNotIntersecting` will be called with an `IntersectionObserverEntry` object, which describes the state of the intersection. The component also accepts a few additional props that correspond to the options used to construct an `IntersectionObserver`:
+### `useIntersection(options: Options)`
 
-- `threshold`: a number or array of number indicating the `intersectionRatio` that must be met before the observer is triggered
+The `useIntersection` hook takes in `IntersectionObserver` options, and returns a tuple of:
+
+- The most recent `IntersectionObserverEntry` value, which details the most recent state of the observer.
+- An object that you can pass as a `ref` to the DOM element you wish to track.
+
+```tsx
+function MyComponent() {
+  const [intersection, intersectionRef] = useIntersection();
+
+  return (
+    <div ref={intersectionRef}>Intersection: {intersection.isIntersecting}</div>
+  );
+}
+```
+
+You can pass additional options to this hook that are used to create the underlying `IntersectionObserver`:
+
+- `threshold`: a number or array of number indicating the `intersectionRatio` that must be met before the observer is triggered.
 - `root`: a string or element that is used as the viewport for visibility testing. If a string is passed, it will be treated as a selector.
 - `rootMargin`: a string representing the margins by which to shrink the root’s bounding box before computing intersections.
 
-This component also allows you to customize the rendered markup. You can pass `children`, which will be rendered inside of the node being observed for intersections. You can also pass a `wrapperComponent` prop that changes the DOM node being wrapped around those children (defaults to a `div`). Regardless of the passed `wrapperComponent`, the `IntersectionObserver` component will always add a `display: contents` style to that node in order to reduce the styling impact of the additional nesting.
+### `<IntersectionObserver />`
+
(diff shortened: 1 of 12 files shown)


packages/react-intersection-observer/src/hooks.ts
```diff
@@ -0,0 +1,147 @@
...
+      }
+
+      lastObserver.current.unobserve(lastNode.current);
+    };
+  });
+
+  return [intersectionEntry, node];
+}
+
+export function useValueTracking<T>(
```
cartogram: this makes sense as a shared hook https://github.com/Shopify/quilt/pull/609/files#diff-99a7b9f41d8c46ef5eb67c2b0da03b2aR3

TheMallen: agreed

packages/react-intersection-observer/README.md
```diff
@@ -13,13 +13,37 @@ $ yarn add @shopify/react-intersection-observer
 
 ## Usage
 
-This package exports an `IntersectionObserver` component. This component will create its own DOM node that will be observed, with optional `onIntersecting` and `onNotIntersecting` props being called as appropriate. `onIntersecting` and `onNotIntersecting` will be called with an `IntersectionObserverEntry` object, which describes the state of the intersection. The component also accepts a few additional props that correspond to the options used to construct an `IntersectionObserver`:
+### `useIntersection(options: Options)`
 
-- `threshold`: a number or array of number indicating the `intersectionRatio` that must be met before the observer is triggered
+The `useIntersection` hook takes in `IntersectionObserver` options, and returns a tuple of:
+
+- The most recent `IntersectionObserverEntry` value, which details the most recent state of the observer.
```
TheMallen: This feels like a convoluted way to describe what it returns, maybe just `The state of the observer`?

packages/react-intersection-observer/README.md
```diff
@@ -35,15 +59,15 @@ This component also allows you to customize the rendered markup. You can pass `c
 
 ### Lifecycle
 
-You may change any prop on an `IntersectionObserver` component, and the component will do the minimum amount of work to unobserve/ re-observe with the new fields. The most expensive updates to make are changing `threshold`, `root`, `rootMargin`, and `wrapperComponent`, as these require disconnecting the old observer and recreating a new one.
+You may change any option on a `useIntersection` hook, or any prop on an `IntersectionObserver` component, and the minimum amount of work will be performed to unobserve/ re-observe with the new fields. The most expensive updates to make are changing `threshold`, `root`, `rootMargin`, and, in the case of the component version, `wrapperComponent`, as these require disconnecting the old observer and recreating a new one.
```
TheMallen: IMO `You may change` feels weird, I think it would flow better to describe this in terms of `When useIntersection is passed new values or IntersectionObserver receives new props`

packages/react-intersection-observer/src/hooks.ts
```diff
@@ -0,0 +1,147 @@
...
+}: Options = {}): [IntersectionObserverEntry, Ref<HTMLElement | null>] {
+  const node = useRef<HTMLElement | null>(null);
+  const lastNode = useRef<HTMLElement | null>(null);
+  const observer = useRef<IntersectionObserver | null>(null);
+  const lastObserver = useRef<IntersectionObserver | null>(null);
+
+  const [intersectionEntry, setIntersectingEntry] = useState<
+    IntersectionObserverEntry
+  >(() => ({
+    boundingClientRect: emptyBoundingClientRect,
```
TheMallen: should each of these maybe be clones instead of a reference to the same object?

packages/react-intersection-observer/src/test/IntersectionObserver.test.tsx
```diff
@@ -27,42 +28,52 @@ describe('<IntersectionObserver />', () => {
...
-        node,
+        child.domNode,
       );
     });
   });
 
   describe('children', () => {
     it('includes the children inside the top-level element', () => {
-      const children = <div>Hello world</div>;
+      const children = <div id="HelloWorld" />;
```
TheMallen: I don't really get why we declare this outside of the jsx, it makes me expect that we will reference it down below but we don't. Maybe just have `HelloWorld` in a string we use in both places, but have the children inline?

packages/react-intersection-observer/src/hooks.ts
lemonmade: Yes, agreed, we just don't have a place for them yet, and I didn't want to throw that in here too.

packages/react-intersection-observer/src/hooks.ts
lemonmade: I don't think it matters, all the fields are meant to be readonly

packages/react-intersection-observer/src/test/IntersectionObserver.test.tsx
lemonmade: 👍

PR #625, opened 2019-04-01
✨ Add `dimension` mock to `jest-dom-mocks

donnielrt:
Brings over dimension mocking utilities [from Web](https://github.com/Shopify/web/commit/957427ac4781f5d7d6dbfe7b6a2dd083ac126fd3#diff-2ef125a18ab2b913aba925a64386d260R1) per @TheMallen's [recommendation](https://github.com/Shopify/web/pull/8822#discussion_r238352729).

This allows mocking `scrollWidth`, `scrollHeight`, `offsetWidth`, and `offsetHeight` in tests.

Once released, I'll remove the corresponding mocks in Web and use these instead.

### DIFF
--- a/packages/jest-dom-mocks/src/dimension.ts
+++ b/packages/jest-dom-mocks/src/dimension.ts
@@ -1,0 +1,102 @@
+import {memoize} from '@shopify/javascript-utilities/decorators';
+
+const enum SupportedDimension {
+  OffsetWidth = 'offsetWidth',
+  OffsetHeight = 'offsetHeight',
+  ScrollWidth = 'scrollWidth',
+  ScrollHeight = 'scrollHeight',
+}
+
+type MockedGetter = (element: HTMLElement) => number;
+type Mock = MockedGetter | number;
+type Mocks = Partial<Record<string, Mock>>;
+
+type AugmentedElement = Element & {[key: string]: Mock};
+
+interface NativeImplentationMap {
+  [key: string]: Element;
+}
+
+function isGetterFunction(mock?: Mock): mock is MockedGetter {
+  return mock != null && typeof mock === 'function';
+}
+
+export default class Dimension {
+  private isUsingMock = false;
+  private overwrittenImplementations: string[] = [];
+
+  mock(mocks: Mocks) {
+    if (this.isUsingMock) {
+      throw new Error(
+        'Dimensions are already mocked, but you tried to mock them again.',
+      );
+    } else if (Object.keys(mocks).length === 0) {
+      throw new Error('No dimensions provided for mocking');
+    }
+
+    this.mockDOMMethods(mocks);
+    this.isUsingMock = true;
+  }
+
+  restore() {
+    if (!this.isUsingMock) {
+      throw new Error(
+        "Dimensions haven't been mocked, but you are trying to restore them.",
+      );
+    }
+
+    this.restoreDOMMethods();
+    this.isUsingMock = false;
+  }
+
+  isMocked() {
+    return this.isUsingMock;
+  }
+
+  @memoize()
+  private get nativeImplementations(): NativeImplentationMap {
+    return {
+      [SupportedDimension.OffsetWidth]: HTMLElement.prototype,
+      [SupportedDimension.OffsetHeight]: HTMLElement.prototype,
+      [SupportedDimension.ScrollWidth]: Element.prototype,
+      [SupportedDimension.ScrollHeight]: Element.prototype,
+    };
+  }
+
+  private mockDOMMethods(mocks: Mocks) {
+    Object.keys(mocks).forEach(method => {
+      const nativeSource = this.nativeImplementations[method];
+      const mock: Mock | undefined = mocks[method];
+
+      this.overwrittenImplementations.push(method);
+
+      if (isGetterFunction(mock)) {
+        Object.defineProperty(nativeSource, method, {
+          get() {
+            return mock.call(this, this);
+          },
+          configurable: true,
+        });
+      } else {
+        Object.defineProperty(nativeSource, method, {
+          value: mocks[method],
+          configurable: true,
+        });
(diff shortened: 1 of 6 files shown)


packages/jest-dom-mocks/src/dimension.ts
```diff
@@ -0,0 +1,68 @@
...
+  ScrollHeight = 'scrollHeight',
+}
+
+type DimensionMap = {[T in SupportedDimension]: HTMLElement | Element};
+type MockedDimensions = {[T in SupportedDimension]: number};
+
+export default class Dimension {
+  private isUsingMock = false;
+  private mocks: Partial<MockedDimensions> = {};
+  private dimensionMap: DimensionMap = {
```
lemonmade: It is probably safer not to actually reference `Element`/ `HTMLElement` until after `mock` is called, otherwise just initializing this can crash tests that bring in one of the other mocks but aren't using the browser jest environment.

packages/jest-dom-mocks/src/dimension.ts
```diff
@@ -0,0 +1,68 @@
...
+    this.isUsingMock = false;
+  }
+
+  isMocked() {
+    return this.isUsingMock;
+  }
+
+  private assignDimensionProperties(properties: Partial<MockedDimensions>) {
+    let key: SupportedDimension;
+    for (key in properties) {
```
lemonmade: We usually do loops over `Object.keys(obj)` instead of `for in` loops

packages/jest-dom-mocks/src/dimension.ts
```diff
@@ -0,0 +1,68 @@
...
+        });
+      }
+    }
+  }
+
+  private removeDimensionProperties(properties: Partial<MockedDimensions>) {
+    let key: SupportedDimension;
+    for (key in properties) {
+      if (properties.hasOwnProperty(key)) {
+        delete properties[key];
```
lemonmade: Does this actually restore? If there was anything there originally, won't this completely remove it?

packages/jest-dom-mocks/src/dimension.ts
```diff
@@ -0,0 +1,68 @@
...
+  private isUsingMock = false;
+  private mocks: Partial<MockedDimensions> = {};
+  private dimensionMap: DimensionMap = {
+    [SupportedDimension.OffsetWidth]: HTMLElement.prototype,
+    [SupportedDimension.OffsetHeight]: HTMLElement.prototype,
+    [SupportedDimension.ScrollWidth]: Element.prototype,
+    [SupportedDimension.ScrollHeight]: Element.prototype,
+  };
+
+  mock(mocks: Partial<MockedDimensions>) {
```
lemonmade: Do you think it's worth making this mock a little more flexible at this point? Something like also accepting a function that takes an element and the property being measured and returns a number (for cases where multiple elements are measured)?

donnielrt: Yeah, not a bad idea, will change 👍

packages/jest-dom-mocks/src/dimension.ts
donnielrt: None of these dimension methods actually exist in JSDom, but it's still probably safer to only delete our version of the mocks 👍

packages/jest-dom-mocks/src/dimension.ts
```diff
@@ -0,0 +1,102 @@
+import {memoize} from '@shopify/javascript-utilities/decorators';
+
+const enum SupportedDimension {
+  OffsetWidth = 'offsetWidth',
+  OffsetHeight = 'offsetHeight',
+  ScrollWidth = 'scrollWidth',
+  ScrollHeight = 'scrollHeight',
+}
+
+type MockedGetter = (element: HTMLElement) => number;
```
donnielrt: Excuse my TS-noobness, better typing suggestions welcome.

packages/jest-dom-mocks/src/test/dimension.test.ts
```diff
@@ -0,0 +1,116 @@
...
+      const dimension = new Dimension();
+
+      const targetEl = document.createElement('div');
+      const otherEl = document.createElement('div');
+
+      targetEl.id = 'testId';
+
+      dimension.mock({
+        scrollWidth(element: HTMLElement) {
+          return element.id === 'testId' ? 200 : 0;
```
donnielrt: I'm not too sure about this API, maybe it feels a little un-react-ey and clunky? 

Sample implementation of how we'd use this in Web: https://github.com/Shopify/web/pull/13540/files.

packages/jest-dom-mocks/src/dimension.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  private restoreDOMMethods() {
+    this.overwrittenImplementations.forEach(method => {
+      const nativeSource = this.nativeImplementations[method];
+
+      if (nativeSource == null) {
+        return;
+      }
+
+      delete (nativeSource as AugmentedElement)[method];
```
donnielrt: 🙈

donnielrt: Oh, and regarding your earlier comment re: `delete`ing instead of restoring. We can't really access the native implementations, by design, so I think it makes sense to just nuke our implementation.

TheMallen:
@donnielrt does this still need a review? What's the state of it?

donnielrt:
@TheMallen yes, please :) I've made some changes after Chris' last comments, would love some :eyes: on whether those look good. Happy to rework as needed!

donnielrt:
@TheMallen I agree that the name `restore` is a bit inaccurate, but I wonder if it's better to keep the naming consistent with other mocking utilities?

donnielrt:
@lemonmade are you happy with the changes? Good to merge?

PR #627, opened 2019-04-03
`useShortcut()` tests

cartogram:
See https://github.com/Shopify/quilt/pull/615#issuecomment-479252968 for more details.

This PR is a follow up to https://github.com/Shopify/quilt/pull/615 that begins to add additional hook-specific tests.

Question: Are these useful or just repetitive given we already[ test the `Shortcut` component](https://github.com/Shopify/quilt/blob/4f2da6ac543f92efb8008a2080d33f5c52fdcfa1/packages/react-shortcuts/src/tests/ShortcutManager.test.tsx)? Is there a rule of thumb for when/ when not to test a hook? (Ie: the [component version](https://github.com/Shopify/quilt/blob/4f2da6ac543f92efb8008a2080d33f5c52fdcfa1/packages/react-shortcuts/src/Shortcut/Shortcut.tsx) doesn't add any new features.) 

I stopped shortly after beginning to add these tests because I believe they add almost no value given the additional time it would take to maintain/run them. Would like at least one second opinion on this though.

### DIFF
--- a/packages/react-shortcuts/src/tests/useShortcut.test.tsx
+++ b/packages/react-shortcuts/src/tests/useShortcut.test.tsx
@@ -1,0 +1,83 @@
+import * as React from 'react';
+import {mount} from '@shopify/react-testing';
+import Key, {ModifierKey, HeldKey} from '../keys';
+import {useShortcut} from '../Shortcut';
+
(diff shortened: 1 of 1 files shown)


cartogram:
closing

Issue #628, opened 2019-04-03
Allowing the GraphQL proxy to support API version


alexandcote:
## Overview
We should probably support API versioning in the GraphQL Proxy. 

You can find more context on [API Versioning here](https://development.shopify.io/engineering/developing_at_Shopify/api/api_version/api_version_tldr)

## Type

- [ ] New feature
- [x] Changes to existing features

## Motivation
If you want your web app to use a specific API version you can't

## Checklist

- [x] Please delete the labels section before submitting your issue
- [x] I have described this issue in a way that is actionable (if possible)

PR #629, opened 2019-04-03
Add an option to configure the api version on the graphql proxy

alexandcote:
Fix: #628

✨With this PR, we can now specify an API version when you create the proxy.

You can do something like:
```ts
...
proxy({version: ApiVersion.Unstable})(ctx, next),
```

Waiting for your feedback

### DIFF
--- a/packages/koa-shopify-graphql-proxy/src/shopify-graphql-proxy.ts
+++ b/packages/koa-shopify-graphql-proxy/src/shopify-graphql-proxy.ts
@@ -5,13 +5,13 @@ export const PROXY_BASE_PATH = '/graphql';
 export const GRAPHQL_PATH_PREFIX = '/admin/api';
 
 export enum ApiVersion {
-  Unstable = 'unstable',
   '2019-04' = '2019-04',
   '2019-09' = '2019-09',
+  Unstable = 'unstable',
 }
 
 interface DefaultProxyOptions {
-  apiVersion?: ApiVersion;
+  apiVersion: ApiVersion;
 }
 
 interface PrivateShopOption extends DefaultProxyOptions {
@@ -21,20 +21,17 @@ interface PrivateShopOption extends DefaultProxyOptions {
 
 type ProxyOptions = PrivateShopOption | DefaultProxyOptions;
 
-export default function shopifyGraphQLProxy(proxyOptions?: ProxyOptions) {
+export default function shopifyGraphQLProxy(proxyOptions: ProxyOptions) {
   return async function shopifyGraphQLProxyMiddleware(
     ctx: Context,
     next: () => Promise<any>,
   ) {
     const {session = {}} = ctx;
 
-    const shop =
-      proxyOptions && 'shop' in proxyOptions ? proxyOptions.shop : session.shop;
+    const shop = 'shop' in proxyOptions ? proxyOptions.shop : session.shop;
     const accessToken =
-      proxyOptions && 'password' in proxyOptions
-        ? proxyOptions.password
-        : session.accessToken;
-    const apiVersion = proxyOptions ? proxyOptions.apiVersion : null;
+      'password' in proxyOptions ? proxyOptions.password : session.accessToken;
+    const apiVersion = proxyOptions.apiVersion;
 
     if (ctx.path !== PROXY_BASE_PATH || ctx.method !== 'POST') {
       await next();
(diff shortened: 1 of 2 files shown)


packages/koa-shopify-graphql-proxy/src/shopify-graphql-proxy.ts
```diff
@@ -2,24 +2,34 @@ import proxy from 'koa-better-http-proxy';
...
 export default function shopifyGraphQLProxy(proxyOptions?: ProxyOptions) {
   return async function shopifyGraphQLProxyMiddleware(
     ctx: Context,
     next: () => Promise<any>,
   ) {
     const {session = {}} = ctx;
 
-    const shop = proxyOptions ? proxyOptions.shop : session.shop;
-    const accessToken = proxyOptions
+    const shop = isPrivateShopOption(proxyOptions)
```
lemonmade: I usually just use inline `'shop' in proxyOptions` for this kind of check

packages/koa-shopify-graphql-proxy/src/shopify-graphql-proxy.ts
```diff
@@ -41,7 +51,11 @@ export default function shopifyGraphQLProxy(proxyOptions?: ProxyOptions) {
         'X-Shopify-Access-Token': accessToken,
       },
       proxyReqPathResolver() {
-        return `https://${shop}${GRAPHQL_PATH}`;
+        if (apiVersion) {
+          return `https://${shop}${GRAPHQL_PATH_PREFIX}${apiVersion}/${GRAPHQL_PATH_SUFFIX}`;
```
lemonmade: Might want to just have `GRAPHQL_PATH_PREFIX` not have a trailing `/` since it's a bit unclear just from reading this bit what strings the path segments are part of (that is, explicitly put the path separators in between each part)

packages/koa-shopify-graphql-proxy/src/shopify-graphql-proxy.ts
```diff
@@ -2,24 +2,36 @@ import proxy from 'koa-better-http-proxy';
 import {Context} from 'koa';
 
 export const PROXY_BASE_PATH = '/graphql';
-export const GRAPHQL_PATH = '/admin/api/graphql.json';
+export const GRAPHQL_PATH_PREFIX = '/admin/api';
 
-export interface ProxyOptions {
+export enum ApiVersion {
+  '2019-04' = '2019-04',
+  '2019-09' = '2019-09',
```
lemonmade: Can we use nicer names for these? `April19`/ `September19` maybe?

packages/koa-shopify-graphql-proxy/src/shopify-graphql-proxy.ts
```diff
@@ -2,24 +2,36 @@ import proxy from 'koa-better-http-proxy';
...
+  Unstable = 'unstable',
+}
+
+interface DefaultProxyOptions {
+  apiVersion: ApiVersion;
```
lemonmade: Maybe just `version`?

packages/koa-shopify-graphql-proxy/src/shopify-graphql-proxy.ts
alexandcote: You make me realize the API version is not September but July. So I did the changes you propose 👍

packages/koa-shopify-graphql-proxy/src/shopify-graphql-proxy.ts
```diff
@@ -2,24 +2,36 @@ import proxy from 'koa-better-http-proxy';
 import {Context} from 'koa';
 
 export const PROXY_BASE_PATH = '/graphql';
-export const GRAPHQL_PATH = '/admin/api/graphql.json';
+export const GRAPHQL_PATH_PREFIX = '/admin/api';
 
-export interface ProxyOptions {
+export enum Version {
```
alexandcote: Do you prefer Version or ApiVersion for the enum?

alexandcote: We chat and will keep `ApiVersion` 👍🏼

tylerball:
Can we add something to the README about this? Currently you have to dive into the code to see what options it takes

alexandcote:
@tylerball Thanks for the reminder! I forgot about that.
https://github.com/Shopify/quilt/pull/651

PR #635, opened 2019-04-08
Bump hot-shots for @Shopify/koa-metrics

haeky:
After debugging why statsd events are not getting firing with `@Shopify/koa-metrics`, I've narrowed it down to `hot-shots`. My assumption is that the socket is getting closed before the data is sent. 

https://github.com/brightcove/hot-shots/commit/c49197c2adb451bbd8e62641c1918c1b4ee97009

### DIFF
--- a/packages/koa-metrics/package.json
+++ b/packages/koa-metrics/package.json
@@ -23,7 +23,7 @@
   },
   "homepage": "https://github.com/shopify/quilt/blob/master/packages/koa-metrics/README.md",
   "dependencies": {
-    "hot-shots": "^5.4.1",
+    "hot-shots": "^6.1.1",
     "tslib": "^1.9.3"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


packages/koa-metrics/package.json
```diff
@@ -23,7 +23,7 @@
   },
   "homepage": "https://github.com/shopify/quilt/blob/master/packages/koa-metrics/README.md",
   "dependencies": {
-    "hot-shots": "^5.4.1",
+    "hot-shots": "^6.1.1",
```
ismail-syed: We generally update our dependencies using `yarn` so the corresponding `yarn.lock`s resolve properly. 

```
yarn add hot-shots
```

ismail-syed:
The brightcove/hot-shots@c49197c changeset looks like it was shipped for `v6.1.1` and `v6.1.0`. /web seems to be using [`5.9.1`](https://github.com/Shopify/web/blob/master/yarn.lock#L8948). Are we able to test this to verify that the bump fixes the issue?

packages/koa-metrics/package.json
haeky: I didn't include the root `yarn.lock` when working on `koa-metrics`. Should be fixed now.

PR #636, opened 2019-04-09
🐛 useServerDomEffect now returns a result

GoodForOneFare:
Previously, `perform`'s return value was discarded, causing `useServerEffect` to add `undefined` to the effect queue.  With no pending promises in the queue, react-html erroneously flagged its render as complete, and finished extraction after the first pass.

(My use case is `<Serialize data={() => data || getDataFromNetwork()} />`, with the network call never having a chance to resolve).


GoodForOneFare:
Added a `useSerialized` test, which indirectly verifies the fixed behaviour.  Thanks for helping to come up with a sane test, @lemonmade 

Will merge on 🍏

Issue #586, opened 2019-03-21
Add "blessed" hooks for GraphQL queries and mutations


lemonmade:

lemonmade:
This library will give us a pretty good idea of how to do it: https://github.com/trojanowski/react-apollo-hooks. In addition to what that library does, though, we need to make it support query component generated by `@shopify/react-graphql`'s `createAsyncQueryComponent`. So, for queries, I'd expect the following to all work:

```tsx
import myQuery from './graphql/MyQuery.graphql';

const MyQuery = createAsyncQueryComponent({
  load: () => import('./graphql/MyQuery.graphql'),
});

const {data, loading, error} = useQuery(myQuery);
const {data, loading, error} = useQuery(MyQuery);
```

We do not need it to do any special suspense stuff, as it won't work with SSR (they have a hack in the above library to make it work, but let's ignore for now).

On the mutation side, I think just this is needed:

```tsx
const mutate = useMutation(myMutation);

// eventually:

mutate({variables: {}});
```

I could be convinced that it is better to let them also pass in mutation options up front, and maybe just merge them together at the end?

In all cases, we should have 100% amazing typing that is automatically inferred (including not letting them pass variables if there are none, forcing them to pass variables if there are non-optional ones, etc).

PR #642, opened 2019-04-09
Add a set of apollo hooks

michelleyschen:
Close https://github.com/Shopify/quilt/issues/586

🎩 instructions:

1. get [this branch](https://github.com/Shopify/web-proving-ground/tree/apollo-query-hook) on `web-proving-ground` locally
2. run `yarn` in `web-proving-ground`
3. in `quilt`, do `yarn && yarn tophat react-graphql ../web-proving-ground`
4. once it complied, go to `web-proving-ground` and do `yarn dev`
5. goto `localhost:8081` to see the query & mutation working properly

### DIFF
--- a/packages/react-graphql/src/async.tsx
+++ b/packages/react-graphql/src/async.tsx
@@ -1,7 +1,11 @@
 import * as React from 'react';
 
 import {LoadProps, DeferTiming} from '@shopify/async';
-import {Async, AsyncPropsRuntime} from '@shopify/react-async';
+import {
+  Async,
+  AsyncPropsRuntime,
+  resolve as resolver,
+} from '@shopify/react-async';
 import {Omit} from '@shopify/useful-types';
 import {DocumentNode} from 'graphql-typed';
 
@@ -28,6 +32,23 @@ export function createAsyncQueryComponent<Data, Variables, DeepPartial>({
   Variables,
   DeepPartial
 >): AsyncQueryComponentType<Data, Variables, DeepPartial> {
+  // set it to the resolved of tryRequire
+  let resolvedDocument: DocumentNode<
+    Data,
+    Variables,
+    DeepPartial
+  > | null = null;
+
+  async function resolve() {
+    try {
+      const response = await resolver(load);
+      resolvedDocument = response;
+      return response;
+    } catch (error) {
+      throw Error('error loading Query document');
+    }
+  }
+
   function AsyncQuery(
     props: Omit<QueryProps<Data, Variables>, 'query'> & ConstantProps,
   ) {
@@ -109,6 +130,13 @@ export function createAsyncQueryComponent<Data, Variables, DeepPartial>({
   FinalComponent.Preload = Preload;
   FinalComponent.Prefetch = Prefetch;
   FinalComponent.KeepFresh = KeepFresh;
+  FinalComponent.resolve = resolve;
+
+  Object.defineProperty(FinalComponent, 'resolved', {
+    get: () => resolvedDocument,
+    // accessor descriptor don't need this
+    // writable: false,
+  });
 
   return FinalComponent;
 }
--- a/packages/react-graphql/src/hooks/utilities.ts
+++ b/packages/react-graphql/src/hooks/utilities.ts
@@ -1,0 +1,27 @@
+import isPlainObject from 'lodash/isPlainObject';
+
+export function objectToKey<T extends Record<string, any>>(obj: T): T | string {
+  if (!isPlainObject(obj)) {
+    return obj;
+  }
+  const sortedObj = Object.keys(obj)
+    .sort()
+    .reduce((result: Record<string, any>, key) => {
+      result[key] = objectToKey(obj[key]);
+      return result;
+    }, {});
+  return JSON.stringify(sortedObj);
+}
+
+export function compact(obj: any) {
+  return Object.keys(obj).reduce(
+    (acc, key) => {
+      if (obj[key] !== undefined) {
+        acc[key] = obj[key];
+      }
+
+      return acc;
+    },
+    {} as any,
+  );
+}
--- a/packages/react-graphql/src/types.ts
+++ b/packages/react-graphql/src/types.ts
@@ -58,4 +58,6 @@ export interface AsyncQueryComponentType<Data, Variables, DeepPartial>
   KeepFresh(
     props: VariableOptions<Variables> & {pollInterval?: number} & ConstantProps,
   ): React.ReactElement<{}>;
+  resolve(): Promise<DocumentNode<Data, Variables, DeepPartial>>;
+  resolved: DocumentNode<Data, Variables, DeepPartial> | null;
 }
--- a/packages/react-graphql/README.md
+++ b/packages/react-graphql/README.md
@@ -173,6 +173,71 @@ function App() {
 render(<App />, document.getElementById('root'));
 ```
 
+### `useQuery`
+
+Here is an example of how to use `useQuery` with a query document.
+
+```tsx
+import React from 'react';
+import {useQuery} from '@shopify/react-graphql';
+
+import customerListQuery from './graphql/CustomerListQuery.graphql';
+
+function CustomerList() {
+  const {data, loading, refetch} = useQuery(customerListQuery);
+  if (loading) {
+    return <div>Loading...</div>;
+  }
+
+  const customers = data && data.customers ? data.customers : [];
+  return (
+    <ul>
+      {customers.map(customer => (
+        <li key={customer.id}>{customer.displayName}</li>
+      ))}
+    </ul>
+    <button onClick={async () => {
+        await refetch();
+    }}>Refetch Customers Data</button>
+  );
+}
+```
+
+You can also use `useQuery` with an async query component created from `@shopify/react-graphql`
+
+Note that because the `.graphql` document is being asynchronously loaded, `refetch` and other helper methods may not be available until the query document is loaded.
+
+```tsx
+import React from 'react';
+import {createAsyncQueryComponent, useQuery} from '@shopify/react-graphql';
+
+const CustomerListQuery = createAsyncQueryComponent({
+  load: () => import('./graphql/CustomerListQuery.graphql'),
+});
+
+function CustomerList() {
+  const {data, loading, refetch} = useQuery(CustomerListQuery);
+  if (loading) {
+    return <div>Loading...</div>;
+  }
+
+  const customers = data && data.customers ? data.customers : [];
+  return (
+    <ul>
+      {customers.map(customer => (
+        <li key={customer.id}>{customer.displayName}</li>
+      ))}
+    </ul>
+    <button onClick={async () => {
+        if (!refetch) {
+            return;
+        }
+        await refetch();
+    }}>Refetch Customers Data</button>
+  );
+}
+```
+
 ### `useMutation`
 
 This hook accepts two arguments: the mutation document, and optionally, a set of options to pass to the underlying mutation. It will return a function that will trigger the mutation when invoked.
--- a/packages/react-graphql/src/hooks/graphql-document.ts
+++ b/packages/react-graphql/src/hooks/graphql-document.ts
@@ -1,0 +1,55 @@
+import {useState, useEffect, useCallback} from 'react';
+import {OperationVariables} from 'apollo-client';
+import {DocumentNode} from 'graphql-typed';
+
+import {AsyncQueryComponentType} from '..';
+
+export default function useGraphQLDocument<
+  Data = any,
+  Variables = OperationVariables,
+  DeepPartial = {}
+>(
+  documentOrComponent:
+    | DocumentNode<Data, Variables>
+    | AsyncQueryComponentType<Data, Variables, DeepPartial>,
+) {
+  const [document, setDocument] = useState<DocumentNode<
+    Data,
+    Variables
+  > | null>(() => {
+    if (isDocumentNode(documentOrComponent)) {
+      return documentOrComponent;
+    } else {
+      return documentOrComponent.resolved || null;
+    }
+  });
+
+  const loadDocument = useCallback(
+    async () => {
+      if (!isDocumentNode(documentOrComponent)) {
+        try {
+          const resolved = await documentOrComponent.resolve();
+          setDocument(resolved);
+        } catch (error) {
+          throw Error('error loading GraphQL document');
+        }
+      }
+    },
+    [documentOrComponent],
+  );
+
+  useEffect(
+    () => {
+      if (!document) {
+        loadDocument();
+      }
+    },
+    [document, documentOrComponent, loadDocument],
+  );
+
+  return document;
+}
+
+function isDocumentNode(arg: any): arg is DocumentNode {
+  return Boolean(arg && arg.kind && arg.kind === 'Document');
+}
--- a/packages/react-graphql/src/hooks/index.ts
+++ b/packages/react-graphql/src/hooks/index.ts
@@ -1,4 +1,5 @@
 export {ApolloProvider} from './ApolloProvider';
 export {useApolloClient} from './apollo-client';
+export {default as useQuery} from './query';
 export {default as useMutation} from './mutation';
 export * from './types';
--- a/packages/react-graphql/src/hooks/query-cache.ts
+++ b/packages/react-graphql/src/hooks/query-cache.ts
@@ -1,0 +1,48 @@
+import ApolloClient, {ObservableQuery, WatchQueryOptions} from 'apollo-client';
+import {print} from 'graphql/language/printer';
+
+import {objectToKey} from './utilities';
+
+const cachedQueriesByClient = new WeakMap<
+  ApolloClient<any>,
+  Map<any, ObservableQuery<any, any>>
+>();
+
+export function getCachedObservableQuery<TData, TVariables>(
+  client: ApolloClient<any>,
+  options: WatchQueryOptions<TVariables>,
+): ObservableQuery<TData, TVariables> {
+  const queriesForClient = getCachedQueriesForClient(client);
+  const cacheKey = getCacheKey(options);
+  let observableQuery = queriesForClient.get(cacheKey);
+  if (observableQuery == null) {
+    observableQuery = client.watchQuery(options);
+    queriesForClient.set(cacheKey, observableQuery);
+  }
+  return observableQuery;
+}
+
+export function invalidateCachedObservableQuery<TVariables>(
+  client: ApolloClient<any>,
+  options: WatchQueryOptions<TVariables>,
+): void {
+  const queriesForClient = getCachedQueriesForClient(client);
+  const cacheKey = getCacheKey(options);
+  queriesForClient.delete(cacheKey);
+}
+
+function getCachedQueriesForClient(client: ApolloClient<any>) {
+  let queriesForClient = cachedQueriesByClient.get(client);
+  if (queriesForClient == null) {
+    queriesForClient = new Map();
+    cachedQueriesByClient.set(client, queriesForClient);
+  }
+  return queriesForClient;
+}
+
+function getCacheKey<TVariables>({
+  query,
+  ...options
+}: WatchQueryOptions<TVariables>): string {
+  return `${print(query)}@@${objectToKey(options)}`;
+}
--- a/packages/react-graphql/src/hooks/query.ts
+++ b/packages/react-graphql/src/hooks/query.ts
@@ -1,0 +1,157 @@
+import {
+  OperationVariables,
+  ApolloError,
+  WatchQueryOptions,
+} from 'apollo-client';
+import {DocumentNode} from 'graphql-typed';
+import {useEffect, useMemo, useState} from 'react';
+
+import {AsyncQueryComponentType} from '..';
+
+import {QueryHookOptions, QueryHookResult} from './types';
+import {useApolloClient} from './apollo-client';
+import useGraphQLDocument from './graphql-document';
+import {
+  getCachedObservableQuery,
+  invalidateCachedObservableQuery,
+} from './query-cache';
+import {compact, objectToKey} from './utilities';
+
+export default function useQuery<
+  Data = any,
+  Variables = OperationVariables,
+  DeepPartial = {}
+>(
+  queryOrComponent:
+    | DocumentNode<Data, Variables>
+    | AsyncQueryComponentType<Data, Variables, DeepPartial>,
+  options: QueryHookOptions<Data, Variables> = {},
+): QueryHookResult<Data, Variables> {
+  const {
+    skip = false,
+    pollInterval,
+    notifyOnNetworkStatusChange = false,
+    client: overrideClient,
+    context,
+    variables,
+    fetchPolicy,
+    errorPolicy,
+  } = options;
+
+  const [responseId, setResponseId] = useState(0);
+  const client = useApolloClient(overrideClient);
+  const query = useGraphQLDocument(queryOrComponent);
+
+  const watchQueryOptions: WatchQueryOptions<Variables> = useMemo(
+    () =>
+      compact({
+        context,
+        errorPolicy,
+        fetchPolicy,
+        notifyOnNetworkStatusChange,
+        pollInterval,
+        query,
+        variables,
+      }),
+    [
+      context,
+      errorPolicy,
+      fetchPolicy,
+      notifyOnNetworkStatusChange,
+      pollInterval,
+      query,
+      variables,
+    ],
+  );
+
+  const observableQuery = useMemo(
+    () => {
+      if (!query) {
+        return;
+      }
+
+      return getCachedObservableQuery<Data, Variables>(
+        client,
+        watchQueryOptions,
+      );
+    },
+    [query, client, watchQueryOptions],
+  );
+
+  useEffect(
+    () => {
+      if (skip || !observableQuery) {
+        return;
+      }
+
+      const invalidateCurrentResult = () => {
+        setResponseId(x => x + 1);
+      };
+      const subscription = observableQuery.subscribe(
+        invalidateCurrentResult,
+        invalidateCurrentResult,
+      );
+
+      invalidateCachedObservableQuery(client, watchQueryOptions);
+
+      return () => {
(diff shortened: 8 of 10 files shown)


packages/react-apollo-hooks/src/useQuery.ts
```diff
@@ -0,0 +1,202 @@
...
+
+  if (!observableQuery) {
+    return {
+      loading: true,
+      networkStatus: undefined,
+    };
+  }
+
+  return {
+    ...getDefaultQueryResult(client, observableQuery),
```
lemonmade: Feels to me like we should do this in the `useMemo` where we create `currentResult` instead?

packages/react-apollo-hooks/src/utils.ts
```diff
@@ -0,0 +1,31 @@
+import isPlainObject from 'lodash/isPlainObject';
```
lemonmade: No Lodash dependencies 🙈

packages/react-graphql/src/async.tsx
```diff
@@ -109,6 +117,7 @@ export function createAsyncQueryComponent<Data, Variables, DeepPartial>({
   FinalComponent.Preload = Preload;
   FinalComponent.Prefetch = Prefetch;
   FinalComponent.KeepFresh = KeepFresh;
+  FinalComponent.resolve = resolve;
```
lemonmade: Do we not need a `resolved` key?

packages/react-graphql/src/types.ts
```diff
@@ -58,4 +58,5 @@ export interface AsyncQueryComponentType<Data, Variables, DeepPartial>
   KeepFresh(
     props: VariableOptions<Variables> & {pollInterval?: number} & ConstantProps,
   ): React.ReactElement<{}>;
+  resolve(): Promise<DocumentNode<Data, Variables, DeepPartial> | null>;
```
lemonmade: Shouldn't this always resolve to a real document node?

packages/react-apollo-hooks/src/useQuery.ts
```diff
@@ -0,0 +1,202 @@
...
+      variables && objToKey(variables),
+      fetchPolicy,
+      errorPolicy,
+    ],
+  );
+
+  const observableQuery = useMemo(
+    () => {
+      if (!query) {
+        return;
```
lemonmade: I would return `null` here personally

packages/react-apollo-hooks/src/useQuery.ts
```diff
@@ -0,0 +1,202 @@
...
+        partial: result.partial,
+      };
+    },
+    [skip, responseId, observableQuery],
+  );
+
+  if (!observableQuery) {
+    return {
+      loading: true,
+      networkStatus: undefined,
```
lemonmade: I feel like this result should be basically indistinguishable from a resolved query in the loading state (so, has all the same methods; it's fine of the methods are noops)

packages/react-apollo-hooks/README.md
```diff
@@ -0,0 +1,125 @@
+# `@shopify/react-apollo-hooks`
```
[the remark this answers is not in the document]
michelleyschen: that actually makes things easier, so 👍

packages/react-apollo-hooks/src/useQuery.ts
```diff
@@ -0,0 +1,202 @@
...
+export default function useQuery<
+  Data = any,
+  Variables = OperationVariables,
+  DeepPartial = {}
+>(
+  queryOrComponent: AsyncQueryComponentType<Data, Variables, DeepPartial>,
+  options?: QueryHookOptions<Data, Variables>,
+): QueryDocumentResolving<Data, Variables> | QueryHookResult<Data, Variables>;
+
+export default function useQuery<
```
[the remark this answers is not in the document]
michelleyschen: I mostly did the overriding so that if you pass in a queryDocument, you won't need to check to see if all the helper methods (like `refetch`), since I can infer the return type from what's being pass in.

packages/react-graphql/src/async.tsx
michelleyschen: I am not using it yet. I still believe you that I will need it for test that I am trying to figure out now.

packages/react-graphql/src/types.ts
michelleyschen: not if the promise fail right?

lemonmade: If it fails you should throw an error, there isn't a very good recovery case.

packages/react-apollo-hooks/src/useQuery.ts
```diff
@@ -0,0 +1,202 @@
...
+>(
+  queryOrComponent:
+    | DocumentNode<Data, Variables>
+    | AsyncQueryComponentType<Data, Variables, DeepPartial>,
+  options: QueryHookOptions<Data, Variables> = {},
+): QueryHookResult<Data, Variables> | QueryDocumentResolving<Data, Variables> {
+  const {
+    skip = false,
+    pollInterval,
+    notifyOnNetworkStatusChange = false,
```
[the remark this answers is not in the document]
michelleyschen: 10 uses actually, wow

packages/react-apollo-hooks/src/useQuery.ts
```diff
@@ -0,0 +1,202 @@
...
+    client: overrideClient,
+    context,
```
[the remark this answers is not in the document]
michelleyschen: > Shared context between your Query component and your network interface (Apollo Link). Useful for setting headers from props or sending information to the `request` function of Apollo Boost.

we may not be using this in web actually, it's a little harder to search for.

packages/react-graphql/src/hooks/useQuery.ts
```diff
@@ -0,0 +1,202 @@
...
+    | DocumentNode<Data, Variables>
+    | AsyncQueryComponentType<Data, Variables, DeepPartial>,
+  options: QueryHookOptions<Data, Variables> = {},
+): QueryHookResult<Data, Variables> | QueryDocumentResolving<Data, Variables> {
+  const {
+    skip = false,
+    pollInterval,
+    notifyOnNetworkStatusChange = false,
+    client: overrideClient,
+    context,
```
michelleyschen: carry over things that's not resolved yet but lost because refactor

> What's this for?

michelleyschen: > Shared context between your Query component and your network interface (Apollo Link). Useful for setting headers from props or sending information to the request function of Apollo Boost.

we may not be using this in web actually, it's a little harder to search for.

packages/react-graphql/src/hooks/useQuery.ts
```diff
@@ -0,0 +1,202 @@
...
+} from './types';
+import {useApolloClient} from './ApolloContext';
+import useGraphQLDocument from './useGraphQLDocument';
+import {
+  getCachedObservableQuery,
+  invalidateCachedObservableQuery,
+} from './queryCache';
+import {compact, objectToKey, actHack} from './utilities';
+
+export default function useQuery<Data = any, Variables = OperationVariables>(
```
michelleyschen: _carry over things that's not resolved yet but lost due to moving files_

https://github.com/Shopify/quilt/pull/642#discussion_r273720608
> Do you need overrides for this? I would have assumed you'd be OK without them since the thing you pass the queryOrComponent to accepts both types?

> Also, random, but I've been getting in the habit of only using named exports. Want to join the club? :P

michelleyschen: I mostly did the overriding so that if you pass in a queryDocument, you won't need to check to see if all the helper methods (like refetch), since I can infer the return type from what's being pass in.

packages/react-graphql/src/hooks/useQuery.ts
```diff
@@ -0,0 +1,202 @@
...
+        return;
+      }
+
+      const invalidateCurrentResult = () => {
+        // A hack to get rid React warnings during tests. The default
+        // implementation of `actHack` just invokes the callback immediately.
+        // In tests, it's replaced with `act` from react-testing-library.
+        // A better solution welcome.
+        actHack(() => {
+          setResponseId(x => x + 1);
```
michelleyschen: _carry over things that's not resolved yet but lost due to moving files_

https://github.com/Shopify/quilt/pull/642#discussion_r273721557
> Hmm seems like a bad idea, not sure why we don't just use a useRef for the current response ID instead of state?

packages/react-graphql/src/hooks/useQuery.ts
```diff
@@ -0,0 +1,202 @@
...
+
+  if (!observableQuery) {
+    return {
+      loading: true,
+      networkStatus: undefined,
+    };
+  }
+
+  return {
+    ...getDefaultQueryResult(client, observableQuery),
```
michelleyschen: _carry over things that's not resolved yet but lost due to moving files_

https://github.com/Shopify/quilt/pull/642#discussion_r273722289
> Feels to me like we should do this in the `useMemo` where we create currentResult instead?

packages/react-graphql/src/hooks/utilities.ts
```diff
@@ -0,0 +1,31 @@
+import isPlainObject from 'lodash/isPlainObject';
```
michelleyschen: _carry over things that's not resolved yet but lost due to moving files_

https://github.com/Shopify/quilt/pull/642#discussion_r273723029
> No Lodash dependencies 🙈

packages/react-graphql/src/hooks/useQuery.ts
```diff
@@ -0,0 +1,202 @@
...
+        partial: result.partial,
+      };
+    },
+    [skip, responseId, observableQuery],
+  );
+
+  if (!observableQuery) {
+    return {
+      loading: true,
+      networkStatus: undefined,
```
michelleyschen: _carry over things that's not resolved yet but lost due to moving files_

https://github.com/Shopify/quilt/pull/642#discussion_r273724153
> I feel like this result should be basically indistinguishable from a resolved query in the loading state (so, has all the same methods; it's fine of the methods are noops)

> Also, random, but I've been getting in the habit of only using named exports. Want to join the club? :P

packages/react-testing/src/mount.ts
```diff
@@ -78,7 +78,7 @@ type CustomMountResult<
   ? Promise<CustomRoot<Props, Context>>
   : CustomRoot<Props, Context>;
 
-class CustomRoot<Props, Context extends object> extends Root<Props> {
+export class CustomRoot<Props, Context extends object> extends Root<Props> {
```
michelleyschen: if I don't export this I get the following error

```
Exported variable 'mountWithGraphQL' has or is using name 'CustomRoot' from external module "Shopify/quilt/packages/react-testing/src/mount" but cannot be named.
```

packages/react-testing/README.md
```diff
@@ -375,7 +375,7 @@ export const mountWithGraphQL = createMount<Options, Context, true>({
   async afterMount(root, {skipInitialGraphQL}) {
     // This is a temporary hack to make GraphQL resolution behave pseudo-synchronously
     // to avoid warnings about setting state outside of act() blocks.
-    root.graphQL.on('pre-resolve', () => {
+    root.context.graphQL.on('pre-resolve', () => {
```
michelleyschen: fix example

packages/react-graphql/src/hooks/useQuery.ts
```diff
@@ -0,0 +1,198 @@
...
+  );
+
+  const currentResult = useMemo<
+    QueryHookResult<Data, Variables> | QueryDocumentResolving<Data, Variables>
+  >(
+    () => {
+      if (!observableQuery) {
+        return {
+          loading: true,
+          networkStatus: undefined,
```
michelleyschen: _carry over things that's not resolved yet but lost due to moving files_

https://github.com/Shopify/quilt/pull/642#discussion_r273724153
> I feel like this result should be basically indistinguishable from a resolved query in the loading state (so, has all the same methods; it's fine of the methods are noops)

packages/react-graphql/src/hooks/README.md
```diff
@@ -0,0 +1,174 @@
...
+```tsx
+import React from 'react';
+import {createAsyncQueryComponent, useQuery} from '@shopify/react-graphql';
+
+const CustomerListQuery = createAsyncQueryComponent({
+  load: () => import('./graphql/CustomerListQuery.graphql'),
+});
+
+function CustomerList() {
+  const {data, loading, refetch} = useQuery(customerListQuery);
```
tleunen: nit: Should be capitalized `CustomerListQuery`

packages/react-graphql/src/hooks/README.md
```diff
@@ -0,0 +1,174 @@
...
+import {Form, TextField, Button, Banner} from '@shopify/polaris';
+import {useQuery} from '@shopify/react-graphql';
+
+import createCustomerMutation from './graphql/CreateCustomerMutation.graphql';
+
+function CustomerDetail() {
+  const [name, setName] = React.useState('');
+
+  let resultBanner= null;
+  const [createCustomer, {loading}] = useMutation(createCustomerMutation, {
```
tleunen: Could we use a promise instead if these `onCompleted` and `onError` callbacks?
```js
const createCustomer = useMutation(createCustomerMutation, options);

try {
  const data = await createCustomer({ variables: { ... } });
} catch{
}
```

packages/react-graphql/src/hooks/useQuery.ts
```diff
@@ -0,0 +1,165 @@
...
+    errorPolicy,
+  } = options;
+
+  const [responseId, setResponseId] = useState(0);
+  const client = useApolloClient(overrideClient);
+  const query = useGraphQLDocument(queryOrComponent);
+
+  const watchQueryOptions: WatchQueryOptions<Variables> = useMemo(
+    () =>
+      compact({
```
utkarshsaxena93: Why do we want to call `compact`? From what I understand from the implementation of `compact`, it removes keys that are `undefined`

packages/react-testing/README.md
utkarshsaxena93: what is this fixing?

packages/react-apollo-hooks/src/ApolloContext.tsx
```diff
@@ -0,0 +1,43 @@
+import ApolloClient from 'apollo-client';
+import React, {ReactElement, ReactNode, useContext} from 'react';
+
+const ApolloContext = React.createContext<null | ApolloClient<any>>(null);
+
+export interface ApolloProviderProps<TCacheShape> {
```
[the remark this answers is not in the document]
utkarshsaxena93: > Don't let your examples show through

What does this mean?

packages/react-apollo-hooks/src/useGraphQLDocument.ts
```diff
@@ -0,0 +1,52 @@
+import {useState, useEffect} from 'react';
```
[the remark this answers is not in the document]
utkarshsaxena93: > we prefer hooks to be nested under a hooks directory

Are you referring to all the custom hooks being nested under a hooks directory?

packages/react-apollo-hooks/src/useQuery.ts
```diff
@@ -0,0 +1,202 @@
...
+      }
+
+      const invalidateCurrentResult = () => {
+        // A hack to get rid React warnings during tests. The default
+        // implementation of `actHack` just invokes the callback immediately.
+        // In tests, it's replaced with `act` from react-testing-library.
+        // A better solution welcome.
+        actHack(() => {
+          setResponseId(x => x + 1);
+        });
```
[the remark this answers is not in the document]
utkarshsaxena93: Can you tell me more about what you mean here and why you are suggesting using `useRef` for setting response ID? Is it because we want to value to be shared and incremented across all instances of `useQuery`? Since it is a state, it will start from `0` for every instance and defeat the purpose of `setResponseId` being unique. Am I understanding this correctly?

packages/react-apollo-hooks/src/useQuery.ts
utkarshsaxena93: What's the rationale behind that?

lemonmade: To get consistency in using either `null` or `undefined` to mean "not present yet"

packages/react-testing/README.md
michelleyschen: that `graphQL` is actually part of the `context` object, not `root`
the example won't work basically.

packages/react-graphql/src/hooks/README.md
michelleyschen: so there are two separate strategy here

1. give you back the promise to handle yourself => the old HoC way
2. allow the use of `onCompleted` & `onError` so I can take care of the promise for you => the `<Mutation />` way

IMO, I like 2, because the amount of boiler plate for the try catch around mutation.

I am however, curious in what scenario you would prefer to handle this yourself instead of just using the callback?

packages/react-graphql/src/hooks/tests/useMutation.test.tsx
```diff
@@ -0,0 +1,197 @@
...
+  beforeEach(() => {
+    prepareAsyncReactTasks();
+  });
+
+  afterEach(() => {
+    teardownAsyncReactTasks();
+  });
+
+  describe('default', () => {
+    const MockMutation = ({children}) => {
```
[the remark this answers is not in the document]
michelleyschen: It is actually really difficult to test `useMutation` by itself because all the react hook doesn't work outside of a component. I do believe we have a doc that talks about it more.

https://github.com/Shopify/web-foundation/blob/ca2d34ff491eca187f9627a261ee0e784819854c/Best%20practices/React/Testing.md#hooks

packages/react-graphql/src/hooks/useGraphQLDocument.ts
```diff
@@ -0,0 +1,48 @@
...
+      } catch (error) {
+        throw Error('error loading GraphQL document');
+      }
+    }
+  }
+
+  useEffect(
+    () => {
+      if (document === null) {
+        loadDocument();
```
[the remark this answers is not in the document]
michelleyschen: Because `loadDocument` is async, it actually cannot be a callback to `useEffect`.
Since I HAVE to create a separate method for it, it just feel better if the method is not indented in this giant block.

Thought I am now wondering if my variable would be different because of this...mm.... 🤔

packages/react-graphql/src/hooks/useGraphQLDocument.ts
```diff
@@ -0,0 +1,48 @@
...
+      }
+    },
+    [documentOrComponent],
```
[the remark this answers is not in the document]
michelleyschen: `setDocument` is a setter so assuming that won't change.
I could include `document` but I also don't see a scenario where I would want to reload the document because `query` changed.

packages/react-graphql/src/hooks/useMutation.tsx
```diff
@@ -0,0 +1,101 @@
...
+      if (!onError) throw apolloError;
+    }
+  }
+
+  const results = {
+    ...state,
+    client,
+  };
+
+  return [runMutation, results];
```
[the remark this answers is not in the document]
michelleyschen: those are really good questions and I don't have good answers 🤔 

tbh I am also waiting for Chris to tell me I am doing these ALL WRONG

packages/react-graphql/src/hooks/useGraphQLDocument.ts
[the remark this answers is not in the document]
utkarshsaxena93: >  it actually cannot be a callback

True. It can still be inlined on line 37 though, no?

>  if my scope would be different and cause problems down

Should be fine IMO

packages/react-graphql/src/hooks/useQuery.ts
```diff
@@ -0,0 +1,165 @@
...
+  );
+
+  useEffect(
+    () => {
+      if (skip || !observableQuery) {
+        return;
+      }
+
+      const invalidateCurrentResult = () => {
+        // A hack to get rid React warnings during tests. The default
```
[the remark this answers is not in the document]
michelleyschen: as @lemonmade pointed out earlier, he want me to get rid of these.

thou testing is still quite difficult because of this type of hack we needed to do.
https://github.com/facebook/react/issues/14769

packages/react-graphql/src/hooks/README.md
utkarshsaxena93: My preference will be `give you back the promise to handle yourself`.

It makes the hook/component more flexible/robust by delegating error handling to the consumer.

packages/react-graphql/src/hooks/useGraphQLDocument.ts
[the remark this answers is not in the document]
utkarshsaxena93: > setDocument is a setter so assuming that won't change

True, however, I was under the assumption is it a good practice all dependencies in cases like these. c/c @lemonmade 

Same goes for `loadDocument`. Thoughts?

I can see though how we can think of this as `componentDidMount` where we want to run this effect only on a particular change (which, in this case, is `documentOrComponent` )

packages/react-graphql/src/hooks/useMutation.tsx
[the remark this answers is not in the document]
utkarshsaxena93: lol c/c @lemonmade

packages/react-graphql/src/hooks/useQuery.ts
michelleyschen: ¯\_(ツ)_/¯ 

this is part of the few things I copied from `react-apollo-hooks`, I don't completely understand it either but am also too scare to have it remove

- compact
- objectToKey
- actHack
- setResponseId
^^ all of the things I still need to fully understand / remove before I put this PR up for official review.

michelleyschen:
Reminder to self: update all the file names from `useX` to `x-y-x`
don't want to do this till the end otherwise I am going to lose all the comments again!

packages/react-graphql/src/hooks/README.md
michelleyschen: ok, spoke with @lemonmade to offer just the mutation to try yourself.
I still think its nice to have a way of handling all mutation error the same way, but something that's better done as additional hooks in web only.

michelleyschen:
Todo:

- [ ] move any changes other than `react-graqphql` to separate PR
- [ ] remove `onComplete`, `onError` and only return the trigger function in `useMutation`
- [ ] also move `useMutation` to a separate PR, likely can be merge in earlier than `useQuery`
- [ ] document the web-proving ground 🎩 method in above & current PR

- [ ] update to `react-dom v16.9.0-alpha.0` to use `async act` in test. Or skip the test in question.
- [ ] remove any `QueryHookResult` not use in web
- [ ] read into `setResponseId` in `useQuery` and how to do those better? maybe?

packages/react-apollo-hooks/src/useQuery.ts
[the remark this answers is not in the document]
michelleyschen: talked to Chris IRL to resolved this.

Basically, this is a mechanism to ensure if a few query with the same options are running back to back, the result that's being return is actually the one being called. 

Since currentResult is being memoize, it won't change unless the dependency change, which mean it won't get change if you call two exactly the same query twice (without responseId)

There are other ways to accomplish this without it being as confusing, (ie. store `currentResult` in state instead), I am still weighting the options.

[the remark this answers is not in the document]
utkarshsaxena93: Not sure if this is relevant:
In the return of the `useEffect`, we unsubscribe. I wonder if it is possible to cancel the pending request (as part of this unsubscribe) if a particular dependency changes (maybe request id, maybe something else)

In any case, thanks for the context. :)

packages/react-apollo-hooks/src/ApolloContext.tsx
```diff
@@ -0,0 +1,43 @@
+import ApolloClient from 'apollo-client';
+import React, {ReactElement, ReactNode, useContext} from 'react';
```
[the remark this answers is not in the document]
tleunen: What is our style? What's wrong doing that?
Using `React.X` all the time? Why?

packages/react-graphql/src/hooks/README.md
tleunen: @michelleyschen - Maybe I'm biased because I've been using apollo-react-hooks for the past couple weeks and that's how mutations are done with this package. But I feel it's a better API overall.

I hate callbacks in general. It hides the execution flow compared to a "simpler" promise/async-await.

I agree that abstracting the error handling is better to keep consistency. How hard would it be to have something similar to the `useQuery`, with a response like `[createCustomer, {loading, errors, data}]`?

The main reason why I don't like having a `onCompleted` explicitly is because you'll have to manage errors in it too, since the `onError` will be specific to http/apollo errors if I'm not mistaken. User errors will have to live inside `onCompleted`, which means you'll have quite a bit of logic there compared to your example. 

I also think that the callback option has a bad API. Based on the example again, you'll have rendering which happen to be in a hook, so consumers will have rendering in several places in the component instead of having it grouped at the end of it (when you return the rendered component). And this will get even worse if the user create a custom hook which uses `useMutation`

packages/react-graphql/src/hooks/useGraphQLDocument.ts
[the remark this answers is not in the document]
tleunen: Btw. This is not really safe (as explained here: https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies).

The reason is because `loadDocument` uses `documentOrComponent`, which comes from the outside. What if this value changes?

packages/react-graphql/src/hooks/useGraphQLDocument.ts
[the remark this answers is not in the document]
tleunen: The `setter` function from `useState` is not required no.

packages/react-graphql/src/hooks/useGraphQLDocument.ts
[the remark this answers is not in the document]
utkarshsaxena93: When you say “this” is not safe. What is this referring to?

PR #643, opened 2019-04-10
Memoize DateTimeFormat to speed up format date

jim-chou-shopify:
**Problem:**
Components that uses date utilities feel a bit slow on marketing section of the web repo.
Example: [StartEndDates](https://github.com/Shopify/web/blob/master/app/components/StartEndDates/StartEndDates.tsx).

Profiling performance on chrome showed following for `applyTimeZoneOffset`:
The bottleneck appears to be [formatDate](https://github.com/Shopify/quilt/blob/36a437d3da67cdb3e930b5298413954b8b1a9055/packages/dates/src/utilities/formatDate.ts#L10) where `Intl.DateTimeFormat` creates new object.

Screen Shot 2019-04-09 at 8 27 49 PM
Screen Shot 2019-04-09 at 8 27 58 PM
Screen Shot 2019-04-09 at 8 28 20 PM
Screen Shot 2019-04-09 at 8 31 02 PM
Screen Shot 2019-04-09 at 8 31 08 PM

**Solution**
Memoize the creation of new `DateTimeFormat` object by `locales` and `options`.


packages/dates/src/utilities/formatDate.ts
```diff
@@ -1,16 +1,38 @@
...
 }
+
+function dateTimeFormatCacheKey(
+  locales: string | string[],
+  options?: Intl.DateTimeFormatOptions,
+) {
+  return `${locales}-${JSON.stringify(options)}`;
+}
+
+const memoizedGetDateTimeFormat = memoize(
```
macdonaldr93: Should this `const` be defined before it's used?

jim-chou-shopify: It works but harder for reading.
I will update.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Functions#Function_scope
> a function can access all variables and functions defined inside the scope in which it is defined. In other words, a function defined in the global scope can access all variables defined in the global scope.

packages/dates/CHANGELOG.md
```diff
@@ -6,7 +6,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 <!-- Unreleased changes should go to UNRELEASED.md -->
 
----
+## Memoize `Intl.DateTimeFormat` [#643](https://github.com/Shopify/quilt/pull/643)
```
darrenhebner: I don't think this would be a header?

I think the header would either be `Added`, `Changed`, or `Fixed`. Hard to tell in this case. https://keepachangelog.com/en/1.0.0/#how

jim-chou-shopify: Maybe use `Changed formatDate to use memoized getDateTimeFormat()`.

jim-chou-shopify: Added to `UNRELEASED.md`
https://github.com/Shopify/quilt/pull/643/files#diff-ffa836aa672cf05f1ca30747cd245f92

packages/dates/src/utilities/formatDate.ts
```diff
@@ -1,16 +1,38 @@
+import memoize from 'lodash/memoize';
```
lemonmade: No Lodash please, we have a @shopify/function-enhancers package to use instead

jim-chou-shopify: I run into a problem with `quilt/packages/function-enhancers/src/memoize.ts` where it expect the enhanced function and resolver to return the same type.
The resolver returns string and the enhanced function returns Object.

I am thinking of returning string because same option may not always be same reference.

https://github.com/Shopify/quilt/pull/643/files#diff-36b8ddd38f1ae318116959216b4a4c2bR28

jim-chou-shopify: @michelleyschen Is it possible to return different type between resolver and enhanced function?

michelleyschen: It was intended to allow this but I might have typed it wrong. 

You can updated these two line to restrict the argument of the resolver but not the return value.
https://github.com/Shopify/quilt/blob/master/packages/function-enhancers/src/memoize.ts#L7
https://github.com/Shopify/quilt/blob/use-mounted-ref/packages/decorators/src/memoize.ts#L3

michelleyschen: I will try to see if I have time to fix it today actually

jim-chou-shopify: Thanks!

michelleyschen: https://github.com/Shopify/quilt/pull/685

PR #645, opened 2019-04-10
[React-i18n] Allow translate function to return object

nimzco:
Allow translate function to return object. For example, if I have:

```json
{
  "selectFieldOptions": {
    "volume": {
      "label": "Volume",
      "units": {
        "ml": "ml",
        "cl": "cl",
        "L": "L",
        "m3": "m3"
      }
    }
  }
}
```
```tsx
i18n.translate("selectFieldOptions.volume.units")
// => { "ml": "ml", "cl": "cl", "L": "L", "m3": "m3" }
```

This makes the behavior of the translate function to be similar to the one we have in Rails.

We could use it for countries as well for example, see: https://github.com/Shopify/web/blob/master/app/locales/en.json#L2784

@adjnor has a similar use case on a feature he is building currently.

### DIFF
--- a/packages/react-i18n/src/i18n.ts
+++ b/packages/react-i18n/src/i18n.ts
@@ -26,6 +26,7 @@ import {
 import {
   getCurrencySymbol,
   translate,
+  getTranslationTree,
   TranslateOptions as RootTranslateOptions,
 } from './utilities';
 
@@ -161,6 +162,15 @@ export class I18n {
 
     try {
       return translate(id, normalizedOptions, this.translations, this.locale);
+    } catch (error) {
+      this.onError(error);
+      return '';
+    }
+  }
+
+  getTranslationTree(id: string): string | object {
+    try {
+      return getTranslationTree(id, this.translations);
     } catch (error) {
       this.onError(error);
       return '';
--- a/packages/react-i18n/src/test/i18n.test.ts
+++ b/packages/react-i18n/src/test/i18n.test.ts
@@ -9,10 +9,13 @@ import {MissingTranslationError} from '../errors';
 
 jest.mock('../utilities', () => ({
   translate: jest.fn(),
+  getTranslationTree: jest.fn(),
   getCurrencySymbol: jest.fn(),
 }));
 
 const translate: jest.Mock = require('../utilities').translate;
+const getTranslationTree: jest.Mock = require('../utilities')
+  .getTranslationTree;
 const getCurrencySymbol: jest.Mock = require('../utilities').getCurrencySymbol;
 
 describe('I18n', () => {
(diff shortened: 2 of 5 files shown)


lemonmade:
Hmm I am not sure I am super happy with this because it is very bad for type safety. It ruins the one small guarantee we could give (if you passed no replacements, or replacements where all values are strings, we will always return a string; otherwise, we return JSX). Given that this is a fairly rare use case (presumably this would only be needed once for every "list type" in an app), I'd rather people use the locale and do a dynamic import for a JSON file based on that instead, unless there is some version of this use case that absolutely 100% requires it to be synchronous.

nimzco:
Hmm, that's a fair argument. We can definitely do things differently, we don't absolutely need this change. It's more that we had a convo IRL with @adjnor and wanted to have extra pair of eyes.

That said, I'm not sure if checking the type to know if it's JSX or string is a good thing to do neither though. 🤔

adjnor:
In the example provided (and I suppose in many cases) `i18n.translate` would end up being used within a component's render method so it would need to be synchronous. If not, doing a dynamic import in `componentDidMount` works but it's not as straight forward. I'm probably lacking context on where and how `i18n.translate` is used but for what it's worth, the behaviour I expected was the one introduced by this PR rather than it failing to find translations

lemonmade:
Maybe @TheMallen could chime in on this one, my feeling on this is that stronger type checking is more in line with our general principles than allowing this behaviour. I think allowing this might actually break many existing type checks, since everywhere currently assumes either string or JSX coming from the method.

We have lots of tools to help with async imports, and it could conceivably be synchronous if the data is small enough.

TheMallen:
IMO there's no reason we couldn't have a second function for returning a subtree like this and type it with a broader type. That way you can generally use the safer function but also opt into the less-safe version for specific cases where it makes sense.

TheMallen:
We could even call it `i18n.getUnsafeTranslationTree('key')`

lemonmade:
Yeah, I'd be more on board if this was a separate method, and it would check and only return an object (no replacements, can't be a "leaf" key, etc)

TheMallen:
It also makes it really readable 
```
i18n.translate('product.helpText')
=> 'some text what helps'
i18n.translate('products.variantsCard')
=> { title: 'Variants', /*whatever*/}
```
vs
```
i18n.translate('product.helpText')
=> 'some text what helps'
i18n.getTranslationTree('product.variantsCard')
=> { title: 'Variants', /*whatever*/}
```
seems a bit clearer for what to expect (having not really used the rails one)

nimzco:
Thanks for your feedbacks. I've updated the code, let me know if that's what you have in mind 😄

nimzco:
🎉

PR #648, opened 2019-04-11
Eslint updates

cartogram:
This PR updates `eslint` from `5.0.0` to `^5.15.0` and `eslint-plugin-shopify` from `^23.1.0` to `27.0.1`. This latter was 4 major releases so there were a lot of new rules, and the big things were all the new @typescript-eslint and react-hooks stuff.


packages/dates/src/get-date-time-parts.ts
```diff
@@ -57,7 +57,7 @@ function getWeekdayValue(weekday: string) {
   return weekdays[weekday];
 }
 
-// eslint-disable-next-line shopify/no-fully-static-classes
+// eslint-disable-next-line shopify/no-fully-static-classes, @typescript-eslint/no-extraneous-class
```
cartogram: I think we should just 🔥 our rule and go with the typescript one. It checks more than just fully static classes but also classes that are just a constructor or completely empty.

packages/enzyme-utilities/src/test/fixtures/Toggle.tsx
```diff
@@ -30,7 +30,7 @@ export function Toggle({onToggle, deferred}: Props) {
       onToggle();
       return Promise.resolve();
     },
-    [setActive],
+    [active, deferred, onToggle],
```
cartogram: second pair of  👀 on this would be good.

packages/graphql-testing/src/test/createGraphQLFactory.test.tsx
```diff
@@ -66,9 +69,9 @@ describe('graphql-testing', () => {
 
     await graphQL.resolveAll();
     somePage.update();
-
     expect(
       somePage.containsMatchingElement(
+        // eslint-disable-next-line shopify/jsx-no-hardcoded-content
```
cartogram: couldn't get this one to pass /shrug

packages/jest-koa-mocks/src/create-mock-context/test/create-mock-context.test.ts
```diff
@@ -90,15 +90,6 @@ describe('create-mock-context', () => {
     expect(protocol).toBe('https');
   });
 
-  it('determines protocol based on `encrypted`', () => {
```
cartogram: this was a duplicate test

packages/jest-mock-router/src/test/index.test.tsx
```diff
@@ -11,16 +11,6 @@ function TestPage({location}: WithRouterProps) {
 const SomePage = withRouter(TestPage) as React.ComponentType;
 
 describe('jest-mock-router', () => {
-  describe('createRouter', () => {
```
cartogram: Decided this test was not worth keeping.

packages/koa-shopify-auth/src/auth/test/oauth-callback.test.ts
```diff
@@ -118,24 +118,7 @@ describe('OAuthCallback', () => {
 
     oAuthCallback(ctx);
 
-    expect(ctx.throw).toBeCalledWith(403, Error.NonceMatchFailed);
-  });
-
-  it('throws a 403 if the nonce query param does not match the cookie', () => {
```
cartogram: Another duplicate test

packages/koa-shopify-webhooks/src/test/register.test.ts
```diff
@@ -81,19 +81,6 @@ describe('registerWebhook', () => {
     expect(registerResponse.result.data).toMatchObject({});
   });
 
-  it('returns a result with success set to true when the server returns a webhookSubscription field', async () => {
```
cartogram: Another duplicate, not sure why there are so many of these

packages/react-graphql/src/tests/async.test.tsx
```diff
@@ -154,6 +156,7 @@ describe('createAsyncQueryComponent()', () => {
         <AsyncQueryComponent.KeepFresh pollInterval={pollInterval} />,
       );
 
+      // eslint-disable-next-line jest/prefer-strict-equal
       expect(trigger(keepFresh.find(Async), 'render', query)).toEqual(
```
cartogram: The only way I could get these to pass was by stringifying the 2 values, I figured just using `toEqual` was safe enough.

packages/react-html/src/hooks.ts
```diff
@@ -58,7 +58,7 @@ export function useClientDomEffect(
     () => {
       perform(manager);
     },
-    [manager, ...inputs],
```
cartogram: the react-hooks rule does not like this spread.

packages/react-i18n/src/e2e/server.test.tsx
```diff
@@ -55,21 +55,21 @@ function WithAsyncI18nComponent({children}: {children?: React.ReactNode}) {
...
+    </>
   );
 }
 
 function WithoutOwnI18nComponent({children}: {children?: React.ReactNode}) {
   const [i18n] = useI18n();
 
   return (
-    <div>
+    <>
```
cartogram: @lemonmade do we need these in a div for any reason?

packages/react-i18n/src/hooks.tsx
```diff
@@ -24,7 +24,7 @@ export function useI18n({
   const parentIds = parentI18n ? parentI18n.ids || [] : [];
   const ids = React.useMemo(() => (id ? [id, ...parentIds] : parentIds), [
     id,
-    ...parentIds,
+    parentIds,
```
cartogram: I think this is okay.

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "1.0.0",
   "description": "Shopify's javascript application utilities",
   "engines": {
-    "node": ">=8.10.0",
+    "node": ">=10.15.3",
```
lemonmade: don't do this, it will break shipit :(

packages/dates/src/get-date-time-parts.ts
lemonmade: For the TypeScript config, sure, it's still good for the JS config.

packages/graphql-testing/src/test/createGraphQLFactory.test.tsx
TheMallen: We probably never actually care about this rule in tests anyway /shrug

lemonmade: We should disable this rule for the whole repo, or at least for tests, it doesn't make sense in our context.

packages/jest-dom-mocks/src/location.ts
```diff
@@ -24,9 +24,9 @@ export default class Location {
 
     this.originalAssign = window.location.assign;
     this.originalReload = window.location.reload;
-    window.location.assign = jest.fn();
-    window.location.reload = jest.fn();
-    window.location.replace = jest.fn();
+    jest.spyOn(window.location, 'assign');
+    jest.spyOn(window.location, 'reload');
+    jest.spyOn(window.location, 'replace');
```
lemonmade: If we are doing it this way we don't need the assignment of the original versions, we can just restore after we are done

packages/jest-koa-mocks/package.json
```diff
@@ -4,6 +4,9 @@
   "license": "MIT",
   "description": "",
   "main": "dist/index.js",
+  "engines": {
+    "node": ">=10.15.3"
+  },
```
lemonmade: Same as above, please avoid doing this, shipit is stuck on an 8.x version of Node :/

packages/react-async/src/tests/component.test.tsx
```diff
@@ -156,8 +158,9 @@ describe('createAsyncComponent()', () => {
       const renderKeepFresh = () => <KeepFresh />;
       const AsyncComponent = createAsyncComponent({load, renderKeepFresh});
 
-      // Enzyme doesn't know how to handle components that only return fragments.
       const keepFresh = mount(
+        // Enzyme doesn't know how to handle components that only return fragments.
+        // eslint-disable-next-line shopify/jsx-prefer-fragment-wrappers
```
TheMallen: I'd just disable this rule for the whole file, I don't think it adds any value for tests.

test/typescript-version.test.ts
```diff
@@ -31,9 +32,10 @@ describe('typescript version', () => {
   });
 
   it('the version in the plop file matches the root version', () => {
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
     const rootPackageJSON = require('../package.json');
```
TheMallen: Also not sure this rule is useful for tests, does it get mad about `require.requireActual` / `require.requireMock`?

packages/react-html/src/hooks.ts
lemonmade: Do you know if there is an issue for this on the react repo?

packages/react-i18n/src/constants/index.ts
```diff
@@ -28,7 +28,7 @@ export const dateStyle = {
   },
 };
 
-export enum Weekdays {
+export enum Weekday {
```
lemonmade: I think this will technically be a breaking change (probably fine not to treat it as breaking since we just did 1.0 and we can sneak this in as a patch to that :P

lemonmade: But make sure to reference it on the changelog

packages/react-i18n/src/hooks.tsx
lemonmade: No, please keep the spread. This is not equivalent and means it will recompute this when there is no parentI18n.

packages/react-shopify-app-route-propagator/src/index.ts
```diff
@@ -38,7 +29,20 @@ export default class ReactShopifyAppRoutePropagator extends React.PureComponent<
     this.propagateLocation();
   }
 
-  propagateLocation() {
+  render() {
+    return null;
+  }
+
+  private get history(): History.History {
```
lemonmade: Is this actually where our linting rule wants getters to go? Seems bleck to me

cartogram:
@lemonmade this is ready for another look.

PR #649, opened 2019-04-12
small refactor and document update to support apollo-hook PR coming up

michelleyschen:
This is to prepare for https://github.com/Shopify/quilt/pull/642/files

Mostly I would like to publish the `react-async` changes, so I can start 🎩 `react-graphql`.
Right now I can't 🎩 properly since I changed two separate packages 😬

### DIFF
--- a/packages/react-async/src/utilities.ts
+++ b/packages/react-async/src/utilities.ts
@@ -1,0 +1,19 @@
+import {Import} from '@shopify/async';
+
+export function normalize(module: any) {
+  if (module == null) {
+    return null;
+  }
+
+  const value = 'default' in module ? module.default : module;
+  return value == null ? null : value;
+}
+
+export async function resolve<T>(load: () => Promise<Import<T>>): Promise<T> {
+  try {
+    const resolved = await load();
+    return normalize(resolved);
+  } catch (error) {
+    throw error;
+  }
+}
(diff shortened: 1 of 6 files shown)


packages/react-testing/src/mount.ts
```diff
@@ -78,7 +78,7 @@ type CustomMountResult<
   ? Promise<CustomRoot<Props, Context>>
   : CustomRoot<Props, Context>;
 
-class CustomRoot<Props, Context extends object> extends Root<Props> {
+export class CustomRoot<Props, Context extends object> extends Root<Props> {
```
michelleyschen: if I don't export this I get the following error

```
Exported variable 'mountWithGraphQL' has or is using name 'CustomRoot' from external module "Shopify/quilt/packages/react-testing/src/mount" but cannot be named.
```
Reason why
https://github.com/Microsoft/TypeScript/issues/5711

packages/react-testing/README.md
```diff
@@ -387,7 +387,7 @@ export const mountWithGraphQL = createMount<Options, Context, true>({
 
     // Here's the important bit: resolve the GraphQL so our first queries are
     // in use for the component under test
-    await root.graphQL.resolveAll();
+    await root.context.graphQL.resolveAll();
```
michelleyschen: that `graphQL` is actually part of the `context` object, not `root`

packages/react-async/src/utilities.ts
```diff
@@ -0,0 +1,19 @@
...
+  const value = 'default' in module ? module.default : module;
+  return value == null ? null : value;
+}
+
+export async function resolve<T>(load: () => Promise<Import<T>>): Promise<T> {
+  try {
+    const resolved = await load();
+    return normalize(resolved);
+  } catch (error) {
+    throw Error('error loading module');
```
lemonmade: Should we bother with this or just let the original error through?

michelleyschen: ya I am fine with that too

PR #650, opened 2019-04-12
Adds useShortcut to Readme

cartogram:
Does what it says in the title. I forgot to do this when I added the hook initially.

### DIFF
--- a/packages/react-shortcuts/README.md
+++ b/packages/react-shortcuts/README.md
@@ -163,6 +163,43 @@ class MyComponent extends React.Component {
 }
 ```
 
+### useShortcut
+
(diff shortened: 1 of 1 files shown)


packages/react-shortcuts/README.md
```diff
@@ -163,6 +163,43 @@ class MyComponent extends React.Component {
 }
 ```
 
+### useShortcut
+
+`<Shortcut />` invokes a hook, `useShortcut()`, under the hood. This hook is also available for use from this package. It will also register a new keyboard shortcut to the `ShortcutManager` and the API is very similar.
```
TheMallen: I'm not sure it needs to mention that the component uses it.

PR #653, opened 2019-04-15
✨ add `useMutation`

michelleyschen:
Part of #586

I have skipped the test due to `act` not supporting async operation.
https://github.com/Shopify/quilt/pull/654
show the test passing if we update to react-dom v16.9.0-alpha.0

🎩 instructions:
- get this branch & [this test branch](https://github.com/Shopify/web-proving-ground/tree/apollo-mutation-hook) on web-proving-ground locally
- run `yarn` in web-proving-ground
- in quilt, do `yarn && yarn tophat react-graphql ../web-proving-ground`
- once it complied, go to `web-proving-ground` and do `yarn dev`
- goto localhost:8081 to see the mutation working properly on customer details page

### DIFF
--- a/packages/react-graphql/README.md
+++ b/packages/react-graphql/README.md
@@ -142,3 +142,78 @@ const MyQuery = createAsyncQueryComponent({
 <MyQuery.Prefetch async={{defer: undefined}} />
 <MyQuery.KeepFresh pollInterval={20_000} async={{defer: undefined}} />
 ```
+
+## Using Apollo Hooks
+
+Using Apollo Hooks assume the usage of [`react-apollo`](https://github.com/apollographql/react-apollo))
+
+### `ApolloProvider`
+
+Before using the individual hooks, you will need to wrap your application with `ApolloProvider` at root of your React component tree.
+
+You can it instead of `react-apollo`'s [`ApolloProvider`](https://www.apollographql.com/docs/react/api/react-apollo#ApolloProvider).
+
+```tsx
+import React from 'react';
+import {render} from 'react-dom';
+
+import ApolloClient from 'apollo-client';
+import {ApolloProvider} from 'react-graphql';
+
+const client = new ApolloClient();
+
+function App() {
+  return (
+    <ApolloProvider client={client}>
+      <MyRootComponent />
+    </ApolloProvider>
+  );
+}
+
+render(<App />, document.getElementById('root'));
+```
+
+### `useMutation`
+
+This hook accepts two arguments: the mutation document, and optionally, a set of options to pass to the underlying mutation. It will return a function that will trigger the mutation when invoked.
+
+Note the set of options can be pass directly into the hook, or pass in while triggering the mutation function.
+If options exist in both places, they will be shallowly merge together with per-mutate options being the priority.
+
+```tsx
+import React from 'react';
+import {Form, TextField, Button, Banner} from '@shopify/polaris';
+import {useQuery} from '@shopify/react-graphql';
+
+import createCustomerMutation from './graphql/CreateCustomerMutation.graphql';
+
+function CustomerDetail() {
+  const [name, setName] = React.useState('');
+  const createCustomer = useMutation(createCustomerMutation, {
+    fetchPolicy: 'network-only',
+  });
+
+  async function handleFormSubmit() {
+    try {
+      await createCustomer({
+        variables: {name},
+      });
+
+      // do something when the mutation is successful
+    } catch (error) {
+      // do something when the mutation fails
+    }
+  }
+
+  return (
+    <Form onSubmit={handleCreateCustomer}>
+      <TextField label="Name" value={name} onChange={(value) => {
+        setName(value);
+      }}>
+      <Button submit>
+        Create Customer
+      </Button>
+    </Form>
+  );
+}
+```
--- a/packages/react-graphql/package.json
+++ b/packages/react-graphql/package.json
@@ -39,6 +39,6 @@
     "dist/*"
   ],
   "peerDependencies": {
-    "react": "^16.6.0"
+    "react": ">=16.8.0 <17.0.0"
   }
 }
--- a/packages/react-graphql/src/hooks/mutation.ts
+++ b/packages/react-graphql/src/hooks/mutation.ts
@@ -1,0 +1,63 @@
+import {OperationVariables} from 'apollo-client';
+import {MutationOptions} from 'react-apollo';
+import {DocumentNode} from 'graphql-typed';
+import {useCallback} from 'react';
+
+import {MutationHookOptions, MutationHookResult} from './types';
+import {useApolloClient} from './apollo-client';
+
+export default function useMutation<Data = any, Variables = OperationVariables>(
+  mutation: DocumentNode<Data, Variables>,
+  options: MutationHookOptions<Data, Variables> = {},
+): MutationHookResult<Data, Variables> {
+  const {
+    client: overrideClient,
+    variables,
+    optimisticResponse,
+    refetchQueries,
+    awaitRefetchQueries,
+    update,
+    context,
+    fetchPolicy,
+  } = options;
+
+  const client = useApolloClient(overrideClient);
+
+  const runMutation = useCallback(
+    (perMutationOptions: MutationOptions<Data, Variables> = {}) => {
+      const mutateVariables = {
+        ...(variables || {}),
+        ...(perMutationOptions.variables || {}),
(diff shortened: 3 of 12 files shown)


packages/react-graphql/src/hooks/apollo-context.tsx
```diff
@@ -0,0 +1,43 @@
...
+  // constant, despite the Apollo client read from context being swapped for
+  // one passed directly as prop.
+  if (overrideClient) {
+    return overrideClient;
+  }
+
+  if (!client) {
+    // https://github.com/apollographql/react-apollo/blob/5cb63b3625ce5e4a3d3e4ba132eaec2a38ef5d90/src/component-utils.tsx#L19-L22
+    throw new Error(
+      'Could not find "client" in the context or passed in as a prop. ' +
```
cartogram: Prefer template literal or array.join() here over `string + string`

packages/react-graphql/src/hooks/tests/mutation.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+const mockMutationData = {
+  updatePet: {id: 'new-pet-id-123', __typename: 'Cat'},
+};
+
+// This is skip because current `act` wrapper does not support async operation
+// The test will pass if we update to `react-dom` v16.9.0-alpha.0
+// and wrap `graphQL.resolveAll()` in an `act`
+// https://github.com/facebook/react/issues/14769#issuecomment-481251431
+// eslint-disable-next-line jest/no-disabled-tests
+describe.skip('useMutation', () => {
```
cartogram: feels like you are missing tests for overrideClient, options and the per-mutation options, I guess they would all just be skipped as well.

packages/react-graphql/README.md
```diff
@@ -142,3 +142,8 @@ const MyQuery = createAsyncQueryComponent({
 <MyQuery.Prefetch async={{defer: undefined}} />
 <MyQuery.KeepFresh pollInterval={20_000} async={{defer: undefined}} />
 ```
+
+### Hooks
+
+Read about the set of customized React hooks [here](src/hooks/README.md)
```
cartogram: Just the one readme IMO

packages/react-graphql/src/hooks/README.md
```diff
@@ -0,0 +1,94 @@
+# Using Apollo Hooks
+
+## `ApolloProvider`
+
+Before using the individual hooks, you will need to wrap your application with `ApolloProvider` at root of your React component tree.
+
+You can replace `react-apollo`'s [`ApolloProvider`](https://www.apollographql.com/docs/react/api/react-apollo#ApolloProvider) if you are only going to use Hooks.
```
lemonmade: Should ours just render the "real" `ApolloProvider` so that you only need one no matter what?

packages/react-graphql/src/hooks/mutation.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export default function useMutation<Data = any, Variables = OperationVariables>(
+  mutation: DocumentNode<Data, Variables>,
+  options: MutationHookOptions<Data, Variables> = {},
+): MutationHookResult<Data, Variables> {
+  const {client: overrideClient, ...mutationOptions} = options;
+
+  const client = useApolloClient(overrideClient);
+
+  const runMutation = useMemo(
```
lemonmade: `useCallback` instead

packages/react-graphql/src/hooks/mutation.ts
```diff
@@ -0,0 +1,29 @@
...
+    () => (perMutationOptions?: MutationOptions<Data, Variables>) => {
+      return client.mutate({
+        mutation,
+        ...mutationOptions,
+        ...perMutationOptions,
+      });
```
lemonmade: Should we do a deep merge here instead of a shallow merge? I think things like `variables` are nested objects, do we think people will pass a subset of options?

packages/react-graphql/src/hooks/mutation.ts
```diff
@@ -0,0 +1,29 @@
...
+    },
+    [client, mutation, mutationOptions],
```
lemonmade: I am a bit worried about putting `mutationOptions` here, since it is pretty unlikely the consumer will know to memoize them. Maybe we should try to turn this into a set of primitives instead?

packages/react-graphql/src/hooks/tests/mutation.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+  mutation UpdatePetMutation($name: String!) {
+    updatePet(name: $name) {
+      id
+    }
+  }
+`;
+
+const createGraphQL = createGraphQLFactory();
+
+const MockMutation = ({children}: {children: Function}) => {
```
lemonmade: Function declarations :D

packages/react-graphql/src/hooks/mutation.ts
michelleyschen: funny enough, when I was playing with the example, I wanted to...

packages/react-graphql/src/hooks/mutation.ts
```diff
@@ -0,0 +1,29 @@
...
+export default function useMutation<Data = any, Variables = OperationVariables>(
+  mutation: DocumentNode<Data, Variables>,
+  options: MutationHookOptions<Data, Variables> = {},
+): MutationHookResult<Data, Variables> {
+  const {client: overrideClient, ...mutationOptions} = options;
+
+  const client = useApolloClient(overrideClient);
+
+  const runMutation = useMemo(
+    () => (perMutationOptions?: MutationOptions<Data, Variables>) => {
```
TheMallen: We don't mention `perMutationOptions` in the docs as far as I can tell, can we do so?

packages/react-graphql/src/hooks/README.md
```diff
@@ -0,0 +1,94 @@
...
+import {Form, TextField, Button, Banner} from '@shopify/polaris';
+import {useQuery} from '@shopify/react-graphql';
+
+import createCustomerMutation from './graphql/CreateCustomerMutation.graphql';
+
+function CustomerDetail() {
+  const [name, setName] = React.useState('');
+  const createCustomer = useMutation(createCustomerMutation);
+
+  async function handleFormSubmit({fields}: FormData<Fields>) {
```
TheMallen: I don't think this example actually works (`Fields` is not defined here right?) could we either get rid of the generic or set it up so that it will actually compile?

packages/react-graphql/src/hooks/apollo-context.tsx
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export function ApolloProvider<CacheShape = any>({
+  client,
+  children,
+}: ApolloProviderProps<CacheShape>): React.ReactElement<
+  ApolloProviderProps<CacheShape>
+> {
+  return (
+    <ApolloContext.Provider value={client}>{children}</ApolloContext.Provider>
```
TheMallen: If we are already making a custom component for this, can we render the actual one from apollo here? If not, then can we just export `ApolloContext` directly? I'd like to be more consistent with this and generally the full context object is the most useful export since it lets you do stuff with `useContext` as well as giving access to both the provider and consumer.

packages/react-graphql/src/hooks/mutation.ts
michelleyschen: I didn't do a full deep merge but I did carry over what apollo did, which is one level deeper for variables.

https://github.com/apollographql/react-apollo/blob/master/src/Mutation.tsx#L222-L223

packages/react-graphql/package.json
```diff
@@ -39,6 +38,7 @@
     "dist/*"
   ],
   "peerDependencies": {
-    "react": "^16.6.0"
+    "react": ">=16.8.0 <17.0.0",
+    "react-apollo": ">=2.2.3 <3.0.0"
```
michelleyschen: move this to `peerDependencies`, in both case of using `<Query />` and `useMutation`, you really need to had setup `ApolloProvider` for it to work.

michelleyschen:
I future proof a little bit with `react-apollo`'s [`ApolloContext`](https://github.com/apollographql/react-apollo/blob/master/src/ApolloContext.ts) and [`ApolloProvider`](https://github.com/apollographql/react-apollo/blob/master/src/ApolloProvider.tsx)
(`ApolloContext` exist in v2.5.3, but they did not export it for consumer to use)

PR #657, opened 2019-04-16
✨ react-form-hooks

TheMallen:
## Summary
A library for building out forms using React Hooks. If wondering why a new library instead of rebuilding `<FormState />` to use hooks under the hood, or adding hooks to that package, check out the FAQ included in this PR.

### Stuff to look critically at
- Hook dependencies are hard for something like this, any suggestions for better ways to do them would be appreciated.
- Docs are not entirely done. I originally wanted to export them via TSDoc but it was not super workable with the setup of this repo. We can fix that eventually, but for now I more or less duplicated between the README and the tsdoc comments :/
- Tests are incomplete. I fought with them a bit but overall we just need to wait for React to fix `act` with async functions and then get it updated in `@shopify/react-testing`.
- No special handling for adding / removing from a list. Easy enough to add as a feature later though.


packages/react-form-hooks/src/hooks/form.ts
```diff
@@ -0,0 +1,93 @@
...
+  const dirty = useDirty(fields);
+  const reset = useReset(fields);
+  const validate = useValidateAll(fields);
+  const [submit, submitting, submitErrors, setSubmitErrors] = useSubmit(
+    onSubmit,
+    fields,
+  );
+  useErrorPropagation(fields, submitErrors);
+
+  return {
```
TheMallen: possibly this should be memoized, I'm not sure it's super worth it though since it's not really meant to be passed onto a component directly and the object itself is not expensive to recompute

packages/react-form-hooks/src/hooks/list/list.ts
```diff
@@ -0,0 +1,264 @@
...
+    () =>
+      mapObject<NormalizedValidationDictionary<any>>(
+        validates,
+        normalizeValidation,
+      ),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [validates, ...validationDependencies],
+  );
+
+  const handlers = useMemo(
```
TheMallen: Originally this thing was based on `useField` in a loop but we can't actually do that because of how hooks work. As a result we have to do some weird stuff with building objects in a loop that call reducer actions.

packages/react-form-hooks/src/hooks/validateAll.ts
```diff
@@ -0,0 +1,36 @@
+import {useCallback} from 'react';
+import {FieldDictionary, FieldOutput, FormError, Field} from '../types';
+import {isField} from '../utilities';
+
+export function useValidateAll(fieldBag: {[key: string]: FieldOutput<any>}) {
```
TheMallen: I'm worried about performance of these traversal functions in very big forms, but I sort of think we solve that issue when it becomes a problem

packages/react-form-hooks/src/validation.ts
```diff
@@ -0,0 +1,47 @@
...
+  };
+}
+
+export function lengthMoreThan(length: number, error: ErrorContent<string>) {
+  return validate(input => {
+    return input.length > length;
+  })(error);
+}
+
+export function notEmpty(error: ErrorContent<string>) {
```
TheMallen: The API for `validate` is slightly different that form-state's. The new API let's you more easily create reusable validators, since there's no real purpose to using `validate` for one offs (it's almost always easier to just pass a function directly to the field).

```
// old way
validate(matcher, errorCreator);

// eg
const title = useField(() => {
  value: '',
  validates: validate((value) => value.includes('✨'), 'Title must include sparkles!')
});

// new way
const validator = validate(matcher)
const error = validator(errorCreator)

// eg
const hasSpakles = validate((value) => value.includes('✨'))

const title = useField({
  value: '',
  validates: hasSparkles("Titles must have sparkles"),
});
```

packages/react-form-hooks/src/validation/predicates.ts
```diff
@@ -0,0 +1,35 @@
+export function lengthMoreThan(length: number) {
```
TheMallen: These are exactly the same as the ones in `form-state`, and they aren't really specific to forms. I feel like we probably just want a `predicates` package in quilt that they both use. I'll make a PR for that too.

packages/react-form-hooks/src/hooks/errorPropagation.ts
```diff
@@ -0,0 +1,28 @@
+import {useEffect} from 'react';
+import get from 'get-value';
```
lemonmade: This tiny?

packages/react-form-hooks/src/hooks/errorPropagation.ts
```diff
@@ -0,0 +1,28 @@
...
+      remoteErrors.forEach(error => {
+        if (error.fieldPath == null) {
+          return;
+        }
+
+        const got = get(fieldBag, error.fieldPath);
+
+        if (isField(got)) {
+          got.setError(error.message);
+        }
```
lemonmade: Seems bad to silently throw away errors you can't associate to fields?

packages/react-form-hooks/src/hooks/field/field.ts
```diff
@@ -0,0 +1,196 @@
...
+      if (error && error.length > 0) {
+        const [firstError] = error;
+        dispatch(updateErrorAction(firstError));
+        return firstError;
+      }
+
+      dispatch(updateErrorAction(undefined));
+    },
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [state.value, ...dependencies],
```
lemonmade: Hmm, seems a bit strange to me we'd need to invalidate this on every state change.

packages/react-form-hooks/src/hooks/field/field.ts
```diff
@@ -0,0 +1,196 @@
...
+
+  const onBlur = useCallback(
+    () => {
+      if (state.touched === false && state.error == null) {
+        return;
+      }
+
+      runValidation();
+    },
+    [runValidation, state.touched, state.error],
```
lemonmade: Do you actually need to reference both fields on `state`, or would it be enough to just invalidate when `state` changes?

packages/react-form-hooks/src/hooks/field/field.ts
```diff
@@ -0,0 +1,196 @@
...
+  );
+
+  return field as Field<Value>;
+}
+
+function normalizeFieldConfig<Value>(
+  input: FieldConfig<Value> | Value,
+): FieldConfig<Value> {
+  if (input == null || typeof input !== 'object') {
+    return {value: input, validates: () => undefined};
```
lemonmade: Weird that `validates` isn't just optional?

packages/react-form-hooks/src/hooks/field/field.ts
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  return input as FieldConfig<Value>;
+}
```
lemonmade: This hook reads remarkably well for how much it does 👍

packages/react-form-hooks/README.md
```diff
@@ -0,0 +1,837 @@
+# `@shopify/react-form-hooks`
```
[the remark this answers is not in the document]
TheMallen: lol

packages/react-form-hooks/README.md
```diff
@@ -0,0 +1,837 @@
...
+    fields: {
+      title: useField('some default title'),
+    },
+    onSubmit: fieldValues => {
+      return {status: 'fail', errors: [{message: 'bad form data'}]};
+    },
+  });
+
+  return (
+    <form onSubmit={submit}>
```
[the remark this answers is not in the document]
TheMallen: nah this works, it cancels the event if there is one
https://github.com/Shopify/quilt/blob/use-form-hooks/packages/react-form-hooks/src/hooks/submit.ts#L26

packages/react-form-hooks/README.md
```diff
@@ -0,0 +1,837 @@
...
+  const fields = {title, description};
+
+  // track whether any field has been changed from it's initial values
+  const dirty = useDirty(fields);
+
+  // generate a reset callback
+  const reset = useReset(fields);
+
+  // handle submission state
+  const [submit, submitting, submitErrors, setSubmitErrors] = useSubmit(
```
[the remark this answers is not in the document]
TheMallen: kee

packages/react-form-hooks/README.md
```diff
@@ -0,0 +1,837 @@
...
+
+A custom hook for handling the state and validations of an input field.
+
+##### Signature
+
+```tsx
+const field = useField(config, validationDependencies);
+```
+
+`config` - The default value of the input, or a configuration object specifying both the value and validation config.
```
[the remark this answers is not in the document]
TheMallen: I was mimicking the output from the TSDoc so it was easier to keep them in synch, but yeah a list seems fine

packages/react-form-hooks/README.md
```diff
@@ -0,0 +1,837 @@
...
+              id={`title-${index}`}
+              name={`title-${index}`}
+              value={fields.title.value}
+              onChange={fields.title.onChange}
+              onBlur={fields.title.onBlur}
+            />
+          </label>
+          {field.title.error && <p>{field.title.error}</p>}
+          <label htmlFor={`description-${index}`}>
+            description{' '}
```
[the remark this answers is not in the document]
TheMallen: because it's technically correct, I guess it's ok to not bother with it for examples though

packages/react-form-hooks/README.md
```diff
@@ -0,0 +1,837 @@
...
+#### Examples
+
+`useForm` wraps one or more fields and return an object with all of the fields you need to manage a form.
+
+```tsx
+import React from 'react';
+import {useField, useForm} from '@shopify/react-form-hooks';
+
+function MyComponent() {
+  const {fields, submit, submitting, dirty, reset, submitErrors} = useForm({
```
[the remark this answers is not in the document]
TheMallen: banners

packages/react-form-hooks/src/hooks/errorPropagation.ts
TheMallen: V tiny
image

packages/react-form-hooks/src/hooks/errorPropagation.ts
TheMallen: We don't throw them away, they stay in `submitErrors` so they can still be displayed in the banner or whatever.

packages/react-form-hooks/src/hooks/field/field.ts
TheMallen: we could simplify this stuff if we made the signature always
```
(config: {value: Value, validates?: Validator<Value>}, dependencies?: unknown[]): Field<Value>;
```
but it would mean that for trivial fields you'd need to do `useField({value: 'foo'});` which is a bit annoying

packages/react-form-hooks/src/hooks/field/field.ts
TheMallen: I think this is better because changing `state.value` would change the whole object but we don't actually need regenerate this in that case

packages/react-form-hooks/README.md
```diff
@@ -0,0 +1,837 @@
...
+  const price = useField(
+    {
+      value: '0.00',
+      validates: value => {
+        if (title.value.includes('expensive') && parseFloat(value) < 1000) {
+          return 'Expensive items must cost more than 1000 dollars';
+        }
+      },
+    },
+    [title.value],
```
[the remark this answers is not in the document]
TheMallen: we don't care if the rest of it changes though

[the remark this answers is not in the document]
lemonmade: It just feels more awkward from the consumer perspective this way. What ever changes in a field object but leaves the value unchanged?

packages/react-form/README.md
```diff
@@ -0,0 +1,873 @@
...
+    title,
+    description,
+  });
+
+  return <form onSubmit={form.submit}>{children(form)}</form>;
+}
+```
+
+**Q: I want to dynamically change how many fields I render, how can I do that?**
+**A:** its tough to do this without breaking the rules of hooks. A possible solution is building a component encapsulating your hook calls, such as above, and then using a `key` to reset it when you need to. Another is 'faking' it by simply not rendering all of your fields despite always initializing the same number. Finally, you could build a custom reducer-based hook similar to `useList` that can support adding and removing fields dynamically. As long as it returns `Field` objects matching the API that the hooks in this library do, you will be able to use it with `useForm` and friends just fine.
```
cartogram: `its` => `It's`

packages/react-form/README.md
```diff
@@ -0,0 +1,873 @@
...
+  );
+}
+```
+
+##### Remarks
+
+- **Building your own:** Internally, `useForm` is a convenience wrapper over `useDirty`, `useReset`, `useValidateAll`, `useSubmit`, and `useErrorPropagation`.If you only need some of its functionality, consider building a custom hook combining a subset of them.
+- **Subforms:** You can have multiple `useForm`s wrapping different subsets of a group of fields. Using this you can submit subsections of the form independently and have all the error and dirty tracking logic "just work" together.
+
+####useSubmit
```
cartogram: You need a line break between these otherwise they end up on one unformatted line.

packages/react-form/README.md
```diff
@@ -0,0 +1,873 @@
...
+          </label>
+          {field.description.error && <p>{field.description.error}</p>}
+        </li>
+      ))}
+    </ul>
+  );
+}
+```
+
+If using `@shopify/polaris` or other custom components that support `onChange`, `onBlur`, `value`, and `error` props then you can accomplish the above more tersely by using the ES6 `spread` (...) operator.
```
cartogram: same comment about moving the backticks around the code part of the spread, not the word part.

packages/react-form/README.md
```diff
@@ -0,0 +1,873 @@
...
+}
+```
+
+## API
+
+This section details the individual functions exported by `@shopify-react-form`. For more detailed typing information see the `.d.ts` files.
+
+### Hooks
+
+#### `useFiel()d()`
```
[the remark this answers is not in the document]
TheMallen: wh()at()?

PR #659, opened 2019-04-17
react-i18n perf optimizations

lemonmade:
This PR addresses a few performance concerns, one of which was the cause of the issue we had in Web when we tried to roll these out:

1. Calling subscriptions when async translations are resolved now queues subscriptions to run in a single animation frame
2. When async translations are resolved to `undefined`, we no longer call the subscription (this is technically a bit dangerous because the `loading` property can get out of date, but we don't currently use this anywhere anyways, and this basically only happens for more detailed locales (e.g., `fr-CA`), where the consumer will get an update on the `fr` part resolving just fine
3. I removed a hacky `useSimpleI18n` hook and, instead, made `useI18n` bail out early when there are no options passed. This violates the rules of hooks, but we get around that by throwing an error if they try to switch between the two modes
4. I split apart the context for i18n objects and the one for the nested IDs in the parent. By having `useI18n` hooks linked up to the `I18nContext`, any update to one `i18n` object higher in the tree would cause all descendants' hooks to run again, which would be useless, because the IDs in question are still the same.
5. Made it so that the `ShareTranslations` component that gets returned is always the same function. This was the main cause of the incident we had: when the "highest" i18n-connected component had some translations that resolved async (even for locales like `en-CA`, which would do an async lookup for the `-CA` languages), its `ShareTranslations` would change. When that happened, the entire subtree below it got unmounted (this is how React's reconciliation works), then remounted. For a component like our `AdminNextFallback`, which sometimes makes non-idempotent get requests on mount, this could cause issues (like multiple GETs eventually causing full-page auth redirects).

With these in place, the performance of i18n on load is much better. Here's the before/ after for a shop in `en-CA`:

Screen Shot 2019-04-17 at 10 46 14 AM

Screen Shot 2019-04-17 at 10 52 15 AM

The `hydrate` bit is about the same length in both, So this basically shaves off 500–600ms.

### DIFF
--- a/packages/react-hooks/README.md
+++ b/packages/react-hooks/README.md
@@ -40,3 +40,19 @@ function MyComponent() {
   return <div>{foo}</div>;
 }
 ```
+
+### `useLazyRef()`
+
+This hook creates a ref object like React’s `useRef`, but instead of providing it the value directly, you provide a function that returns the value. The first time the hook is run, it will call the function and use the returned value as the initial `ref.current` value. Afterwards, the function is never invoked. You can use this for creating refs to values that are expensive to initialize.
+
+```tsx
+function MyComponent() {
+  const ref = useLazyRef(() => someExpensiveOperation());
+
+  React.useEffect(() => {
+    console.log('Initialized expensive ref', ref.current);
+  });
+
+  return null;
+}
+```
--- a/packages/react-i18n/src/hooks.tsx
+++ b/packages/react-i18n/src/hooks.tsx
@@ -25,11 +25,16 @@ export function useI18n(options?: Partial<RegisterOptions>): Result {
     );
   }
 
+  // Yes, this would usually be dangerous. But just above this line, we check to make
+  // sure that they never switch from the case where `options == null` to `options != null`,
+  // so we know that a given use of this hook will only ever hit one of these two cases.
+  /* eslint-disable react-hooks/rules-of-hooks */
   if (options == null) {
     return useSimpleI18n(manager);
   } else {
-    return useComplexI18n(options!, manager);
+    return useComplexI18n(options, manager);
   }
+  /* eslint-enable react-hooks/rules-of-hooks */
 }
 
 function useComplexI18n(
@@ -50,7 +55,7 @@ function useComplexI18n(
 
   const [i18n, setI18n] = React.useState(() => {
     const {translations} = manager.state(ids.current);
-    return new I18n(translations, manager.details, ids.current);
+    return new I18n(translations, manager.details);
   });
 
   const i18nRef = React.useRef(i18n);
@@ -58,12 +63,12 @@ function useComplexI18n(
   React.useEffect(
     () => {
       return manager.subscribe(ids.current, ({translations}, details) => {
-        const newI18n = new I18n(translations, details, ids.current);
+        const newI18n = new I18n(translations, details);
         i18nRef.current = newI18n;
         setI18n(newI18n);
       });
     },
-    [manager],
+    [ids, manager],
   );
 
   // We use refs in this component so that it never changes. If this component
@@ -74,7 +79,7 @@ function useComplexI18n(
   // this component to re-render, so no descendants will get the new ids/ i18n
   // value. Because we don't actually have any such cases, we're OK with this
   // for now.
-  const ShareTranslations = React.useMemo(
+  const shareTranslationsComponent = useLazyRef(
     () =>
       function ShareTranslations({children}: {children: React.ReactNode}) {
         return (
@@ -85,10 +90,9 @@ function useComplexI18n(
           </I18nIdsContext.Provider>
         );
       },
-    [i18nRef, ids],
   );
 
-  return [i18n, ShareTranslations];
+  return [i18n, shareTranslationsComponent.current];
 }
 
 function useSimpleI18n(manager: I18nManager): Result {
--- a/packages/react-i18n/src/hooks.tsx
+++ b/packages/react-i18n/src/hooks.tsx
@@ -1,13 +1,14 @@
 import * as React from 'react';
 
+import {useLazyRef} from '@shopify/react-hooks';
+
 import {I18n} from './i18n';
 import {I18nManager, RegisterOptions} from './manager';
 import {I18nContext, I18nIdsContext, I18nParentContext} from './context';
 
-export function useI18n(options?: Partial<RegisterOptions>): [
-  I18n,
-  React.ComponentType<{children: React.ReactNode}>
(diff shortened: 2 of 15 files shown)


packages/react-hooks/README.md
```diff
@@ -40,3 +40,19 @@ function MyComponent() {
   return <div>{foo}</div>;
 }
 ```
+
+### `useLazyRef()`
```
TheMallen: isn't this basically just `useMemo` with no dependencies?

packages/react-i18n/src/hooks.tsx
```diff
@@ -20,59 +17,94 @@ export function useI18n({
     );
   }
 
-  const parentI18n = React.useContext(I18nParentsContext);
-  const parentIds = parentI18n ? parentI18n.ids || [] : [];
-  // eslint-disable-next-line react-hooks/exhaustive-deps
-  const ids = React.useMemo(() => (id ? [id, ...parentIds] : parentIds), [
```
TheMallen: I feel like this would have been ok as just `useMemo(() => /*stuff*/, [])`

packages/react-i18n/src/hooks.tsx
```diff
@@ -20,59 +17,94 @@ export function useI18n({
...
+        const newI18n = new I18n(translations, details, ids.current);
+        i18nRef.current = newI18n;
+        setI18n(newI18n);
       });
     },
-    [ids, manager],
+    [manager],
   );
 
+  // We use refs in this component so that it never changes. If this component
```
TheMallen: Could we call this out in the docs as a thing that you should not do?

packages/react-i18n/src/manager.ts
```diff
@@ -192,11 +203,27 @@ export class I18nManager {
...
+    this.idsToUpdate.add(id);
+
+    if (this.enqueuedUpdate != null) {
+      return;
     }
+
+    const isBrowser = typeof window !== 'undefined';
+    const enqueue = isBrowser ? window.requestAnimationFrame : setImmediate;
+
+    this.enqueuedUpdate = enqueue(() => {
```
TheMallen: neat

packages/react-i18n/src/hooks.tsx
lemonmade: It could have, but for a value we know for sure that never changes it's safe to use a ref; `useMemo` is not a guarantee (I will do that for the component, too)

packages/react-hooks/README.md
lemonmade: `useMemo` is not guaranteed to actually be the same forever (they might ditch some stuff if there is memory pressure). This one is a guarantee

yarn.lock
```diff
@@ -572,14 +572,22 @@
...
-"@types/react@*", "@types/react@16.8.10", "@types/react@16.8.2", "@types/react@>=16.4.0", "@types/react@^16.0.2":
+"@types/react@*", "@types/react@>=16.4.0", "@types/react@^16.0.2":
   version "16.8.10"
   resolved "https://registry.yarnpkg.com/@types/react/-/react-16.8.10.tgz#1ccb6fde17f71a62ef055382ec68bdc379d4d8d9"
   integrity sha512-7bUQeZKP4XZH/aB4i7k1i5yuwymDu/hnLMhD9NjVZvQQH7ZUgRN3d6iu8YXzx4sN/tNr0bj8jgguk8hhObzGvA==
   dependencies:
     "@types/prop-types" "*"
     csstype "^2.2.0"
 
+"@types/react@16.8.2":
```
GoodForOneFare: My yarn seems to be producing different results to everyone else lately, but it's producing a diff without `@types/react` duplicates, which makes more sense to me:

```diff
--- a/yarn.lock
+++ b/yarn.lock
@@ -550,7 +550,7 @@
   resolved "https://registry.yarnpkg.com/@types/prop-types/-/prop-types-15.5.8.tgz#8ae4e0ea205fe95c3901a5a1df7f66495e3a56ce"
   integrity sha512-3AQoUxQcQtLHsK25wtTWIoIpgYjH3vSDroZOUr7PpCHw/jLY1RB9z9E8dBT/OSmwStVgkRNvdh+ZHNiomRieaw==

-"@types/react-dom@^16.0.11", "@types/react-dom@^16.8.3":
+"@types/react-dom@16.8.3", "@types/react-dom@^16.0.11", "@types/react-dom@^16.8.3":
   version "16.8.3"
   resolved "https://registry.yarnpkg.com/@types/react-dom/-/react-dom-16.8.3.tgz#6131b7b6158bc7ed1925a3374b88b7c00481f0cb"
   integrity sha512-HF5hD5YR3z9Mn6kXcW1VKe4AQ04ZlZj1EdLBae61hzQ3eEWWxMgNLUbIxeZp40BnSxqY1eAYLsH9QopQcxzScA==
@@ -572,7 +572,7 @@
     "@types/history" "^3"
     "@types/react" "*"

-"@types/react@*", "@types/react@>=16.4.0", "@types/react@^16.0.2":
+"@types/react@*", "@types/react@16.8.10", "@types/react@16.8.2", "@types/react@>=16.4.0", "@types/react@^16.0.2":
   version "16.8.10"
   resolved "https://registry.yarnpkg.com/@types/react/-/react-16.8.10.tgz#1ccb6fde17f71a62ef055382ec68bdc379d4d8d9"
   integrity sha512-7bUQeZKP4XZH/aB4i7k1i5yuwymDu/hnLMhD9NjVZvQQH7ZUgRN3d6iu8YXzx4sN/tNr0bj8jgguk8hhObzGvA==
@@ -580,14 +580,6 @@
     "@types/prop-types" "*"
     csstype "^2.2.0"

-"@types/react@16.8.2":
-  version "16.8.2"
-  resolved "https://registry.yarnpkg.com/@types/react/-/react-16.8.2.tgz#3b7a7f7ea89d1c7d68b00849fb5de839011c077b"
-  integrity sha512-6mcKsqlqkN9xADrwiUz2gm9Wg4iGnlVGciwBRYFQSMWG6MQjhOZ/AVnxn+6v8nslFgfYTV8fNdE6XwKu6va5PA==
-  dependencies:
-    "@types/prop-types" "*"
-    csstype "^2.2.0"
-
 "@types/safe-compare@^1.1.0":
   version "1.1.0"
   resolved "https://registry.yarnpkg.com/@types/safe-compare/-/safe-compare-1.1.0.tgz#47ed9b9ca51a3a791b431cd59b28f47fa9bf1224"
```

packages/react-hooks/README.md
TheMallen: 👍

PR #662, opened 2019-04-18
🎉 create @shopify/predicates

TheMallen:
This package pulls out all the predicates from `react-form-state` into their own package, and sets it up as our one stop shop for predicate functions. Once #657 and this are merged I will go back and make both form libraries depend on this.

### DIFF
--- a/packages/predicates/README.md
+++ b/packages/predicates/README.md
@@ -1,0 +1,31 @@
+# `@shopify/predicates`
+
+[![Build Status](https://travis-ci.org/Shopify/quilt.svg?branch=master)](https://travis-ci.org/Shopify/quilt)
+[![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](LICENSE.md) [![npm version](https://badge.fury.io/js/%40shopify%2Fpredicates.svg)](https://badge.fury.io/js/%40shopify%2Fpredicates.svg)
+
+A set of common Javascript predicates. The functions in this library either take one input and return `true` / `false`, or return a customized function that does so.
+
+## Installation
+
+```bash
+$ yarn add @shopify/predicates
+```
+
+## API
+
+### Predicates
+
+- `isPositiveNumericString` Returns `true` when its input is a positive numeric string and `false` otherwise.
+- `isNumericString` Returns `true` when its input is a numeric string and `false` otherwise.
+- `isEmpty` Returns `true` when its input is `null`, `undefined`, or has `length` 0.
+- `isEmptyString` Returns `true` when it's input is an empty string or contains only whitespace.
+- `notEmpty` Returns `true` when its input is not `null`, `undefined` or has `length` 0.
+
+### Predicate creators
+
+- `lengthMoreThan` Given a number, returns a function that returns true when that input has length more than that number and false otherwise.
+- `lengthLessThan` Given a number, returns a function that returns true when that input has length less than that number and false otherwise.
+
+### Helpers
+
+- `not` Given a function that returns a boolean, returns a function that returns a boolean in the opposite cases.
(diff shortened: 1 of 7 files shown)


packages/predicates/src/test/predicates.test.ts
```diff
@@ -0,0 +1,118 @@
+import {
+  lengthMoreThan,
+  lengthLessThan,
+  isPositiveNumericString,
+  isNumericString,
+  isEmpty,
+  isEmptyString,
+  notEmpty,
```
cartogram: These are not used

packages/predicates/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+- `isPositiveNumericString` Returns `true` when its input is a positive numeric string and `false` otherwise.
+- `isNumericString` Returns `true` when its input is a numeric string and `false` otherwise.
+- `isEmpty` Returns `true` when its input is `null`, `undefined`, or has `length` 0.
+- `isEmptyString` Returns `true` when it's input is an empty string or contains only whitespace.
+- `notEmpty` Returns `true` when its input is not `null`, `undefined` or has `length` 0.
+
+### Predicate creators
+
+- `lengthMoreThan` Returns a function which returns `true` when it's input has length more than the given number and `false` otherwise.
```
cartogram: Take it or leave it, but it _might_ be a bit more accurate to say:

`lengthMoreThan` Invoked with a number and returns a function that returns `true` when that input has length more than the number and `false` otherwise.

packages/predicates/README.md
```diff
@@ -0,0 +1,31 @@
...
+- `lengthLessThan` Returns a function which returns `true` when it's input has length less than the given number and `false` otherwise.
+
```
cartogram: Same as above

`lengthLessThan ` Invoked with a number and returns a function that returns `true` when that input has length less than the number and `false` otherwise.

packages/predicates/README.md
```diff
@@ -0,0 +1,31 @@
+# `@shopify/predicates`
+
+[![Build Status](https://travis-ci.org/Shopify/quilt.svg?branch=master)](https://travis-ci.org/Shopify/quilt)
+[![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](LICENSE.md) [![npm version](https://badge.fury.io/js/%40shopify%2Fpredicates.svg)](https://badge.fury.io/js/%40shopify%2Fpredicates.svg)
+
+A set of common predicates for JS/TS apps. Most of the functions in this library either take one input and return `true` / `false`, or return a customized function that does so.
```
cartogram: Remove `most of`

packages/predicates/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "name": "@shopify/predicates",
+  "version": "0.0.0",
+  "license": "MIT",
+  "description": "A set of common predicates for JS/TS apps",
```
cartogram: I'd drop the acronyms and the "apps". What about `A set of common Javascript predicates`?

packages/predicates/README.md
cartogram: Also if go with the description change I suggested, replace this first sentence.

packages/predicates/src/test/predicates.test.ts
```diff
@@ -0,0 +1,175 @@
...
+
+    expect(lengthLessThanTwo('hi')).toBe(false);
+    expect(lengthLessThanTwo({length: 2})).toBe(false);
+    expect(lengthLessThanTwo([1, 2])).toBe(false);
+  });
+
+  it('returns a function that returns false for inputs with length more than the given value', () => {
+    const lengthLessThanThree = lengthLessThan(3);
+
+    expect(lengthLessThanThree('I am so very cool')).toBe(false);
```
cartogram: 😎

PR #665, opened 2019-04-19
⬆️ Add 'unversioned' version to 'koa-shopify-graphql-proxy'

alexandcote:
Related to: https://github.com/Shopify/shopify/issues/196085

Add a new `unversioned` version in the ApiVersion enum of the `koa-shopify-graphql-proxy` package.

### DIFF
--- a/packages/koa-shopify-graphql-proxy/src/shopify-graphql-proxy.ts
+++ b/packages/koa-shopify-graphql-proxy/src/shopify-graphql-proxy.ts
@@ -8,6 +8,7 @@ export enum ApiVersion {
   April19 = '2019-04',
   July19 = '2019-07',
   Unstable = 'unstable',
+  Unversioned = 'unversioned',
 }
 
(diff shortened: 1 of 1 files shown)


packages/koa-shopify-graphql-proxy/src/shopify-graphql-proxy.ts
```diff
@@ -8,6 +8,7 @@ export enum ApiVersion {
   April19 = '2019-04',
   July19 = '2019-07',
   Unstable = 'unstable',
+  Unversioned = 'unversioned',
```
utkarshsaxena93: For my understanding, what is the difference between `Unstable` and `Unversioned` ?

Is it worth including comments to provide some context into what each means?

alexandcote: `Unversioned` is more stable than `Unstable` and will be only available to 1p app.

PR #675, opened 2019-04-26
Form hook tweaks

TheMallen:
This PR carries our a few tweaks to the hooks from `react-form` I noticed needed doing while building out some forms in web using the package. It also switches `react-form-state` and `react-form` to use `@shopify/predicates` instead of duplicating predicate logic.

The details of the changes to `react-form`'s hook behaviour are in `CHANGELOG.md`


packages/react-form/src/hooks/submit.ts
```diff
@@ -16,24 +16,21 @@ export function useSubmit<T extends FieldBag>(
   const [submitting, setSubmitting] = useState(false);
   const [remoteErrors, setRemoteErrors] = useState([] as FormError[]);
 
-  const submit = useCallback(
-    async (event?: React.FormEvent) => {
-      if (event && event.preventDefault && !event.defaultPrevented) {
-        event.preventDefault();
-      }
+  async function submit(event?: React.FormEvent) {
```
TheMallen: Rerendering a `Form` element is not expensive, so this being reference equal between renders is not a big deal. In addition, checking `fieldBag` meant that in most cases this would be regenerated each render anyway.

packages/predicates/CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [1.1.0]
+
+### Added
+
+- `notNumericString`
```
lemonmade: This is not a good changelog

packages/react-form/src/hooks/field/field.ts
```diff
@@ -188,9 +194,17 @@ export function useField<Value = string>(
...
-    return {value: input, validates: () => undefined};
+  if (isFieldConfig(input)) {
+    return input;
   }
 
-  return input as FieldConfig<Value>;
+  return {value: input, validates: () => undefined};
+}
+
+function isFieldConfig<Value>(input: any): input is FieldConfig<Value> {
```
lemonmade: `unknown` is a better type in these cases

packages/react-form/src/hooks/field/field.ts
```diff
@@ -188,9 +194,17 @@ export function useField<Value = string>(
...
+  return (
+    typeof input === 'object' &&
```
lemonmade: `input` can't be null, right?

packages/react-form/src/hooks/field/test/field.test.tsx
```diff
@@ -14,7 +14,7 @@ describe('useField', () => {
           <input
             id="test-field"
             name="test-field"
-            value={field.value}
+            value={field.value as any}
```
lemonmade: Bummer, why was this needed?

packages/react-form/src/hooks/submit.ts
lemonmade: I am not sure that it is safe to say that the thing this callback gets passed to is cheap to re-render, in the end you really have no idea what the cost is. Ideally we would memoize because the consumer can't, but if that's impossible, then that's fine.

packages/react-form/src/hooks/field/field.ts
TheMallen: nope, at least not if they are using TS :P

packages/react-form/src/hooks/field/test/field.test.tsx
TheMallen: Because in one of the tests we use an object as the value and the HTML input haaaates it.  Maybe I'll just go back and make a different test component for that.

packages/predicates/CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [1.1.0]
+
+### Added
+
+- added the `notNumericString` predicate, which returns `true` when its input is not a numeric string and `false` otherwise.
```
cartogram: Should we add a PR link?

packages/react-form/src/hooks/field/test/field.test.tsx
```diff
@@ -194,6 +194,42 @@ describe('useField', () => {
...
+                onBlur={field.onBlur}
+              />
+            </label>
+            {field.error && <p>{field.error}</p>}
+          </>
+        );
+      }
+
+      const wrapper = mount(<FooBarTestField config={{foo: 'bar'}} />);
+      const newValue = faker.commerce.product();
```
cartogram: that's handy

PR #678, opened 2019-04-29
♻️ refactor i18n errors to enforce the error message that should be t…

michelleyschen:
…hrow

This should fix https://github.com/Shopify/web/issues/13037 once its web is updated.

- add back the `MissingTranslationError` error message that was removed from https://github.com/Shopify/quilt/pull/547/files#diff-a1fd1ab286680c4b32b5676c4e8f4d54L77-R77
- refactor it in a way where all of the specific i18n errors will provided consistent useful message

### DIFF
--- a/packages/react-i18n/src/errors.ts
+++ b/packages/react-i18n/src/errors.ts
@@ -1,12 +1,49 @@
-export class MissingTranslationError extends Error {}
-export class MissingReplacementError extends Error {}
-export class MissingCurrencyCodeError extends Error {}
-export class MissingCountryError extends Error {}
-export class InvalidI18nConnectionError extends Error {}
+import {Replacements} from './types';
+
+export class MissingTranslationError extends Error {
+  constructor(key: string) {
+    super(`Missing translation for key: ${key}`);
+  }
+}
+export class MissingReplacementError extends Error {
+  constructor(replacement: string, replacements: Replacements = {}) {
+    let errorMessage = '';
+    const replacementKeys = Object.keys(replacements);
+
+    if (replacementKeys.length < 1) {
+      errorMessage = `No replacement found for key '${replacement}' (and no replacements were passed in).`;
+    } else {
+      errorMessage = `No replacement found for key '${replacement}'. The following replacements were passed: ${replacementKeys
+        .map(key => `'${key}'`)
+        .join(', ')}`;
+    }
+
+    super(errorMessage);
+  }
+}
+export class MissingCurrencyCodeError extends Error {
+  constructor(additionalMessage = '') {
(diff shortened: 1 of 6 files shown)


packages/react-i18n/src/errors.ts
```diff
@@ -1,12 +1,45 @@
...
+    super(`Missing translation for key: ${key}`);
+  }
+}
+export class MissingReplacementError extends Error {
+  constructor(replacement: string, replacements: Replacements = {}) {
+    const errorMessages = [`No replacement found for key '${replacement}'.`];
+    const replacementKeys = Object.keys(replacements);
+
+    if (replacementKeys.length < 1) {
+      errorMessages.push('And no replacements were passed in.');
```
lemonmade: I think putting this in parens instead of a separate sentence reads better

packages/react-i18n/src/errors.ts
```diff
@@ -1,12 +1,45 @@
...
+    } else {
+      errorMessages.push(
+        `The following replacements were passed: ${replacementKeys
+          .map(key => `'${key}'`)
+          .join(', ')}`,
+      );
+    }
+
+    super(errorMessages.join(' '));
```
lemonmade: Seems weird to do it this way versus string concatenation to me /shrug

packages/react-i18n/src/errors.ts
```diff
@@ -1,12 +1,45 @@
...
+  }
+}
+export class MissingCurrencyCodeError extends Error {
+  constructor(scopeName: string) {
+    super(
+      `No currency code provided. ${scopeName} cannot be called without a currency code.`,
```
lemonmade: This feels too specific to where we currently throw the error, I’d rather the consumer just pass the whole message.

michelleyschen: Why would it be too specific? I feel that if the error being thrown is not meant to be specific with a consistent message, we can just throw a generic `Error` with the customized message instead of `MissingCurrencyCodeError`.

packages/react-i18n/src/errors.ts
michelleyschen: sure

packages/react-i18n/src/errors.ts
lemonmade: Because now it is tied to only things that the user actually called. What if the logic for throwing the error goes into a function that doesn't know about the call site? What if we move checking for the timezone to initialization instead of call time?

michelleyschen: Alternatively, I can enforce only the `No currency code provided.` message, and concat it with whatever the consumer want to display as the second sentence.

PR #679, opened 2019-04-29
add doc & test for `useMountedRef`

michelleyschen:


packages/react-hooks/README.md
```diff
@@ -56,3 +56,32 @@ function MyComponent() {
   return null;
 }
 ```
+
+### `useMountedRef()`
+
+This hook keep track of a component's mounted / un-mounted status and returns a ref object like React’s [`useRef`](https://reactjs.org/docs/hooks-reference.html#useref) with a boolean value representing said status. This is often use when a component contains async task that set state after the task resolved.
```
lemonmade:
```suggestion
This hook returns a ref object (like React’s [`useRef`](https://reactjs.org/docs/hooks-reference.html#useref)) that tracks whether a component is mounted. This is often use when a component contains async tasks that set state after the task resolved.
```

packages/react-hooks/README.md
```diff
@@ -56,3 +56,32 @@ function MyComponent() {
...
+
+```tsx
+import * as React from 'react';
+import {useMountedRef} from '@shopify/react-hooks';
+const promise = new Promise();
```
lemonmade: Newline after imports

packages/react-hooks/README.md
```diff
@@ -56,3 +56,32 @@ function MyComponent() {
...
+
+function MockComponent() {
+  const [resolved, setResolved] = React.useState(false);
+  const mounted = useMountedRef();
+
+  async function handleOnClick() {
+    await promise;
```
lemonmade: I'd usually show a slightly more realistic example, like `await waitForSomethingToLoad()`

packages/react-hooks/src/hooks/tests/mounted-ref.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+      return (
+        <button onClick={handleOnClick} type="button">
+          Resolved Promise
+        </button>
+      );
+    }
+
+    const mockComponent = mount(<MockComponent />);
+    mockComponent.find('button')!.trigger('onClick', undefined as any);
```
lemonmade: Do you need the second argument? It should now be optional.

PR #689, opened 2019-05-06
Fix issues with error propagation in react-form

TheMallen:
Previously we expected `fieldPath` keys from failed submissions, but this does not work automagically with our GraphQL APIs, so this PR changes the expected shape to match them.

In addition, after making this change I encountered issues with infinite rerenders in components on submit. This seemed to be because `useErrorPropagation` was causing a rerender of each field whenever it iterated through them, which would then cause it to run again and cause them to rerender. I did not see this in my earlier tests due to some convenient facts around the data I was returning. 

To make these things fool-proof (and remove some unnecessary looping happening each render) I made `useValidateAll` and `useErrorPropagation` into regular old non-hook functions and just called them in our other hooks when we need to. They are still exported so that users can use them with their own custom hooks if they want.


packages/react-form/src/hooks/submit.ts
```diff
@@ -33,13 +42,14 @@ export function useSubmit<T extends FieldBag>(
...
+      setSubmitErrors(result.errors);
+      propagateErrors(fieldBag, result.errors);
     } else {
-      setRemoteErrors([]);
+      setSubmitErrors([]);
     }
   }
 
-  return {submit, submitting, errors: remoteErrors, setErrors: setRemoteErrors};
+  return {submit, submitting, errors: submitErrors, setErrors: setSubmitErrors};
```
michelleyschen: so I wonder if the `setErrors` should be a wrapper of both `setSubmitErrors` & `propagateErrors` ? 
```
setSubmitErrors(result.errors);
propagateErrors(fieldBag, result.errors);
```

packages/react-form/src/utilities.ts
```diff
@@ -40,4 +49,51 @@ export function isChangeEvent(
   );
 }
 
+export function propagateErrors(
+  fieldBag: {[key: string]: FieldOutput<any>},
+  remoteErrors: FormError[],
```
michelleyschen: nit pick, it is named `remoteErrors` here, but we also pass in `clientError` above?

TheMallen: It should really just be called `errors` here, as it gets used both for errors from running all the local validations and for errors from the submit callback.

PR #691, opened 2019-05-08
Add ArgumentAtIndex type

lemonmade:
Adds an `ArgumentAtIndex` type for grabbing arbitrary argument types from a function.

Screen Shot 2019-05-08 at 10 14 06 AM

### DIFF
--- a/packages/useful-types/src/types.ts
+++ b/packages/useful-types/src/types.ts
@@ -3,7 +3,12 @@ import {ComponentType, HTMLAttributes} from 'react';
 export type ThenType<T> = T extends Promise<infer U> ? U : T;
 
 export type Arguments<T> = T extends (...args: infer U) => any ? U : never;
-export type FirstArgument<T> = T extends ((arg: infer U) => any) ? U : never;
+export type ArgumentAtIndex<
+  Func,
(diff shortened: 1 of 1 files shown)


packages/useful-types/README.md
```diff
@@ -25,7 +25,16 @@ The following type aliases are provided by this library:
   type P = ThenType<typeof promise>; // string
   ```
 
-- `FirstArgument<T>`: Resolves to the type of the first argument to the passed function. Useful for cases where you wish to extract the type of arguments without actually exporting the argument types, and is a nice complement to TypeScript’s built-in `ReturnType`.
+- `ArgumentAtIndex<T, Index>`: Resolves to the type of the argument of the passed function at the passed index. Useful for cases where you wish to extract the type of arguments without actually exporting the argument types, and is a nice complement to TypeScript’s built-in `ReturnType`.
+
+  ```ts
+  const func = (foo: Promise<any>, bar: string) => foo.then(() => bar);
+  type Arg1 = FirstArgument<typeof func, 0>; // Promise<any>
```
GoodForOneFare: I think `FirstArgument` => `ArgumentAtIndex`  for all but line 37.

PR #694, opened 2019-05-09
✨ add the css-utilities lib

michelleyschen:
Close https://github.com/Shopify/react-utilities/issues/24

The plan is to release this package, and we can deprecate `@shopify/react-utilities` entirely.

### DIFF
--- a/packages/css-utilities/README.md
+++ b/packages/css-utilities/README.md
@@ -1,0 +1,69 @@
+# `@shopify/css-utilities`
+
+[![Build Status](https://travis-ci.org/Shopify/quilt.svg?branch=master)](https://travis-ci.org/Shopify/quilt)
+[![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](LICENSE.md) [![npm version](https://badge.fury.io/js/%40shopify%2Fcss-utilities.svg)](https://badge.fury.io/js/%40shopify%2Fcss-utilities.svg) [![npm bundle size (minified + gzip)](https://img.shields.io/bundlephobia/minzip/@shopify/css-utilities.svg)](https://img.shields.io/bundlephobia/minzip/@shopify/css-utilities.svg)
+
+A set of css styling related utilities.
+
+## Installation
+
+```bash
+$ yarn add @shopify/css-utilities
+```
+
+Both `classNames` and `variationName` are helper functions that are use to generate class names.
+
+### `classNames`
+
+Is a straight export of [`classnames`](https://github.com/JedWatson/classnames). It returns a classNames string that are conditionally joined together.
+
+### `variationName`
+
+Is a utility that returns a camelCase string combining the name and value passed in.
+
+## Usage
+
+Given this CSS file for a `Square` React component:
+
+```scss
+.Square {
+  color: transparent;
+}
+
+.sizeSmall {
+  height: 20px
+  width: 20px
+}
+
+.sizeLarge {
+  height: 44px
+  width: 44px
+}
+
+.colorWhite {
+  background-color: white;
+}
+
+.colorBlack {
+  background-color: black;
+}
+```
+
+The following can be use to generate different class names for the component based on its props:
+
+```tsx
+interface Props {
+  size: 'small' | 'large';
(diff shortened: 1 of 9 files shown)


packages/react-css-utilities/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Installation
+
+```bash
+$ yarn add @shopify/react-css-utilities
+```
+
+## Usage
+
+Both `classnames` and `variationName` utilities should ideally be use in combination with a `scss` styles sheet and [`css module`](https://github.com/css-modules/css-modules) setup to generate class name for react component.
```
lemonmade: A few comments:

- You jump right into this without even saying that these are two functions exported by the package
- `scss` here doesn't really matter
- `CSS` (uppercase)
- Again, I don't think React is relevant here.

packages/react-css-utilities/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+Given a scss file for a simple Square
```
lemonmade:
```suggestion
Given this CSS file for a `Square` React component:
```

packages/react-css-utilities/README.md
```diff
@@ -0,0 +1,72 @@
...
+.colorWhite {
+  background-color: white;
+}
+
+.colorBlack {
+  background-color: black;
+}
+```
+
+The following can be use to generate different class name from Props in a react component
```
lemonmade:
```suggestion
The following can be use to generate different class names for the component based on its props:
```

packages/react-css-utilities/README.md
```diff
@@ -0,0 +1,72 @@
...
+    styles[variationName('size', size)],
+  );
+
+  return <div className={className} />;
+}
+```
+
+### Resulting className
+
+The final class name in the DOM will depend on your css module set up.
```
lemonmade: Don't bother saying this, the library has no impact on the built style sheet.

packages/react-css-utilities/README.md
lemonmade: Also it's `classNames` from the library

lemonmade: I think you should also link to the `classname` library, you don't really say what that function even does (removes falsy values, etc)

packages/react-css-utilities/package.json
```diff
@@ -0,0 +1,36 @@
...
+  },
+  "bugs": {
+    "url": "https://github.com/shopify/quilt/issues"
+  },
+  "homepage": "https://github.com/Shopify/quilt/blob/master/packages/react-css-utilities/README.md",
+  "dependencies": {
+    "classnames": "^2.2.6"
+  },
+  "devDependencies": {
+    "@types/classnames": "^2.2.7",
```
lemonmade: The types actually need to be a dependency here, otherwise the consumer won't necessarily have them installed.

packages/css-utilities/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+A set of css styling related utilities.
+
+## Installation
+
+```bash
+$ yarn add @shopify/css-utilities
+```
+
+Both `classnames` and `variationName` are helper functions that are use to generate classNames.
```
lemonmade: Still should be classNames for the first one, and maybe classname for the second one?

packages/css-utilities/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+### `classNames`
+
+Is a straight export of [`classnames` by JedWatson](https://github.com/JedWatson/classnames). It returns a classNames string that are conditionally joined together.
```
lemonmade: I don’t think you need to list the name of the maintainer.

packages/css-utilities/README.md
```diff
@@ -0,0 +1,77 @@
...
+}
+```
+
+### Resulting className
+
+```tsx
+<Square size="small" color="white" />
+// in browser
+// <div class="Square Square--sizeSmall Square--colorWhite" />
+```
```
lemonmade: Still think you should remove this whole section.

PR #697, opened 2019-05-10
Fix nested wrappers bug in `react-testing`

cartogram:
This PR fixes a bug in react-testing where an unresolved promise required an additional `act()` around the state change.

### DIFF
--- a/packages/react-testing/src/root.tsx
+++ b/packages/react-testing/src/root.tsx
@@ -79,7 +79,9 @@ export class Root<Props> {
   }
 
   act<T>(action: () => T, {update = true} = {}): T {
+    const updateWrapper = update ? this.update.bind(this) : noop;
     let result!: T;
+
     if (this.acting) {
       return action();
     }
@@ -87,9 +89,7 @@ export class Root<Props> {
     this.acting = true;
 
     const afterResolve = () => {
-      if (update) {
-        this.update();
-      }
+      updateWrapper();
       this.acting = false;
 
       return result;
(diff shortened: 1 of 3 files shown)


packages/react-testing/src/root.tsx
```diff
@@ -107,6 +107,8 @@ export class Root<Props> {
     });
 
     if (isPromise(result)) {
+      updateWrapper();
```
lemonmade: Hmm, I'm not sure I understand why you would do it here, also. if the result is a promise, there shouldn't be a reason to update until after the act block has finished, is there?

TheMallen: If the action you take looks something like

```
async function doASubmit() { 
   setSomeLoadingState(true);
   await sendSomeRequest();
   setSomeLoadingState(false);
}
```

Then after synchronous execution pauses on the await there has been a state update, if we don't call `updateWrapper` then any test for loading states in cases will fail.

lemonmade: For this to make sense, `setSomeLoadingState` has to run synchronously. I don't think that's guaranteed for async functions?

cartogram:
@lemonmade @TheMallen updated the PR to do 2 things:
1) made the update option for act is an enum of `Pending`, `Complete` or `Never`. The default is still `Complete` as before, but now we can offer somewhere for users to configure this from their tests.
2) expose the update method publicly on the root so that users can call the `wrapper.update()` after a loading state is triggered and then assert on the updated UI.

packages/react-testing/src/root.tsx
lemonmade: Resolved in Slack, we agreed that this is probably the expected behaviour, and if they need to wait until any other update is finished and then update the tree, they can do so with `act`

Issue #571, opened 2019-03-11
FormState.List is broken for nested lists


jrobusto:
## Overview

Related to #519, the recent fix #531 fixed the issue for lists a single level deep. However, if there's a `FormState.List` within a `FormState.List`, the component crashes when an item is added to the nested list, because `initialValue` is `undefined` instead of an empty array.

I believe adding an additional guard of `initialValue ||` before `intialValue[index]` to line 49 of `List.tsx` will solve this issue and it seems to work with local testing, but I do not know the package well enough to thoroughly test for unintended side-effects.

## Consuming repo

Shopify/quilt

## Checklist

- [x] Please delete the labels section before submitting your issue
- [x] I have described this issue in a way that is actionable (if possible)

PR #698, opened 2019-05-13
Fix Formstate.Nested in Formstate.List

axelerator:
Using a nested property in a list would result in a "Cannot read property
'brand' of undefined" error when adding new items to the list.

The error looks something like this:
```
Cannot read property 'brand' of undefined
TypeError: 
    at eval (webpack:///./node_modules/@shopify/react-form-state/dist/components/Nested.js?:33:49)
    at eval (webpack:///./node_modules/@shopify/react-form-state/dist/utilities.js?:9:28)
    at Array.reduce (<anonymous>)
    at Object.mapObject (webpack:///./node_modules/@shopify/react-form-state/dist/utilities.js?:7:34)
    at Nested.render (webpack:///./node_modules/@shopify/react-form-state/dist/components/Nested.js?:32:39)
```

This may also address #571

<details><summary>Minimal broken example of broken</summary>
<p>

```jsx
function App() {
  return (
    <AppProvider>
    <div className="App">
      <header className="App-header">
        <FormState
        initialValues={{
          variants: [
            {
              description: {
                brand: 'brandName1',
              },
            },
          ],
        }}
      >
        {({fields}) => {
          const {variants} = fields;
          function addVariant() {
            variants.onChange(
              arrayUtils.push(variants.value, {
                description: {
                  brand: 'brandName3',
                },
              }),
            );
          }
          return (
            <form>
              <div>
                  <div>
                    <FormState.List field={fields.variants}>
                      {(fields) => {
                        return (
                          <FormState.Nested field={fields.description}>
                            {(nestedFields) => {
                              return (
                                <TextField
                                  label=""
                                  {...nestedFields.brand}
                                />
                              );
                            }}
                          </FormState.Nested>
                        );
                      }}
                    </FormState.List>

                    <ButtonGroup>
                      <Button onClick={addVariant}>Add variant</Button>
                    </ButtonGroup>
                  </div>
              </div>
            </form>
          );
        }}
      </FormState>
        <p>
          Edit <code>src/App.js</code> and save to reload.
        </p>
        <a
          className="App-link"
          href="https://reactjs.org"
          target="_blank"
          rel="noopener noreferrer"
        >
          Learn React
        </a>
      </header>
    </div>
    </AppProvider>
  );
}
```
</p>
</details>

### DIFF
--- a/packages/react-form-state/src/components/Nested.tsx
+++ b/packages/react-form-state/src/components/Nested.tsx
@@ -23,7 +23,8 @@ export default class Nested<Fields> extends React.PureComponent<
     const innerFields: FieldDescriptors<Fields> = mapObject(
       value,
       (value, fieldPath) => {
-        const initialFieldValue = initialValue[fieldPath];
+        const initialFieldValue = initialValue && initialValue[fieldPath];
+
         return {
           value,
           onBlur,
--- a/packages/react-form-state/src/components/tests/List.test.tsx
+++ b/packages/react-form-state/src/components/tests/List.test.tsx
@@ -6,7 +6,8 @@ import {trigger} from '@shopify/enzyme-utilities';
 // eslint-disable-next-line shopify/strict-component-boundaries
 import {Input} from '../../tests/components';
 import {lastCallArgs} from '../../tests/utilities';
-import FormState from '../..';
+import FormState, {arrayUtils} from '../..';
+import {FieldDescriptor} from '../../types';
 
 describe('<FormState.List />', () => {
   it('passes form state into child function for each index of the given array', () => {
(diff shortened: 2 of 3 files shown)


packages/react-form-state/src/components/Nested.tsx
```diff
@@ -23,7 +23,10 @@ export default class Nested<Fields> extends React.PureComponent<
     const innerFields: FieldDescriptors<Fields> = mapObject(
       value,
       (value, fieldPath) => {
-        const initialFieldValue = initialValue[fieldPath];
+        let initialFieldValue: any;
+        if (initialValue) {
+          initialFieldValue = initialValue[fieldPath];
+        }
```
lemonmade: Would probably be a bit cleaner to do `const initialFieldValue = initialValue && initialValue[fieldPath]`

lemonmade: I am a bit confused how this could ever happen if the field type is not an optional type; that is, in the example case you provided, why is `initialValue` undefined? It looks like `initialValue` should be `{brand: 'brandName1'}`, no?

axelerator: I am not sure - but I think it's du to the fact that an added list item is not present in the initial data passed to the FormState.

lemonmade: In the example you provided (very helpful btw), the initial values are definitely present. I wonder if this is really more of a bug in `List` not passing down the value to the render prop correctly? Can you log out what `fields.description` is?

axelerator: The error happens only when you add a new element to the list - this new, second item is not present in the data initially passed to `<FormState initialValues=..`, which has only one element in the variants` list

packages/react-form-state/src/components/tests/List.test.tsx
```diff
@@ -339,4 +339,52 @@ describe('<FormState.List />', () => {
     expect(updatedFields.title.value).toBe(newTitle);
     expect(updatedFields.department.value).toBe(newDepartment);
   });
+
+  it.only('does not raise an exceptions when list items are nested', () => {
```
marina101: you should remove the only

axelerator: done

axelerator:
I added a test and a changelog entry

PR #699, opened 2019-05-13
🐛 react-i18n Babel plugin no longer fails on babel-preset-typescript ASTs

GoodForOneFare:
### The problem
When integrating react-i18n + jest@24, the i18n Babel plugin fails with messages about manipulating removed AST paths.

### The technical problem
* `babel-preset-typescript` removes all type references from the AST
* If an `import` pulls in _only_ type information, `@babel/typescript` sees it as having no purpose, and strips it out, too
* Our react-i18n/babel plugin pulls out the "last import" node at the start of processing, and stores it in state
* By the time the plugin gets to injecting a translations import, the last import state may be pointing at a removed AST path

### The solution
Figure out the last import immediately before injecting the translations import.

### DIFF
--- a/packages/react-i18n/src/babel-plugin.ts
+++ b/packages/react-i18n/src/babel-plugin.ts
@@ -6,7 +6,7 @@ import {NodePath} from '@babel/traverse';
 import stringHash from 'string-hash';
 
 interface State {
-  lastImport: NodePath<Types.ImportDeclaration>;
+  program: NodePath<Types.Program>;
 }
 
 export default function injectWithI18nArguments({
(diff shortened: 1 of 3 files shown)


packages/react-i18n/src/babel-plugin.ts
```diff
@@ -124,9 +120,17 @@ export default function injectWithI18nArguments({
               filename,
               fallbackID,
               insertImport() {
-                lastImport.insertAfter(
-                  fallbackTranslationsImport({id: fallbackID}),
-                );
+                const lastImport = program
```
GoodForOneFare: Now that I've written the PR, I'm wondering why we don't just insert after the `withI18n` path?  It seems like that'd avoid storing `program` in state.

lemonmade: There is no guarantee the `withI18n` path will be in a location that you can insert an import (and there is a guarantee that `useI18n` will *not* be in a place you can insert an import. I do think it should be fine to just add it as the first line in the program, though.

PR #701, opened 2019-05-14
Progressive hydration helper packages

lemonmade:
This PR adds two helper libraries that will be needed for progressive hydration:

- `@shopify/react-idle`, which adds some basic primitives for registering idle callbacks
- `@shopify/react-hydrate`, which provides the tools to have client and server markup differ without throwing away the server work

All of the details are in the READMEs. I also had to update our `requestIdleCallback` mock a bit more, and fix a bug in `react-testing` with `findAll` not finding `memo`-ed elements.

### DIFF
--- a/packages/react-hydrate/README.md
+++ b/packages/react-hydrate/README.md
@@ -1,0 +1,127 @@
+# `@shopify/react-hydrate`
+
+[![Build Status](https://travis-ci.org/Shopify/quilt.svg?branch=master)](https://travis-ci.org/Shopify/quilt)
+[![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](LICENSE.md) [![npm version](https://badge.fury.io/js/%40shopify%2Freact-hydrate.svg)](https://badge.fury.io/js/%40shopify%2Freact-hydrate.svg) [![npm bundle size (minified + gzip)](https://img.shields.io/bundlephobia/minzip/@shopify/react-hydrate.svg)](https://img.shields.io/bundlephobia/minzip/@shopify/react-hydrate.svg)
+
+Utilities for hydrating server-rendered React apps.
+
+## Installation
+
+```bash
+$ yarn add @shopify/react-hydrate
+```
+
+## Usage
+
+This library is intended to assist with "progressive hydration", a pattern where you fully render an application on the server, but wait to hydrate parts of it when it reaches the client. Typically, doing different work on the server and client would result in the server markup being thrown out. This library avoids that issue by rendering a wrapping `div` around the content on the server, adding an ID to that element, and setting the `dangerouslySetInnerHTML` prop on the client to the resulting server markup in order to avoid mismatches. Once you have done whatever work on the client to load the necessary components, this hardcoded markup is removed, allowing the React tree to take over.
+
+### Server
+
+There are two key pieces to making this work. First, your server must render a `HydrationContext.Provider` element around your React tree (to be clear: this is only required for the server, not on the client). This element will provide a `HydrationManager` to the three, which manages the identifiers that map server markup to client markup.
+
+```tsx
+import React from 'react';
+import {render} from 'react-dom';
+import {HydrationContext, HydrationManager} from '@shopify/react-hydration';
+import App from '../app';
+
+export async function middleware(ctx, next) {
+  const hydrationManager = new HydrationManager();
+
+  ctx.body = render(
+    <HydrationContext.Provider value={hydrationManager}>
+      <App />
+    </HydrationContext.Provider>,
+  );
+
+  await next();
+}
+```
+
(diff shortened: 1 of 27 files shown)


lemonmade:
@Shopify/xss-extermination-squad this is a safe use of `dangerouslySetInnerHTML` because it only sets the value to whatever the original server markup was (which would have been safe).

packages/react-hydrate/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Installation
+
+```bash
+$ yarn add @shopify/react-hydrate
+```
+
+## Usage
+
+This library is intended to assist with "progressive hydration", a pattern where you fully render an application on the server, but wait to hydrate parts of it when it reaches the client. Typically, doing different work on the server and client would result in the server markup being thrown out. This library avoids that issue by rendering a wrapping `div` around the content on the server, adding an ID to that element, and setting the `dangerouslySetInnerHTML` prop on the client to the resulting server markup in order to avoid mismatches. Once you have done whatever work on the client to load the necessary components, this hardcoded markup is removed, allowing the React tree to take over.
```
cartogram: > Typically, doing different work on the server and client would result in the server markup being thrown out.

A tiny note about "why" after this sentence would help understand the problem these utilities are solving.

packages/react-hydrate/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Server
+
+There are two key pieces to making this work. First, your server must render a `HydrationContext.Provider` element around your React tree (to be clear: this is only required for the server, not on the client). This element will provide a `HydrationManager` to the three, which manages the identifiers that map server markup to client markup.
```
cartogram: `three => tree`

packages/react-hydrate/README.md
```diff
@@ -0,0 +1,127 @@
...
+```tsx
+import React, {useState, useEffect} from 'react';
+import {Hydrator} from '@shopify/react-hydrate';
+
+// This is a hypothetical component that knows how to load a component
+// asynchronously, but can also potentially access it synchronously on the
+// server. So, the server will immediately have access to the component and
+// can render it for the initial page load, but the client will have to wait
+// until it is loaded asynchronously. The Hydrator component stands in the middle
+// to bridge the gap and prevent the server markup from being thrown away.
```
cartogram: I like this comment.

packages/react-hydrate/src/test/e2e.test.tsx
```diff
@@ -0,0 +1,91 @@
+import React from 'react';
```
cartogram: Are we calling these `e2e` now?

packages/react-hydrate/src/test/e2e.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+    return (
+      <HydrationContext.Provider value={manager}>
+        {element}
+      </HydrationContext.Provider>
+    );
+  },
+});
+
+describe('react-hydrate', () => {
+  it('renders a wrapping element with a hydration ID by no innerHTML when there are children', () => {
```
cartogram: `by` => `but`

packages/react-idle/src/test/OnIdle.test.tsx
```diff
@@ -0,0 +1,115 @@
...
+      mount(<OnIdle perform={spy} />);
+
+      expect(spy).not.toHaveBeenCalled();
+
+      animationFrame.runFrame();
+
+      expect(spy).toHaveBeenCalled();
+    });
+
+    it('runs perform in immediately when requestIdleCallback is not supported and unsupportedBehavior is set to animation frame', () => {
```
cartogram: `in immediately`, remove the `in`

packages/react-idle/src/test/OnIdle.test.tsx
```diff
@@ -0,0 +1,115 @@
...
+      );
+
+      expect(spy).not.toHaveBeenCalled();
+
+      animationFrame.runFrame();
+
+      expect(spy).toHaveBeenCalled();
+    });
+
+    it('runs perform in immediately when requestIdleCallback is not supported and unsupportedBehavior is set to immediate', () => {
```
cartogram: same here

packages/react-idle/src/test/OnIdle.test.tsx
```diff
@@ -0,0 +1,115 @@
...
+          unsupportedBehavior={UnsupportedBehavior.Immediate}
+        />,
+      );
+
+      expect(spy).toHaveBeenCalled();
+    });
+
+    it('cancels the animation frame when the component unmounts', () => {
+      const spy = jest.fn();
+      const onIdle = mount(<OnIdle perform={spy} />);
```
cartogram: I think you mean to have `unsupportedBehavior={UnsupportedBehavior.AnimationFrame}` here?

packages/react-hydrate/src/test/e2e.test.tsx
lemonmade: I dunno if it's a real convention, but it wasn't a test of any particular part so it felt appropriate here.

PR #702, opened 2019-05-15
Support SSR in react-graphql hooks

lemonmade:
When rolling out the new stuff to `web-proving-ground`, I noticed that server rendering was broken. There were actually two separate SSR-breaking issues at play:

- When you render an `asyncQueryComponent`, it renders an `Async` component, which in turn registers the ID of that async asset as "used" for the purposes of adding the script tag. In the query hook with an async component, we manually call `resolve`, which does not have any way of registering the asset as used, so the script for the query is not in the initial payload
- When you render an `asyncQueryComponent`, it also eventually renders an Apollo `Query` component, which, through a convoluted set of steps involving old context and our `react-effect-apollo` library, eventually causes those promises to be waited on when calling `extract` from `react-effect`. No such behaviour exists when using the hook version of the query.

This PR solves the two issues by:

- Adding a new hook for registering an async asset as used, and passing through the ID of the original query document if it was an async query
- Adding a `useServerEffect` hook that waits for a query to resolve on the server (which, best of all, operates completely independently of `react-effect-apollo`!)

Here's the before showing the script tags and serialization (note that the script tag for the query is after `main`, so it wasn't part of the original payload):

Screen Shot 2019-05-15 at 12 12 28 PM
Screen Shot 2019-05-15 at 12 11 50 PM

After:

Screen Shot 2019-05-15 at 12 11 20 PM
Screen Shot 2019-05-15 at 12 11 00 PM

### DIFF
--- a/packages/react-graphql/src/hooks/query.ts
+++ b/packages/react-graphql/src/hooks/query.ts
@@ -1,10 +1,12 @@
+import {useEffect, useMemo, useState, useRef} from 'react';
 import {
   OperationVariables,
   ApolloError,
   WatchQueryOptions,
 } from 'apollo-client';
 import {DocumentNode} from 'graphql-typed';
-import {useEffect, useMemo, useState, useRef} from 'react';
+import {useServerEffect} from '@shopify/react-effect';
+import {useAsyncAsset} from '@shopify/react-async';
 
 import {AsyncQueryComponentType} from '../types';
 import {QueryHookOptions, QueryHookResult} from './types';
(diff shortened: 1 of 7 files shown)


packages/react-graphql/src/hooks/query.ts
```diff
@@ -1,10 +1,12 @@
+import {useEffect, useMemo, useState, useRef} from 'react';
 import {
   OperationVariables,
   ApolloError,
   WatchQueryOptions,
 } from 'apollo-client';
 import {DocumentNode} from 'graphql-typed';
-import {useEffect, useMemo, useState, useRef} from 'react';
+import {useServerEffect} from '@shopify/react-effect';
+import {useAsyncAsset} from '@shopify/react-async';
```
michelleyschen: `QueryHookOptions` does include a variable call `ssr`, 
and there are some special case the [`react-apollo-hooks`](https://github.com/trojanowski/react-apollo-hooks/blob/master/src/useQuery.ts#L86-L87) does do and I didn't take into account originally.

should `useServerEffect` &  `useAsyncAsset` always applied or should it be dependent on the `ssr` variable & has it affect the `fetchPolicy`?

lemonmade:
@michelleyschen opened an issue: https://github.com/Shopify/quilt/issues/703

Issue #687, opened 2019-05-06
[react-graphql] Render loading when query isn’t available


annasfrnv:
## Overview

If the async query is not available, there’s going to be a gap between the query being loaded and mounting the component.

This is happening on this line:
https://github.com/Shopify/quilt/blob/master/packages/react-graphql/src/async.tsx#L57

Shouldn’t return `null`, instead, it’d make sense to continue loading.

## Consuming repo

In `shopify/web`, here’s a visual reference:

![loading](https://user-images.githubusercontent.com/6495272/57241992-5f23bf80-7000-11e9-9d63-dab8bb255e5f.gif)

1. Navigate to a details page 
2. See a skeleton rendered in `renderLoading` 
3. See `null` (blank page in UI) being rendered when loading async query
4. See Details component mounting

## Checklist

- [x] Please delete the labels section before submitting your issue
- [x] I have described this issue in a way that is actionable (if possible)

PR #707, opened 2019-05-20
Render children when query not present

lemonmade:
Fixes https://github.com/Shopify/quilt/issues/687

This PR calls the `children` of an async query component with an object that looks like the loading state. This prevents unnecessary effort to keep a single loading state throughout the entire loading process of a query.

### DIFF
--- a/packages/react-graphql/src/async.tsx
+++ b/packages/react-graphql/src/async.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 
 import {DocumentNode} from 'graphql-typed';
+import {NetworkStatus} from 'apollo-client';
 import {LoadProps, DeferTiming} from '@shopify/async';
 import {
   Async,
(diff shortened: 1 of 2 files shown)


packages/react-graphql/src/async.tsx
```diff
@@ -47,14 +49,31 @@ export function createAsyncQueryComponent<Data, Variables, DeepPartial>({
...
+            componentProps.children({
+              data: undefined,
+              loading: true,
+              variables: (componentProps as any).variables,
+              client,
+              networkStatus: NetworkStatus.loading,
+              refetch: noop as any,
+              fetchMore: noop as any,
+              updateQuery: noop as any,
+              startPolling: noop as any,
```
TheMallen: my only real concern is that with these functions as no-ops it's possible you could get into weird situations where a component tries to call `startPolling` or `subscribeToMore` immediately and that just fails silently because the initial render / `componentDidMount`  was given a `noop`

lemonmade: This is true but unfortunately there isn't anything we can do about that here, hopefully they would correctly update their poll when this function changes.

PR #711, opened 2019-05-21
🐛 fix doc

michelleyschen:
`graphQL.operations.all()` documentation is in correct.

I do wonder if we should expose `pendingRequests` if people have uses for them.
cc @lemonmade what do you think?

### DIFF
--- a/packages/graphql-testing/README.md
+++ b/packages/graphql-testing/README.md
@@ -37,7 +37,7 @@ To resolve all pending graphQL operations:
 await graphQL.resolveAll();
 ```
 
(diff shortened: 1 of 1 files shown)


lemonmade:
I'm not sure if it makes ton of sense to expose the pending requests, it kind of pushes people to do a weird sloppy thing where they start GraphQL mutations but don't resolve them. I think this clarification is covered by this slightly larger PR: https://github.com/Shopify/quilt/pull/706

michelleyschen:
great! I am going to close this PR in preference to the larger one @lemonmade is working on.

PR #714, opened 2019-05-22
adding validateWithFields() to support consuming fields in a validator

patsissons:
in #690 the `validate()` signature was updated to support consumption of `fields: Fields`, but this change also made it a required parameter of generated validator function. This PR _effectively_ reverts #690 and creates a new `validateWithFields()` wrapper that exposes the generated validator from `validate()` with a signature that supports consuming the `fields` param.

### DIFF
--- a/packages/react-form-state/src/validators.ts
+++ b/packages/react-form-state/src/validators.ts
@@ -10,7 +10,7 @@ import {
 } from '@shopify/predicates';
 import {mapObject} from './utilities';
 
-interface Matcher<Input, Fields> {
+interface Matcher<Input, Fields = any> {
   (input: Input, fields: Fields): boolean;
 }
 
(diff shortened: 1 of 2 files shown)


patsissons:
I'm running into further issues with testing this. I'm going to look into this a bit deeper.

packages/react-form-state/src/validators.ts
```diff
@@ -73,11 +73,26 @@ export function validateList<Input extends object, Fields>(
   };
 }
 
+export function validateWithFields<Input, Fields>(
+  matcher: Matcher<Input, Fields>,
+  errorContent: ErrorContent,
+) {
+  return validate(matcher, errorContent) as (
```
patsissons: is there any benefit in writing a test to cover this wrapper? it's just messing with typing information so it's probably not worth while. 

Is there a better way to wrap `validate()`? i wanted to just use typing information, but my earlier naive approach doesn't work

packages/react-form-state/src/validators.ts
```diff
@@ -73,11 +73,26 @@ export function validateList<Input extends object, Fields>(
...
+    input: Input,
+    fields: Fields,
+  ) => ErrorContent | undefined | void;
+}
+
+export function validate<Input>(
+  matcher: Matcher<Input>,
+  errorContent: ErrorContent,
+): (input: Input) => ErrorContent | undefined | void;
```
patsissons: this is just the old signature reverted from #690

packages/react-form-state/src/validators.ts
TheMallen: eh I'd not bother.

patsissons:
yea i though about the `const` approach but scrunched my nose at it 😄

PR #716, opened 2019-05-22
⚡ prevent needless rerenders of pages using `useForm`

TheMallen:
This PR uses a cool trick thought of by @lemonmade  to keep functions returned by hooks reference equal despite the values they close over changing. It rolls this out to all of the callbacks returned by `useForm` and adds some tests to make sure we don't regress.

### DIFF
--- a/packages/react-form/src/hooks/form.ts
+++ b/packages/react-form/src/hooks/form.ts
@@ -1,3 +1,4 @@
+import {useCallback, useRef} from 'react';
 import {SubmitHandler, FormMapping, FieldBag, Form} from '../types';
 import {validateAll} from '../utilities';
 import {useDirty} from './dirty';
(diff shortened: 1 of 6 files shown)


packages/react-form/src/hooks/form.ts
```diff
@@ -61,21 +62,34 @@ export function useForm<T extends FieldBag>({
...
+    },
+    [basicReset, setErrors],
+  );
+
+  const fieldsRef = useRef(fields);
+  fieldsRef.current = fields;
+
+  const validate = useCallback(
+    () => {
+      return validateAll(fieldsRef.current);
```
TheMallen: I suppose we could bring back `validateAll`  as a discrete hook now possibly, but I'm not sure that actually is much better.

packages/react-form/src/hooks/visitFields.ts
```diff
@@ -10,9 +10,12 @@ export default function useVisitFields(
   fieldBag: {[key: string]: FieldOutput<any>},
   visitor: FieldVisitor,
 ) {
+  const fieldBagRef = useRef(fieldBag);
+  fieldBagRef.current = fieldBag;
```
lemonmade: Could be a useful generic hook?

TheMallen: true, though I do like that if you are using actual `useRef` the linting rule knows not to force you to add it to deps arrays

PR #721, opened 2019-05-28
[react-graphql] skip watchQuery

ken3650:
Fixes an issue we discovered in `web` (https://github.com/Shopify/web/issues/13938).
(Happening through `AdjacentResources` in `web`.)

The query was being run when we wanted to skip it; variables in the query may not be valid yet, so we need to wait until we don't want to skip to `watchQuery` (https://www.apollographql.com/docs/react/api/apollo-client#ApolloClient.watchQuery).

### DIFF
--- a/packages/react-graphql/src/hooks/query.ts
+++ b/packages/react-graphql/src/hooks/query.ts
@@ -71,13 +71,13 @@ export default function useQuery<
 
   const queryObservable = useMemo(
     () => {
-      if (!watchQueryOptions) {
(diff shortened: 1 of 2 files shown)


ken3650:
Unsure how to write a test for this 🤷‍♂

ken3650:
> This makes sense. A test you could write here would be to spy on `watchQuery` and make sure it was never called.

@lemonmade Good suggestion; added tests that do exactly that 👌

PR #724, opened 2019-05-29
Initial router docs

lemonmade:
This PR shows off the API of the router I've been working on. I have all the code basically ready to go, but I wanted to get some feedback at a slightly higher level first. Looking to hear from people how the API feels, whether we think things are missing, and what other documentation we should provide, particularly about comparisons with react-router.

### DIFF
--- a/packages/react-router/README.md
+++ b/packages/react-router/README.md
@@ -1,0 +1,378 @@
+# `@shopify/react-router`
+
+[![Build Status](https://travis-ci.org/Shopify/quilt.svg?branch=master)](https://travis-ci.org/Shopify/quilt) [![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](LICENSE.md) [![npm version](https://badge.fury.io/js/%40shopify%2Freact-router.svg)](https://badge.fury.io/js/%40shopify%2Freact-router)
+
+A universal router for React with first-class support for the [WHATWG `URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL) and prefetching.
+
+## Installation
+
+```bash
+$ yarn add @shopify/react-router
+```
+
+## Usage
+
+### `<Router />`
+
+Creates a router that is provided to the rest of the React tree. This router powers the `useCurrentUrl()` and `useRouter()` hooks and, by extension, the rest of the features of this library. On the client, you should render a `Router` with no props. On the server, you should pass in the `url` prop, which is a `URL` object that represents the current location to simulation.
+
+```tsx
+// Example in Koa...
+
+const app = new Koa();
+app.use((ctx) => {
+  ctx.body = renderToString(
+    <Router url={ctx.URL}>
+      <App />
+    </Router>,
+  );
+});
+```
+
+This component basically just composes the `Prefetcher` component, serialization for the `NoMatch` component, alongside a number of context providers (`RouterContext`, `CurrentUrlContext`). If you wish to omit some features from your final bundle, you can skip the `Router` component and render those primitives instead.
+
+### `<Route />`
+
+Interacts with the `Router` instance to register behavior when the location matches a URL. There is only one mandatory prop, `match`, which specifies the URL to match. This prop can be one of the following:
+
+- a `string`, in which case it will match any URL with a pathname that exactly matches, or that exactly matches except for a trailing `/`,
+- a `RegExp`, in which case it will match any URL with a pathname that matches the regex, or,
+- a function that takes an [`EnhancedURL` object](#enhancedurl), and returns a boolean indicating whether or not to match.
+
+A `<Route />` can then specify the behavior to use for the matching URLs by specifying any set of the following props:
+
+- `render`: a function that is called with an `EnhancedURL` object, and returns JSX to use when the route matches.
+- `renderPrefetch`: a function that is called with a `URL` object, and returns JSX to render when the route is [prefetched](#prefetcher).
+- `redirect`: any value accepted by the [`Redirect` component’s `to` prop](#redirect), which will be used to render a `Redirect` component when the route matches (should not be provided if `render` is also specified).
+
+The following example shows a component created by [`@shopify/react-async`](https://github.com/Shopify/quilt/tree/master/packages/react-async) that is rendered and prefetched when the current path is exactly `/products` or `/products/`:
+
+```tsx
+import React from 'react';
+import {Route} from '@shopify/react-router';
+import {createAsyncComponent} from '@shopify/react-async';
+
+const Products = createAsyncComponent({
+  load: () => import('./Products'),
+});
+
+export function ProductsRoute() {
+  return (
+    <Route
+      match="/products"
+      render={() => <Products />}
+      renderPrefetch={() => <Products.Prefetch />}
+    />
+  );
+}
+```
+
+This example shows how to extract a part of the path from the URL that your component needs to know (in this case, an ID in the form `/products/ID`):
+
+```tsx
+import React from 'react';
+import {Route} from '@shopify/react-router';
+import Product from './Product';
+
+const PRODUCT_MATCH = /^\/products\/(\d+)(\/|$)/;
+
+export function ProductRoute() {
+  return (
+    <Route
+      match={PRODUCT_MATCH}
+      render={(url) => {
+        const match = url.pathname.match(PRODUCT_MATCH);
+        return <Product id={match ? match[1] : undefined} />;
+      }}
+    />
+  );
+}
+```
+
+This example shows how to redirect a user from `/` to `/home`:
+
+```tsx
+import React from 'react';
+import {Route} from '@shopify/react-router';
+
+export function RootRoute() {
+  return <Route match="/" redirect="/home" />;
+}
+```
+
+### `<Link />`
+
+The `Link` component renders a native anchor tag pointing at a page in your application. When this anchor is activated, a client-side navigation will be performed instead of a full-page navigation. However, should JavaScript fail to load, or if JavaScript is still being downloaded, it will still let the user navigate the application. This component also knows when the user is intending to open a link in a new tab or window, and will allow that navigation to behave normally.
+
+The `Link` component accepts all props you could pass to an anchor element, except for `href`. In place of `href`, this component accepts a `to` prop, which can be any type allowed by [`router.navigate`](#router-navigate).
+
+```tsx
+import {Link} from '@shopify/react-router';
+
+export function MyComponent() {
+  return <Link to={{pathname: '/products/new'}}>Create product</Link>;
+}
+```
+
+### `useCurrentUrl()`
+
+Components in your application will often want to know the current URL. This can be particularly useful in cases like analytics, where you want to trigger an event every time the URL changes. The `useCurrentUrl()` hook gives components access to the current `EnhancedURL` (a `URL` representing the current location with an additional `state` field for the current location state). This hook will re-render your component whenever the current URL changes.
+
+```tsx
+import {useEffect} from 'react';
+import {useCurrentUrl} from '@shopify/react-router';
+
+function useNavigationTracking() {
+  const currentUrl = useCurrentUrl();
+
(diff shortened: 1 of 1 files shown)


packages/react-router/README.md
```diff
@@ -0,0 +1,378 @@
...
+app.use((ctx) => {
+  ctx.body = renderToString(
+    <Router url={ctx.URL}>
+      <App />
+    </Router>,
+  );
+});
+```
+
+This component basically just composes the `Prefetcher` component, serialization for the `NoMatch` component, alongside a number of context providers (`RouterContext`, `CurrentUrlContext`). If you wish to omit some features from your final bundle, you can skip the `Router` component and render those primitives instead.
```
TheMallen: Might be nice to see what the equivalent code would look like to compose them

packages/react-router/README.md
```diff
@@ -0,0 +1,378 @@
...
+
+Interacts with the `Router` instance to register behavior when the location matches a URL. There is only one mandatory prop, `match`, which specifies the URL to match. This prop can be one of the following:
+
+- a `string`, in which case it will match any URL with a pathname that exactly matches, or that exactly matches except for a trailing `/`,
+- a `RegExp`, in which case it will match any URL with a pathname that matches the regex, or,
+- a function that takes an [`EnhancedURL` object](#enhancedurl), and returns a boolean indicating whether or not to match.
+
+A `<Route />` can then specify the behavior to use for the matching URLs by specifying any set of the following props:
+
+- `render`: a function that is called with an `EnhancedURL` object, and returns JSX to use when the route matches.
```
TheMallen: thoughts on `render` vs `children` for the primary render prop? Personally I like `children` better

packages/react-router/README.md
```diff
@@ -0,0 +1,378 @@
...
+
+```tsx
+import {useEffect} from 'react';
+import {useCurrentUrl} from '@shopify/react-router';
+
+function useNavigationTracking() {
+  const currentUrl = useCurrentUrl();
+
+  useEffect(() => {
+    trackButNotInACreepyWayPlease(currentUrl.href);
```
TheMallen: nice

packages/react-router/README.md
```diff
@@ -0,0 +1,378 @@
...
+  );
+}
+```
+
+#### `EnhancedURL`
+
+This library is driven by a the `URL` object. In order to support features like location state, however, most of this library operates on `EnhancedURL` objects. These objects are identical to a `URL` object except for the following additions:
+
+- All members should be considered immutable
+- An additional `state` field is provided, which is the location state of the location the object represents
```
TheMallen: we've mentioned `location state` twice at this point without explaining what it is or linking to the section about it, as someone who still feels like it's a weird thing in the first place I think that's not great

packages/react-router/README.md
```diff
@@ -0,0 +1,378 @@
...
+}
+```
+
+### `<Redirect />`
+
+The `Redirect` component is essentially a component wrapper around running `router.navigate(to, {replace: true})`. It accepts a `to` prop, which can be anything you can pass to `router.navigate`. This component also works with [`@shopify/react-network`](https://github.com/Shopify/quilt/tree/master/packages/react-network) to perform a network redirect when rendered on the server.
+
+### `<NoMatch />`
+
+The `NoMatch` component keeps track of whether any nested routes have been matched and, if no matches are found, renders the result of calling `renderFallback`. Note that the `Route` components do not need to be direct children; they can be nested or composed however you like.
```
TheMallen: I think this is the only place we mention nested routes, how do you invision those working?

packages/react-router/README.md
```diff
@@ -0,0 +1,378 @@
...
+The `createTestRouter` function creates a mock router that matches up the expected shape of a normal router, but that does not actually navigate when `navigate`, `go`, `back`, or `forward` are called. This function accepts an optional `url` argument, which should be a `URL` object representing the location you wish to simulate in your test. If it is not provided, a URL will be constructed that represents the value of `window.location.href`.
+
+```tsx
+import React from 'react';
+import {createMount} from '@shopify/react-testing';
+import {TestRouter, createTestRouter} from '@shopify/react-testing';
+
+export const mount = createMount({
+  render: (element, _, {pathname = '/test'}) => (
+    <TestRouter router={createRouter(new URL(pathname))}>{element}</TestRouter>
```
TheMallen: `createRouter` should be `createTestRouter` right?

packages/react-router/README.md
```diff
@@ -0,0 +1,378 @@
...
+```
+
+This example shows how to extract a part of the path from the URL that your component needs to know (in this case, an ID in the form `/products/ID`):
+
+```tsx
+import React from 'react';
+import {Route} from '@shopify/react-router';
+import Product from './Product';
+
+const PRODUCT_MATCH = /^\/products\/(\d+)(\/|$)/;
```
TheMallen: I get the intention of using more language primitives instead of arcane weakly typed DSLs, but I think a static url with one or more dynamic segments is probably common enough that a specific abstraction would be preferable than a bunch of boilerplate. Maybe something like

```
  <DynamicSegmentRoute
     match="/product/:id" 
     render{(url, segments) => {
        const [id] = segments;
        return <Product id={id} />;
     }}
  />
```

packages/react-router/README.md
```diff
@@ -0,0 +1,378 @@
...
+```
+
+The router instance has the following methods:
+
+#### `router.navigate()`
+
+`router.navigate()` is the main method of the router. It allows you to add or replace an entry in the navigation stack. The first argument is a description of the location to push to, and can be one of the following:
+
+- A string, which should be a relative or absolute pathname, search, and hash,
+- A `URL` object (**note:** must be on the same origin as the current URL),
```
TheMallen: Why can't the router be smart enough to fall back to setting `location` when it encounters a different base URL? The behaviour could even be configurable. I've always found it a wart of RR that you have to remember when to use it vs `window.location`, especially in server rendered apps where you are generally avoiding `window.location` in other places.

packages/react-router/README.md
```diff
@@ -0,0 +1,378 @@
...
+
+Creates a router that is provided to the rest of the React tree. This router powers the `useCurrentUrl()` and `useRouter()` hooks and, by extension, the rest of the features of this library. On the client, you should render a `Router` with no props. On the server, you should pass in the `url` prop, which is a `URL` object that represents the current location to simulation.
+
+```tsx
+// Example in Koa...
+
+const app = new Koa();
+app.use((ctx) => {
+  ctx.body = renderToString(
+    <Router url={ctx.URL}>
```
TheMallen: how does this look with redirects?

packages/react-router/README.md
TheMallen: Or maybe it's built into the basic `Route` /shrug

packages/react-router/README.md
```diff
@@ -0,0 +1,378 @@
+# `@shopify/react-router`
```
ismail-syed: It would be good to document the need for our own router as a decision record.

packages/react-router/README.md
```diff
@@ -0,0 +1,378 @@
...
+}
+
+function MyComponentRoute() {
+  return (
+    <Route
+      match={/^\/resource(\/new)?/}
+      render={({pathname, searchParams}) => (
+        <MyComponent
+          create={pathname.contains('/new')}
+          prefillTitle={searchParams.get('prefillTitle')}
```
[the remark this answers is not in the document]
lemonmade: I specifically avoided anything like this partially because, as you say, it is based on the native URL, but also because a) an object is not a good representation of query params, because they can be more complex, and b) it means we need to be able to thread through custom types for the query params, which are incredibly difficult to do in a natural way (and are never actually type safe).

packages/react-router/README.md
vsumner: ➕To all of the above. It would make a refactor from `react-router` to this much gentler.

lemonmade: Hmm, I am definitely open to this idea, but my feeling is that it's better in this case for the library to be the primitive, and consuming projects can explore ideas that work better for them personally if they care. I specifically designed this library (`NoMatch` in particular) to work well with composition (unlike react-router, where `Switch`s break composition).

If you wanted the API Mat suggested, you could do it super easily in userspace:

```tsx
function DynamicSegmentRoute({match, render}, {match: string, render: (url: URL, segments: string[]) => React.ReactNode}) {
  return <Route match={toRegex(match)} render={(url) => render(url, getDynamicSegments(match, url)} />
}
```

If you wanted the exact React Router PR (even just for rolling it out incrementally), you can get that too:

```tsx
function DynamicSegmentRoute({path, render}: ReactRouterRouteProps) {
  return <Route match={toRegex(path)} render={(url) => render(url, getParamsObject(match, url)} />
}
```

packages/react-router/README.md
lemonmade: You just do what web does right now: check for a redirect after applying the network context to the request, and bail before setting the body.

packages/react-router/README.md
lemonmade: Good idea 👍

packages/react-router/README.md
lemonmade: I find `children` to not be a great fit for this case because you will very likely also have `renderPrefetch` (at least for our case), and having them be in very different spots feels wrong to me.

packages/react-router/README.md
lemonmade: It's the entire contents of the `Router` component file, do you think it's worth duplicating?

packages/react-router/README.md
tleunen: Don't you think that we'll always use `toRegex()` and because of that it's better to have it in the library itself?
I mean, it's a lot easier to read and understand what `/product/:id` does than `/^\/products\/(\d+)(\/|$)/`

lemonmade: As soon as you have the idea of placeholders you are back in the land of being unable to provide any type guarantees about those extracted bits.  You will not always end up using a regency - many routes across our apps do not have dynamic pieces to them. The above were just for illustrating that you can use the regex/ function form here to create any API you like.

I don’t disagree that the regex is more awkward, but it also supports the full power of JS, rather than the confusing subset allowed by the path to regex stuff (https://github.com/pillarjs/path-to-regexp)

packages/react-router/README.md
TheMallen: Maybe just a link then :/

packages/react-router/README.md
TheMallen: I guess I more meant could we add that to the example :P

packages/react-router/README.md
TheMallen: I don't think there's an issue with allowing pure regexes for cases where we need them, I just feel like the path notation is a common enough convention that there's not that good of a reason to not support it. It's only less type safe than a regex if we pass it down as a dictionary. As an array of strings it's not actually any different than the value returned from regex matching.

It's likely that we'll end up using a tiny wrapper route component everywhere as is, which is maybe fine but begs the question of why not just support the common usecase directly.

packages/react-router/README.md
lemonmade: They can work however the developer likes, just like nested components. In theory, this is now totally OK:

```tsx
<NoMatch renderFallback={() => <NotFound />}>
  <Route match="/" render={() => <Home />
  <ProductRoutes />
</NoMatch>

function ProductRoutes() {
  return <>
    <Route match="/products" render={() => <ProductList />} />
    <Route match="/products/new" render={() => <ProductDetails />} />
  </>;
}
```

So there really isn't anything new to learn, and you don't need to worry about keeping the routes on a single level like with react-router.

packages/react-router/README.md
lemonmade: How would people feel if `render` got called with two arguments: the current URL and an array of strings, which in the case of a string `match` would be any segment matching an `:foo` pattern (restricted to just the simplest case), and in the case of a regular expression would be the capture groups?

PR #726, opened 2019-05-30
react-graphql improvements

lemonmade:
This PR makes a couple of small improvements to react-graphql:

1. When you pass `skip` to `useQuery` on the server, it no longer attempts to fetch anything at all (before, it would try to resolve an async GraphQL document, mark it as used, and actually run the query)
2. When variables stay deep equal, the result of `useQuery` is referentially equal (it always returned the same data, but the object itself was changed between renders, which is a bummer because it's a common one to use as an input to other `useMemo`s)

In fixing (2), I also made an improvement to `react-testing`. Previously, calling `setProps` on a custom mounted element would set props on whatever the component was wrapped in, not the component itself. The changes to `react-testing` makes it so that only the element itself is updated with new props, which removes the need for hacks like https://github.com/Shopify/web/blob/master/tests/modern/AppContext.tsx#L63-L66.

### DIFF
--- a/packages/react-graphql/src/hooks/graphql-document.ts
+++ b/packages/react-graphql/src/hooks/graphql-document.ts
@@ -2,6 +2,7 @@ import {useState, useEffect, useCallback} from 'react';
 import {OperationVariables} from 'apollo-client';
 import {DocumentNode} from 'graphql-typed';
 import {useMountedRef} from '@shopify/react-hooks';
+import {useAsyncAsset} from '@shopify/react-async';
 
 import {AsyncQueryComponentType} from '../types';
 
@@ -13,7 +14,7 @@ export default function useGraphQLDocument<
   documentOrComponent:
     | DocumentNode<Data, Variables>
     | AsyncQueryComponentType<Data, Variables, DeepPartial>,
-): [DocumentNode<Data, Variables> | null, string | undefined] {
+): DocumentNode<Data, Variables> | null {
   const [document, setDocument] = useState<DocumentNode<
     Data,
     Variables
@@ -52,10 +53,11 @@ export default function useGraphQLDocument<
     [document, loadDocument],
   );
 
-  return [
-    document,
+  useAsyncAsset(
     isDocumentNode(documentOrComponent) ? undefined : documentOrComponent.id,
-  ];
+  );
+
+  return document;
 }
 
 function isDocumentNode(arg: any): arg is DocumentNode {
--- a/packages/react-graphql/src/hooks/query.ts
+++ b/packages/react-graphql/src/hooks/query.ts
@@ -39,6 +39,7 @@ export default function useQuery<
 
   useAsyncAsset(id);
 
+  const serializedVariables = variables && JSON.stringify(variables);
   const watchQueryOptions = useMemo<WatchQueryOptions<Variables> | null>(
     () => {
       if (!query) {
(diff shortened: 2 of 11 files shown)


packages/react-graphql/src/hooks/graphql-document.ts
```diff
@@ -52,10 +53,11 @@ export default function useGraphQLDocument<
     [document, loadDocument],
   );
 
-  return [
-    document,
+  useAsyncAsset(
```
lemonmade: I also moved this back into this hook, where it feels more appropriate than the `useQuery` hook

packages/react-graphql/src/hooks/query.ts
```diff
@@ -33,12 +36,15 @@ export default function useQuery<
...
     context,
   } = options;
-
   const client = useApolloClient(overrideClient);
-  const [query, id] = useGraphQLDocument(queryOrComponent);
 
-  useAsyncAsset(id);
+  if (typeof window === 'undefined' && skip) {
+    return createDefaultResult(client, variables);
+  }
```
lemonmade: This is the early bailout

packages/react-graphql/src/hooks/query.ts
```diff
@@ -90,33 +95,9 @@ export default function useQuery<
...
-        : (noop as any),
-      subscribeToMore: queryObservable
-        ? queryObservable.subscribeToMore.bind(queryObservable)
-        : (noop as any),
-      client,
-    }),
-    [queryObservable, client, variables],
+    () => createDefaultResult(client, variables, queryObservable),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [queryObservable, client, serializedVariables],
```
lemonmade: This is the fix for the result changing when variables were deep equal

packages/react-graphql/src/hooks/tests/query.test.tsx
```diff
@@ -33,14 +28,6 @@ const mockData = {
...
 
 describe('useQuery', () => {
-  beforeEach(() => {
-    prepareAsyncReactTasks();
-  });
-
-  afterEach(() => {
-    teardownAsyncReactTasks();
-  });
-
```
lemonmade: No longer needed

packages/react-testing/src/mount.ts
```diff
@@ -105,11 +105,11 @@ export function createMount<
...
     const context = createContext(options);
-    const rendered = render(element, context, options);
 
-    const wrapper = new CustomRoot(rendered, context, root =>
-      root.find(element.type),
-    );
+    const wrapper = new CustomRoot(element, context, {
+      render: element => render(element, context, options),
+      resolveRoot: root => root.find(element.type),
+    });
```
lemonmade: The basic gist of this fix is that instead of a `Root` around the result of `render(element)`, the `Root` just gets the raw element under test, and we thread through the `render` to the component that eventually renders the component to the DOM, since it is in charge of handling updated props.

packages/react-testing/src/tests/mount.test.tsx
```diff
@@ -72,6 +73,29 @@ describe('createMount()', () => {
     expect(div).not.toContainReactComponent('span', {id: 'ShouldNotBeFound'});
   });
 
+  it('can props on a nested element even if it is wrapped in providers', () => {
```
GoodForOneFare:
```suggestion
  it('can set props on a nested element even if it is wrapped in providers', () => {
```

PR #729, opened 2019-05-31
Update Apollo dependencies for latest typedefs

GoodForOneFare:
Trying to upgrade Apollo dependencies in web, and type-check is erroring out on `FetchPolicy` no longer accepting `cache-and-network`.  This change is described in the [apollo-client docs](https://github.com/apollographql/apollo-client/blob/master/CHANGELOG.md):

> The FetchPolicy type has been split into two types, so that passing cache-and-network to ApolloClient#query is now forbidden at the type level, whereas previously it was forbidden by a runtime invariant assertion:

As far as I can see, Quilt is always using `watchQuery`, which is a valid use case for `cache-and-network`.  So let's just allow it in our types.

### DIFF
--- a/packages/react-graphql/README.md
+++ b/packages/react-graphql/README.md
@@ -222,7 +222,7 @@ This hook accepts two arguments:
 interface QueryHookOptions<Variables = OperationVariables> {
   ssr?: boolean;
   variables?: Variables;
-  fetchPolicy?: FetchPolicy;
+  fetchPolicy?: FetchPolicy | WatchQueryFetchPolicy;
   errorPolicy?: ErrorPolicy;
   pollInterval?: number;
   client?: ApolloClient<any>;
--- a/packages/react-graphql/README.md
+++ b/packages/react-graphql/README.md
@@ -222,7 +222,7 @@ This hook accepts two arguments:
 interface QueryHookOptions<Variables = OperationVariables> {
   ssr?: boolean;
   variables?: Variables;
-  fetchPolicy?: FetchPolicy | WatchQueryFetchPolicy;
+  fetchPolicy?: WatchQueryFetchPolicy;
   errorPolicy?: ErrorPolicy;
   pollInterval?: number;
   client?: ApolloClient<any>;
--- a/packages/react-graphql/src/types.ts
+++ b/packages/react-graphql/src/types.ts
@@ -8,7 +8,6 @@ import {
 } from 'graphql-typed';
 import {QueryResult} from 'react-apollo';
 import {
-  FetchPolicy,
   ErrorPolicy,
   OperationVariables,
   ApolloError,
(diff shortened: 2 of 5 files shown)


packages/react-graphql/README.md
```diff
@@ -222,7 +222,7 @@ This hook accepts two arguments:
 interface QueryHookOptions<Variables = OperationVariables> {
   ssr?: boolean;
   variables?: Variables;
-  fetchPolicy?: FetchPolicy;
+  fetchPolicy?: FetchPolicy | WatchQueryFetchPolicy;
```
lemonmade: Seems like we can just use `WatchQueryFetchPolicy` since, at least based on the changelog, it is `FetchPolicy | other policies`

packages/react-graphql/src/types.ts
```diff
@@ -27,7 +28,7 @@ export type VariableOptions<Variables> = IfEmptyObject<
 
 export type QueryProps<Data = any, Variables = OperationVariables> = {
   children: (result: QueryResult<Data, Variables>) => React.ReactNode;
-  fetchPolicy?: FetchPolicy;
+  fetchPolicy?: FetchPolicy | WatchQueryFetchPolicy;
```
lemonmade: Same

yarn.lock
```diff
@@ -883,21 +883,13 @@ apollo-link-dedup@^1.0.0:
...
-  integrity sha512-iL9yS2OfxYhigme5bpTbmRyC+Htt6tyo2fRMHT3K1XRL/C5IQDDz37OjpPy4ndx7WInSvfSZaaOTKFja9VWqSw==
-  dependencies:
-    apollo-utilities "^1.0.0"
-    zen-observable-ts "^0.8.10"
-
-apollo-utilities@1.2.1, apollo-utilities@^1.0.0, apollo-utilities@^1.2.1:
-  version "1.2.1"
-  resolved "https://registry.yarnpkg.com/apollo-utilities/-/apollo-utilities-1.2.1.tgz#1c3a1ebf5607d7c8efe7636daaf58e7463b41b3c"
-  integrity sha512-Zv8Udp9XTSFiN8oyXOjf6PMHepD4yxxReLsl6dPUy5Ths7jti3nmlBzZUOxuTWRwZn0MoclqL7RQ5UEJN8MAxg==
+apollo-utilities@1.3.0, apollo-utilities@^1.2.1, apollo-utilities@^1.3.0:
```
lemonmade: Nice

yarn.lock
```diff
@@ -7217,28 +7197,16 @@ rc@^1.0.1, rc@^1.1.6, rc@^1.1.7:
     minimist "^1.2.0"
     strip-json-comments "~2.0.1"
 
-"react-apollo@>=2.2.3 <3.0.0":
-  version "2.5.4"
-  resolved "https://registry.yarnpkg.com/react-apollo/-/react-apollo-2.5.4.tgz#4b5ad2e9e38001521d072445acdb5b3b5489b22d"
-  integrity sha512-olH9zYijOXVfj14hD7bQlZ0POBJchxg2e+mfnxEiEdqZra4+58SfIY0KPhmM9jqbeeusc6J/P4zzWIHt5DdNDg==
+"react-apollo@>=2.2.3 <3.0.0", react-apollo@^2.2.3:
```
lemonmade: Also nice

PR #731, opened 2019-06-02
MockApolloLink propagates unmocked data errors via immediate throws

GoodForOneFare:
This prevents Apollo's internals from throwing unrelated errors (it tries to reach into undefined results, and throws non-descriptive results before it can propagate MockApolloLink's actional feedback).

### Before
mock-apollo-errors-before

### After
More lines, but at least there's actionable content in here:

mock-apollo-link-after

### DIFF
--- a/packages/jest-mock-apollo/src/MockApolloLink.ts
+++ b/packages/jest-mock-apollo/src/MockApolloLink.ts
@@ -60,8 +60,7 @@ export default class MockApolloLink extends ApolloLink {
             ' (you provided a function that did not return a valid mock result)';
         }
 
-        const error = new Error(message);
-        result = error;
+        throw new Error(message);
       } else if (response instanceof Error) {
         result = {errors: [new GraphQLError(response.message)]};
       } else {
--- a/packages/jest-mock-apollo/src/test/MockApolloLink.test.ts
+++ b/packages/jest-mock-apollo/src/test/MockApolloLink.test.ts
@@ -27,14 +27,19 @@ describe('MockApolloLink', () => {
   it('returns error message with empty mock object', async () => {
     const mockApolloLink = new MockApolloLink({}, schema);
 
-    const result = await new Promise(resolve => {
-      mockApolloLink.request(mockRequest).subscribe(resolve);
-    });
-
-    expect(result).toMatchObject({
-      message:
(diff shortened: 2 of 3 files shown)


packages/jest-mock-apollo/src/MockApolloLink.ts
```diff
@@ -60,8 +60,7 @@ export default class MockApolloLink extends ApolloLink {
             ' (you provided a function that did not return a valid mock result)';
         }
 
-        const error = new Error(message);
-        result = error;
+        throw new Error(message);
```
lemonmade: Should we be doing the same down on line 76?

packages/jest-mock-apollo/src/test/MockApolloLink.test.ts
```diff
@@ -27,14 +27,19 @@ describe('MockApolloLink', () => {
...
+        mockApolloLink.request(mockRequest).subscribe({
+          next() {},
+          error(err) {
+            reject(err);
+          },
+          complete() {},
+        });
+      }),
+    ).rejects.toThrow(
+      "Can’t perform GraphQL operation 'Pets' because no valid mocks were found (it looks like you tried to provide data directly to the mock GraphQL client. You need to provide your fixture on the key that matches its operation name. To fix this, simply change your code to read 'mockGraphQLClient({Pets: yourFixture}).'",
```
lemonmade: In your travels, did you notice if it would be possible anywhere to have these kinds of things actually fail the test? I wonder if we can maybe propagate it to the method where we `resolveAllGraphQL` so that that part rejects?

packages/jest-mock-apollo/src/test/MockApolloLink.test.ts
```diff
@@ -59,13 +69,18 @@ describe('MockApolloLink', () => {
...
+      new Promise((_, reject) => {
+        mockApolloLink.request(mockRequest).subscribe({
+          next() {},
+          error(err) {
+            reject(err);
+          },
+          complete() {},
+        });
+      }),
+    ).rejects.toThrow(
```
lemonmade: Are these same changes not needed in `graphql-testing`?

GoodForOneFare:
After digging in a bit more, there's a better way to do this.

packages/jest-mock-apollo/src/test/MockApolloLink.test.ts
GoodForOneFare: I'm certain there's a way to fail fast, but I haven't found it yet :/

Unfortunately, many of these errors come from tests that never attempt to `resolveAllGraphQL`, so that wouldn't provide comprehensive coverage.

PR #735, opened 2019-06-04
Add toProvideReactContext matcher

lemonmade:
Adds a new matcher to make it easier to assert on context. While I was here, I also improved the printing of context providers and consumers.

Screen Shot 2019-06-04 at 11 53 27 AM

### DIFF
--- a/packages/react-testing/src/matchers/utilities.ts
+++ b/packages/react-testing/src/matchers/utilities.ts
@@ -62,14 +62,14 @@ export function diffs(
   props: object,
   expand?: boolean,
 ) {
-  return element.reduce<string>(
-    (diffs, element, index) =>
-      `${diffs}${index === 0 ? '' : '\n\n'}${normalizedDiff(element, props, {
-        expand,
-        showLegend: index === 0,
-      })}`,
-    '',
-  );
+  return element.reduce<string>((diffs, element, index) => {
+    const separator = index === 0 ? '' : '\n\n';
+
+    return `${diffs}${separator}${normalizedDiff(element, props, {
+      expand,
+      showLegend: index === 0,
+    })}`;
+  }, '');
 }
 
 function normalizedDiff(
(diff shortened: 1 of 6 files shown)


packages/react-testing/src/matchers/utilities.ts
```diff
@@ -57,6 +57,49 @@ export function assertIsNode(
...
 }
 
+export function diffs(
+  element: Element<any>[],
+  props: object,
+  expand?: boolean,
+) {
+  return element.reduce<string>(
+    (diffs, element, index) =>
+      `${diffs}${index === 0 ? '' : '\n\n'}${normalizedDiff(element, props, {
```
TheMallen: I guess this was just moved, but with the way prettier makes these kind of functiony ternary-y templates awful to look at I'd almost rather just do something like

```
element.map((e) => normalizedDiff(e, props, {/*stuff*/})).join('\n\n')
```

packages/react-testing/src/matchers/utilities.ts
```diff
@@ -57,6 +57,49 @@ export function assertIsNode(
...
+
+  return showLegend ? result : result.split('\n\n')[1];
+}
+
+export function printType(type: string | React.ComponentType<any>) {
+  if (typeof type === 'object' && '_context' in type) {
+    const context = (type as any)._context;
+    const componentName = type === context.Provider ? 'Provider' : 'Consumer';
+    const displayName = context.displayName || 'Context';
+    return `<${displayName}.${componentName} />`;
```
TheMallen: Neat, this is the part that fixes `<undefined />` I guess? I find these intermediary variables also much nicer than the ternary inline below, could we make it consistent?

packages/react-testing/src/matchers/utilities.ts
lemonmade: I'm not sure I understand, the index is relevant here, I can't just throw it away, and it doesn't make sense to pollute the `normalizedDiff` function with this (since its purpose is specific and different from this)

lemonmade: I cleaned it up a bit with an intermediate variable, I think you'll like it better.

PR #738, opened 2019-06-04
Don’t load translations if they don't exist

lemonmade:
Our current Babel transform attempts to return the result of importing `import('./translations/${locale}')`. This works fine, but means that for any locale other than the fallback, we always try to load that file. This is useless for locales like `en-CA`, where we do not provide explicit translations and instead just use an `en.json`. For those, it's just wasted bandwidth.

This PR fixes this issue by reading the translations that are available for a component, and bailing immediately if the requested locale is not one of those.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -590,7 +590,7 @@
   resolved "https://registry.yarnpkg.com/@types/prop-types/-/prop-types-15.5.8.tgz#8ae4e0ea205fe95c3901a5a1df7f66495e3a56ce"
   integrity sha512-3AQoUxQcQtLHsK25wtTWIoIpgYjH3vSDroZOUr7PpCHw/jLY1RB9z9E8dBT/OSmwStVgkRNvdh+ZHNiomRieaw==
 
-"@types/react-dom@^16.0.11", "@types/react-dom@^16.8.3":
+"@types/react-dom@16.8.3", "@types/react-dom@^16.0.11", "@types/react-dom@^16.8.3":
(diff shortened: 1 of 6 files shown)


lemonmade:
cc/ @rexmac and @qq99 who brought this up 🥇

yarn.lock
```diff
@@ -598,14 +612,22 @@
...
-"@types/react@*", "@types/react@16.8.10", "@types/react@16.8.2", "@types/react@>=16.4.0", "@types/react@^16.0.2":
+"@types/react@*", "@types/react@>=16.4.0", "@types/react@^16.0.2":
   version "16.8.10"
   resolved "https://registry.yarnpkg.com/@types/react/-/react-16.8.10.tgz#1ccb6fde17f71a62ef055382ec68bdc379d4d8d9"
   integrity sha512-7bUQeZKP4XZH/aB4i7k1i5yuwymDu/hnLMhD9NjVZvQQH7ZUgRN3d6iu8YXzx4sN/tNr0bj8jgguk8hhObzGvA==
   dependencies:
     "@types/prop-types" "*"
     csstype "^2.2.0"
 
+"@types/react@16.8.2":
```
GoodForOneFare: Not sure why yarn added this :/  If I delete `@types/react` lines and run yarn, the duplicate disappears.

(using yarn 1.10.1)

lemonmade: So weird, no one depends on this exact version...

PR #739, opened 2019-06-05
Unmocked mutation errors now propagate to the test console

GoodForOneFare:
This is an alternative fix to https://github.com/Shopify/quilt/pull/731.  It fixes a more specific issue, and doesn't step outside of Apollo's link chain.

### What?
Better feedback for unmocked test mutations.

#### Before

mutation-unmocked-before

### After
A little more output, but there's actionable feedback in here:

mutation-unmocked-after

### How tho?

Previously, Apollo would try to dig into `result.errors`, find nothing, and assume that it should proceed with unpacking mutation data.  This caused Apollo to throw exceptions when trying to dig into the `undefined` result.

Now, the expected mutation result shape is used, and Apollo correctly interprets our errors as... errors :)

### DIFF
--- a/packages/jest-mock-apollo/CHANGELOG.md
+++ b/packages/jest-mock-apollo/CHANGELOG.md
@@ -8,6 +8,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ---
 
+## Unreleased
+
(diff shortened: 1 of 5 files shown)


lemonmade:
@GoodForOneFare can you either update or close this?

Issue #741, opened 2019-06-06
[react-i18n] Pluralization, special case "zero" in English?


movermeyer:
## Overview

Add special handling for using the "zero" pluralization key in English, similar to [how Rails i18n does](https://guides.rubyonrails.org/i18n.html#pluralization):

```ruby
lookup_key = :zero if count == 0 && entry.has_key?(:zero)
lookup_key ||= count == 1 ? :one : :other
entry[lookup_key]
``` 

Today:
```javascript
new Intl.PluralRules('en-US').select(0)
<- "other"
```

## Motivation

We claim that:

>@shopify/react-i18n handles pluralization similarly to Rails’ default i18n utility.

However, users who have Rails code are not getting the same results as they migrate to React.

---

## Checklist

- [ ] I have described this issue in a way that is actionable (if possible)

movermeyer:
Thoughts? cc @nimzco

nimzco:
Seems like a legitimate use case. The problem should be solved on [`react-i18n`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/PluralRules) level since `Intl.PluralRules` is a [Draft Spec](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/PluralRules) (currently solved through a polyfill).

nimzco:
@TheMallen @lemonmade thoughts on this? Here's a quick patch that would solve this: https://github.com/Shopify/quilt/pull/742/files

lemonmade:
Is `zero` a valid entry to come from plural rules, per the spec? I couldn't actually see a formal list anywhere. Is this actually a thing that applies across locales (the suggested patch does not scope it to any particular language or locale)?

movermeyer:
> Is `zero` a valid entry to come from plural rules, per the spec?

I don't think so.

However, Rails has special casing for `zero`, and people use the "zero" key in English in Rails to handle cases like:

```zero: 'no messages',
one: '%{count} message',
other: '%{count} messages'```

https://guides.rubyonrails.org/i18n.html#pluralization

We claim that:

> @shopify/react-i18n handles pluralization similarly to Rails’ default i18n utility.

However, users who have Rails code are not getting the same results as they migrate to React.

lemonmade:
[...]

PR #742, opened 2019-06-06
[WIP] [React-i18n] Handle zero count

nimzco:
Fixes https://github.com/Shopify/quilt/issues/741.

If count is `0` and `zero` key exists in the dictionary, use it.


movermeyer:
@nimzco Add a comment that explains why?

nimzco:
We don't need to handle zero key. See conversation here: https://github.com/Shopify/quilt/issues/741
`zero` key exists in certain languages, see: https://github.com/Shopify/quilt/issues/741#issuecomment-499902981

PR #746, opened 2019-06-07
🐛 add in missing dependency

michelleyschen:
The two module are being use here https://github.com/Shopify/quilt/blob/master/packages/react-i18n/src/i18n.ts#L1-L2

but was never include in the dependency of the project.

### DIFF
--- a/packages/react-i18n/package.json
+++ b/packages/react-i18n/package.json
@@ -34,6 +34,8 @@
     "typescript": "~3.2.1"
   },
   "dependencies": {
+    "@shopify/decorators": "^1.1.4",
+    "@shopify/function-enhancers": "^1.0.4",
     "@shopify/i18n": "^0.1.5",
     "@shopify/react-effect": "^3.0.4",
     "@shopify/react-hooks": "^1.2.1",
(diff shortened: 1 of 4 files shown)


packages/react-i18n/package.json
```diff
@@ -35,12 +35,14 @@
   },
   "dependencies": {
     "@shopify/i18n": "^0.1.5",
+    "@shopify/javascript-utilities": "^2.4.0",
```
lemonmade: Weren't these actually missing dependencies of `i18n`? Also, can we just remove these in favour of our own stuff?

michelleyschen: it is actually in `react-i18n`,  not `i18n`. The error message is as follow.
> Unable to resolve module `@shopify/javascript-utilities/decorators` from `/Users/path/to/node_modules/@shopify/react-i18n/dist/i18n.js: Module `@shopify/javascript-utilities/decorators` does not exist in the Haste module map`

but you are right,  I can switch this one out to use our own decorator. 
thou our version of memoize skip server execution, do you think that would be a problem here?
(if not I will switch both the decorator & memoizeFn)

PR #748, opened 2019-06-10
Add support for body/html attributes

lemonmade:
This PR adds support for setting `body`/ `html` attributes from application code. It does so by augmenting the existing `HtmlManager` and adding the `useBodyAttributes`/ `<BodyAttributes />`/ `useHtmlAttributes`/ `<HtmlAttributes />` exports. Web needs this in order to properly set the HTML `lang` attribute, now that we've moved the construction of the i18n manager out of the entry points.

### DIFF
--- a/packages/react-html/src/manager.ts
+++ b/packages/react-html/src/manager.ts
@@ -64,34 +64,23 @@ export class HtmlManager {
   }
 
   addTitle(title: string) {
-    const titleObject = {title};
-    this.titles.push(titleObject);
-    this.updateSubscriptions();
-    return this.removeTitle.bind(this, titleObject);
+    return this.addDescriptor({title}, this.titles);
   }
 
   addMeta(meta: React.HTMLProps<HTMLMetaElement>) {
-    this.metas.push(meta);
-    this.updateSubscriptions();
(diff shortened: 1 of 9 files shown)


packages/react-html/src/manager.ts
```diff
@@ -123,4 +141,24 @@ export class HtmlManager {
       this.updateSubscriptions();
     }
   }
+
+  private removeHtmlAttributes(
+    attributes: React.HtmlHTMLAttributes<HTMLHtmlElement>,
+  ) {
+    const index = this.htmlAttributes.indexOf(attributes);
+
+    if (index >= 0) {
```
TheMallen: We have a lot of functions that do more or less the same thing with different collections, took me a minute to understand what the differences were and why we did it this way, might it make sense to stick this into a `removeAttributes` helper?

Probably the same deal for adding attributes too I suppose.

Alternatively attribute lists could be a class that has an `add` and `remove` function of their own and have the `updateSubscriptions` callback given to them on initialization, but that might be a little overkill.

lemonmade: DRY'ed up the methods that just did push to array + splice later on 👍

PR #749, opened 2019-06-10
Fix useQuery's side effect wrapping result under 'current' key

gavrix:
The issue I reported in *web-foundation-tech* channel a while ago. 

Screen Shot 2019-06-10 at 11 29 53 AM


gavrix:
@michelleyschen wasn't sure about version number in CHANGELOG, added under 'unreleased' — is that OK?

michelleyschen:
rebase with master and merge away!

Issue #750, opened 2019-06-10
[react-graphql] Typings for `data` is incorrect


tleunen:
Since we have our own wrapper on top of Apollo, maybe we can find a way to have better typings.

When importing the graphql file and using it inside the `useQuery` hook, the data object seems to always be `TData | {}`. But what typescript knows is more `TData | undefined`. 

```js
import SectionQuery from './graphql/SectionQuery.graphql';
const {data} = useQuery(SectionQuery);
```

![Screen Shot 2019-06-10 at 12 31 46 PM](https://user-images.githubusercontent.com/1972567/59210756-4f0d7b80-8b7c-11e9-902f-1db9ee053df7.jpg)

In the previous example, typescript tells me the type of `data` is `SectionQueryData | undefined`. But then, if I inspect the type of `SectionQueryData`, the typescript generated file gives me this:
```
export interface SectionQueryData {
  shop: SectionQueryData.Shop;
}
```

Because of that, and the wrong typing on `data`, typescript doesn't say that `data.shop` might not exist. It says that data might not exist. But if `data` exists, `shop` should exist and it's currently not the case because `data` is never undefined. Instead, it's an empty object.

Would there be a way to improve the typings for the `data` returned value?

cc @michelleyschen

PR #751, opened 2019-06-10
Fix useQuery data returned value

tleunen:
Fixes https://github.com/Shopify/quilt/issues/750.

So I was debugging my issue that I described in #750 and I noticed that we called `.currentResult` on the observable query and this public function is actually deprecated and will be removed in v3. And this function was actually replacing the `undefined` value with an empty object as described here:
https://github.com/apollographql/apollo-client/blob/547d9f518d6b3c5b27edbdb2365691498d9e5a8d/packages/apollo-client/src/core/ObservableQuery.ts#L154-L162

`.getCurrentResult` actually returns the data value without changing the undefined value into an empty object.


tleunen:
Thanks! Added changelog. Let me know if the wording is not right ;)

michelleyschen:
rebase with master and merge away!

PR #758, opened 2019-06-18
[react-i18n] Adding `unformatNumber`

lukashlavacka:
This PR adds a function to convert locale formatted number to a plain javascript number number in a string form.

*Example*

`"1,23,456.789"` in `en-IN` locale will be unformatted to `"123456.789"`

Very similar to `unformatCurrency` just without the currency stuff.

This utility will be used to unformat decimal number in https://github.com/Shopify/web/issues/14837

### DIFF
--- a/packages/react-i18n/src/i18n.ts
+++ b/packages/react-i18n/src/i18n.ts
@@ -201,38 +201,32 @@ export class I18n {
     }).format(amount);
   }
 
+  unformatNumber(input: string): string {
+    const decimalSymbol = this.numberDecimalSymbol();
+
+    const normalizedValue = normalizedNumber(input, decimalSymbol);
+
+    return normalizedValue === '' ? '' : parseFloat(normalizedValue).toString();
+  }
+
   formatCurrency(amount: number, options: Intl.NumberFormatOptions = {}) {
     return this.formatNumber(amount, {as: 'currency', ...options});
   }
 
   unformatCurrency(input: string, currencyCode: string): string {
-    const nonDigits = /\D/g;
-
     // This decimal symbol will always be '.' regardless of the locale
     // since it's our internal representation of the string
-    const symbol = this.decimalSymbol(currencyCode);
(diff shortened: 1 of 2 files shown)


lukashlavacka:
During testing I found out that `Intl` calls from test do not work as expected. It seems that calling `new Intl.NumberFormat('de-DE')` from test does not actually create the expected number format.

I found similar issue on StackOverflow https://stackoverflow.com/questions/52329629/intl-numberformat-behaves-incorrectly-in-jest-unit-test suggesting different `Intl` build config for node(?) https://github.com/nodejs/node/blob/master/BUILDING.md#build-with-full-icu-support-all-locales-supported-by-icu

packages/react-i18n/src/i18n.ts
```diff
@@ -332,6 +324,39 @@ export class I18n {
 
     return decimal.length === 0 ? DECIMAL_NOT_SUPPORTED : decimal;
   }
+
+  private numberDecimalSymbol() {
+    return this.formatNumber(1.1, {
```
lukashlavacka: This should be replaces with [Intl.NumberFormat.prototype.formatToParts(number)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NumberFormat/formatToParts) once available.

packages/react-i18n/src/test/i18n.test.ts
```diff
@@ -511,6 +511,60 @@ describe('I18n', () => {
...
+
+    afterAll(() => {
+      formatNumber.mockRestore();
+    });
+
+    const input = 123456.7891;
+
+    it('handles number with period decimal symbol', () => {
+      formatNumber.mockImplementationOnce(() => '1.1');
+      const i18n = new I18n(defaultTranslations, defaultDetails);
```
ken3650: any way to pull up this initialization? in `beforeEach`?

PR #759, opened 2019-06-20
[react-form] notEmptyString validator now rejects empty strings

patsissons:
Similar to the `notEmpty` validator, we want to disable the `skipOnEmpty` flag so that an empty string is rejected.

### DIFF
--- a/packages/react-form/src/validation/validators.ts
+++ b/packages/react-form/src/validation/validators.ts
@@ -14,7 +14,7 @@ export function notEmpty(error: ErrorContent<string>) {
 }
 
 export function notEmptyString(error: ErrorContent<string>) {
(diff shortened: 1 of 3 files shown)


packages/react-form/src/validation/validators.ts
```diff
@@ -14,7 +14,7 @@ export function notEmpty(error: ErrorContent<string>) {
 }
 
 export function notEmptyString(error: ErrorContent<string>) {
-  return validator(predicates.notEmptyString)(error);
+  return validator(predicates.notEmptyString, {skipOnEmpty: false})(error);
```
patsissons: I think this will require updating the `notEmptyString` predicate to chain off of the `notEmpty` predicate in order to protect against nullable inputs.

PR #760, opened 2019-06-20
[react-form] adding new positiveIntegerString validator

patsissons:
This validator is similar to the `positiveNumericString` validator, except that it rejects fractional numbers. This is particularly useful for validating ID numbers (such as banking numbers).

### DIFF
--- a/packages/react-form/src/validation/validators.ts
+++ b/packages/react-form/src/validation/validators.ts
@@ -17,14 +17,14 @@ export function notEmptyString(error: ErrorContent<string>) {
   return validator(predicates.notEmptyString, {skipOnEmpty: false})(error);
 }
 
+export function positiveIntegerString(error: ErrorContent<string>) {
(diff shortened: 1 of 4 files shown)


packages/react-form/src/validation/validators.ts
```diff
@@ -17,6 +17,12 @@ export function notEmptyString(error: ErrorContent<string>) {
   return validator(predicates.notEmptyString)(error);
 }
 
+export function positiveIntegerString(error: ErrorContent<string>) {
+  return validator(input => {
+    return input !== '' && (input.match(/[^0-9]/g) || []).length === 0;
```
TheMallen: might make sense to move both this and `positiveNumericString` 's predicate into `predicates` and import them here :)

patsissons: hmm yea that's a good idea

PR #762, opened 2019-06-20
react-async 2

lemonmade:
This PR basically redoes `@shopify/react-async` to add a bunch of useful features, and to provide a lot more flexibility in prefetching. The key changes/ additions are:

- Preload/ prefetch/ keep fresh are now based on hooks, not components. This allows them to be much more flexible, because the hooks simply return functions to preload/fetch/keep fresh, which can be used wherever you need a callback. This brings the possibility of doing this like prefetching on hover of some element that will bring in the async component, which up until now would have been extremely awkward.
- There is now a notion of *when* an async asset is needed. Before, we only tracked the assets that were needed for the server render. Now, we also track (and expose, through `asyncAssets.used(AssetTiming.NextPage/ AssetTiming.CurrentPage)`) assets that are not needed immediately, but will be needed soon, either because they are declaratively being preloaded (`usePreload(AsyncComponent)`/ `<AsyncComponent.Preload />`), or because they are deferred in some way. Support was also added to `@shopify/react-html` to make it easy to get `link[rel=prefetch]` tags for these assets.
- The library now supports a new form of deferring: progressive hydration. This technique works by rendering the component synchronously on the server, but *not* including its assets in the initial payload. On the client, the server-rendered markup is picked up and used until the assets load, which can be delayed until (for example) the component is in the viewport. I updated `@shopify/react-html` to use `@shopify/react-hydrate` to make this all work automatically.

The migration guide and updated `recipes` guide provide a good overview of these new features.

This PR also made it trivially easy to add things that have been missing for awhile, like `renderError` for when a promise fails to resolve, and better behaviour for async components that have already been resolved in the past.

Unfortunately, this PR is quite massive. A lot of it is tests and docs, but because `@shopify/react-graphql` is tied into this library pretty heavily, there are a lot of changes across those two + react-html.


packages/react-async/documentation/recipes.md
```diff
@@ -157,33 +147,124 @@ First, we create the async component for our order list. This component will pre
...
+
+export default function CustomerOrdersModal() {
+  const [showModal, setShowModal] = useState(false);
+  const prefetch = usePrefetch(CustomerOrdersModal);
+
+  return (
     <>
-      <OrderListEmptyStateQuery.KeepFresh />
-      <OrderListQuery.KeepFresh />
+      <CustomerOrdersModal.Preload />
```
loic-d: This will call `usePreload(CustomerOrdersModalQuery)`, right?

What does it mean to `preload` an async query component?  

`prefetch` will send the API call. `preload` will fetch the bundle for the query component? (what was generated by https://github.com/Shopify/graphql-tools-web/tree/master/packages/graphql-mini-transforms)

qq99:
I don't know that I have time to review the code changes, but I'll add my 2 cents with awkwardness on the old API.

I had a popover that wanted to fetch remote data & render that data inside it upon hover/click/focus.  Ideally, if clicked before data was ready, the popover activator spins loading.  If clicked after the data was ready, it just opens up like a typical popover.  It never opens up while data is prefetching (because it goes against my UX desires, but also because the popover can't handle changes in child height nicely).

You can hook up the asynchronous query component easily enough, and using `setState` to know to render `<TheQuery.Prefetch>` in my parent was achievable (though it doesn't look as pretty in a hook component)

The things I found myself needing:
- ability to prefetch from a parent/somewhere functionally (pretty much just a cosmetic need, it feels nicer to have my handler call a fnc than to setState & rerender), and it looks like your PR provides that
- ability to know whether the prefetch is stillLoading, to be able to render the button as `loading`
- ability to know whether the prefetch is doneLoading, to allow the popover to finally open `active={popoverOpen && doneLoading && hasData}` kinda deal

Are those latter 2 possible?

packages/react-async/README.md
```diff
@@ -135,54 +162,89 @@ const MyComponentOnIdle = createAsyncComponent({
...
+const MyModalComponent = createAsyncComponent({
+  load: () => import('./MyModalComponent'),
+  defer: ({open}) => open,
+});
+```
+
+#### Progressive hydration
 
-The library always allows you to pass an `async` prop with some custom options for the underlying `<Async />` loader component (**note**: this library reserves the `async` prop name, so you can’t use that name for any of your component’s own props, or for the props you specify in the `renderPreload`, `renderPrefetch`, or `renderKeepFresh` options).
+It can sometimes be useful to server render a component, but to wait to load its assets until later in the page lifecycle. This is particularly relevant for large, mostly static components, components that are very likely to be outside the viewport on load, and expensive components that would cause significant layout shifts if only rendered on the client. This library supports this pattern through the `deferHydration` option, and with the help from the [`@shopify/react-hydrate` package](../react-hydrate).
```
loic-d: Not sure how progressive hydration can solve layout shifts for this scenario:
- User loads a page with a component outside of the viewport at the bottom of the page. 
- Markup for this component is rendered on the server but CSS and JS assets are not loaded
- User scrolls to the bottom of the page. Assets are loaded.
- Component is re-rendered with its styles which might create a layout shift (CSS `height` is now set for example)

packages/react-async/documentation/recipes.md
```diff
@@ -5,6 +5,8 @@ This document covers a few common scenarios where `@shopify/react-async` and rel
 ## Table of contents
 
 1. [Index pages with empty states](#index-pages-with-empty-states)
+1. [UI that is initially hidden](#ui-that-is-initially-hidden)
+1. [Below-the-fold content](#below-the-fold-content)
```
loic-d: You can increment numbers.

lemonmade: Markdown auto-increments them, I prefer this in case I rearrange/ add more/ remove some.

packages/react-async/README.md
lemonmade: Yeah, you're right, I need to make sure the styles load, but not the JS.

packages/react-async/documentation/recipes.md
loic-d: TIL

packages/react-async/documentation/recipes.md
lemonmade: Yes, that's correct. Preload just brings in the assets, Prefetch runs the query, KeepFresh runs the query and polls.

lemonmade:
@qq99 I don't think this library really makes much difference one way or the other with what you are trying to achieve. What you want is basically a way for a child to communicate that it has finished loading to its parent, and to have the parent respond to that. Without suspense, the best way to do that is simply to add a prop to the component. You could then use this async stuff if you want to add prefetching.

```tsx
function MyPopover({active, open, onLoaded}) {
  const {loading} = useQuery(AsyncQuery, {skip: !active});
  useEffect(() => {
    if (!loading) { onLoaded(); }
  });
  return <Popover open={open} />;
}

function MyComponent() {
  const [active, setActive] = useState(false);
  const [open, setOpen] = useState(false);
  const prefetch = usePrefetch(MyPopover);
  
  return <>
    <button onMouseEnter={prefetch} onClick={() => setActive(true)} />
    <MyPopover active={active} open={open} onLoaded={() => setOpen(true)} />
  </>;
}
```

packages/react-async/documentation/migrations.md
```diff
@@ -0,0 +1,96 @@
...
++  useKeepFresh,
+} from '@shopify/react-async';
+import {MyComponentsQuery} from './components';
+
+const MyComponent = createAsyncComponent({
+  load: () => import('./MyComponent'),
+-  renderPreload: () => <MyComponentsQuery.Preload />,
+-  renderPrefetch: () => <MyComponentsQuery.Prefetch />,
+-  renderKeepFresh: () => <MyComponentsQuery.KeepFresh />,
++  usePreload: () => usePreload(MyComponentsQuery),
```
TheMallen: O.o this api is super weird. I feel like the `use` prefix is confusing here since these are properties not hooks?

I'd maybe drop the use prefix and have one property you can pass a component to that will automatically call the corresponding hook for each?

packages/react-async/documentation/migrations.md
```diff
@@ -0,0 +1,96 @@
...
+  return (
+    <MyComponent
+      onClick={onClick}
+-      async={{renderLoading: () => <Loading onClick={onClick} />}}
+    />
+  );
+}
+```
+
+If you were using the `async` prop for any other reason, you will need to drop down into the primitive `useAsync` instead.
```
TheMallen: Could we get an example of this?

packages/react-async/documentation/migrations.md
```diff
@@ -0,0 +1,96 @@
...
+
+### `AsyncAssetManager#used` is now a method
+
+`AsyncAssetManager#used` was a getter in the 1.x branch of `@shopify/react-async`. In order to support distinct timing for when to load assets, this getter has been turned into a method. Affected code can simply add the call expression; `AsyncAssetManager#used` without arguments will default to returning only the assets to immediately load, which was the only kind of asset before 2.x:
```
TheMallen: > Affected code can simply add the call expression

"add the call expression" is a bit of an esoteric way of saying this, might be a bit clearer as something like

> Code which was using the getter should be updated to call the method instead

packages/react-async/documentation/recipes.md
```diff
@@ -157,33 +147,124 @@ First, we create the async component for our order list. This component will pre
...
+
+```tsx
+// in /app/sections/Orders/OrderList/OrderList.tsx
+
+import React from 'react';
+import {usePrefetch} from '@shopify/react-async';
+import {CustomerOrdersModal} from './components';
+
+export default function CustomerOrdersModal() {
+  const [showModal, setShowModal] = useState(false);
```
TheMallen: This example is super nice

packages/react-async/documentation/recipes.md
```diff
@@ -157,33 +147,124 @@ First, we create the async component for our order list. This component will pre
...
+Using this technique is easy with `@shopify/react-async`. Create an async component using `createAsyncComponent`, and set the `deferHydration` option. In the case of content that is below-the-fold, you can use the `InViewport` defer timing to wait until the server-rendered UI is scrolled into view before loading its scripts.
+
+```tsx
+// in /app/sections/Orders/components/ExpensiveComponent/ExpensiveComponent/index.tsx
+
+import {createAsyncComponent, DeferTiming} from '@shopify/react-async';
+
+export default createAsyncComponent({
+  load: () => import('./ExpensiveComponent'),
+  deferHydration: DeferTiming.InViewport,
```
TheMallen: neat

packages/react-async/src/component.tsx
```diff
@@ -1,135 +1,330 @@
...
+      immediate: deferred == null,
+    });
+
+    const {current: startedHydrated} = useRef(useHydrationManager().hydrated);
+
+    if (error) {
+      return renderError(error);
+    }
+
+    let loadingMarkup: ReactNode | null = null;
```
TheMallen: These conditionals are kinda gnarly, I'd really like to see them in their own functions, or at least putting the `let` for `contentMarkup` closer to where it is actually used, as it's hard to tell at a glance whether the first couple branches change it this way.

packages/react-async/src/component.tsx
```diff
@@ -1,135 +1,330 @@
...
   defer,
-  renderLoading,
-  renderPreload = noopRender,
-  renderPrefetch = noopRender,
-  renderKeepFresh = noopRender,
+  deferHydration,
+  displayName,
+  renderLoading = noopRender,
+  renderError = defaultRenderError,
+  usePreload: useCustomPreload = noopUse,
```
TheMallen: Ah so they are named with `use` because they are actually invoked as hooks in here. 🤔

Can we add some like JSDoc to this function or something that specifies that they are directly used as hooks?

packages/react-async/src/component.tsx
```diff
@@ -1,135 +1,330 @@
...
+  Prefetch.displayName = `Async.Prefetch(${displayName})`;
+
+  function KeepFresh(options: KeepFreshOptions) {
+    useIdleCallback(useKeepFresh(options));
+    return null;
   }
 
+  KeepFresh.displayName = `Async.KeepFresh(${displayName})`;
+
   // Once we upgrade past TS 3.1, this will no longer be necessary,
```
TheMallen: I know this isn't added in this PR, but we are past 3.1 in web, is this still relevant here?

packages/react-async/src/context/assets.ts
```diff
@@ -1,23 +1,71 @@
...
-  markAsUsed(id: string) {
-    this.used.add(id);
+  private assets = new Map<string, AssetOptions>();
+
+  used(timing: AssetTiming | AssetTiming[] = AssetTiming.Immediate) {
+    const timingArray = Array.isArray(timing) ? timing : [timing];
+
+    const assets: AssetSelector[] = [];
+
+    for (const [asset, {scripts, styles}] of this.assets) {
```
TheMallen: spent a while parsing this for why it can't be a map/filter pair, not a big deal but might be more consistent with our other code /shrug

packages/react-async/src/component.tsx
```diff
@@ -1,135 +1,330 @@
...
+
+function defaultRenderError(error: Error) {
+  if (error) {
+    throw error;
+  }
+
+  return null;
+}
+
+function Loader<T>({
```
TheMallen: This thing might be worth testing in isolation / being in it's own file /shrug

packages/react-async/src/component.tsx
```diff
@@ -1,135 +1,330 @@
...
+      return <OnIdle perform={load} />;
+    case DeferTiming.InViewport:
+      return (
+        <IntersectionObserver
+          threshold={0}
+          onIntersectionChange={handleIntersection}
+        />
+      );
+    default:
+      return null;
```
TheMallen: so if I understand correctly, if I pass `defer={"Some madeup value" as any}` this will fail silently / never load?

Could we throw an error here?

packages/react-async/src/context/assets.ts
```diff
@@ -1,23 +1,71 @@
...
+    const styles = typeof timing === 'object' ? timing.styles : timing;
+
+    if (current == null) {
+      this.assets.set(id, {
+        scripts: scripts || AssetTiming.Immediate,
+        styles: styles || AssetTiming.Immediate,
+      });
+    } else {
+      this.assets.set(id, {
+        scripts: Math.max(scripts || current.scripts, current.styles),
```
TheMallen: This max/min stuff is made me go reread a bunch because I suddenly felt like I had no idea what this is doing.

Might be a good spot for a 'why' comment.

packages/react-async/src/hooks.ts
```diff
@@ -1,13 +1,121 @@
...
+        const resolved = await resolver.resolve();
+
+        if (mounted.current) {
+          // It's important to use the function form of setValue here.
+          // Resolved is going to be a function in most cases, since it's
+          // a React component. If you do not set it using the function form,
+          // React treats the component as the function that returns state,
+          // so it sets state with the result of manually calling the component
+          // (so, usually JSX).
+          setValue(() => resolved);
```
TheMallen: I hate that we have to do this :/

packages/react-async/src/hooks.ts
```diff
@@ -1,13 +1,121 @@
...
+  {assets, scripts = assets, styles = assets, immediate = true}: Options = {},
+) {
+  const [value, setValue] = useState<T | Error | null>(
+    () =>
+      immediate || typeof window !== 'undefined' ? resolver.resolved : null,
+  );
+
+  const mounted = useMountedRef();
+
+  const load = useCallback(
```
TheMallen: I guess we have to `useCallback` for this stuff because we have `useEffect`s that depend on it?

packages/react-async/src/hooks.ts
```diff
@@ -1,13 +1,121 @@
...
+  const async = useContext(AsyncAssetContext);
 
   useServerEffect(() => {
-    if (assets && id) {
-      assets.markAsUsed(id);
+    if (async && id) {
+      async.markAsUsed(id, {scripts, styles});
     }
-  });
+  }, async ? async.effect : undefined);
```
TheMallen: 🤔 this seems fishy, should we add our own throw for cases where this changes length so that it's a bit clearer if it gets hit?

packages/react-async/src/provider.tsx
```diff
@@ -15,34 +17,123 @@ interface ConsumerProps<Value> {
...
 }
 
 export function createAsyncContext<Value>({
   id,
   load,
 }: Options<Value>): AsyncContextType<Value> {
+  const resolver = createResolver({id, load});
   const Context = React.createContext<Value | null>(null);
 
+  function Root() {
```
TheMallen: Are we making this return a component exclusively for nicer errors? I don't hate that idea but we should add a comment mentioning that is why if so. (As far as I can tell this function could just return an object literal)

packages/react-async/src/tests/component.test.tsx
```diff
@@ -1,176 +1,872 @@
...
+        {element}
+      </HydrationContext.Provider>
+    );
   },
-}));
+});
 
-function MockComponent({name = 'Tobi'}: {name?: string}) {
-  return <>Hello, {name}</>;
+function ResolvedComponent({name = 'friend'}: {name?: string}) {
```
TheMallen: friend :)

packages/react-async/src/tests/component.test.tsx
```diff
@@ -1,176 +1,872 @@
...
+        expect(load).not.toHaveBeenCalled();
+      });
+
+      it('starts loading the component when it is in the viewport', () => {
+        const load = jest.fn(
+          () => createResolvablePromise(ResolvedComponent).promise,
+        );
+        const AsyncComponent = createAsyncComponent({
+          load,
+          deferHydration: DeferTiming.InViewport,
```
TheMallen: Not for this PR, but thinking about usecases for loading on scroll I wonder if we might want `DeferTiming.AlmostInViewPort` that starts loading when the screen is a certain distance away from the content? or `DeferTiming.ScrollContainer` that starts loading when the container is scrolled

packages/react-async/documentation/migrations.md
lemonmade: They are hooks, though. Whatever gets called here gets called as part of the final useX, and must obey the rules of hooks.

packages/react-async/documentation/migrations.md
lemonmade: I left the common case example above, I am not aware of any other specific examples, but I just know they’d all be achievable with the primitive API

packages/react-async/src/component.tsx
lemonmade: Isn’t that what useX signals everywhere else?

packages/react-async/src/component.tsx
lemonmade: No, it’s an implementation detail. All of this is tested in the tests for the actual public API

packages/react-async/src/component.tsx
lemonmade: Do we usually guard against people intentionally ignoring the type system for no reason?

packages/react-async/src/hooks.ts
lemonmade: Indeed

packages/react-async/src/tests/component.test.tsx
lemonmade: The API is pretty unworkable once you get more complex, I think cases like that would be better off using useAsync directly.

packages/react-async/src/component.tsx
TheMallen: yeah, I would just weird when they are parameters to a method

packages/react-async/src/component.tsx
TheMallen: We also don't usually have a code path for it /shrug

packages/react-async/src/tests/component.test.tsx
TheMallen: fair

packages/react-async/src/component.tsx
lemonmade: Yeah I agree with that, I'll add some JSDoc to it.

packages/react-async/src/context/assets.ts
lemonmade: Personally, I'd like to see this style become a bit more common, it's totally legit and works regardless of what type `assets` is, assuming it is utterable.

packages/react-async/src/component.tsx
lemonmade: I guess so, I think if this were gone there would be the inverse question of "what happens if it doesn't match the provided cases?"

packages/react-async/src/context/assets.ts
lemonmade: Agreed :+1:

packages/react-async/src/hooks.ts
lemonmade: lol agreed

packages/react-async/src/hooks.ts
lemonmade: not sure what you mean by changes length here, the second argument to this function is an effect, not an array.

packages/react-async/src/provider.tsx
lemonmade: Will add a comment, because yeah, it's just for errors.

PR #766, opened 2019-06-25
Reinstate translation key exists method

CameronGorrie:
[Legacy-i18n had a `translationKeyExists` method](https://github.com/Shopify/web/blob/master/packages/@shopify/legacy-i18n/src/decorators.tsx#L294). This PR adds it to the new library. Some alternatives are discussed [here](https://github.com/Shopify/web/pull/15137#discussion_r296862478).

📔will update the README if this is deemed the correct approach

### DIFF
--- a/packages/react-i18n/README.md
+++ b/packages/react-i18n/README.md
@@ -241,6 +241,16 @@ i18n.translate(key, {scope: 'MyComponent.option'});
 i18n.translate(key, {scope: ['MyComponent', 'option']});
 ```
 
+It may be necessary to check dynamic keys. You can use the `translationKeyExists` method to do so:
+
+```ts
+const keyExists = i18n.translationKeyExists(key);
+
+if (keyExists) {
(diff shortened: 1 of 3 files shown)


packages/react-i18n/README.md
```diff
@@ -241,6 +241,16 @@ i18n.translate(key, {scope: 'MyComponent.option'});
 i18n.translate(key, {scope: ['MyComponent', 'option']});
 ```
 
+It may be necessary to check dynamic keys. You can use the `translationKeyExists` method to do so:
+
+```ts
+const keyExists = i18n.translationKeyExists(key);
```
tleunen: Just discovered the function, and docs. And it seems to be an error here. The `key` here must be `MyComponent.option.${key}`, right?

Also, any reason not to have the same syntax as `.translate`? Kinda weird if you have `.translate(xxx, {scope: yyy})` on one side, and something else on the other side.

cc @CameronGorrie @TheMallen

PR #769, opened 2019-06-27
Don't call betweenEachPass on any kind of last pass

lemonmade:
This PR fixes some logical errors in `react-effect`’s `betweenEachPass`. We use this hook as a way to clear out any results in managers when there is another pass to perform. This prevents stale data from a previous pass ending up in a subsequent one. However, we were calling this even when we had reached the `maxPasses`. In practice, this meant that any code reaching the maximum number of passes in Web would run the `betweenEachPass` of `react-async`, which clears out the cache of used assets, so no async assets would ever be reported 😬.

cc/ @rox163

### DIFF
--- a/packages/react-effect/CHANGELOG.md
+++ b/packages/react-effect/CHANGELOG.md
@@ -7,7 +7,7 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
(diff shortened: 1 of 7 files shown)


GoodForOneFare:
Oh, `react-network` tests are failing?  I guess that passing and a CHANGELOG entry would be nice.

Issue #763, opened 2019-06-20
Current code break react hook rules


michenly:
## Overview

https://github.com/Shopify/quilt/blob/master/packages/react-graphql/src/hooks/query.ts#L41-L43

The above code break [react hook rules](https://reactjs.org/docs/hooks-rules.html) and need to be corrected.

This bug was introduced in [#726](https://github.com/Shopify/quilt/pull/726) and the fix should contain two parts investigations.

1. Figure out why the [`react-hooks` linter](https://www.npmjs.com/package/eslint-plugin-react-hooks) did not break when this was introduced 

2. The fix should not have early return but also does not attempt to resolve an async GraphQL document on the server when skip.

## Consuming repo

> What repo were you working in when this issue occurred?

`Shopify/web`

## Labels

- [x] Add the `Type: Bug` label to this issue.

### Area

- [x] Add any relevant `Area: <area>` labels to this issue

### Scope

- Is this issue related to a specific package?

  - [x] Tag it with the `Package: <package_name>` label.

---

## Checklist

- [ ] Please delete the labels section before submitting your issue
- [ ] I have described this issue in a way that is actionable (if possible)

PR #770, opened 2019-06-27
Fix cache-less queries in SSR

lemonmade:
There are two types of `fetchPolicies` that ignore the cache: `network-only` and `no-cache`. We aren't currently handling either of them well: AFAICT, in both cases queries lead to infinite looping while trying to resolve during SSR, because the client will make a fresh query every time. This PR is a stab at addressing them that I *think* makes sense. In the case of `network-only`, we can transform it to be `cache-first` in SSR, so that it will save to cache and re-use for all subsequent server renders. `no-cache` has a promise about not putting stuff in the cache, so for those, I've updated the code to consider them equivalent to `skip`ped queries.

cc/ @rox163 while I was playing with this locally, it reduced the number of render loops needed for products pages to just 3 (one for the first set of queries, one for the queries rendered by the product show page, and one last one where it finds no more queries to resolve).

### DIFF
--- a/packages/react-graphql/src/hooks/query.ts
+++ b/packages/react-graphql/src/hooks/query.ts
@@ -44,6 +44,10 @@ export default function useQuery<
 
   const query = useGraphQLDocument(queryOrComponent);
 
+  const normalizedFetchPolicy =
+    typeof window === 'undefined' && fetchPolicy === 'network-only'
+      ? 'cache-first'
+      : fetchPolicy;
   const serializedVariables = variables && JSON.stringify(variables);
   const watchQueryOptions = useMemo<WatchQueryOptions<Variables> | null>(
     () => {
@@ -55,7 +59,7 @@ export default function useQuery<
         query,
         context,
         variables,
-        fetchPolicy,
+        fetchPolicy: normalizedFetchPolicy,
         errorPolicy,
         pollInterval,
         notifyOnNetworkStatusChange,
@@ -67,7 +71,7 @@ export default function useQuery<
       // eslint-disable-next-line react-hooks/exhaustive-deps
       context && JSON.stringify(context),
       serializedVariables,
-      fetchPolicy,
+      normalizedFetchPolicy,
       errorPolicy,
       pollInterval,
       notifyOnNetworkStatusChange,
--- a/packages/react-graphql/src/hooks/query.ts
+++ b/packages/react-graphql/src/hooks/query.ts
@@ -38,8 +38,8 @@ export default function useQuery<
   } = options;
   const client = useApolloClient(overrideClient);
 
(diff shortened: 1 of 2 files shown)


packages/react-graphql/src/hooks/query.ts
```diff
@@ -38,12 +38,16 @@ export default function useQuery<
   } = options;
   const client = useApolloClient(overrideClient);
 
-  if (typeof window === 'undefined' && skip) {
+  if ((typeof window === 'undefined' && skip) || fetchPolicy === 'no-cache') {
```
michelleyschen: this early return actually breaks [react hook rule](https://github.com/Shopify/quilt/issues/763). Is there anyway not to do this?

packages/react-graphql/src/hooks/query.ts
```diff
@@ -38,12 +38,16 @@ export default function useQuery<
...
     return createDefaultResult(client, variables);
   }
 
   const query = useGraphQLDocument(queryOrComponent);
 
+  const normalizedFetchPolicy =
+    typeof window === 'undefined' && fetchPolicy === 'network-only'
+      ? 'cache-first'
```
GoodForOneFare: A test explaining this behaviour would be grand.

packages/react-graphql/src/hooks/query.ts
GoodForOneFare: I think I'm just not getting it, but should this be:
```
if (typeof window === 'undefined' && (skip || fetchPolicy === 'no-cache')) {
```

lemonmade: Yup, you're absolutely right

lemonmade: The rules of hooks don't matter for server rendering. They only exist so that a single component, when updated, calls hooks in the same order, but in server renders, the entire component is thrown out each time.

packages/react-graphql/src/hooks/query.ts
lemonmade: I'm a little conflicted, because it's actually very hard to test this given that Jest's default environment includes the DOM.

GoodForOneFare: A few Quilt tests use `@jest-environment node`.  Is that not possible here?

lemonmade: Hit is, and I did that. I was just being lazy before, thanks for calling it out.

packages/react-graphql/src/hooks/query.ts
michelleyschen: that make sense to me, so this is really a dev experience thing to for consumer of this package to understand why they would be getting this warning.

Issue #728, opened 2019-05-31
Add back loading prop for async translations. 


Fionoble:
## Overview

This is a request to add back the loading prop for async translations, so that the partners dashboard doesn't display the fallback locale for a split second while fetching the current translations. This is happening because partner is still using client side rendering, which makes sense why other apps like web didn't need this feature since they are using ssr.

## Type

- [x] New feature
- [ ] Changes to existing features

## Motivation

> What inspired this feature request? What problems were you facing,
> or what else makes you think this should be in `quilt`?

I believe it was a thing at one point but wasn't very used so it got 🔥

Partner's dashboard recently updated our translations to be async. This was mostly so that we wouldn't have to update a ton of files when we introduced a new language (which happened a few times recently), on top of the performance gains. 

However, now we're getting that split second view of the fallback translation before the current translation content is loaded.

The loading prop would help us know that the app should still be in the loading state when etching the translations.

## Checklist

- [x] Please delete the labels section before submitting your issue
- [x] I have described this issue in a way that is actionable (if possible)

PR #771, opened 2019-06-27
[react-i18n] Pass loading state to I18n class

Fionoble:
Closes: https://github.com/Shopify/quilt/issues/728

Add the loading state to the I18n class. 

This property is already being determined by the manager.state function but isn't being read anywhere. This PR just passes it along to the I18n class and exposes it publicly.

**Note**: I made the loading prop mandatory in the I18nDetails. This lead to me having to change several tests to include this. Another option I thought of was defaulting to false in the class itself `this.loading = loading || false;`. If this is a more desirable approach I can change it. 👍

**EDIT:** I ended up going with the default approach for less code to be changed.

### DIFF
--- a/packages/react-i18n/src/i18n.ts
+++ b/packages/react-i18n/src/i18n.ts
@@ -63,6 +63,7 @@ export class I18n {
   readonly defaultCurrency?: string;
   readonly defaultTimezone?: string;
   readonly onError: NonNullable<I18nDetails['onError']>;
+  readonly loading: boolean;
 
   get language() {
     return languageFromLocale(this.locale);
@@ -102,7 +103,8 @@ export class I18n {
       country,
       pseudolocalize = false,
       onError,
-    }: I18nDetails,
+      loading,
+    }: I18nDetails & {loading?: boolean},
   ) {
     this.locale = locale;
     this.defaultCountry = country;
@@ -110,6 +112,7 @@ export class I18n {
     this.defaultTimezone = timezone;
     this.pseudolocalize = pseudolocalize;
     this.onError = onError || defaultOnError;
+    this.loading = loading || false;
   }
 
   translate(
--- a/packages/react-i18n/src/test/i18n.test.ts
+++ b/packages/react-i18n/src/test/i18n.test.ts
@@ -35,6 +35,20 @@ describe('I18n', () => {
     });
   });
 
+  describe('#loading', () => {
+    it('is exposed publicly', () => {
+      const locale = 'en';
+      const i18n = new I18n(defaultTranslations, {locale, loading: true});
+      expect(i18n).toHaveProperty('loading', true);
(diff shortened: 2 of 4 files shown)


packages/react-i18n/src/i18n.ts
```diff
@@ -102,6 +103,7 @@ export class I18n {
       country,
       pseudolocalize = false,
       onError,
+      loading,
     }: I18nDetails,
```
lemonmade: Does this type have a `loading` property on it?

packages/react-i18n/src/test/i18n.test.ts
```diff
@@ -1011,25 +1025,29 @@ describe('I18n', () => {
 
   describe('#weekStartDay()', () => {
     it('uses the defaultCountry to get the week start day', () => {
-      const i18n = new I18n(defaultTranslations, {locale: 'en', country: 'FR'});
+      const i18n = new I18n(defaultTranslations, {
+        locale: 'en', country: 'FR'});
```
lemonmade: ?

packages/react-i18n/src/test/i18n.test.ts
```diff
@@ -1011,25 +1025,36 @@ describe('I18n', () => {
 
   describe('#weekStartDay()', () => {
     it('uses the defaultCountry to get the week start day', () => {
-      const i18n = new I18n(defaultTranslations, {locale: 'en', country: 'FR'});
+      const i18n = new I18n(defaultTranslations, {
+        locale: 'en',
+        country: 'FR',
+      });
```
Fionoble: Linter?

packages/react-i18n/src/test/i18n.test.ts
Fionoble: I ran `yarn eslint` with the `--fix` option, I believe this is the result of that? I'll revert the formatting updates.

packages/react-i18n/src/i18n.ts
Fionoble: Yeah, I added it here: https://github.com/Shopify/quilt/pull/771/files#diff-a87e8a7d929dffc3c1a320c5490eeee5R10

lemonmade: Ah, I didn't see that. I'd rather it not be added there, because it doesn't actually exist there for the manager, and it's meant to represent the overall i18n "context" in which the app currently exists (whereas loading is a temporary state that applies to individual components). Instead, I'd rather augment this argument type (so, something like `I18nDetails & {loading?: boolean}`)

Fionoble: Ah I see, will do 👍

Fionoble:
@lemonmade I added a change log entry and addressed the type concern 👍

PR #777, opened 2019-07-04
✨ @shopify/koa-react-sidebar Assets, logger and graphql features

cartogram:
## Description

This Package is still not ready for prime-time, but this PR adds Assets and GraphQL set up in the SSR render.

It also adds a simple logger.


packages/koa-react-sidecar/src/render/render.tsx
```diff
@@ -1,46 +1,106 @@
...
-    const app = renderFunction(ctx);
+  return async function renderFunction(ctx: RenderContext) {
+    const logger = getLogger(ctx);
+    const assets = getAssets(ctx);
     const networkManager = new NetworkManager();
     const htmlManager = new HtmlManager();
+    const graphQLClient =
+      graphQLClientOptions &&
+      createGraphQLClient({host: ctx.host, ...graphQLClientOptions});
+    const ApolloBridge = createApolloBridge();
```
michelleyschen: I think we should remove this if `graphQLClientOptions ` doesn't exist

packages/koa-react-sidecar/example/index.tsx
```diff
@@ -1,11 +1,23 @@
...
 }
 
 createServer({
   port: 4444,
-  render: () => <MockApp />,
+  graphQLClientOptions: {},
+  render: ({graphQLClient}) => {
+    if (graphQLClient) {
+      return (
+        <ApolloProvider client={graphQLClient}>
```
TheMallen: I'll elaborate on this more down the line but this feels like it's a bit too in the middle of too opinionated and not opinionated enough. Specifically, it feels weird that we put in a tonne of work to set up the apollo client but then the user still needs to import apollo and render the provider themselves

packages/koa-react-sidecar/src/render/render.tsx
```diff
@@ -1,46 +1,106 @@
...
+                </NetworkContext.Provider>
+              </AsyncAssetContext.Provider>
+            </HtmlContext.Provider>
           );
         },
+        afterEachPass({renderDuration, resolveDuration, index, finished}) {
+          const pass = `Pass number ${index} ${
+            finished ? ' (this was the final pass)' : ''
+          }`;
+          const rendering = `Rendering ${renderDuration}ms`;
```
TheMallen: These are kind of confusing / misleading since we just told them a pass finished but `Rendering` makes it sound like we are talking about one that is ongoing.  If I was reading I might think that it was talking about the next one.

packages/koa-react-sidecar/src/render/render.tsx
```diff
@@ -1,46 +1,106 @@
...
 
     applyToContext(ctx, networkManager);
-    const response = renderToString(<Html manager={htmlManager}>{app}</Html>);
+
+    const [styles, scripts] = await Promise.all([
+      assets.styles(),
+      assets.scripts(),
+    ]);
+
+    const serializedInitialApolloData = graphQLClient ? (
```
TheMallen: This is more of what I mean, we handle setting up a bunch of stuff for them here but it feels awkward that they need to grab a bunch of very specific libraries to use that and know about exactly how we've set up this serialization. With `extract` and the self-serializer pattern, everything else their app wants to setup they can just do with their own components, I'm wondering why this should be different. I have two possible directions I'd prefer to see this go:

- Build a self-serializer for Apollo clients / data and just let the user use that if they want. Then this library continues to only care about being a minimal setup for `react-html` apps.

- Make this even more opinionated about apollo stuff. Render the provider for them, maybe provide a simple client api that gives them access to the serialized data.

Personally, after sleeping on it and looking over this I think I prefer the first approach. Encouraging self-serializers and keeping this lightweight / leveraging `@shopify/react-html`'s superpowers of letting apps handle this stuff within their universal code seems like a big win for needing to update this thing less.

packages/koa-react-sidecar/src/graphql-client/createGraphQLClient.ts
```diff
@@ -0,0 +1,27 @@
+import {ApolloClient} from 'apollo-client';
+import {InMemoryCache, NormalizedCacheObject} from 'apollo-cache-inmemory';
+import {createApolloLink, Options} from './link';
+
+export type GraphQLClientOptions = Options & {
+  initialData?: NormalizedCacheObject;
+};
+
+export function createGraphQLClient({
```
michelleyschen: since this setup is only to connect to Shopify admin, and most of the rails + react actually connect to their own API, I dont think this setup will be useful for everyone.

I think we should instead let consumer pass their own `graphQLClient` or `createGraphQLClient`.

PR #781, opened 2019-07-08
✨ React testing / Add 'toContainReactComponentTimes' matcher 

loic-d:
## Description

This PR adds a new Jest matcher `toContainReactComponentTimes` to `react-testing`.

```
.toContainReactComponentTimes(type: string | React.ComponentType, times: number, props?: object)
```

It asserts that a component matching `type` is in the descendants of the passed node a number of times.

A similar matcher `toContainComponentTimes` is available in Shopify Web and is widely used. Adding it here so other developers could benefit from it.

Screenshots of errors messages when assertions are failing:

Screen Shot 2019-07-08 at 11 48 36 AM
Screen Shot 2019-07-08 at 11 47 48 AM

## Type of change

- [x] React-testing - Minor: New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/packages/react-testing/README.md
+++ b/packages/react-testing/README.md
@@ -438,6 +438,18 @@ expect(myComponent).toContainReactComponent('div', {
 });
 ```
 
+### `.toContainReactComponentTimes(type: string | React.ComponentType, times: number, props?: object)`
+
+Asserts that a component matching `type` is in the descendants of the passed node a number of times. If the third argument is passed, this expectation will further filter the matches by components whose props are equal to the passed object (again, asymmetric matchers are fully supported). To assert that one component is or is not the descendant of the passed node use `.toContainReactComponent` or `.not.toContainReactComponent`.
+
+```tsx
+const myComponent = mount(<MyComponent />);
+
+expect(myComponent).toContainReactComponentTimes('div', 5, {
+  'aria-label': 'Hello world',
+});
+```
+
 ### `.toProvideReactContext<T>(context: Context<T>, value?: T)`
 
 Asserts that at least one `context.Provider` is in the descendants of the passed node. If the second argument is passed, this expectation will further filter the matches by providers whose value is equal to the passed object (again, asymmetric matchers are fully supported).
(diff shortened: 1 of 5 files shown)


packages/react-testing/src/matchers/components.ts
```diff
@@ -66,3 +66,46 @@ export function toContainReactComponent<
...
+          Object.keys(props).every(key =>
+            this.equals(props[key], element.props[key]),
+          ),
+        );
+
+  const pass = foundByProps.length === times;
+
+  const message = pass
+    ? () =>
+        `${matcherHint('.not.toContainReactComponentTimes')}\n\n` +
```
TheMallen: nit: imo slightly cleaner to do 
```
[
  `my`,
  `strings`,
].join('\n')
```

packages/react-testing/README.md
```diff
@@ -438,6 +438,18 @@ expect(myComponent).toContainReactComponent('div', {
 });
 ```
 
+### `.toContainReactComponentTimes(type: string | React.ComponentType, times: number, props?: object)`
```
TheMallen: 3 positional params feels a bit :/ but I can't really think of a better way to do this API

packages/react-testing/src/matchers/components.ts
```diff
@@ -66,3 +66,46 @@ export function toContainReactComponent<
...
+        `Expected the React element:\n  ${receivedColor(node.toString())}\n` +
+        `Not to contain component:\n  ${expectedColor(printType(type))}\n` +
+        `${times} ${pluralize('time', times)}, but it did.\n`
+    : () =>
+        `${matcherHint('.toContainReactComponentTimes')}\n\n` +
+        `Expected the React element:\n  ${receivedColor(node.toString())}\n` +
+        `To contain component:\n  ${expectedColor(printType(type))}\n` +
+        `${times} ${pluralize('time', times)}, but it was found ${
+          foundByProps.length
+        }.`;
```
lemonmade: A little weird to end a sentence with a number.

loic-d: I'll add 'time(s)' at the end of the sentence.

packages/react-testing/README.md
loic-d: Same here. I think we should still expose prop matching as third param (optional).

PR #792, opened 2019-07-15
Refactor polyfills package

lemonmade:
## Description

This PR reworks our polyfills setup. This, alongside its updated usage in sewing-kit (PR coming soon), should finally fulfill the promise of full shakeable polyfills by environment. When using this alongside the sewing-kit branch, `web-proving-ground` can import the following:

```tsx
import '@shopify/polyfills/base';
import '@shopify/polyfills/intl';
import '@shopify/polyfills/fetch';
import '@shopify/polyfills/url';
```

Which generates the following for the two builds:

Screen Shot 2019-07-15 at 9 23 26 AM
Screen Shot 2019-07-15 at 9 23 15 AM

In baseline we have a larger chunk of corejs (21K) than in latest (11K). This is about what we'd expect 👍. Similarly, using these polyfills for node/ Jest correctly adds the relevant globals that are otherwise missing from those environments.


packages/polyfills/package.json
```diff
@@ -1,12 +1,14 @@
 {
-  "name": "@shopify/polyfills",
-  "version": "0.0.14",
+  "name": "lemon-polyfills",
+  "version": "0.0.16",
```
lemonmade: Will clean up before shipping

yarn.lock
```diff
@@ -612,14 +612,22 @@
...
-"@types/react@*", "@types/react@16.8.10", "@types/react@16.8.2", "@types/react@>=16.4.0", "@types/react@^16.0.2":
+"@types/react@*", "@types/react@>=16.4.0", "@types/react@^16.0.2":
   version "16.8.10"
   resolved "https://registry.yarnpkg.com/@types/react/-/react-16.8.10.tgz#1ccb6fde17f71a62ef055382ec68bdc379d4d8d9"
   integrity sha512-7bUQeZKP4XZH/aB4i7k1i5yuwymDu/hnLMhD9NjVZvQQH7ZUgRN3d6iu8YXzx4sN/tNr0bj8jgguk8hhObzGvA==
   dependencies:
     "@types/prop-types" "*"
     csstype "^2.2.0"
 
+"@types/react@16.8.2":
```
GoodForOneFare: This can be deduplicated.

packages/polyfills/src/url.node.ts
```diff
@@ -1,13 +1 @@
...
-    export interface Global {
-      URL: typeof URL;
-      URLSearchParams: typeof URLSearchParams;
-    }
-  }
-}
-
-global.URL = URL;
-global.URLSearchParams = URLSearchParams;
+// URL and URLSearchParams available as globals by default.
```
GoodForOneFare: Add `"node": ">=10.0.0"` to this package's `package.json#engines`.

packages/polyfills/.gitignore
```diff
@@ -1,2 +0,0 @@
-*.js
```
ismail-syed: Why is this being removed?

ismail-syed: Ah, is it because `dist` is already `.gitignore`'d

packages/polyfills/src/url.node.ts
lemonmade: Discussed on Slack: going to leave this off for now because both Travis and Shipit are on Node 8.x :(

PR #805, opened 2019-07-23
✨ Semaphore

atesgoral:
## Description

Introduces a counting semaphore implementation.

## Checklist

### DIFF
--- a/packages/semaphore/README.md
+++ b/packages/semaphore/README.md
@@ -1,0 +1,64 @@
+# `@shopify/semaphore`
+
+[![Build Status](https://travis-ci.org/Shopify/quilt.svg?branch=master)](https://travis-ci.org/Shopify/quilt)
+[![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](LICENSE.md) [![npm version](https://badge.fury.io/js/%40shopify%2Fsemaphore.svg)](https://badge.fury.io/js/%40shopify%2Fsemaphore.svg)  [![npm bundle size (minified + gzip)](https://img.shields.io/bundlephobia/minzip/@shopify/semaphore.svg)](https://img.shields.io/bundlephobia/minzip/@shopify/semaphore.svg)
+
+Counting [semaphore](https://en.wikipedia.org/wiki/Semaphore_(programming))
+
+## Installation
+
+```bash
+$ yarn add @shopify/semaphore
+```
+
+## Usage
+
+### Instantiation
+
+Create a semaphore instance by calling the `Semaphore` constructor with a count argument:
+
+```typescript
+const semaphore = new Semaphore(3);
+```
+
+If you need a mutex, a semaphore with a count of 1 will effectively act as one:
+
+```typescript
+const mutex = new Semaphore(1);
+```
+
+### Acquiring permits
+
+Call `.acquire()` on a Semaphore instance to acquire a permit. The result is a promise that gets resolved with a Permit instance when a permit is available:
+
+```typescript
+const permit = semaphore.acquire();
+```
+
+### Releasing permits
+
+Call the `.release()` method on a Permit instance to release it:
+
+```typescript
+permit.release();
+```
+
+## Example
+
+```typescript
+const MAX_SIMULTANEOUS_FETCHES = 2;
+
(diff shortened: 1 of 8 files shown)


packages/semaphore/README.md
```diff
@@ -0,0 +1,64 @@
+# `@shopify/semaphore`
+
+[![Build Status](https://travis-ci.org/Shopify/quilt.svg?branch=master)](https://travis-ci.org/Shopify/quilt)
+[![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](LICENSE.md) [![npm version](https://badge.fury.io/js/%40shopify%2Fsemaphore.svg)](https://badge.fury.io/js/%40shopify%2Fsemaphore.svg)  [![npm bundle size (minified + gzip)](https://img.shields.io/bundlephobia/minzip/@shopify/semaphore.svg)](https://img.shields.io/bundlephobia/minzip/@shopify/semaphore.svg)
+
+Counting [semaphore](https://en.wikipedia.org/wiki/Semaphore_(programming))
```
michelleyschen: This explanation is just a little too generic.
If you can add more explaining on what this package does.
Specially things that can answer `In x situation I would want to use this package` type question.

packages/semaphore/README.md
```diff
@@ -0,0 +1,64 @@
...
+```
+
+## Example
+
+```typescript
+const MAX_SIMULTANEOUS_FETCHES = 2;
+
+const fetchSemaphore = new Semaphore(MAX_SIMULTANEOUS_FETCHES);
+
+async function throttledFetch(path) {
```
michelleyschen: does semaphore create a throttle?

atesgoral: No, not really. I couldn't find a better term to use for this made-up example.

atesgoral: I've renamed the function to `callApi`

packages/semaphore/src/Semaphore.ts
```diff
@@ -0,0 +1,43 @@
+export class Permit {
```
macdonaldr93: I think this would be more readable with one class per file. The tests could be split this way as well

packages/semaphore/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Example
+
+```typescript
+const MAX_SIMULTANEOUS_FETCHES = 2;
+
+const fetchSemaphore = new Semaphore(MAX_SIMULTANEOUS_FETCHES);
+
+async function callApi(path) {
+  const permit = await fetchSemaphore.acquire();
```
macdonaldr93: Just to make sure I'm understanding correctly. A semaphore doesn't actually do any of the work. Essentially you're just awaiting permission to do work and then the semaphore doesn't have a concept of the work that's going to take place. It's simply there to hold until the permit is released?

packages/semaphore/src/Semaphore.ts
```diff
@@ -0,0 +1,43 @@
...
+      return new Promise(resolve => this.deferreds.push({resolve}));
+    }
+  }
+
+  private createPermit(): Permit {
+    return new Permit(() => {
+      this.availablePermits++;
+
+      if (this.deferreds.length) {
+        this.deferreds.shift()!.resolve(this.createPermit());
```
macdonaldr93: Would you mind explaining this line? Also, I'm noticing the use of `!` which would basically ignore or bypass typechecking for `null` or `undefined`. It's because you're checking `.length` right before?

packages/semaphore/src/Semaphore.ts
atesgoral: I didn't want to elevate Permit to a top-level concept. The only reason I ended up defining this as a class here was to make TypeScript-ifying things easier. In my original implementation this is just a lightweight YOLO object.

packages/semaphore/README.md
atesgoral: Correct! A semaphore is a gatekeeper.

packages/semaphore/src/Semaphore.ts
atesgoral: "If there are pending requests to acquire a permit, pop the first one (first in is first out) and resolve the pending promise of that request."

Yes, `!` is to silence TypeScript since I'm already checking length.

PR #809, opened 2019-07-24
✨New package `react-server`

cartogram:
## Description

This PR adds a new package for providing a server that uses our `react-html` and `react-effect`, giving both rails and node apps a turn-key SSR solution.

I will do a documentation update once the general API is approved.


packages/react-self-serializers/README.md
```diff
@@ -92,3 +92,15 @@ function App({locale}: {locale?: string}) {
   return <I18n locale={locale}>{/* rest of the app */}</I18n>;
 }
 ```
+
```
cartogram: this got onto the wrong branch, ignore

packages/react-server/src/render/render.tsx
```diff
@@ -0,0 +1,85 @@
...
+    applyToContext(ctx, networkManager);
+
+    const immediateAsyncAssets = asyncAssetManager.used(AssetTiming.Immediate);
+    const [styles, scripts] = await Promise.all([
+      assets.styles({name: 'main', asyncAssets: immediateAsyncAssets}),
+      assets.scripts({name: 'main', asyncAssets: immediateAsyncAssets}),
+    ]);
+
+    const response = renderMarkup(
+      <Html locale="fr" manager={htmlManager} styles={styles} scripts={scripts}>
```
TheMallen: @GoodForOneFare  the version of this in [the rails proving ground](https://github.com/Shopify/web-rails-proving-ground/blob/master/packages/%40shopify/react-server/render/render.tsx#L80) is more up-to-date and has the locale functionality completed.

packages/react-server/README.md
```diff
@@ -0,0 +1,57 @@
+# `@shopify/react-server`
```
TheMallen: I still like "Loom" :P

packages/react-server/src/render/render.tsx
GoodForOneFare: Ah, cool cool.  I'll fold that into this PR. Thanks! 😸

PR #815, opened 2019-07-25
🐛 fix version number and broken link

michelleyschen:


packages/react-html/CHANGELOG.md
```diff
@@ -8,7 +8,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## Unreleased
 
-## [3.0.1] - 2019-07-04
+## [9.0.1] - 2019-07-04
```
ismail-syed: The square bracketed text generate hyperlinks when the text is referenced to that link. In this case, we'd need something like the following

```
[9.0.1]: https://github.com/Shopify/quilt/compare/@shopify/react-html@9.0.0...@shopify/react-html@9.0.1
```

Which links to changes from `@shopify/react-html@9.0.0` to @shopify/react-html@9.0.1`. 

You can also reference sewing-kit's, https://github.com/Shopify/sewing-kit/blame/master/CHANGELOG.md#L1902-L1907

michelleyschen: I thought the square bracket is more of a stylistic choice tbh 😂

michelleyschen: so I tried it out and didn't love it.

because of the monorepo nature of quilt, comparing between tag will also pick up changes thats does not belong to `react-html` at all 👇 

I think I am just going to remove all the square bracketed

Screen Shot 2019-07-26 at 11 39 43 AM

PR #816, opened 2019-07-25
✨ ResourcePool

atesgoral:
## Description

ResourcePool class: Uses `@shopify/semaphore` to manage access to a pool of resources.

## Checklist


atesgoral:
Closing. Will let this simmer a bit in actual use and later make the call whether we should publish it.

Issue #713, opened 2019-05-21
[react-i18n] Component not getting translations


jgodson:
## Overview
This happens frequently when doing in app navigation through our app (react-router).

The first time a component is rendered it stays using the `en` translations and the other translations never show up on the `i18n` object under `translations` even through many re-renders of the component. The same component rendered a second time on the page _does_ have the translations though. Navigating (in-app) and returning will render with the translations properly. Doing a page refresh _usually_ works, but also sometimes causes this same issue. It's definitely much more noticeable when doing in-app navigation though.
<details>
<summary>Screenshot</summary>

![alt](https://screenshot.click/21-45-a180x-i9xmz.jpg)
</details>

## Consuming repo
https://github.com/Shopify/parsnip
PR - https://github.com/Shopify/parsnip/pull/3143

andrewlo:
I'm seeing this issue after updating from v0.11.2 to v1.2.3. I was using the babel plugin on v0.11.2, so that doesn't seem to be the problem.

I see that around 25-50% of the time it renders in English instead of the language set, when doing refresh or navigations within the app

cc @lemonmade

lemonmade:
I need a specific page and reproduction steps to diagnose.

jgodson:
@lemonmade for parsnip, install the app, pick `Esty` as a platform, grab a file from the [fixtures](https://github.com/Shopify/parsnip/tree/master/spec/fixtures/files/etsy) and drag it in. You'll see this behaviour through the regular app flow (run an import by continuing once the file is parsed).

PR #824, opened 2019-08-07
Fix subscription updates being dropped

lemonmade:
## Description

Fixes https://github.com/Shopify/quilt/issues/713 (and hopefully a few project-specific issues, like https://github.com/Shopify/fbs/issues/5714). This PR fixes an issue where components with async translations were not always updating once the translations were loaded.

This issue was fairly intermittent on every example I had, but I finally managed to catch it in the act with a debugger. What I found was that the async translation was being loaded just fine, as evidenced by the fact that, on the second time the component loaded, its translations were present. However, in some cases, the async translation finished loading so quickly that it happened before the `useEffect` hook where the `useI18n` hook subscribes for updates to translations it cares about. So, the hook had not yet subscribed, and the manager had nothing to update. The loaded translation effectively got "lost" for the mounted component.

This PR addresses the problem by moving the subscription to be done synchronously, so it is guaranteed to be performed before the async translation is marked as loaded. This fixed the issue in FBS when I applied the changes locally.

cc/ @lsit @andrewlo @jgodson since I think you have all indicated this affected your projects.

## Checklist

### DIFF
--- a/packages/react-i18n/src/hooks.tsx
+++ b/packages/react-i18n/src/hooks.tsx
@@ -41,6 +41,10 @@ function useComplexI18n(
   {id, fallback, translations}: Partial<RegisterOptions>,
   manager: I18nManager,
 ): Result {
+  const managerRef = React.useRef<I18nManager | null>(null);
+  const unsubscribeRef = React.useRef<ReturnType<I18nManager['subscribe']>>(
+    noop,
+  );
   const parentIds = React.useContext(I18nIdsContext);
 
   // Parent IDs can only change when a parent gets added/ removed,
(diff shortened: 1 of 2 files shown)


packages/react-i18n/src/hooks.tsx
```diff
@@ -49,8 +53,30 @@ function useComplexI18n(
...
-    manager.register({id, translations, fallback});
+  // When the manager changes, we need to do the following IMMEDIATELY (i.e.,
+  // not in a useEffect callback):
+  //
+  // 1. Register the component’s translations. This ensures that the first render gets
+  //    the synchronous translations, if available.
+  // 2. Unsubscribe from changes to a previous manager.
+  // 3. Subscribe to changes from the new manager. This ensures that if the subscription
+  //    is updated between render and `useEffect`, the state update is not lost.
+  if (manager !== managerRef.current) {
```
TheMallen: Is there any way we can add a regression test for this?

lemonmade: I can't think of how to test the actual issue this PR addresses. I could definitely test the "manager changing" bit, but I don't think that's really a common case to worry about. The problem here is one of timing, and I am not really sure how I can force an async translation to be marked as complete (and have the resulting enqueued update happen) before `useEffect` is called.

jgodson:
Just FYI. I've tested this in https://github.com/Shopify/parsnip/pull/3735 and it seems to have fixed the issue. Thanks Chris!

andrewlo:
We've also updated to grab this fix, thanks Chris!

PR #829, opened 2019-08-08
Added debounce and throttle

AndrewMusgrave:
## Description

Add debounce and throttle to function enhancers [from polaris-rails](https://github.com/Shopify/polaris-rails/blob/master/admin/lib/utilities.js) and the [tests](https://github.com/Shopify/polaris-rails/blob/258b0d0660b92eecfb6c7b94b829f1bfc3e8d21c/test/javascripts/unit/admin/utilities_test.js) 🙇‍♂️ and typescriptify it

## Type of change

## Checklist

- [ ] I have prefixed my pull request title with the corresponding emoji from [this guide](https://gitmoji.carloscuesta.me/)


packages/function-enhancers/src/debounce.ts
```diff
@@ -0,0 +1,75 @@
+export interface Options {
+  leading?: boolean;
+  trailing?: boolean;
+  maxWait?: number;
+}
+
+export default function debounce<F extends (...args: any[]) => void>(
```
TheMallen: A TSDoc comment might be nice here to get some tooltip explanations of what the options here do. We do some of that in Quilt libraries such as `react-form` and I think here it could be helpful.

TheMallen:
PS @AndrewMusgrave  sorry for the long wait

Issue #825, opened 2019-08-07
🔥 Deprecate / Delete koa-react-sidecar


TheMallen:
`@shopify/react-server` is the successor of `koa-react-sidecar`. With it's release we can go ahead and deprecate + delete the code for `koa-react-sidecar`.

PR #830, opened 2019-08-08
🔥 koa-react-sidecar

michelleyschen:
Fixes https://github.com/Shopify/quilt/issues/825


Becojo:
Could one of the collaborators mark the package as deprecated on NPM?

PR #831, opened 2019-08-08
♻️ move dependencies to peerDependencies

michelleyschen:
I discover this while adding it to `web-proving-ground`, where the project actually have two version of `@shopify/react-html` and we run into error.

To avoid this going forward, I move all `dependencies` into `peerDependencies` and provided a greater range of support.

### DIFF
--- a/packages/react-self-serializers/package.json
+++ b/packages/react-self-serializers/package.json
@@ -23,11 +23,6 @@
     "url": "https://github.com/shopify/quilt/issues"
   },
   "homepage": "https://github.com/Shopify/quilt/blob/master/packages/react-self-serializers/README.md",
-  "dependencies": {
-    "@shopify/react-graphql": "^4.0.3",
-    "@shopify/react-html": "^8.1.3",
-    "@shopify/react-i18n": "^1.5.1"
-  },
   "devDependencies": {
     "@shopify/react-testing": "^1.7.3",
     "@types/faker": "4.1.5",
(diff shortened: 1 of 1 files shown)


packages/react-self-serializers/package.json
```diff
@@ -23,19 +23,17 @@
...
     "@shopify/react-testing": "^1.7.3",
     "@types/faker": "4.1.5",
     "faker": "^4.1.0",
     "typescript": "~3.2.1"
   },
   "peerDependencies": {
-    "react": ">=16.8.0"
+    "react": ">=16.8.0",
+    "@shopify/react-graphql": ">=4.0.3",
+    "@shopify/react-html": ">=2.0.0",
```
michelleyschen: when `useSerialized` is added

packages/react-self-serializers/package.json
```diff
@@ -23,19 +23,17 @@
...
   "devDependencies": {
     "@shopify/react-testing": "^1.7.3",
     "@types/faker": "4.1.5",
     "faker": "^4.1.0",
     "typescript": "~3.2.1"
   },
   "peerDependencies": {
-    "react": ">=16.8.0"
+    "react": ">=16.8.0",
+    "@shopify/react-graphql": ">=4.0.3",
```
michelleyschen: when `createGraphQLClient` was added

packages/react-self-serializers/package.json
```diff
@@ -23,19 +23,17 @@
...
+    "@shopify/react-html": ">=2.0.0",
+    "@shopify/react-i18n": ">=1.2.3"
```
michelleyschen: when `I18nDetails` was added

PR #832, opened 2019-08-09
🐛 [react-graphql] Add missing dependencies

tleunen:
## Description

A couple of direct dependencies were introduced in https://github.com/Shopify/quilt/pull/810 but were missing in package.json

## Checklist


packages/react-graphql/package.json
```diff
@@ -27,7 +27,9 @@
     "@shopify/async": "^2.0.0",
     "@shopify/react-async": "^3.0.2",
     "@shopify/react-hooks": "^1.2.1",
+    "@shopify/react-network": "^3.1.3",
     "@shopify/useful-types": "^1.3.0",
+    "apollo-cache-inmemory": "^1.6.2",
```
michelleyschen: this package is only for type, we can do dev dependency

tleunen: The `InMemoryCache` class is also used
https://github.com/Shopify/quilt/blob/c3b3a09b0953d2090a56fd4e597cbc9c54932248/packages/react-graphql/src/createClient.ts#L23

michelleyschen: nvm!

Issue #787, opened 2019-07-12
[I18N] Create a name formatter module


mehdi-loup:
## Overview

This module will offer you a way to display user name regarding locale preferences.

## Type

- [x] New feature
- [ ] Changes to existing features

## Motivation

In country like 🇯🇵 for example, it's offensive to display first_name before last_name. 

### Area

- [x] Add any relevant `Area: <area>` labels to this issue

### Scope

- Is this issue related to a specific package?

  - [x] Tag it with the `Package: <package_name>` label.

- Is it related to a new package?
  - [ ] Tag it with the `New package` Label

---

## Checklist

- [x] Please delete the labels section before submitting your issue
- [x] I have described this issue in a way that is actionable (if possible)

mateus:
@goldenson do you know if the I18n team is looking towards adding something like this to Web? Based on this PR https://github.com/Shopify/shopify/pull/207667

goldenson:
@nimzco @adjnor do you know if somebody is working on that?

adjnor:
Not that I know of 🤔

nimzco:
That could be a good first task for Benoit maybe.
cc @trishrempel

benoitzohar:
Looking into it 👍

PR #834, opened 2019-08-09
 ✨ Add ability to format names based on locale

benoitzohar:
## Description

Fixes https://github.com/Shopify/quilt/issues/787

Gives the ability to format a person's name properly based on their locale.

In country like 🇯🇵 for example, it's offensive to display first_name before last_name. We provide the following method to solve this problem:

- Add `formatName(firstName, lastName, {full: true})` in the react-i18n package

## Notes

The locales using the most frequent usage (`firstName lastName`) are not listed in the constant file on purpose: reduce the amount of unnecessary data and reduce the run time of going through the Map.

## Type of change

- [X] `react-i18n` Minor: New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/packages/react-i18n/README.md
+++ b/packages/react-i18n/README.md
@@ -123,8 +123,8 @@ export default withI18n()(NotFound);
 The provided `i18n` object exposes many useful methods for internationalizing your apps. You can see the full details in the [`i18n` source file](https://github.com/Shopify/quilt/blob/master/packages/react-i18n/src/i18n.ts), but you will commonly need the following:
 
 - `formatNumber()`: formats a number according to the locale. You can optionally pass an `as` option to format the number as a currency or percentage; in the case of currency, the `defaultCurrency` supplied to the i18n `I18nContext.Provider` component will be used where no custom currency code is passed.
-- `formatCurrency()`: formats a number as a currency according ot the locale. Convenience function that simply _auto-assigns_ the `as` option to `currency` and calls `formatNumber()`.
-- `formatPercentage()`: formats a number as a percentage according ot the locale. Convenience function that simply _auto-assigns_ the `as` option to `percent` and calls `formatNumber()`.
+- `formatCurrency()`: formats a number as a currency according to the locale. Convenience function that simply _auto-assigns_ the `as` option to `currency` and calls `formatNumber()`.
+- `formatPercentage()`: formats a number as a percentage according to the locale. Convenience function that simply _auto-assigns_ the `as` option to `percent` and calls `formatNumber()`.
 - `formatDate()`: formats a date according to the locale. The `defaultTimezone` value supplied to the i18n `I18nContext.Provider` component will be used when no custom `timezone` is provided. Assign the `style` option to a `DateStyle` value to use common formatting options.
   - `DateStyle.Long`: e.g., `Thursday, December 20, 2012`
   - `DateStyle.Short`: e.g., `Dec 20, 2012`
@@ -132,6 +132,9 @@ The provided `i18n` object exposes many useful methods for internationalizing yo
   - `DateStyle.Time`: e.g., `11:00 AM`
 - `weekStartDay()`: returns start day of the week according to the country.
 - `getCurrencySymbol()`: returns the currency symbol according to the currency code and locale.
+- `formatName()`: formats a name (first name and/or last name) according to the locale. e,g
+  - `formatName('John', 'Smith')` will return `John` in Germany and `Smith様` in Japan
+  - `formatName('John', 'Smith', {full: true})` will return `John Smith` in Germany and `SmithJohn` in Japan
 
 Most notably, you will frequently use `i18n`’s `translate()` method. This method looks up a key in translation files that you supply based on the provided locale. This method is discussed in detail in the next section.
 
--- a/packages/react-i18n/src/constants/index.ts
+++ b/packages/react-i18n/src/constants/index.ts
@@ -180,3 +180,21 @@ export {
   default as currencyDecimalPlaces,
   DEFAULT_DECIMAL_PLACES,
 } from './currency-decimal-places';
+
+export const CUSTOM_NAME_FORMATTERS = new Map([
+  [
+    'ja',
+    (firstName: string, lastName: string, full: boolean) =>
+      full ? `${lastName}${firstName}` : `${lastName}様`,
+  ],
+  [
+    'zh-CN',
+    (firstName: string, lastName: string, full: boolean) =>
+      full ? `${lastName}${firstName}` : lastName,
+  ],
+  [
+    'zh-TW',
+    (firstName: string, lastName: string, full: boolean) =>
+      full ? `${lastName}${firstName}` : lastName,
+  ],
+]);
--- a/packages/react-i18n/src/i18n.ts
+++ b/packages/react-i18n/src/i18n.ts
@@ -17,6 +17,7 @@ import {
   Weekday,
   currencyDecimalPlaces,
   DEFAULT_DECIMAL_PLACES,
+  CUSTOM_NAME_FORMATTERS,
 } from './constants';
 import {
   MissingCurrencyCodeError,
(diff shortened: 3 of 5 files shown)


packages/react-i18n/src/constants/index.ts
```diff
@@ -180,3 +180,17 @@ export {
...
+
+export enum FullNameFormat {
+  StartsWithLastNameNoSpace = 'StartsWithLastNameNoSpace',
+}
+
+export const CUSTOM_FULL_NAME_FORMATS = new Map([
+  ['ja', FullNameFormat.StartsWithLastNameNoSpace],
+  ['zh-CN', FullNameFormat.StartsWithLastNameNoSpace],
+  ['zh-TW', FullNameFormat.StartsWithLastNameNoSpace],
+]);
```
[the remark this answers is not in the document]
lemonmade: Can we get to a single method with an options object instead? I think the formatters could likewise be used for both if they took first name/ last name/ `full`

packages/react-i18n/src/test/i18n.test.ts
```diff
@@ -25,6 +27,7 @@ describe('I18n', () => {
   beforeEach(() => {
     translate.mockReset();
     getCurrencySymbol.mockReset();
+    getFallbackLocale.mockReset();
```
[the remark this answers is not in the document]
lemonmade: I think the fact that you are only mocking it out to return the actual results you would get with the real version means it would be better to just use the real function instead. You can do that in the mock by including the real module:

```ts
jest.mock('../utilities', () => {
  ...require.requireActual('../utilities'),
  mockOne: jest.fn(),
});
```

packages/react-i18n/src/utilities/locale.ts
```diff
@@ -0,0 +1,6 @@
+export function getFallbackLocale(locale: string) {
+  if (locale && locale.length > 3 && locale.includes('-')) {
+    return locale.split('-')[0];
```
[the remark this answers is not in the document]
lemonmade: You're fine to import it from that package 👍

packages/react-i18n/src/i18n.ts
```diff
@@ -308,6 +312,53 @@ export class I18n {
...
+
+    const {locale} = this;
+
+    let nameFormat = CUSTOM_FULL_NAME_FORMATS.get(locale);
+    if (!nameFormat) {
+      const fallBackLocale = getFallbackLocale(locale);
+      if (fallBackLocale) {
+        nameFormat = CUSTOM_FULL_NAME_FORMATS.get(fallBackLocale);
+      }
+    }
```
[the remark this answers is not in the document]
benoitzohar: As it turns it, the `fallbackLocale`, or "`language`", was already available in this class using an existing function from the `@shopify/i18n` package.
Hence I managed to shrink it down to a one-liner:
```ts
const nameFormat = CUSTOM_FULL_NAME_FORMATS.get(this.locale) || CUSTOM_FULL_NAME_FORMATS.get(this.language);
```

Same goes for `formatFullName()`  🎉

benoitzohar:
As per @lemonmade suggestion, I merged both `formatName()` and `formatFullName()` into a single function which accept an `options` object: `{full?: boolean}` and returns `firstName  lastName` if `full === true`, and `firstName` if `full !== true`.

packages/react-i18n/README.md
```diff
@@ -123,15 +123,18 @@ export default withI18n()(NotFound);
...
 - `formatDate()`: formats a date according to the locale. The `defaultTimezone` value supplied to the i18n `I18nContext.Provider` component will be used when no custom `timezone` is provided. Assign the `style` option to a `DateStyle` value to use common formatting options.
   - `DateStyle.Long`: e.g., `Thursday, December 20, 2012`
   - `DateStyle.Short`: e.g., `Dec 20, 2012`
   - `DateStyle.Humanize`: e.g., `December 20, 2012`, `Today`, or `Yesterday`
   - `DateStyle.Time`: e.g., `11:00 AM`
 - `weekStartDay()`: returns start day of the week according to the country.
 - `getCurrencySymbol()`: returns the currency symbol according to the currency code and locale.
+- `formatName()`: formats a name (first name and/or last name) according to the locale. e,g
+  - `formatName('John', 'Smith')` will return `John` in Germany and `Smith様` in Japan
+  - `formatName('John', 'Smith', {full: true})` will return `John Smith` in Germany and `SmithJohn` in Japan
```
loic-d: Really nice API 👍

packages/react-i18n/src/i18n.ts
```diff
@@ -314,6 +315,29 @@ export class I18n {
     return getCurrencySymbol(locale, {currency});
   }
 
+  formatName(firstName: string, lastName?: string, options?: {full?: boolean}) {
+    if (!firstName) {
+      return lastName;
```
loic-d: Should it be `return lastName || ''`?
If `lastName` is missing, your function will return `undefined`.

packages/react-i18n/src/i18n.ts
```diff
@@ -314,6 +315,29 @@ export class I18n {
     return getCurrencySymbol(locale, {currency});
   }
 
+  formatName(firstName: string, lastName?: string, options?: {full?: boolean}) {
```
loic-d: Nitpick but if your list of options will grow, you could extract it into a type.

packages/react-i18n/src/test/i18n.test.ts
```diff
@@ -1065,6 +1066,92 @@ describe('I18n', () => {
...
+    });
+
+    it('defaults to firstName for unknown locale', () => {
+      const i18n = new I18n(defaultTranslations, {locale: 'unknown'});
+
+      expect(i18n.formatName('first', 'last')).toStrictEqual('first');
+    });
+
+    it('uses fallback locale value for unknown locale', () => {
+      const i18n = new I18n(defaultTranslations, {locale: 'fr-CA'});
```
loic-d: `fr-CA` is not an 'unknown' locale?

packages/react-i18n/src/test/i18n.test.ts
```diff
@@ -1065,6 +1066,92 @@ describe('I18n', () => {
...
+      expect(i18n.formatName('first', 'last')).toStrictEqual('first');
+    });
+
+    it('returns firstName for English', () => {
+      const i18n = new I18n(defaultTranslations, {locale: 'en'});
+
+      expect(i18n.formatName('first', 'last')).toStrictEqual('first');
+    });
+
+    it('returns custom name for Japan', () => {
```
loic-d: Should `Japan` be `Japanese`?

packages/react-i18n/src/test/i18n.test.ts
```diff
@@ -1065,6 +1066,92 @@ describe('I18n', () => {
...
+    });
+
+    it('returns lastName first and no space for Japan', () => {
+      const i18n = new I18n(defaultTranslations, {locale: 'ja'});
+
+      expect(i18n.formatName('first', 'last', {full: true})).toStrictEqual(
+        'lastfirst',
+      );
+    });
+  });
```
loic-d: Should `zh-CN` / `zh-TW` be tested as well?

packages/react-i18n/src/test/i18n.test.ts
```diff
@@ -1065,6 +1066,92 @@ describe('I18n', () => {
...
+      expect(i18n.formatName('first', undefined, {full: true})).toStrictEqual(
+        'first',
+      );
+    });
+
+    it('returns only the lastName when firstName is missing using full', () => {
+      const i18n = new I18n(defaultTranslations, {locale: 'en'});
+
+      expect(i18n.formatName('', 'last', {full: true})).toStrictEqual('last');
+    });
```
loic-d: Missing test for `i18n.formatName('', undefined, {full: true})`?

packages/react-i18n/src/i18n.ts
benoitzohar: I don't see any other potential options for this function. I'll keep it as is for now. 👍

packages/react-i18n/src/test/i18n.test.ts
benoitzohar: That label wasn't explicit enough, it's now `uses fallback locale for locale without custom formatter`.

loic-d: I get it now!

packages/react-i18n/src/test/i18n.test.ts
benoitzohar: Yes. I wasn't sure about adding all the options in the tests as it seems a bit unnecessary. But now that we have a custom formatter for each locale, it's probably a good idea!
Thanks!

PR #844, opened 2019-08-13
🐛 fix auth issue with invalid accesstoken

### LINKED ISSUE #727: Koa Session Cookie Blocking Reinstallation
## Overview

When using @shopify/koa-shopify-auth to handle auth for my Shopify App, I was unable to reinstall the app, after uninstalling it, without first clearing the browser cache. After some troubleshooting, it seems that that the browser will pass the expired koa session cookies with the reinstallation attempt, which results in a failed redirect.

I'm wondering if anybody else has run into this issue.

katiedavis:
## Description

Fixes #727 

## Type of change
In the koa-shopify-auth package, users were having issues reinstalling their apps via the partner dashboard. This PR adds a call to the admin API (I tried to find the 'cheapest' call) to check the validity of the access token. This solve the issue of users not being able to reinstall the app on their stores. 

I'm looking for feedback on this. If anyone knows a better way of checking if an access token is valid please let me know! 

## Checklist

### DIFF
--- a/packages/koa-shopify-auth/src/verify-request/verify-request.ts
+++ b/packages/koa-shopify-auth/src/verify-request/verify-request.ts
@@ -1,5 +1,5 @@
 import {Context} from 'koa';
-
+import {Method, Header, StatusCode} from '@shopify/network';
 import {NextFunction} from '../types';
 import {TEST_COOKIE_NAME, TOP_LEVEL_OAUTH_COOKIE_NAME} from '../index';
 
@@ -20,9 +20,28 @@ export default function verifyRequest({
       query: {shop},
       session,
     } = ctx;
+    const redirect = `${authRoute}?shop=${shop}`;
 
     if (session && session.accessToken) {
       ctx.cookies.set(TOP_LEVEL_OAUTH_COOKIE_NAME);
+      // If a user has installed the store previously on their shop, the accessToken can be stored in session.
+      // we need to check if the accessToken is valid, and the only way to do this is by hitting the api.
+      const response = await fetch(
+        `https://${session.shop}/admin/metafields.json`,
+        {
+          method: Method.Post,
+          headers: {
+            [Header.ContentType]: 'application/json',
+            'X-Shopify-Access-Token': session.accessToken,
+          },
+        },
+      );
+
+      if (response.status === StatusCode.Unauthorized) {
+        ctx.redirect(redirect);
+        return;
+      }
+
       await next();
       return;
     }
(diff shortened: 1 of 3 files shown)


packages/koa-shopify-auth/src/verify-request/verify-request.ts
```diff
@@ -23,6 +22,20 @@ export default function verifyRequest({
 
     if (session && session.accessToken) {
       ctx.cookies.set(TOP_LEVEL_OAUTH_COOKIE_NAME);
+      // we need to check if the accessToken is valid.
+      const response = await fetch(`https://${shop}/admin/metafields.json`, {
```
TheMallen: nit: we have constants for these things in  `@shopify/network` under

eg.
```ts
import {Method, Header} from '@shopify/network';
```

```ts
method: Method.Post
headers: {
  [Header.ContentType]: 'application/json',
 ```

File: packages/koa-shopify-auth/src/verify-request/verify-request.ts [hunk 2]
```diff
@@ -23,6 +22,20 @@ export default function verifyRequest({
...
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json',
+          'X-Shopify-Access-Token': session.accessToken,
+        },
+      });
+
+      if (response.status === 401) {
+        ctx.redirect(`${authRoute}?shop=${shop}`);
```
tylerball: let's make this redirect string a shared const

packages/koa-shopify-auth/src/verify-request/verify-request.ts
```diff
@@ -23,6 +22,20 @@ export default function verifyRequest({
 
     if (session && session.accessToken) {
       ctx.cookies.set(TOP_LEVEL_OAUTH_COOKIE_NAME);
+      // we need to check if the accessToken is valid.
```
TheMallen: I'd prefer if this comment was a tiny bit more detailed as to why we have to do it

packages/koa-shopify-auth/src/verify-request/verify-request.ts
```diff
@@ -23,6 +22,20 @@ export default function verifyRequest({
...
+      const response = await fetch(`https://${shop}/admin/metafields.json`, {
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json',
+          'X-Shopify-Access-Token': session.accessToken,
+        },
+      });
+
+      if (response.status === 401) {
```
TheMallen: can use `Status.Unauthorized` here when we bring in the `Status` enum

PR #845, opened 2019-08-13
♻️ renaming Props to PropsFor

patsissons:
## Description

The original `Props` type often collided with pre-existing `interface Props` in components, this new `PropsFor` will not encounter the same problems.

## Type of change

- [x] `@shopify/useful-types` Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)

This is a breaking change because we are not deprecating the old `Props` and instead just renaming.

## Checklist


patsissons:
closing in favour of #846

PR #846, opened 2019-08-13
🔥remove Props<T> from useful-types

patsissons:
## Description

We can now replace `Props<T>` from `@shopify/useful-types` with `ComponentProps`, `ComponentPropsWithRef`, and `ComponentPropsWithoutRef` native to the `react` types.

Typically we would be simply swapping out `Props<T>` for `ComponentPropsWithoutRef<T>` for most cases.

This also allows us to remove the `@types/react` dependency. 🎉

replaces #845 

## Type of change

- [x] `useful-types` Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

### DIFF
--- a/packages/react-testing/src/types.ts
+++ b/packages/react-testing/src/types.ts
@@ -1,9 +1,13 @@
 import * as React from 'react';
-import {
-  Arguments,
-  MaybeFunctionReturnType,
-  Props as PropsForComponent,
-} from '@shopify/useful-types';
+import {Arguments, MaybeFunctionReturnType} from '@shopify/useful-types';
+
+export type PropsFor<
+  T extends string | React.ComponentType<any>
+> = T extends string
+  ? React.HTMLAttributes<T>
+  : T extends React.ComponentType<any>
+    ? React.ComponentPropsWithoutRef<T>
+    : never;
 
 export type FunctionKeys<T> = {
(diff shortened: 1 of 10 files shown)


packages/useful-types/README.md
```diff
@@ -60,7 +60,9 @@ The following type aliases are provided by this library:
   type SelectiveObj = Omit<Obj, 'foo' | 'bar'>; // {baz: number}
   ```
 
-- `Props<T>`: Extracts the prop type from a React component. This allows you to access property types without having to manually export/ import the type.
+- `Props<T>`: ⚠️ No longer availble.
```
BPScott: The list of things that aren't in this library is pretty long ;) I reckon this info is better placed in the changelog

packages/react-testing/src/types.ts
```diff
@@ -1,9 +1,13 @@
 import * as React from 'react';
-import {
-  Arguments,
-  MaybeFunctionReturnType,
-  Props as PropsForComponent,
-} from '@shopify/useful-types';
+import {Arguments, MaybeFunctionReturnType} from '@shopify/useful-types';
+
+export type PropsFor<
```
patsissons: this _duplication_ of the old `Props<T>` is a bit ugly, but it's a lot better than using conditional types in each place where we were using `Props<T>` before.

Issue #838, opened 2019-08-12
[quilt_rails // react-server] Better error experience in development when server rendering fails


TheMallen:
## Overview
Currently, when `react-server` fails to render we serve an ugly and unhelpful error message

![image](https://user-images.githubusercontent.com/4889856/62879810-f1231e80-bcf9-11e9-8275-94a50a966e27.png)

In development we should provide as much context as we can to what has gone wrong. Possible solutions:
- in `react-server` we could respond with the full text and stack of the error in development
- in `quilt_rails` we could use `sewing_kit`'s view helpers to attempt to fall back to client side rendering
- in `react-server` we could attempt to fall back to client-side rendering

PR #850, opened 2019-08-14
[@shopify/react-server] Better error messages in dev mode

ismail-syed:
## Description

Fixes https://github.com/Shopify/quilt/issues/838

This provides better error logging in development. 

**Before:**

Screen Shot 2019-08-14 at 6 16 25 PM

**After:**

image (3)

- Patch bump

## Checklist

### DIFF
--- a/packages/react-server/CHANGELOG.md
+++ b/packages/react-server/CHANGELOG.md
@@ -9,7 +9,7 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ### Changed
 
-- Improve error experience in development when server rending fails [#850](https://github.com/Shopify/quilt/pull/850)
+- Improve error experience in development when server rendering fails [#850](https://github.com/Shopify/quilt/pull/850)
 
 ## 0.1.0
 
--- a/packages/react-server/CHANGELOG.md
+++ b/packages/react-server/CHANGELOG.md
@@ -7,6 +7,12 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Changed
+
+- Improve error experience in development when server rending fails [#850](https://github.com/Shopify/quilt/pull/850)
+
+## 0.1.0
+
 ### Added
 
 - `@shopify/react-server` package
--- a/packages/react-server/src/logger/logger.ts
+++ b/packages/react-server/src/logger/logger.ts
@@ -2,7 +2,7 @@ import {Context} from 'koa';
 import chalk from 'chalk';
 import {KoaNextFunction} from '../types';
 
-const LOGGER = Symbol('logger');
+export const LOGGER = Symbol('logger');
 const PREFIX = chalk`{underline sidecar} ⁓ `;
 
 interface LoggerOptions {
@@ -47,7 +47,7 @@ export function createLogger(options?: LoggerOptions) {
   };
 }
 
-const noopLogger = {
+export const noopLogger = {
   log: noop,
   error: noop,
 };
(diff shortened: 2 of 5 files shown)


packages/react-server/src/render/render.tsx
```diff
@@ -64,7 +64,12 @@ export function createRender(render: RenderFunction) {
       });
     } catch (error) {
       logger.error(error);
-      throw error;
+      if (process.env.NODE_ENV === 'development') {
+        ctx.body = `Internal Server Errror: \n\n${error.message}`;
```
ismail-syed: I attempted to opt for something [cleaner](https://github.com/koajs/koa/wiki/Error-Handling#catching-downstream-errors), but couldn't get it to work.

packages/react-server/src/render/render.tsx
```diff
@@ -64,7 +64,12 @@ export function createRender(render: RenderFunction) {
...
+        return;
+      } else {
+        ctx.throw(500, error);
```
ismail-syed: Explicitly calling the `500` makes this easier to find for future readers.

TheMallen: We should use `@shopify/network`'s `StatusCode` enum.

packages/react-server/src/render/render.tsx
TheMallen: Errror -> Error

packages/react-server/CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Changed
+
+- Improve error experience in development when server rending fails [#850](https://github.com/Shopify/quilt/pull/850)
```
cartogram:
```suggestion
- Improve error experience in development when server rendering fails [#850](https://github.com/Shopify/quilt/pull/850)
```

packages/react-server/src/render/render.tsx
```diff
@@ -35,9 +35,9 @@ export function createRender(render: RenderFunction) {
     const asyncAssetManager = new AsyncAssetManager();
     const acceptsLanguages = ctx.acceptsLanguages && ctx.acceptsLanguages();
     const locale = Array.isArray(acceptsLanguages) ? acceptsLanguages[0] : 'en';
-    const app = render({...ctx, locale});
 
     try {
+      const app = render({...ctx, locale});
```
ismail-syed: Couple things here:

- The `render` call needs to be in the try/catch block error handling
- `return response;` doesn't do anything in koa middlewares. It's been removed cc/ @cartogram in case it does have some weird side-effecty use case.

cartogram: @ismail-syed its okay to remove `return response`.

packages/react-server/src/logger/logger.ts
```diff
@@ -2,7 +2,7 @@ import {Context} from 'koa';
 import chalk from 'chalk';
 import {KoaNextFunction} from '../types';
 
-const LOGGER = Symbol('logger');
+export const LOGGER = Symbol('logger');
```
cartogram: Lets keep the `Symbol` private to this module and instead create a `setLogger()` and export that. It is more consistent with how we've implemented middleware in the past.

Have a look at the example here: https://github.com/Shopify/web/blob/master/server/middleware/handler/handler.ts#L18

packages/react-server/src/render/test/render.test.tsx
```diff
@@ -0,0 +1,61 @@
...
+    expect(ctx.body).toContain(myCoolApp);
+  });
+
+  it('in development the body contains a meaningful error messages', () => {
+    withEnv('development', async () => {
+      const ctx = createMockContext();
+      ctx.state[LOGGER] = noopLogger;
+
+      const renderFunction = createRender(() => <BrokenApp />);
+      await renderFunction(ctx as RenderContext);
```
cartogram: This is fine as is, but you could wrap the `createMockContext` function with the missing properties from `RenderContext` and use that as the `ctx`.

PR #860, opened 2019-08-15
[WIP] 🚀 use incremental build to speed up CI

### LINKED ISSUE #857: Use `--incremental` flag and `tsBuildInfo` to speed up Quilt CI
We should be able to leverage `tsBuildInfo` files to make builds essentially a no-op for libraries that haven't changed, which would help our CI times tremendously.

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html

TheMallen:
closes #857 

I tried to set up incremental builds but am getting lots of OOM, probably we can't continue to rely on `lerna` to run our command this way, likely we would need to also setup project references :/


TheMallen:
Closing in favour of work by @ismail-syed

PR #864, opened 2019-08-16
remove shop and access token

cartogram:
## Description

Migrates some aspects over from `web-rails-proving-ground`:

- Removes the `shop` or `accesstoken` options in `createGraphQLClient`
- Exports the logger middleware from `react-server`. I was testing a few different setups with `webgen` and found that if you want to consume the `createRender` middleware directly (and roll your own koa server in a node app) you need to create the logger in a middleware as well. Right now you can't do this. Alternately, we could remove all the logging from `react-server` (just use `console`), or better, have a simple server logger middleware package that we can share. We have a logger package in this repo, but it is not very good.

### DIFF
--- a/packages/react-server/src/render/test/render.test.tsx
+++ b/packages/react-server/src/render/test/render.test.tsx
@@ -1,8 +1,7 @@
 import React from 'react';
 import {createMockContext} from '@shopify/jest-koa-mocks';
 import withEnv from '@shopify/with-env';
-import {createRender, RenderingContext} from '../render';
-import {Logger, noopLogger, setLogger} from '../../logger';
+import {createRender, RenderContext} from '../render';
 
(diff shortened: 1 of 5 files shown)


packages/react-server/src/render/test/render.test.tsx
```diff
@@ -23,18 +22,16 @@ describe('createRender', () => {
   it('response contains "My cool app"', async () => {
     const myCoolApp = 'My cool app';
     const ctx = createMockContext();
-    setLogger(ctx, new Logger());
```
cartogram: cc/ @ismail-syed ended up just putting a fallback on `console` here cause it was make it hard to consume the `createRender` middleware as a standalone thing.

ismail-syed: Clever!

PR #878, opened 2019-08-20
Alternate GraphQL self-serializer

lemonmade:
This PR is an alternate take on the GraphQL self-serializer. In this way of doing things, we let the user entirely construct their own Apollo client, which lets them do anything they like — custom links, `apollo-boost`, etc. It then mutates the link and cache in order to be able to extract/ rehydrate with initial data.

Would love some thoughts on the approach, and if we like it, whether we would need to go through a full set of major versions, or whether we are early enough on these packages that we can shift the approach this significantly in a minor.


packages/react-self-serializers/README.md
```diff
@@ -99,30 +99,49 @@ A self-serializing provider for initial GraphQL data from Apollo.
...
+
 import {GraphQL} from '@shopify/react-self-serializers';
 
-function App({shop, accessToken}: {shop?: string; accessToken?: string}) {
-  return (
-    <GraphQL shop={shop} accessToken={accessToken}>
-      {/* rest of the app */}
-    </GraphQL>
-  );
+function App({
```
TheMallen: We could show using it with `@shopify/react-graphql`'s `createGraphQLClient` for maximum simplicity + stylepoints

lemonmade: I would delete that function as part of this. It's a weird place for an opinionated Apollo creator, IMO. Thoughts?

packages/react-self-serializers/src/GraphQLComponent.tsx
```diff
@@ -1,30 +1,38 @@
...
-  createGraphQLClient,
-  Options,
-} from '@shopify/react-graphql';
+import {ApolloProvider, createSsrExtractableLink} from '@shopify/react-graphql';
 import {useLazyRef} from '@shopify/react-hooks';
 
-interface Props extends Partial<Options> {
+interface Props {
   children?: React.ReactNode;
+  createClient(): import('apollo-client').ApolloClient<any>;
```
michelleyschen: ❤️  for decoupling this and allow user to pass any client setup they would like. 
If only this was available an hour ago :p

packages/react-graphql/src/links.ts
```diff
@@ -52,18 +51,3 @@ class SsrExtractableLink extends ApolloLink {
     });
   }
 }
-
-export class ApolloClient<CacheShape> extends BaseApolloClient<CacheShape> {
```
michelleyschen: Recently we had been going with the apollo vs apollo-boost approach.
1. Providing a simple base usage
2. Export individual pieces to allow of future customization

for that I wonder if we want to keep this exported as a simple version of `createClient` that the consumer can choose to pass it into the self serializer themselves.

PR #879, opened 2019-08-20
🐛 no longer clobber koa context

### LINKED ISSUE #876: Koa request alias does not work
## Motivation

Usage:
```
const render = (ctx: Context) =>
  React.createElement(App, {
    server: true,
    locale: ctx.locale,
    location: ctx.request.url,
    apiKey: process.env.SHOPIFY_API_KEY,
    shop: ctx.request.get('X-Shopify-Shop-Origin')
  });
```

Notice the use of `ctx.request.get('X-Shopify-Shop-Origin')`
ideally we can use `ctx.get('X-Shopify-Shop-Origin')` but this is not available right now.
likely due to the spread operation here
https://github.com/Shopify/quilt/blob/master/packages/react-server/src/render/render.tsx#L40

## Checklist

TheMallen:
Fixes #876 

## Description
This PR fixes the issue we were having where our use of the spread operator on the `ctx` we got passed by Koa broke a bunch of the proxying / getters on that object, leading things to not work unexpectedly and require delving into `ctx.request` properties directly.
 
This PR is actually a breaking change, switching the API for `render` functions from `(ctx: Koa.Context & {locale: string}) => JSX` to `(ctx: Koa.Context, data: {locale: string})`.

### DIFF
--- a/packages/react-server/README.md
+++ b/packages/react-server/README.md
@@ -26,12 +26,11 @@ To begin using this package, Node apps only require a server entry point that ca
 
 ```tsx
 import React from 'react';
-import {createServer, RenderContext} from '@shopify/react-server';
-
+import {createServer} from '@shopify/react-server';
 import App from '../app';
 
 const app = createServer({
-  render: (ctx: RenderContext) => <App location={ctx.request.url} />,
+  render: ctx => <App location={ctx.request.url} />,
 });
 ```
 
@@ -71,7 +70,7 @@ interface Options {
   assetPrefix?: string;
   // any additional Koa middleware to mount on the server
   serverMiddleware?: compose.Middleware<Context>[];
-  // a function of `(ctx: Koa.Context) => React.ReactNode`
+  // a function of `(ctx: Context, data: {locale: string}): React.ReactElement<any>`
   render: RenderFunction;
   // whether to run in debug mode
   debug?: boolean;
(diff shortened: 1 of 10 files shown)


packages/react-server/src/render/test/render.test.tsx
```diff
@@ -24,11 +24,24 @@ describe('createRender', () => {
     const ctx = createMockContext();
 
     const renderFunction = createRender(() => <>{myCoolApp}</>);
-    await renderFunction(ctx as RenderContext);
+    await renderFunction(ctx);
 
     expect(ctx.body).toContain(myCoolApp);
   });
 
+  // we cannot test this with `createMockContext` because it actually clobbers the proxy methods itself :(
```
michelleyschen: 😂

packages/react-server/README.md
```diff
@@ -31,11 +31,11 @@ To begin using this package, Node apps only require a server entry point that ca
 
 ```
 import React from 'react';
-import {createServer, RenderContext} from '@shopify/react-server';
+import {createServer, Context} from '@shopify/react-server';
 import App from '../app';
 
 const app = createServer({
-  render: (ctx: RenderContext) => <App location={ctx.request.url} />
+  render: (ctx: Context) => <App location={ctx.request.url} />
```
lemonmade: You shouldn't ever need to specify the type explicitly here I don't think?

packages/react-server/src/render/test/render.test.tsx
lemonmade: Is there an issue for this? Seems... problematic

TheMallen: updated the PR, I was just doing it wrong and am a dumdum

PR #896, opened 2019-08-22
🐛 fix breakages in production when used with ShopifySecurityBase

TheMallen:
## Summary
This adds an explicit check for `ShopifySecurityBase ` and if it is defined wraps our call to `reverse_proxy` with a call to `ShopifySecurityBase::HTTPHostRestriction.whitelist`. This should fix the issues we've been having with using the two gems together causing SSR to break.

## Tophatting🎩

I'm not really sure how to tophat this locally to be honest. I wanted to get feedback from security folks quick regardless.

### DIFF
--- a/gems/quilt_rails/lib/quilt_rails/logger.rb
+++ b/gems/quilt_rails/lib/quilt_rails/logger.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module Quilt
   module Logger
     def self.log(string)
--- a/gems/quilt_rails/lib/quilt_rails/logger.rb
+++ b/gems/quilt_rails/lib/quilt_rails/logger.rb
@@ -1,0 +1,11 @@
+module Quilt
+  module Logger
+    def self.log(string)
+      if defined? Rails && Rails.logger.nil?
+        puts string
+      else
+        Rails.logger.info(string)
+      end
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


gems/quilt_rails/lib/quilt_rails/logger.rb
```diff
@@ -0,0 +1,11 @@
+module Quilt
```
TheMallen: I moved this out because it was cluttery

gems/quilt_rails/lib/quilt_rails/react_renderable.rb
```diff
@@ -7,28 +7,31 @@ module ReactRenderable
     include ReverseProxy::Controller
 
     def render_react
+      if defined? ShopifySecurityBase
+        ShopifySecurityBase::HTTPHostRestriction.whitelist([Quilt.configuration.react_server_host]) do
```
JackMc: You could redo this as `ShopifySecurityBase::HTTPHostRestriction.whitelist([Quilt.configuration.react_server_host], &method(:proxy))` (I think)

TheMallen: personally I find this slightly more readable, but that might just be due to my dislike of/tendency to be confused by the unary ampersand operator :P

Issue #894, opened 2019-08-21
React SSR error page


michelleyschen:
When React SSR encounter error, we should have a nice error page with the stack track on where the error came from.

TheMallen:
Inspirations:

## Rails error screen
![image](https://user-images.githubusercontent.com/4889856/63620165-95407b80-c5be-11e9-8e9e-7a54b5629d44.png)

## The one CRA uses
https://github.com/facebook/create-react-app/tree/master/packages/react-error-overlay

PR #901, opened 2019-08-23
✨ include the error stack when presenting SSR errors in dev

TheMallen:
## Description

Part of #894 

This PR makes it so that react-server includes the stack as well as the error message when something goes wrong in SSR. This should make it easier to debug issues. It does not change the presentation at all, so it is only part of the referenced issue.


michelleyschen:
So excited for this!

PR #909, opened 2019-08-26
Adds `useAcceptLanguages()` hook

cartogram:
## Description

Fixes (issue #)

## Type of change

## Checklist

- [ ] I have prefixed my pull request title with the corresponding emoji from [this guide](https://gitmoji.carloscuesta.me/)

### DIFF
--- a/packages/react-network/src/hooks.ts
+++ b/packages/react-network/src/hooks.ts
@@ -1,5 +1,5 @@
 import {useContext} from 'react';
-import {CspDirective, StatusCode} from '@shopify/network';
+import {CspDirective, StatusCode, Header} from '@shopify/network';
 import {useServerEffect} from '@shopify/react-effect';
 
 import {NetworkContext} from './context';
@@ -37,4 +37,11 @@ export function useStatus(code: StatusCode) {
 
 export function useRedirect(url: string, status?: StatusCode) {
   useNetworkEffect(network => network.redirectTo(url, status));
+}
+
+export function useAcceptLanguages(fallback = 'en') {
+  const acceptsLanguages = useRequestHeader(Header.AcceptLanguage);
+  const locales = acceptsLanguages ? acceptsLanguages.split(',') : [fallback];
+
+  return [locales[0]];
 }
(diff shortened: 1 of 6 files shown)


cartogram:
@lemonmade @TheMallen re-did this as a stand-alone hook in `react-network`, how does this look/ feel?

packages/react-network/src/hooks.ts
```diff
@@ -38,3 +38,10 @@ export function useStatus(code: StatusCode) {
 export function useRedirect(url: string, status?: StatusCode) {
   useNetworkEffect(network => network.redirectTo(url, status));
 }
+
+export function useAcceptLanguages(fallback = 'en') {
+  const acceptsLanguages = useRequestHeader(Header.AcceptLanguage);
+  const locales = acceptsLanguages ? acceptsLanguages.split(',') : [fallback];
+
+  return [locales[0]];
```
lemonmade: Why return only the first one?

packages/react-network/src/test/hooks.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+  it('returns the locale from the language header', async () => {
+    const lang = 'it';
+    const wrapper = await mountWithLanguageHeader(<MockComponent />, lang);
+
+    expect(wrapper.find(MockComponent)).toContainReactText(lang);
+  });
+
+  it('returns the first locale from the language header with multiple languages', async () => {
+    const lang = 'it, fr, en-US';
```
lemonmade: IIRC the actual header also has percentage matches that we'll need to strip out of the string (example from MDN: `fr-CH, fr;q=0.9, en;q=0.8, de;q=0.7, *;q=0.5`)

packages/react-network/src/test/hooks.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+  });
+
+  it('returns `en` if no fallback is set and no language header exists', async () => {
+    const wrapper = await mountWithLanguageHeader(<MockComponent />);
+
+    expect(wrapper.find(MockComponent)).toContainReactText('en');
+  });
+});
+
+function mountWithLanguageHeader(
```
lemonmade: If you need a custom mount, use `createMount` instead

packages/react-network/src/test/hooks.test.tsx
cartogram: Yes you are correct. I am going to install a module to do the actual parsing here.

PR #916, opened 2019-08-27
✨ Adds 'short' and 'explicit' currency formatting to react-i18n

cejaekl:
## Description

Implements https://vault.shopify.io/projects/8347

Adds two new methods to `I18n`.

`formatCurrencyShort()` will format a given amount, in a given locale, as a given currency.
If the currency has a symbol, then only that symbol will be used, and the corresponding ISO
three-letter code will be omitted.  (If the currency has no symbol, then the ISO three-letter
code is used instead).
Examples:
  - `(1.25, 'de-CH', 'CHF')` => `CHF 1,25`
  - `(1.25, 'fr-FR', 'EUR')` => `1,25 €`
  - `(125,  'en-US', 'JPY')` => `¥ 125`
  - `(1.25, 'en-US', 'OMR')` => `OMR 1.250`
  - `(1.25, 'en-US', 'USD')` => `$ 1.25`

`formatCurrencyExplicit()` will format a given amount, in a given locale, as a given currency.
If the currency has a symbol, then both the symbol and the ISO code will be used.
If the currency does not have a symbol, then only the ISO code will be used.
Examples:
  - `(1.25, 'de-CH', 'CHF')` => `CHF 1,25`
  - `(1.25, 'fr-FR', 'EUR')` => `1,25 € EUR`
  - `(125,  'en-US', 'JPY')` => `¥ 125 JPY`
  - `(1.25, 'en-US', 'OMR')` => `OMR 1.250`
  - `(1.25, 'en-US', 'USD')` => `$ 1.25 USD`

NOTE:  To unit test this, I have added `full-icu` as a test-time dependency, and now invoke `jest` as `NODE_ICU_DATA=node_modules/full-icu jest`.  This changed the behaviour of the test environment (by actually having internationalization API calls behave like they do in a browser) and necessitated modifications to a handful of other unit tests which were asserting "wrong" results.

## Type of change

- [ ] `react-i18n` Patch: Bug/ Documentation fix (non-breaking change which fixes an issue or adds documentation)
- [X] `react-i18n` Minor: New feature (non-breaking change which adds functionality)
- [ ] `react-i18n` Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

- [X] I have prefixed my pull request title with the corresponding emoji from [this guide](https://gitmoji.carloscuesta.me/)


patsissons:
These changes are a bit outside my usual context, you probably want to tag someone from web foundations for a more comprehensive review. The only thing that comes to mind is whether or not its worth while evaluating the `currencyDisplay` option to reduce some of the formatting code?

cejaekl:
> whether or not its worth while evaluating the `currencyDisplay` option to reduce some of the formatting code?

It would be nice if we could.  

`currencyDisplay` gives us three options:  
  - `symbol` (default):  e.g., `US$`
  - `code`: e.g., `USD`
  - `name`: e.g., `dollars`

The design we're trying to implement, however, uses the "bare" symbol (e.g., `$` instead of `US$`) and, sometimes adds the ISO code as well.  I don't think `currencyDisplay` gets us any closer to that, unfortunately.

(https://vault.shopify.io/projects/8347 , https://docs.google.com/presentation/d/1Qd7d7hesKiVuDCOLunXaFnPVnscqaL1OLHP-gILBfDU/edit#slide=id.g32426ed60e_0_59)

packages/react-i18n/src/i18n.ts
```diff
@@ -44,6 +44,29 @@ const DECIMAL_NOT_SUPPORTED = 'N/A';
 const PERIOD = '.';
 const DECIMAL_VALUE_FOR_CURRENCIES_WITHOUT_DECIMALS = '00';
 
+// Specifies the number of fractional digits for those currencies that don't use two digits.
+const CURRENCY_DIGITS = {
```
ismail-syed: There's already a utility imported for this, [currencyDecimalPlaces](https://github.com/Shopify/quilt/blob/master/packages/react-i18n/src/constants/index.ts#L179-L182).

packages/react-i18n/src/i18n.ts
```diff
@@ -231,6 +254,47 @@ export class I18n {
...
+  }
+
+  formatCurrencyShort(
+    amount: number,
+    options: NumberFormatOptions = {},
+  ): string {
+    const {locale} = this;
+    const sym = this.getShortCurrencySymbol(options.currency);
+    let adjustedPrecision = options.precision;
+    if (adjustedPrecision === undefined) {
```
ismail-syed: I think we can clean this up a bit to prevent the use of let. We always want to use `const` whenever possible. 

Haven't tested this out but something along the lines of 

```
const currency = options.currency || this.defaultCurrency || '';
const currencyPosition = currency in CURRENCY_DIGITS ? CURRENCY_DIGITS[currency] : 2;
const adjustedPrecision = options.precision || currencyPosition;
```

packages/react-i18n/src/i18n.ts
```diff
@@ -231,6 +254,47 @@ export class I18n {
...
+      }
+    }
+    const formattedAmount = memoizedNumberFormatter(locale, {
+      style: 'decimal',
+      minimumFractionDigits: adjustedPrecision,
+      maximumFractionDigits: adjustedPrecision,
+      ...options,
+    }).format(amount);
+
+    if (sym.prefixed) {
```
ismail-syed: We could clean this one up a bit as well. 
```
return sym.prefix ?  sym.symbol + formattedAmount : formattedAmount + sym.symbol;
```

packages/react-i18n/src/i18n.ts
```diff
@@ -299,6 +363,31 @@ export class I18n {
...
+  // For other currencies, e.g. CHF and OMR, the "symbol" is the ISO currency code.
+  // In those cases, we return the full currency code without stripping the country.
+  getShortCurrencySymbol(currencyCode?: string) {
+    const currency = currencyCode || this.defaultCurrency || '';
+    const regionCode = currency.substring(0, 2);
+    const info = this.getCurrencySymbol(currency);
+    const shortSymbol = info.symbol.replace(regionCode, '');
+
+    if (shortSymbol.match(/[A-Z]/)) {
+      // It looks like we have an alphabetic code, not a symbol.  Return the full code.
```
ismail-syed: If we pull out the regex into a meaningful variable, we might be able to reduce the verbosity in these comments 

```
const alphabeticCode = /[A-Z]/;
return shortSymbol.match(alphabeticCode)
   ? info
   : {symbol: shortSymbol, prefixed: info.prefixed};
```

packages/dates/src/tests/utilities.test.ts
```diff
@@ -1,9 +1,21 @@
 import {formatDate} from '../utilities/formatDate';
 
 describe('formatDate()', () => {
-  it('formats dates into YYYY-MM-DD', () => {
+  it('formats dates as expected for various locales', () => {
```
ismail-syed: We should be able to leverage jest `it.each` for things like this. 

```
 it.each([
    ['en-US', '12/31/2017'],
    ['en-GB', '31/12/2017'],
    ['de-DE', '31.12.2017'],
 ])(`formats using the locale $s and date $s`, (locale, date) => {
    const baseDate = new Date(2018, 0, 0);
    const formatedDate = `[${locale}] ${formatDate(baseDate, locale)}`;
    expect(formatedDate).toBe(`[${locale}] ${date}`);
  });
```

packages/react-i18n/src/i18n.ts
ismail-syed: Also, it looks like we joining values with symbols for formatting? Much safer to use template strings in JS instead of concatenating strings like this.
 
```
return sym.prefix ?  `${sym.symbol}  ${formattedAmount}` : `${formattedAmount} ${sym.symbol}`;
```

packages/react-i18n/src/test/i18n.test.ts
```diff
@@ -594,6 +594,83 @@ describe('I18n', () => {
...
+      ['en-US', 'OMR', 'OMR ', true, 'OMR 1,234.560'],
+      ['en-US', 'USD', '$ ', true, '$ 1,234.56 USD'],
+
+      ['fr-FR', 'EUR', ' €', false, '1 234,56 € EUR'],
+      ['fr-FR', 'JPY', ' JPY', false, '1 235 JPY'],
+      ['fr-FR', 'OMR', ' OMR', false, '1 234,560 OMR'],
+      ['fr-FR', 'USD', ' $US', false, '1 234,56 $ USD'],
+    ];
+
+    it('formats as expected in several locales', () => {
```
ismail-syed: Same here, let's leverage `it.each` instead

packages/react-i18n/src/test/i18n.test.ts
```diff
@@ -594,6 +594,83 @@ describe('I18n', () => {
...
+      ['en-US', 'OMR', 'OMR ', true, 'OMR 1,234.560'],
+      ['en-US', 'USD', '$ ', true, '$ 1,234.56'],
+
+      ['fr-FR', 'EUR', ' €', false, '1 234,56 €'],
+      ['fr-FR', 'JPY', ' JPY', false, '1 235 JPY'],
+      ['fr-FR', 'OMR', ' OMR', false, '1 234,560 OMR'],
+      ['fr-FR', 'USD', ' $US', false, '1 234,56 $'],
+    ];
+
+    it('formats as expected in several locales', () => {
```
ismail-syed: `it.each`

packages/react-i18n/src/i18n.ts
```diff
@@ -299,6 +363,31 @@ export class I18n {
...
   }
 
+  // Intl.NumberFormat sometimes annotates the "curerncy symbol" with a country code.
+  // For example, in locale 'fr-FR', 'USD' is given the "symbol" of " $US".
+  // This method strips out the country-code annotation, if there is one.
+  // (So, for 'fr-FR' and 'USD', the return value would be " $").
+  //
+  // For other currencies, e.g. CHF and OMR, the "symbol" is the ISO currency code.
+  // In those cases, we return the full currency code without stripping the country.
+  getShortCurrencySymbol(currencyCode?: string) {
```
ismail-syed: Should `getShortCurrencySymbol` have test coverage as well?

packages/react-i18n/src/i18n.ts
```diff
@@ -231,6 +254,47 @@ export class I18n {
     return this.formatNumber(amount, {as: 'currency', ...options});
   }
 
+  formatCurrencyExplicit(
```
lemonmade: Extra methods feels very awkward compared to an option for the existing `formatCurrency` method

packages/react-i18n/src/i18n.ts
```diff
@@ -231,6 +254,47 @@ export class I18n {
...
+    }
+    return `${value} ${isoCode}`;
+  }
+
+  formatCurrencyShort(
+    amount: number,
+    options: NumberFormatOptions = {},
+  ): string {
+    const {locale} = this;
+    const sym = this.getShortCurrencySymbol(options.currency);
```
lemonmade: Spell it out :)

packages/dates/src/tests/utilities.test.ts
lemonmade: Personally I prefer tests that aren't generated, it's way harder to find the test in the codebase when the name doesn't actually exist anywhere.

packages/react-i18n/src/i18n.ts
lemonmade: You shouldn't use `||` here because `0` is falsy and there are lots of zeros in the equation.

packages/react-i18n/src/i18n.ts
```diff
@@ -299,6 +363,31 @@ export class I18n {
     return WEEK_START_DAYS.get(country) || DEFAULT_WEEK_START_DAY;
   }
 
+  // Intl.NumberFormat sometimes annotates the "curerncy symbol" with a country code.
```
lemonmade:
```suggestion
  // Intl.NumberFormat sometimes annotates the "currency symbol" with a country code.
```

packages/react-i18n/src/i18n.ts
lemonmade: If it's not public API, it should be marked `private` too

packages/react-i18n/src/i18n.ts
```diff
@@ -299,6 +363,31 @@ export class I18n {
...
+  // For example, in locale 'fr-FR', 'USD' is given the "symbol" of " $US".
+  // This method strips out the country-code annotation, if there is one.
+  // (So, for 'fr-FR' and 'USD', the return value would be " $").
+  //
+  // For other currencies, e.g. CHF and OMR, the "symbol" is the ISO currency code.
+  // In those cases, we return the full currency code without stripping the country.
+  getShortCurrencySymbol(currencyCode?: string) {
+    const currency = currencyCode || this.defaultCurrency || '';
```
lemonmade: in general, it's cleaner to provide the defaults in the argument list instead: `getShortCurrencySymbol(currency = this.defaultCurrency || '')`

packages/react-i18n/src/i18n.ts
```diff
@@ -299,6 +363,31 @@ export class I18n {
...
+    const regionCode = currency.substring(0, 2);
```
lemonmade: This seems really prone to being not true in some instances to me, is there something that makes it a guarantee that all currency codes will work this way?

packages/react-i18n/src/i18n.ts
ismail-syed: ^ ahh, good point.

packages/react-i18n/src/i18n.ts
cejaekl: Well, it is the way ISO creates their codes; first two letters are the region (usually the country, but there are some non-country regions, e.g. `EUR` and `XAF`, for currencies that get used in more than one country.).
https://en.wikipedia.org/wiki/ISO_4217#Code_formation

packages/react-i18n/src/i18n.ts
cejaekl: Good point--thank you!  Adjusting as suggested.

packages/react-i18n/src/i18n.ts
cejaekl: Adjusting as suggested--thank you.

packages/react-i18n/src/i18n.ts
cejaekl: Good idea--adjusting as suggested.

packages/dates/src/tests/utilities.test.ts
cejaekl: I'm happy to follow direction here, but need to know whose to follow.  ;)

For the time being, I've reworked these tests to use `it.each()`.  I think that that should not be too hard to track down, because we can always `grep` for the text given to `describe()`.
(Also, the output of `dev test` shows the file and line number where the failure happened, which makes it easy to find the offending test.)

packages/react-i18n/src/i18n.ts
cejaekl: Both good points.  Adjusted as suggested.

packages/react-i18n/src/i18n.ts
cejaekl: OK.  :)

packages/react-i18n/src/test/i18n.test.ts
cejaekl: Done.   :)

packages/react-i18n/src/test/i18n.test.ts
cejaekl: Done.  :)

packages/react-i18n/src/i18n.ts
cejaekl: I'm happy to change this, but could use a bit of direction re: what you'd like to see.
Maybe something like this?

Extend `NumberFormatOptions`:
```
export interface NumberFormatOptions extends Intl.NumberFormatOptions {
  as?: 'number' | 'currency' | 'percent';
  precision?: number;
  verbosity?: 'short' | 'explicit';
}
```
which results in calls like:
```
formatCurrency(1.25, {currency: 'USD', verbosity: 'short') ==> '$1.25'
```
and
```
formatCurrency(1.25, {currency: 'USD', verbosity: 'explicit') ==> '$1.25 USD'
```

packages/react-i18n/src/i18n.ts
cejaekl: > Another concern, is unformatting a potential need? We had to built `i18n.unformatCurrency` for cases where we'd need to unformat strings that we're already formatted via `i18n.formatCurrency`

No plans to support unformat here.
CC: @robinvankampen , @winterweix , @paulstairmand in case they think otherwise.

packages/react-i18n/src/i18n.ts
cejaekl: I've changed the public API to
```
formatCurrency(1.25, {currency: 'CAD', form: 'short'}) ==> '$1.25'
```
and 
```
formatCurrency(1.25, {currency: 'CAD', form: 'explicit'}) ==> '$1.25 CAD'
```
The legacy version
```
formatCurrency(1.25, {currency: 'CAD'}) ==> '$1.25' or 'CA$1.25' depending on browser/locale
```
remains for backward compatibility with old code.

packages/react-i18n/src/i18n.ts
```diff
@@ -349,7 +362,65 @@ export class I18n {
...
+
+    return shortSymbol.match(alphabeticCharacters)
+      ? info
+      : {symbol: shortSymbol, prefixed: info.prefixed};
+  }
+
   private humanizeDate(date: Date, options?: Intl.DateTimeFormatOptions) {
+    const {defaultTimezone} = this;
+    const timeZone =
+      options && options.timeZone ? options.timeZone : defaultTimezone;
```
lacoursieresimon: Very nit picky, but I guess you can simplify this like that: 
```javascript 
options && options.timeZone || defaultTimezone;
```

packages/react-i18n/src/i18n.ts
```diff
@@ -349,7 +362,65 @@ export class I18n {
...
+  private formatCurrencyExplicit(
+    amount: number,
+    options: Intl.NumberFormatOptions = {},
+  ): string {
+    const value = this.formatCurrencyShort(amount, options);
+    const isoCode = options.currency || this.defaultCurrency || '';
+    if (value.includes(isoCode)) {
+      return value;
+    }
+    return `${value} ${isoCode}`;
```
ismail-syed: nit: 

```
return value.includes(isoCode) ? value : `${value} ${isoCode}`;
```

packages/react-i18n/src/i18n.ts
```diff
@@ -9,6 +9,7 @@ import {
   TimeUnit,
   Weekdays,
 } from '@shopify/javascript-utilities/dates';
+import {applyTimeZoneOffset} from '@shopify/dates';
```
ismail-syed: What does this timezone & date related stuff have to do with currency formatting? Is it because of `NODE_ICU_DATA`?

packages/react-i18n/README.md
```diff
@@ -123,7 +123,10 @@ export default withI18n()(NotFound);
...
+- `formatCurrency()`: formats a number as a currency according to the locale. Its behaviour depends on the `form:` option.
+  - if `form: 'short'` is given, then a possibly-ambiguous short form is used, consisting of the bare symbol if the currency has a symbol, or the ISO 4217 code if there is no symbol for that currency.  Examples:  `CHF 1,25`, `€ 1,25 EUR`, `OMR 1.250`, `$ 1.25 USD`
+  - if `form: 'explicit'` is given, then the result will be the same as for `short`, but will append the ISO 4217 code if it is not already present
+  - if `form:` is not given, then behaviour reverts to the legacy (deprecated) `formatCurrency()`, which is a convenience function that simply _auto-assigns_ the `as` option to `currency` and calls `formatNumber()`.  Note that this will resemble `form: 'short'`, but will sometimes extend the symbol with extra information depending on the browser's implementation of `Intl.NumberFormat` and the locale in use.  For example, `formatCurrency(1.25, {currency: 'CAD'})` may return `$ 1.25`, or it might return `CA$ 1.25`.
```
ismail-syed: > then behaviour reverts to the legacy (deprecated)

Is this _really_ being deprecated? It's used in quite a few places so we'll have to be careful with how we deprecate it if we choose to. 

> but will sometimes extend the symbol with extra information depending on the browser's implementation of `Intl.NumberFormat` and the locale in use.

Do browser implementations dictate this? I thought it was just locales passed in?

Eg. 

```
console.log(new Intl.NumberFormat('en-US', { style: 'currency', currency: 'CAD' }).format(123456.789));
// > "CA$123,456.79"

console.log(new Intl.NumberFormat('en-CA', { style: 'currency', currency: 'CAD' }).format(123456.789));
// > "$123,456.79"
```

packages/react-i18n/src/i18n.ts
cejaekl: I think that adding `full_icu` is what triggered this.
At the end of the day, this test was ill-defined:  it was setting a date using a timestamp in a specific timezone, and then asserting that the day of the week was a certain value in the "local" timezone where the unit test was running.  This would work sometimes, depending on what timezone was configured in the environment where the unit test ran.  (I got to the point where it passed in my local `dev test`, but failed in CI.) . So, yes, this has nothing to do with currency formatting--but the unit test was failing, for a bogus reason, which was blocking my PR from being merged.

packages/react-i18n/README.md
cejaekl: The ECMAScript standard says that a currency symbol may be provided for a given combination of (currency, locale), but leaves what symbol is provided (or, even, whether a symbol is given at all, or whether the ISO 4217 code is used) as "implementation dependent":
https://www.ecma-international.org/ecma-402/6.0/index.html#sec-properties-of-intl-numberformat-constructor

Firefox, Safari and Chrome all seem to use the same implementation, but there's no guarantee that that will be the case for all browsers.

cejaekl: Regarding deprecation of the legacy `formatCurrency()`... we should bring @paulstairmand and @winterweix into that conversation.

PR #924, opened 2019-08-28
Remove tophat script

cartogram:
### what is this?

This PR replaces the `tophat` script with the tophat package from `create-shopify`.

### y tho? seriously.

It is a handy script that contributors of this repository are actively using.  It provides a better dev experience than `yarn link`, `yalc` and the like. I originally moved the script (as is) into `webgen` because I wanted a similar tophat experience, I made some adjustments there and thought it would be easy enough to convert it to typescript, release as a package and allow others to install it in their projects.

I made some improvements on the original, making it to be slightly less tied to this repo structure specifically and it now prompts for missing arguments when run without them.

### sure, what's next then?

I do not think `create-shopify` is the right place for this in the longterm, but that it should either be a standalone `@shopify/tophat` package or pulled into sewing-kit(-next) along with webgen.


TheMallen:
@cartogram  Y tho

cartogram:
I quickly pulled it out and into a separate package that gives a bit of a UI and stuff. This isn't quite ready yet though.

cartogram:
@TheMallen it is fully backwards compatible tho, you can pull down this branch and it all still works.

PR #935, opened 2019-08-30
Universal Cookies

cartogram:
## Description

This PR adds logic around setting and getting cookies to `react-network`.

Question: Should we have a separate package for this (`@shopify/universal-cookies`)? At first I felt like cookies fell into the "network" side of things, but on the other hand, they also live in the browser.

I also need to figure out a way to keep the cookie headers on the server in sync with changes (see comment [here](https://github.com/Shopify/quilt/pull/935/files#diff-30b2790b68e33e0fdcd0c1ad0932a5faR30). One way I was thinking would be that there is a way to attach a change watcher on the cookies during the `applyToContext` step. The cookies setter would then call the watcher every time cookies change. This pushes me further in the direction of this being outside of the `react-network` package.

## Type of change

- [x] `react-network` Minor: New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/packages/react-network/src/server.ts
+++ b/packages/react-network/src/server.ts
@@ -22,9 +22,9 @@ export function applyToContext<T extends Context>(
     ctx.set(header, value);
   }
 
-  for (const [cookie, value] of cookies) {
-    console.log(cookie, value);
-    ctx.cookies.set(cookie, value);
+  for (const [cookie, options] of cookies) {
+    const {value, ...rest} = options;
+    ctx.cookies.set(cookie, value, rest);
   }
 
   return ctx;
--- a/packages/react-network/src/server.ts
+++ b/packages/react-network/src/server.ts
@@ -24,7 +24,7 @@ export function applyToContext<T extends Context>(
 
   for (const [cookie, options] of cookies) {
     const {value, ...rest} = options;
-    ctx.cookies.set(cookie, value, rest);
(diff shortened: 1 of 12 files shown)


packages/react-network/src/test/hooks.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+        .find(MockComponent)!
+        .find('button')!
+        .trigger('onClick');
+
+      expect(wrapper.find(MockComponent)).toContainReactText(`baz`);
+    });
+  });
+});
+
+function mountWithCookies(
```
cartogram: Use creatMount

packages/react-network/src/server.ts
```diff
@@ -22,5 +22,14 @@ export function applyToContext<T extends Context>(
     ctx.set(header, value);
   }
 
+  for (const [cookie, options] of cookies) {
+    const {value, ...rest} = options;
+    ctx.cookies.set(cookie, value, rest);
+  }
+
+  // eslint-disable-next-line no-warning-comments
+  // TODO: Add a watcher function that reapplies the cookies
```
TheMallen: If we just used https://www.npmjs.com/package/universal-cookie it would handle updating the cookies for us

cartogram:
Closing in favour of https://github.com/Shopify/quilt/pull/1002

Issue #892, opened 2019-08-21
Move `Router` component to quilt


michelleyschen:
Move this to quilt to reuse

https://github.com/Shopify/web-rails-proving-ground/blob/master/app/ui/foundation/App/components/Router/Router.tsx

PR #936, opened 2019-08-30
Universal Router

cartogram:
## Description

Fixes https://github.com/Shopify/quilt/issues/892

This PR adds the initial `universal-router` package. This package just exports mostly-as-is components from `react-router`, but provides a common location for apps to import router components as we swap out the implementation for something inline with #724.

I'm going to release this pre-`1.0.0` and so we can iterate on docs quickly as we go.


packages/universal-router/src/Redirect.tsx
```diff
@@ -0,0 +1,21 @@
...
+import {withRouter, RouteComponentProps} from 'react-router';
+import {Redirect as NetworkRedirect} from '@shopify/react-network';
+
+interface Props {
+  url: string;
+}
+
+type ComposedProps = Props & RouteComponentProps<{}, {}>;
+
+class Redirect extends React.Component<ComposedProps> {
```
TheMallen: Y U no hook

packages/universal-router/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+[![Build Status](https://travis-ci.org/Shopify/quilt.svg?branch=master)](https://travis-ci.org/Shopify/quilt)
+[![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](LICENSE.md) [![npm version](https://badge.fury.io/js/%40shopify%2Funiversal-router.svg)](https://badge.fury.io/js/%40shopify%2Funiversal-router.svg) [![npm bundle size (minified + gzip)](https://img.shields.io/bundlephobia/minzip/@shopify/universal-router.svg)](https://img.shields.io/bundlephobia/minzip/@shopify/universal-router.svg)
+
+A universal router for React.
+
+## Installation
+
+```bash
+$ yarn add @shopify/universal-router
```
lemonmade: `react-router`?

packages/universal-router/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "@shopify/universal-router",
+  "version": "0.0.0",
```
lemonmade: Let's keep this sub-1.0 until we have our own thing, and make sure the docs are clear about that

packages/universal-router/src/utilities.ts
```diff
@@ -0,0 +1,3 @@
+export function isServer() {
```
lemonmade: I assume this is just for mocking in tests?

packages/universal-router/package.json
cartogram: yeah I'll add a note/banner to the readme.

packages/universal-router/src/utilities.ts
cartogram: Yeah, thats right.

PR #937, opened 2019-08-30
Migrate testing setup to use babel-jest

ismail-syed:
## Description

Related to https://github.com/Shopify/quilt/issues/857

This PR migrates quilts testing setup to use `babel-jest` over `ts-jest`. This is because `ts-jest` has a [hard time](https://github.com/kulshekhar/ts-jest/issues/766) understanding TS project references. This also makes reviewing the subsequent PR to address https://github.com/Shopify/quilt/issues/857 a bit easier. 

In summary: 

- Upgrading to jest 24
- Adding `babel`  & `babel-jest`
- Get CI 🍏


ismail-syed:
CI is failing because `FieldName` is undefined. `@shopify/address-consts` isn't being picked up in this [export](https://github.com/Shopify/quilt/blob/master/packages/address/src/index.ts#L1) and results in undefined. Changing the export to `export * from '../../address-consts';` fixes it. 

When poking around locally, the package is infact there and symlinked to the folder. It's also referenced as expected in the `moduleNameMapper` jest config.

jest.config.js
```diff
@@ -16,15 +16,16 @@ const moduleNameMapper = getPackageNames().reduce(
       '<rootDir>/packages/react-html/src/server/index.ts',
     '@shopify/react-network/server':
       '<rootDir>/packages/react-network/src/server.ts',
+    '@shopify/address-consts': '<rootDir>/packages/address-consts/src/index.ts',
```
ismail-syed: This one drove me up the wall. `@shopify/address-conts` was not getting _module mapped_ properly. 
`require.resolve('@shopify/address-consts')`  was resolving to `quilt/packages/address/src/index.ts`, instead of `quilt/packages/address-consts/src/index.ts`. Moving this up in the list manually got the module resolving correctly.

lemonmade: I think we could also fix this by adding a `$` to the `@shopify/${name}` bit, presumably that part is a regex and `address` matches `address-consts`

Issue #934, opened 2019-08-29
react-server-webpack-plugin's generated code should render providers that are tied to the incoming request we know we always need (CSRF, cookies)


michenly:
Related comment 
https://github.com/Shopify/quilt/issues/891#issuecomment-524438184

PR #941, opened 2019-08-30
Add cookie middleware and default providers to webpack plugin generated files

cartogram:
This pr is a second take of https://github.com/Shopify/quilt/pull/935 taking the third-party approach  and works #934 into the solution.

### DIFF
--- a/packages/react-server/package.json
+++ b/packages/react-server/package.json
@@ -30,6 +30,8 @@
     "@shopify/react-html": "^9.2.1",
     "@shopify/react-network": "^3.2.0",
     "@shopify/sewing-kit-koa": "^6.1.1",
+    "react-cookie": "4.0.1",
+    "@shopify/react-csrf-universal-provider": "1.0.2",
     "chalk": "^2.4.2",
     "koa": "^2.5.0",
     "koa-compose": "^4.1.0",
--- a/packages/react-server/package.json
+++ b/packages/react-server/package.json
@@ -33,7 +33,8 @@
     "chalk": "^2.4.2",
     "koa": "^2.5.0",
     "koa-compose": "^4.1.0",
-    "koa-mount": "^4.0.0"
+    "koa-mount": "^4.0.0",
+    "universal-cookie-koa": "4.0.1"
   },
   "devDependencies": {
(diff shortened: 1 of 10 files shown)


cartogram:
@TheMallen makes sense, I'm working this issue https://github.com/Shopify/quilt/issues/934 into this PR.

packages/react-server/package.json
```diff
@@ -30,10 +30,13 @@
     "@shopify/react-html": "^9.2.1",
     "@shopify/react-network": "^3.2.0",
     "@shopify/sewing-kit-koa": "^6.1.1",
+    "react-cookie": "4.0.1",
+    "@shopify/react-csrf-universal-provider": "1.0.2",
```
michelleyschen: All the dependency that a project may already have install should be a range instead of a specific version.

ie.
```
"@shopify/react-network": ">=3.2.0",
"@shopify/react-html": ">=8.0.0"`
"@shopify/react-csrf-universal-provider": ">=1.0.0"
...all the other ones
```

packages/react-server/src/providers/ConditionalProvider.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+
+interface Props<T extends React.ComponentType<any>> {
+  condition: boolean;
+  provider: T;
+  // props: React.ComponentPropsWithoutRef<T>;
+  props?: any;
```
cartogram: TODO, I'd like to get this to infer props from the component

packages/react-server/src/providers/providers.tsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import {CookiesProvider} from 'react-cookie';
+import {useRequestHeader} from '@shopify/react-network';
+import {CsrfUniversalProvider} from '@shopify/react-csrf-universal-provider';
+import {ConditionalProvider} from './ConditionalProvider';
+
+interface Options {
+  cookies?: boolean;
+  csrf?: boolean;
```
cartogram: Will add more options here as we need them

cartogram:
superseding by https://github.com/Shopify/quilt/pull/1002 and https://github.com/Shopify/quilt/pull/985

PR #942, opened 2019-09-03
[Polyfills] Replace deprecated babel-polyfills with core-js

tleunen:
## Description

`@babel/polyfill` has been deprecated in favor of `core-js` and `regenerator-runtime`.

## Checklist

### DIFF
--- a/packages/polyfills/src/base.ts
+++ b/packages/polyfills/src/base.ts
@@ -1,1 +1,2 @@
-require('@babel/polyfill');
+require('core-js/stable');
+require('regenerator-runtime/runtime');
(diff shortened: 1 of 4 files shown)


tleunen:
@TheMallen - Any idea why the tests fail?

packages/polyfills/src/base.ts
```diff
@@ -1 +1,2 @@
-require('@babel/polyfill');
+require('core-js/stable');
+require('regenerator-runtime/runtime');
```
lemonmade: Just to confirm, this is removed by `@babel/preset-env` when generators are supported?

tleunen: I haven't tested, but it should yes.

`@babel/polyfill` prior to 7.4 was already importing it (and a couple of other things): https://github.com/babel/babel/blob/v7.3.5/packages/babel-polyfill/src/index.js and according to the documentation, it is equivalent to the 2 imports I added.
https://babeljs.io/docs/en/next/babel-polyfill.html

tleunen:
Closing in favor of #1323

PR #944, opened 2019-09-03
React renderable error in tests

### LINKED ISSUE #932: quilt_rails gem should have special handling for test environments
Currently if an app depending on `quilt_rails` tries to do any kind of controller test on their `UiController` they will likely get an unhelpful error about react-server not responding. Since we don't want folks to try to test their react apps from the rails side, we should make this explicit by providing a customized error message in tests that tells the consumer not to test in that way.

TheMallen:
closes #932 

## Problem
Currently if a user tries to use railsy integration tests on their React routes, they will see the following in their tests:

```
"Errno::ECONNREFUSED: Waiting for React server to boot up. If this error presists verify that @shopify/react-server is configured on #{url}"
```

This error is not very helpful for a few reasons:
- makes them think they have misconfigured the server somehow
- does not push them in the right direction for testing React apps (ie. our libraries / jest)

## Solution
This PR adds a special error message to `render_react` which informs the consumer they should not be using Rails integration tests, but rather use Jest and `@shopify/react-testing`

## Possible issues
If we actually *do* want to support integration / E2E tests for Rails devs, we would need a very different solution. If it becomes a common request / pain point we may want to revisit.

### DIFF
--- a/gems/quilt_rails/lib/quilt_rails/react_renderable.rb
+++ b/gems/quilt_rails/lib/quilt_rails/react_renderable.rb
@@ -7,6 +7,8 @@ module Quilt
     include ReverseProxy::Controller
 
     def render_react
+      raise DoNotIntegrationTestError if Rails.env.test?
+
       if defined? ShopifySecurityBase
         ShopifySecurityBase::HTTPHostRestriction.whitelist([Quilt.configuration.react_server_host]) do
           proxy
@@ -37,6 +39,12 @@ module Quilt
       def initialize(url)
         # rubocop:disable LineLength
         super "Errno::ECONNREFUSED: Waiting for React server to boot up. If this error presists verify that @shopify/react-server is configured on #{url}"
+      end
+    end
+
+    class DoNotIntegrationTestError < StandardError
+      def initialize
+        super "Do not try to use Rails integration tests on your quilt_rails app. Instead use Jest and @shopify/react-testing to test your React application directly."
       end
     end
   end
--- a/gems/quilt_rails/test/quilt_rails/react_renderable_test.rb
+++ b/gems/quilt_rails/test/quilt_rails/react_renderable_test.rb
@@ -4,15 +4,27 @@ module Quilt
     include Quilt::ReactRenderable
 
     def test_render_react_calls_reverse_proxy_with_server_uri_and_csrf
+      Rails.env.stubs(:test?).returns(false)
       url = "#{Quilt.configuration.react_server_protocol}://#{Quilt.configuration.react_server_host}"
 
       assert_equal(render_react, reverse_proxy(url, headers: { 'X-CSRF-Token': form_authenticity_token }))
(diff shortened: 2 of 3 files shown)


gems/quilt_rails/test/quilt_rails/react_renderable_test.rb
```diff
@@ -4,15 +4,27 @@ class ReactRenderableTest < Minitest::Test
...
+    def test_render_react_errors_in_tests
+      Rails.env.stubs(:test?).returns(true)
+      assert_raises Quilt::ReactRenderable::DoNotIntegrationTestError do
+        render_react
+       end
+    end
+
+    private
+
+    # Stubbing this method the mixin calls
```
TheMallen: Just added some comments here to make it clearer these are not tests, but an implementation detail of the tests.

gems/quilt_rails/test/quilt_rails/react_renderable_test.rb
```diff
@@ -4,15 +4,27 @@ class ReactRenderableTest < Minitest::Test
     include Quilt::ReactRenderable
 
     def test_render_react_calls_reverse_proxy_with_server_uri_and_csrf
+      Rails.env.stubs(:test?).returns(false)
```
TheMallen: originally had this in the existing error handler, but seemed smarter to just fail fast if it's a test.

gems/quilt_rails/lib/quilt_rails/react_renderable.rb
```diff
@@ -39,5 +41,11 @@ def initialize(url)
         super "Errno::ECONNREFUSED: Waiting for React server to boot up. If this error presists verify that @shopify/react-server is configured on #{url}"
       end
     end
+
+    class DoNotIntegrationTestError < StandardError
+      def initialize
+        super "Do not try to use Rails integration tests on your quilt_rails app. Instead use Jest and @shopify/react-testing to test your react application directly."
```
GoodForOneFare: I'd try to be less DO AS WE SAY:

```suggestion
        super "quilt_rails does not support Rails integration tests. Instead, use Jest and @shopify/react-testing to test your react application components in isolation."
```

gems/quilt_rails/lib/quilt_rails/react_renderable.rb
```diff
@@ -39,5 +41,11 @@ def initialize(url)
         super "Errno::ECONNREFUSED: Waiting for React server to boot up. If this error presists verify that @shopify/react-server is configured on #{url}"
       end
     end
+
+    class DoNotIntegrationTestError < StandardError
+      def initialize
+        super "quilt_rails does not support Rails integration tests. Instead, use Jest and @shopify/react-testing to test your react application components in isolation."
```
ismail-syed: Capitalize `react` since we're capitalizing `Rails`?

Issue #945, opened 2019-09-03
Move templates for `react-server-webpack-plugin` to syntax-highlight-able files


cartogram:
## Overview

If we move the generated code for the `react-server-webpack-plugin` away from inline template strings and into a separate file that is read and transformed, we can provide a better developer experience when editing them.

PR #947, opened 2019-09-04
Pull `react-server-webpack-plugin` files source into separate templates

cartogram:
## Description

This PR pulls the inline template strings used to write the "virtual" files out into templates that are parsed and transformed via babel. This allows those files to be a little more manageable and syntax-highlighted.

Fixes https://github.com/Shopify/quilt/issues/945 

~~This PR requires the `@shopify/ast-utilities` package added here https://github.com/Shopify/quilt/pull/949~~

### DIFF
--- a/packages/react-server-webpack-plugin/src/react-server-webpack-plugin.ts
+++ b/packages/react-server-webpack-plugin/src/react-server-webpack-plugin.ts
@@ -1,6 +1,9 @@
 import {join, resolve} from 'path';
 import {Compiler} from 'webpack';
 import VirtualModulesPlugin from 'webpack-virtual-modules';
+import {readFileSync} from 'fs-extra';
+import {transformSync} from '@shopify/ast-utilities';
+import {fillCallExpressionOptions} from './transforms';
 
 interface Options {
   basePath: string;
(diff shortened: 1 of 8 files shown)


packages/react-server-webpack-plugin/src/templates/server.ts
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import {createServer} from '@shopify/react-server';
+import App from 'index';
+
+const render = ctx =>
```
TheMallen: If we are already using babel we could also use jsx :0

packages/react-server-webpack-plugin/src/react-server-webpack-plugin.ts
```diff
@@ -76,39 +79,28 @@ function serverSource(options: Options) {
...
+    initial.toString(),
+    fillCallExpressionOptions('createServer', {
+      port: serverPort,
+      ip: serverIp,
+      assetPrefix: serverAssetPrefix,
+    }),
+  );
+
   return `
     ${HEADER}
```
TheMallen: Could we just include the header as an interpolation in the template as well?

Edit: I see we aren't doing templates quite how I expected, interesting

packages/react-server-webpack-plugin/src/transforms.ts
```diff
@@ -0,0 +1,25 @@
+import * as t from '@babel/types';
+
```
TheMallen: It feels a bit weird to me to do this as an ast transform rather than just using handlebars or something since it's so simple

packages/react-server-webpack-plugin/src/react-server-webpack-plugin.ts
cartogram: We could pass the header in yeah, but I didn't really see much value in it vs just popping it on the top like this.

cartogram:
@TheMallen can I get another look at one?

cartogram:
@TheMallen oh totally, I'll explore, but gonna merge this for now.

PR #951, opened 2019-09-04
[react-graphql] / Support `ssr=false` for `useQuery`

loic-d:
## Description

Fixes https://github.com/Shopify/quilt/issues/703

Consumers of `useQuery` can now prevent a query to resolve when rendering on the server.

```
  const {data, loading, error} = useQuery(CustomerListQuery, {
    ssr: false,
  });
```

You can 🎩 this change in a consumer project.
In quilt, `cd packages/react-graphql && yarn run build`
`yarn tophat react-graphql <relative-path-consumer-project>`

`ssr: false` in the above code snippet will prevent `CustomerListQuery` to resolve on the server. Instead, it will be resolved on the client at run-time.

Screen Shot 2019-09-04 at 1 05 41 PM

## Type of change

- [x] Minor: New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/packages/react-graphql/src/hooks/query.ts
+++ b/packages/react-graphql/src/hooks/query.ts
@@ -35,11 +35,15 @@ export default function useQuery<
     client: overrideClient,
     notifyOnNetworkStatusChange,
     context,
+    ssr = true,
   } = options;
   const client = useApolloClient(overrideClient);
 
-  if (typeof window === 'undefined' && (skip || fetchPolicy === 'no-cache')) {
-    return {...createDefaultResult(client, variables), loading: !skip};
+  if (
+    typeof window === 'undefined' &&
+    (skip || fetchPolicy === 'no-cache' || !ssr)
+  ) {
+    return createDefaultResult(client, variables);
   }
 
   const query = useGraphQLDocument(queryOrComponent);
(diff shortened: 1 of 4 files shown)


packages/react-graphql/src/hooks/query.ts
```diff
@@ -35,10 +35,14 @@ export default function useQuery<
...
+    ssr = true,
   } = options;
   const client = useApolloClient(overrideClient);
 
-  if (typeof window === 'undefined' && (skip || fetchPolicy === 'no-cache')) {
+  if (
+    typeof window === 'undefined' &&
+    (skip || fetchPolicy === 'no-cache' || !ssr)
+  ) {
     return {...createDefaultResult(client, variables), loading: !skip};
```
lemonmade: Do you know what the behaviour of `loading` should be here?

michelleyschen: loading can be infer from here
https://github.com/trojanowski/react-apollo-hooks/blob/master/src/useQuery.ts#L89

loic-d: Good point.
Looks like [they are setting is to `false` during SSR ](https://github.com/trojanowski/react-apollo-hooks/blob/master/src/useQuery.ts#L160)(`react-apollo-hook`)

Would it be safe to always set `loading` to `false`?

lemonmade: I guess `loading = false` probably makes sense for any kind of skipping, yeah

michelleyschen: false is !ssr? https://github.com/trojanowski/react-apollo-hooks/blob/master/src/useQuery.ts#L89

PR #952, opened 2019-09-04
Fix tophat script

cartogram:
## Description
see https://github.com/Shopify/webgen/pull/235

This PR updates the `tophat` script via a `create-shopify` bump. The main issue this was fixing was caused by the presence of both the `react-router` and `@shopify/react-router` packages.


michelleyschen:
🎩 the 🎩 and it works great. Thanks for the fast turn around.

PR #962, opened 2019-09-05
Use TypeScript project references

ismail-syed:
## Description

Closes https://github.com/Shopify/quilt/issues/857

This PR adds Typescript [Project References](https://www.typescriptlang.org/docs/handbook/project-references.html) to quilt. In doing so we aim to see faster CI builds and faster local rebuilds. 

### Faster CI builds

* Previously [~19mins](https://travis-ci.org/Shopify/quilt/builds/611064917?utm_source=github_status&utm_medium=notification) CI runs
* Currently in this branch I'm getting [~14mins](https://travis-ci.org/Shopify/quilt/builds/611484018?utm_source=github_status&utm_medium=notification)

### Faster local rebuilds

Screen Shot 2019-11-13 at 1 05 37 PM


packages/address-consts/package.json
```diff
@@ -6,10 +6,6 @@
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "sideEffects": false,
-  "scripts": {
-    "build": "tsc --p tsconfig.build.json",
-    "prepublishOnly": "yarn run build"
```
ismail-syed: I'm not 100% sure about this, but I don't believe we need these `prepublishOnly` hooks anymore. sewing-kit monorepo packages don't include any of these so I presume it's not necessary.

packages/tsconfig.json
```diff
@@ -0,0 +1,71 @@
...
+    {"path": "./react-server-webpack-plugin"},
+    {"path": "./react-shopify-app-route-propagator"},
+    {"path": "./react-shortcuts"},
+    {"path": "./react-testing"},
+    {"path": "./react-tracking-pixel"},
+    {"path": "./react-universal-provider"},
+    {"path": "./semaphore"},
+    {"path": "./sewing-kit-koa"},
+    {"path": "./useful-types"},
+    {"path": "./with-env"}
```
TheMallen: It sucks that we have to manually maintain a list like this. It doesn't have to be part of this PR but at the least we should add a task to include adding an entry here in the new package generator, or at least have a separate script that gets run after like we do with adding to the main readme.

packages/react-tracking-pixel/tsconfig.json
```diff
@@ -0,0 +1,16 @@
...
+    "baseUrl": "./src",
+    "rootDir": "./src"
+  },
+  "include": [
+    "../../config/typescript/**/*",
+    "./src/**/*.ts",
+    "./src/**/*.tsx"
+  ],
+  "exclude": ["**/*.test.ts", "**/*.test.tsx"],
+  "references": [{"path": "../react-html"}]
```
TheMallen: We already have issues with remembering to add dependencies to `package.json` when they are local packages. We are going to want to have some way of statically enforcing that dependencies are tracked properly, whether through a linter, a custom script, a repo-level test, or whatever.

script.js
```diff
@@ -0,0 +1,17 @@
+const {resolve, basename} = require('path');
```
TheMallen: I don't think this was meant to be left in

packages/react-async/src/Prefetcher.tsx
```diff
@@ -118,11 +118,13 @@ class ConnectedPrefetcher extends React.PureComponent<Props, State> {
     if (
       timeout != null &&
       urlsEqual(closestUrl, timeoutUrl) &&
+      // @ts-ignore
```
TheMallen: why this

packages/react-i18n-universal-provider/src/I18nUniversalProvider.tsx
```diff
@@ -26,6 +26,7 @@ export function I18nUniversalProvider({
   const [serialized, Serialize] = useSerialized<Serialized>('i18n');
 
   const i18nDetails: I18nDetails = {
+    // @ts-ignore
```
TheMallen: why tho

packages/react-self-serializers/src/I18n.tsx
```diff
@@ -0,0 +1,62 @@
+import React, {useRef} from 'react';
+
+import {useSerialized, useHtmlAttributes} from '@shopify/react-html';
```
TheMallen: what's going on with these added typescript files? do you just need to rebase?

packages/graphql-testing/src/links/mocks.ts
```diff
@@ -7,7 +7,7 @@ export class MockLink extends ApolloLink {
     super();
   }
 
-  request(operation: Operation) {
+  request(operation: Operation): Observable<any> {
```
TheMallen: is `any` right here?

packages/react-import-remote/src/hooks.ts
```diff
@@ -103,9 +103,8 @@ export function useImportRemote<Imported = unknown>(
     () => {
       if (defer === DeferTiming.Idle) {
         if ('requestIdleCallback' in window) {
-          idleCallbackHandle.current = (window as WindowWithRequestIdleCallback).requestIdleCallback(
-            loadRemote,
-          );
+          idleCallbackHandle.current = (window as WindowWithRequestIdleCallback &
```
TheMallen: nit: maybe we should add `type ExtendedWindow<T> = window & typeof globalThis & T` to `@shopify/useful-types` and use it for all these weird window extensiony things

yarn.lock
```diff
@@ -11097,6 +11097,11 @@ typescript-estree@2.1.0:
     lodash.unescape "4.0.1"
     semver "5.5.0"
 
+typescript@^3.6.2:
```
GoodForOneFare: Can the `@typescript@~3.2.1` (line 11105) be removed?

package.json
```diff
@@ -4,14 +4,14 @@
...
-    "build": "lerna run build",
+    "build": "yarn tsc --build packages",
     "lint": "yarn eslint '**/*.{ts,tsx}'",
     "ci:lint-docs": "yarn generate docs && test -z \"$(git status --porcelain)\" || echo 'The root README has not been updated. Run `yarn generate docs` in the root of your quilt directory and try again.'",
+    "pretest": "yarn run build",
     "test": "NODE_ICU_DATA=node_modules/full-icu jest --maxWorkers=3 --coverage=false",
     "check": "lerna run check",
     "release": "lerna publish",
-    "clean": "rimraf './packages/*/dist/**/*.{js,d.ts}'",
+    "clean": "rimraf './packages/*/dist/**/*.{js,d.ts}' './packages/*/*.tsbuildinfo' './packages/*.tsbuildinfo'",
```
GoodForOneFare: I know `dist/src` is annoying, but it's unavoidable for most projects.  So I'd prefer standardising on that, even for projects that currently use `dist`.

GoodForOneFare: More importantly, how are you planning to have `tsbuildinfo` files cached in CI?  I'm still at the top of the PR, so I may be missing the magic that makes it happen.

packages/useful-types/tsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
+  "extends": "../tsconfig_base.json",
+  "compileOnSave": false,
```
GoodForOneFare: `compileOnSave` should be in the base config, and deleted from all others.

packages/useful-types/tsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+  "compilerOptions": {
+    "outDir": "./dist",
+    "baseUrl": "./src",
+    "rootDir": "./src"
+  },
+  "include": [
+    "../../config/typescript/**/*",
```
GoodForOneFare: What is this pulling in?

packages/useful-types/tsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+    "./src/**/*.ts",
+    "./src/**/*.tsx"
```
GoodForOneFare: Why does this need to process tsx files?

packages/useful-types/tsconfig.json
```diff
@@ -0,0 +1,15 @@
...
+  ],
+  "exclude": ["**/*.test.ts", "**/*.test.tsx"]
```
GoodForOneFare: There are no test files, so 🔥

packages/with-env/tsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
+  "extends": "../tsconfig_base.json",
+  "compileOnSave": false,
+  "compilerOptions": {
+    "outDir": "./dist",
+    "baseUrl": "./src",
+    "rootDir": "./src"
+  },
+  "include": [
+    "../../config/typescript/**/*",
```
GoodForOneFare: Same question about pulling in `config` and tsx files as above.

packages/tsconfig_base.json
```diff
@@ -22,8 +26,8 @@
       "es2017",
       "es2018",
       "esnext.asynciterable"
-    ]
+    ],
+    "typeRoots": ["../node_modules/@types"]
   },
-  "include": ["./**/*.ts", "./**/*.tsx"],
-  "exclude": ["./node_modules"]
+  "include": ["../config/typescript/**/*.d.ts"]
```
GoodForOneFare: Does it make sense to have an include here?  I thought all projects should specify their own includes?

ismail-syed:
Thanks for all the feedback folks. I'll get back around to addressing these comments sometime this week. Metrics tracking works is higher on my list of but I'll be been pushing this in-between cycles.

packages/react-async/src/Prefetcher.tsx
ismail-syed: Couple ts-ignores I forgot to address.

packages/react-import-remote/src/hooks.ts
ismail-syed: Added to useful-types

script.js
ismail-syed: Nope, removing

packages/useful-types/tsconfig.json
ismail-syed: Doesn't seem like anything. I wrote a script that converted all `tsconfig.build.json
` => `tsconfig.json` and updated the extended field. Looking at the history of this file this must have been included from the generators.

packages/tsconfig_base.json
ismail-syed: This should be removed

packages/tsconfig.json
ismail-syed: I've added a [test](https://github.com/Shopify/quilt/pull/962/files#diff-09d775e5b319188b96e0907456993c34) to cover for now.

packages/react-tracking-pixel/tsconfig.json
ismail-syed: The compiler does an ok-ish good job at checking this. If a project is missing an explicit reference here then the build fails. That said, I've seen cases of it passing without an explicit reference. I'll open up an issue.

packages/react-self-serializers/src/I18n.tsx
ismail-syed: Yes, needed a rebase.

packages/graphql-testing/src/links/mocks.ts
ismail-syed: I spent a bit of time trying to address these. The changes were significant in some cases. I'd prefer having those changes in a subsequent PR if that's ok with you.

yarn.lock
ismail-syed: Removed

package.json
ismail-syed: > So I'd prefer standardising on that, even for projects that currently use dist

I've updated all the monorepo tsconfigs to reflect this as sewing-kit does. (`'.'` as the `baseUrl` and `rootDir`). Also changed the number of imports from `./dist/foo` to `./dist/src/foo`.

ismail-syed: > how are you planning to have tsbuildinfo files cached in CI?

I tried getting Travis to only cache the `tsbuildinfo` files, but I couldn't find any docs or working examples of caching files. Only directories. How does sewing-kit cache `tsbuildinfo` files? 

In Travis, what if we move all the `tsbuildinfo`s to a `tmp_build_cache` directory of some sort after a build and dump them back into their respective `packages/*` paths before a subsequent build starts.

ismail-syed:
I've rebased the PR and addressed all the comments. Ready for another set of 👀 @GoodForOneFare @TheMallen

PR #966, opened 2019-09-06
Update Lerna

cartogram:
## Description

In order to make the ignore option added in https://github.com/Shopify/quilt/pull/965 work, we need to be on a more recent version of Lerna. This PR brings us to the most recent version, and testing the ignoreChanges config works as expected (where currently it does not). As far as I can, there is no reason for us not to upgrade.

If all goes well, I'll quickly roll this out to our other Lerna repos.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "jest": "^24.9.0",
     "jest-enzyme": "^7.0.1",
     "jest-transform-graphql": "^2.1.0",
-    "lerna": "^2.9.0",
+    "lerna": "^3.16.4",
     "plop": "^2.0.0",
     "prettier": "^1.14.0",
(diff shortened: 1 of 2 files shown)


cartogram:
Closing @michelleyschen noticed that only `s` in `commands` was the difference between v2/v3 so we can stay on v2 and still use this feature.

michelleyschen:
lerna v2 `commands.ignore`
lerna v3 `command. ignoreChanges`

they renamed a bunch of things b/t he 2 version but its good now!
https://github.com/Shopify/quilt/commit/be8453b7f3c3c7a0d0e1683ef07a5cfbbdf74b6b#diff-52d4bf43db96176e77cc9038a78f98e1

PR #968, opened 2019-09-06
🔥 remove all of `@shopify/javascript-utilities` that I could

michelleyschen:
## Description

Part of https://github.com/Shopify/quilt/issues/136
I was able to remove all but one package.

## Type of change

I think dependency update is minor right?

## Checklist

### DIFF
--- a/packages/dates/package.json
+++ b/packages/dates/package.json
@@ -23,7 +23,7 @@
   },
   "homepage": "https://github.com/Shopify/quilt/blob/master/packages/dates/README.md",
   "dependencies": {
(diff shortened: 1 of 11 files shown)


michelleyschen:
> 👍 I've been doing dependency update as minor. What is the remaining package?

Defiantly been debating over patch or minor for dependency update. But since removing @shopify/javascript-utilities does not change any behaviour or API, I was going to go with patch. (Where like a react dependency update will certainly be a minor)

What do you think? @cartogram

cartogram:
@michelleyschen If nothing changes for the consumer, I agree patch is right.

PR #969, opened 2019-09-06
React server fix

michelleyschen:
## Description
This PR comes after https://github.com/Shopify/quilt/pull/973

Widen the range of dependencies `react-server` will accept. Mostly going with the last major update with a few exception.

## Type of change

## Checklist


packages/react-server/package.json
```diff
@@ -24,12 +24,13 @@
   },
   "homepage": "https://github.com/Shopify/quilt/blob/master/packages/react-server/README.md",
   "dependencies": {
-    "@shopify/network": "^1.4.1",
-    "@shopify/react-async": "^3.0.6",
-    "@shopify/react-effect": "^3.2.2",
-    "@shopify/react-html": "^9.2.1",
-    "@shopify/react-network": "^3.2.0",
-    "@shopify/sewing-kit-koa": "^6.1.1",
+    "@shopify/network": ">=1.0.0",
```
cartogram: For my own curiosity/ learning, I am still a little unclear about why these needed to be widened.

GoodForOneFare: Yeah, I'd like some team discussion about making this step.

packages/react-server/src/render/render.tsx
```diff
@@ -45,18 +46,27 @@ export function createRender(render: RenderFunction, options: Options = {}) {
...
+            </NetworkContext.Provider>
+          </HydrationContext.Provider>
+        </AsyncAssetContext.Provider>
+      );
+    }
 
     try {
       const app = render(ctx);
       await extract(app, {
-        decorate(app) {
```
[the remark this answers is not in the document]
michelleyschen: app was confusing to me since app is also used above in a different scope. This is to make it more clear.

packages/react-server/CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## [Unreleased] -->
 
+## [0.4.0] - 2019-09-??
+
+### Fixed
+
+- Widen the range of dependencies react-server will accept. Fix the 🐛 where some provider does not get render in the stream. Add rendering of `HydrationContext` [#969](https://github.com/Shopify/quilt/pull/969)
```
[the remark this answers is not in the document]
michelleyschen: Awesome! I'll add those changes

packages/react-server/package.json
michelleyschen: The reason I widen the range of dependencies is because we are starting to encounter a lot of bugs whenever a repo install 2 different dependency that separately depends on 2 different version of the same library.

A big part of the 🐛 me and Gord saw early this week in `web-rails-proving-ground` is due to 2 version of `react-network`, and 2 version of `react-effect`. These type of bugs are difficult to figure out and not always clear how to resolve.

IMO having a range make sense when including any library that has a central manager / context. It is also more correct since the code does work across a given range and would not force our consumer to need to cyclically update a list of dependency because of one patch update.

What I do currently to find the range, is finding the last time the particular API signature use in the library were change in the dependency, and do the minor release after that. This should guarantee non breaking but include all the bug fixes. In some cases if there is a major release after that I would go with that just because I dont want us to have to support too far back.

This does come with one major down side: lerna would no longer pick up the version and automatically update for us. (Unless it does? That would be sweet) So we would need to be more careful when we release major / minor version of these context including library to manually check where the dependency being use may warrant a range update as well.

Would love to hear everyone's thoughts and what alternatives we may have.

GoodForOneFare: > What I do currently to find the range, is finding the last time the particular API signature use in the library were change in the dependency, and do the minor release after that

If we decide to switch to ranges, I would like to see a document that tells contributors/releasers how to follow this process.

ismail-syed: `>=` will pick up major (potentially breaking) updates too, is that what we want? Widening the range this much feels like it might encourage some bad practices. If a consumer is in one of these cases, doesn't `yarn resolutions`  solve this for us?

>  lerna would no longer pick up the version and automatically update for us

I don't think this is worth the tradeoff given `yarn resolution`'s can solve the other issue.

GoodForOneFare: As well as Ismail's point about breaking changes, I'm concerned that:
- We're trying to fix the problem in the wrong place
  - i.e., can we automate this as a CI check?  Does dependabot fix this?
- We're making it harder to ship Quilt changes

michelleyschen: > `>=` will pick up major (potentially breaking) updates too, is that what we want? 

I can certainly make it `"@shopify/network": ">=1.0.0  <2.0.0",`

michelleyschen: I will give `yarn resolution` a try with `web-rails-proving-ground` now

GoodForOneFare: `yarn resolutions` should be a last resort recommendation.  Those should not become a common solution to Quilt problems.

If I recall correctly, the duplicate libs that we saw earlier this week were fixed by deduping dependencies.  No resolutions were necessary, so I don't think our current setup gets us into a space where resolutions are the only solution.

michelleyschen: Overall this clearly warrant more discussion, I am going to make a separate PR of everything here except for the range change so we can ship the bug fix first.

cartogram: Can we write a shared post-install script that warns about duplicate dependencies and advises about how to run the de-duplicate script? we could include it in sewing-kit.

ismail-syed: > I can certainly make it "@shopify/network": ">=1.0.0 <2.0.0",

Isn't that just the same as `^1.0.0`. It'll pick up the latest minor version.

michelleyschen: purely informational:
I just did a quilt publish and it seems that it forced all the range into a non-range dependency.

So lerna publish does pick up range when a dependency update occur, we just need to manually decided if we want range / no-range each time.

michelleyschen: > > I can certainly make it "@shopify/network": ">=1.0.0 <2.0.0",
> 
> Isn't that just the same as `^1.0.0`. It'll pick up the latest minor version.

now I wonder why we do 
```
"peerDependencies": {
    "react": ">=16.8.0 <17.0.0"
  },
```
everywhere 🤔

michelleyschen: > If I recall correctly, the duplicate libs that we saw earlier this week were fixed by deduping dependencies.

Agree this would be a prefer solution. Although this is difficult because the error always surface in different ways.

What I think we need is a better trip wire for the consumer to know they need to deduping dependencies and which library they need to do it with. Which is inline with what Matt suggested.

A sewing-kit doctor task maybe?

PR #975, opened 2019-09-09
Use root typescript version 

ismail-syed:
## Description

Related to https://github.com/Shopify/quilt/issues/857 - Makes it easier to review the upcoming larger TS refactor PR of this repo

This PR removes all the monorepo package level TS dependencies and defers the use of typescript to be consumed from the root level of the project. 

**Why?**
- We don't really need this with Typescript project references moving forward
- Managing independent TS version for each monorepo package doesn't scale


packages/admin-graphql-api-utilities/package.json
```diff
@@ -22,9 +22,7 @@
     "url": "https://github.com/shopify/quilt/issues"
   },
   "homepage": "https://github.com/Shopify/quilt/blob/master/packages/admin-graphql-api-utilities/README.md",
-  "devDependencies": {
-    "typescript": "~3.2.1"
-  },
+  "devDependencies": {},
```
TheMallen: 🔥  these empty fields IMO

packages/react-router/package.json
```diff
@@ -29,7 +29,7 @@
     "react-router-dom": "^5.0.1"
   },
   "devDependencies": {
-    "typescript": "~3.2.1"
```
michelleyschen: do we need this in `devDependencies` when it is in `dependencies` as well?

packages/react-serialize/package.json
```diff
@@ -29,8 +29,5 @@
   "peerDependencies": {
     "react": "^16.0.0"
   },
-  "devDependencies": {
```
michelleyschen: damn I really gonna kill this packages

packages/react-router/package.json
ismail-syed: Oh snap, good catch. It was from a bad rebase. I'll check through the rest

michelleyschen:
I am not sure if I am understanding this all correctly but is there any chance that we can do the same thing for `tslib` ? and should we?

ismail-syed:
> react-router is the only package I have some concern about.

That was a bad rebase, it's fixed now. 

> ... is there any chance that we can do the same thing for tslib ? and should we?

`tslib` is a dependency whereas `typescript` is a dev dependency. Consumers still need to install the specified `tslib`

Issue #959, opened 2019-09-05
Node error traces are unreadable in Splunk


GoodForOneFare:
## Overview
Each line of a Node error traces appear as a separate Splunk log line.  This makes them unreadable.

[This Splunk forum](https://answers.splunk.com/answers/705902/why-are-my-multi-line-events-getting-split.html) suggests that adding timing info to logs may help to group them.  

## Consuming repo
beta-buddy, web-rails-proving-ground, etc

ismail-syed:
Do we have any splunk queries to view the logs in question?

GoodForOneFare:
No, because of `some Node exceptions aren't logged` (#958).  You can create a service, add uncaught exception handler, change the middleware's logger to `Verbose`, and log test exceptions before app startup / during app render.

ismail-syed:
web-rails-proving-ground  in prod is updated with `@shopify/react-server@0.5.0`. We now have readable splunk logs with for requests going through`@shopify/react-server`.

PR #978, opened 2019-09-10
[@shopify/react-server] Improve logger to provide more readable splunk logs

ismail-syed:
## Description

Fixes https://github.com/Shopify/quilt/issues/959

### Before (Separate logs)
* image
* image
* image

### After (Aggregated logs)

image

This is done by adding a `buffer` property to the logger class and joining logs to it throughout the request lifecycle across middlewares. This buffer is only used in production. In development were just console logging. 

Also, since I'm already in here logging, I added a few other properties to the log object that I think would be useful in debugging: `http_method`, `http_response`

### 🎩 
- This is shipped in `@shopify/react-server@0.5.0-beta.1`. You can pull that into a Rails SSR project and poke around
- I've also pulled this into `web-rails-proving-ground` in production. You can test it out by hitting the `/debug-errors` route and verifying that those logs are readable in splunk. You can find the preconfigured splunk queries for the production machines in `CloudPortal` linked from ServicesDB.

## Checklist

### DIFF
--- a/packages/react-server/src/logger/logger.ts
+++ b/packages/react-server/src/logger/logger.ts
@@ -1,50 +1,87 @@
-import {Context} from 'koa';
+/* eslint-disable no-process-env */
+import {Context, Request} from 'koa';
 import chalk from 'chalk';
 import {KoaNextFunction} from '../types';
 
 export const LOGGER = Symbol('logger');
-const PREFIX = chalk`{underline sidecar} ⁓ `;
-
-interface LoggerOptions {
-  level?: Verbosity;
-}
-
-export enum Verbosity {
-  Off,
-  Debug,
-}
+const PREFIX = chalk`{underline [React Server]}  `;
 
 export class Logger {
-  private logger: Pick<Console, 'log' | 'error'> =
-    this.options.level === Verbosity.Off ? noopLogger : console;
-
-  constructor(private options: LoggerOptions = {level: Verbosity.Off}) {}
+  private buffer = '';
+  private logger = console;
 
   log(message: string) {
-    this.logger.log(`${PREFIX}${message}`);
+    if (process.env.NODE_ENV === 'development') {
+      this.logger.log(`${PREFIX}${message}`);
+    } else {
+      this.buffer = `${this.buffer}\n${message}`;
+    }
+  }
+}
+
+function initialRequestMessage(request: Request): string {
+  const requestMethod = `${request.method.toUpperCase()} "${request.url}"`;
+  return `Started ${requestMethod} for at ${new Date().toISOString()}`;
+}
+
+function endRequestMessage(ctx: Context, requestDuration: number): string {
+  const httpStatus = `${ctx.status} ${ctx.message || ''}`;
+  const duration = `${requestDuration.toFixed(0)}ms`;
+
+  return `Completed ${httpStatus} at ${new Date().toISOString()} in ${duration}`;
+}
+
+function endRequest(ctx: Context, requestDuration: number) {
+  ctx.state.logger.log(endRequestMessage(ctx, requestDuration));
+
+  if (process.env.NODE_ENV === 'development') {
+    return;
   }
 
-  error(message: string) {
-    this.logger.error(`${PREFIX}${message}`);
-  }
-}
-
-export function getLogger(ctx: Context): Logger {
-  return ctx.state[LOGGER];
-}
-
-export function setLogger(ctx: Context, logger: Logger) {
-  ctx.state[LOGGER] = logger;
-}
-
-export function createLogger(options?: LoggerOptions) {
-  const logger = new Logger(options);
-
-  return async function loggerMiddleware(ctx: Context, next: KoaNextFunction) {
-    setLogger(ctx, logger);
-
-    await next();
+  /* eslint-disable babel/camelcase */
+  const logObject: any = {
+    datetime: new Date().toISOString(),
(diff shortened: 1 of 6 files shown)


packages/react-server/src/logger/logger.ts
```diff
@@ -1,50 +1,87 @@
-import {Context} from 'koa';
+/* eslint-disable no-process-env */
+import {Context, Request} from 'koa';
 import chalk from 'chalk';
 import {KoaNextFunction} from '../types';
 
 export const LOGGER = Symbol('logger');
-const PREFIX = chalk`{underline sidecar} ⁓ `;
+const PREFIX = chalk`{underline [React Server]}  `;
```
ismail-syed: IMO, `[React server]` is better suited in logs for this than `sidecar ~`. I can change it back if we don't like it

packages/react-server/src/logger/logger.ts
```diff
@@ -1,50 +1,87 @@
...
 
-interface LoggerOptions {
-  level?: Verbosity;
```
ismail-syed: The `level` option dictates the `verbosity` of the logger. The `verbosity` is kind of a lie, it's more like `noopLogger` or `logger`. Apps should get logging for free both in production and dev environments. When it is actually verbose, we can re-evaluate the need for this.

packages/react-server/src/logger/logger.ts
```diff
@@ -1,50 +1,87 @@
...
-export enum Verbosity {
-  Off,
-  Debug,
+function initialRequestMessage(request: Request): string {
+  const requestMethod = `${request.method.toUpperCase()} "${request.url}"`;
+  return `Started ${requestMethod} for at ${new Date().toISOString()}`;
 }
 
-export class Logger {
-  private logger: Pick<Console, 'log' | 'error'> =
```
ismail-syed: I'm dropping `console.error`. `.error` prints to `stderr` but in production we `console.log` a lob object with a bunch of metadata. If we have errors we generally look at 500s logs anyways.

packages/react-server/src/render/render.tsx
```diff
@@ -102,12 +101,14 @@ export function createRender(render: RenderFunction, options: Options = {}) {
       ctx.set(Header.ContentType, 'text/html');
       ctx.body = response;
     } catch (error) {
-      logger.error(error);
+      const errorMessage = `React SSR Error:\n${error.stack || error.message}`;
+
+      logger.log(errorMessage);
+      ctx.status = StatusCode.InternalServerError;
```
ismail-syed: Koa's being silly. If I don't set this status to 500, then `ctx.throw` actually wouldn't throw a 500, I was getting 404s

packages/react-server/src/logger/logger.ts
```diff
@@ -1,50 +1,87 @@
...
+    ctx.state.logger.log('Error during server execution, see details below.');
+    ctx.state.logger.log(
+      `${error.stack || error.message || 'No stack trace was present'}`,
+    );
+  } finally {
+    endRequest(ctx, requestDuration(requestStartTime));
+  }
 }
 
 export const noopLogger = {
```
cartogram: noticed you removed this below, wondering it is still needed?

packages/react-server/src/logger/logger.ts
```diff
@@ -1,50 +1,87 @@
...
+    }
+  }
 }
 
-export enum Verbosity {
-  Off,
-  Debug,
+function initialRequestMessage(request: Request): string {
+  const requestMethod = `${request.method.toUpperCase()} "${request.url}"`;
+  return `Started ${requestMethod} for at ${new Date().toISOString()}`;
```
cartogram:
```suggestion
  return `Started ${requestMethod} at ${new Date().toISOString()}`;
```

packages/react-server/src/render/render.tsx
```diff
@@ -39,7 +38,7 @@ type Options = Pick<
  */
 export function createRender(render: RenderFunction, options: Options = {}) {
   return async function renderFunction(ctx: Context) {
-    const logger = getLogger(ctx) || console;
+    const logger = ctx.state.logger || console;
```
cartogram: Why not continue to use `getLogger` rather than storing directly on state?

packages/react-server/src/logger/logger.ts
GoodForOneFare:
```suggestion
const PREFIX = chalk.underline('[React Server]');
```

packages/react-server/src/render/render.tsx
```diff
@@ -102,12 +101,14 @@ export function createRender(render: RenderFunction, options: Options = {}) {
       ctx.set(Header.ContentType, 'text/html');
       ctx.body = response;
     } catch (error) {
-      logger.error(error);
+      const errorMessage = `React SSR Error:\n${error.stack || error.message}`;
```
GoodForOneFare:
```suggestion
      const errorMessage = `React server-side rendering error:\n${error.stack || error.message}`;
```

packages/react-server/src/logger/logger.ts
ismail-syed: good catch, not needed anymore either

packages/react-server/src/render/render.tsx
ismail-syed: Ou, true. Also `LOGGER = Symbol('logger');` so better to keep the `getLogger`

Issue #964, opened 2019-09-05
Fix doc generation


michelleyschen:
- run `dev generate docs` right after completing `dev generate package` so the readme will be modify as soon as the new package is added

- `dev generate docs` should not delete the rails doc or the deprecated section

PR #979, opened 2019-09-10
🐛 add gem read me to template so it wont get overwrite

michelleyschen:
## Description

Fixes (issue #)

Close https://github.com/Shopify/quilt/issues/964


ismail-syed:
My understanding was that this PR backfills the missing ones and #977 tries to make sure it doesn't happen again.

cartogram:
@ismail-syed the root Readme is written from scratch every time, so any changes to it in a PR like this will get overwritten on the next `generate`.

michelleyschen:
Its two different approaches to the same thing.

I opt for a manual method of adding it to the README template because I dont think we will have many gem packages in the future and they can be static.

Matt opt for dynamically generating it so it will always be up to date with the shape of the repo.

Since Matt already merged his PR, I will be closing this one.

Issue #963, opened 2019-09-05
For all the packages thats been deprecated in quilt: Delete all code other than readme


michelleyschen:
This is so that lerna will stop picking it up to publish.

ie. `react-self-serializers`

- go through npm registry on what packages are deprecated
- Delete all code other than readme in quilt
- move those a deprecated section in the main readme

PR #980, opened 2019-09-10
Deprecated packages

michelleyschen:
## Description

Close https://github.com/Shopify/quilt/issues/963


ismail-syed:
Should we keep the changelogs there as well and leave a note in the changelog with the date of the deprecation?

michelleyschen:
> Should we keep the changelogs there as well and leave a note in the changelog with the date of the deprecation?

I like the idea of leaving changelog :D

packages/react-self-serializers/CHANGELOG.md
```diff
@@ -0,0 +1,51 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
```
ismail-syed: Replace this line with something like. Same with the other package


## ❗️ Deprecated on 2019-09-10 on version 1.3.15 ❗️

PR #985, opened 2019-09-11
Add DefaultProviders to webpack generated App component

cartogram:
## Description

Fixes (issue #)

This PR adds a mechanism for wrapping the App component in a set of providers using using a new function `createCombinedProviders` and a new `DefaultProviders` component. 

`createCombinedProviders` will accept a config and return a subset of providers based on that config (currently this only provides the `csrf` provider). When called with no arguments, will return all of the Providers available (the same as using `DefaultProviders`).

I also wrap the app generated by webpack with this component.

## Type of change

- [x] `react-server` Minor: New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/packages/react-server/CHANGELOG.md
+++ b/packages/react-server/CHANGELOG.md
@@ -16,6 +16,7 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 ### Added
 
 - Improved logger to provide more readable production logs in Splunk [#978](https://github.com/Shopify/quilt/pull/978)
+- App is now wrapped with a set of default providers
 
 ## [0.4.0] - 2019-09-06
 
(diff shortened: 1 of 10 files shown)


packages/react-server/CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## [Unreleased] -->
 
+## [0.5.0] - 2019-09-11
+
+### Added
+
+- App is now wrapped with a set of default providers
```
cartogram: this not true, will update.

packages/react-server/src/providers/providers.tsx
```diff
@@ -0,0 +1,32 @@
...
+        condition={csrf}
+        props={{value: csrfToken}}
+      >
+        {children}
+      </ConditionalProvider>
+    );
+  };
+}
+
+export const DefaultProvider = createCombinedProvider();
```
TheMallen: can we document this in the README ?

packages/react-server/CHANGELOG.md
TheMallen: Can we also document this in the readme?

PR #990, opened 2019-09-11
`@shopify/react-server` Set X-React-Server-Request-Time header on response 

ismail-syed:
## Description

Related to https://github.com/Shopify/web-rails-proving-ground/issues/14

This PR adds a `metrics` koa middleware to `@shopify/react-server`. This middleware sets a `X-React-Server-Request-Time` response header. This metric captures the duration of a request through the react server in ms. 

Next steps: 
- [ ] Gather metrics React rendering time
- [ ] Gather metrics GraphQL request  times
- [ ] Log these metrics in `quilt_rail` & send them up to statsd.

## Type of change

## Checklist


packages/react-server/src/metrics/test/metric.test.tsx
```diff
@@ -0,0 +1,46 @@
...
+      ip,
+      render: () => <div>Mock app</div>,
+    });
+  });
+
+  afterEach(() => {
+    server.close();
+  });
+
+  it('responds with a X-React-Server-Request-Time header', async () => {
```
ismail-syed: Not sure how useful this is, open to 🔥 if others think it's useless. 

I was going to add another test where the server would contain a `~50ms` delay in the request and the test would assert that the header's value is a reasonable value given the delay. That got a bit gross and didn't seem super useful either.

packages/react-server/src/metrics/metrics.ts
```diff
@@ -0,0 +1,27 @@
...
+}
+
+async function middleware(ctx: Context, next: KoaNextFunction) {
+  try {
+    await next();
+  } finally {
+    const [seconds, nanoseconds] = process.hrtime(
+      ctx.state[START_TIME_STATE_KEY],
+    );
+    const ms = seconds * 1000 + nanoseconds / 1e6;
```
TheMallen: nit: I'd prefer if these magic numbers had descriptive names (eg. `MILLIS_PER_SECOND`)

packages/react-server/src/metrics/metrics.ts
```diff
@@ -0,0 +1,27 @@
...
+    const requestTime = Math.round(ms);
+
+    ctx.set('X-React-Server-Request-Time', requestTime.toString());
```
TheMallen: nit: could we have this as part of an enum or something rather than an inline string? (Then we can also import it in the test and export it from the package)

packages/react-server/src/metrics/test/metric.test.tsx
TheMallen: it's pretty medium, but I think overall I'd keep the test /shrug

packages/react-server/src/metrics/metrics.ts
```diff
@@ -0,0 +1,27 @@
+import {Context} from 'koa';
+import compose from 'koa-compose';
+
+import {KoaNextFunction} from '../types';
+
+const START_TIME_STATE_KEY = Symbol('startTime');
+
+async function startStatsd(ctx: Context, next: KoaNextFunction) {
```
cartogram:
```suggestion
async function startStatsD(ctx: Context, next: KoaNextFunction) {
```

packages/react-server/src/metrics/metrics.ts
```diff
@@ -0,0 +1,27 @@
...
+      ctx.state[START_TIME_STATE_KEY],
+    );
+    const ms = seconds * 1000 + nanoseconds / 1e6;
+    const requestTime = Math.round(ms);
+
+    ctx.set('X-React-Server-Request-Time', requestTime.toString());
+  }
+}
+
+export const metricsMiddleware = compose([startStatsd, middleware]);
```
cartogram:
```suggestion
export const metricsMiddleware = compose([startStatsD, middleware]);
```

packages/react-server/src/metrics/metrics.ts
GoodForOneFare: I'd prefer using [`Server-Timing`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Server-Timing) as a source of truth for all Node=>Rails metrics because it:
- Avoids proliferation of timing header tags every time we think of a new metric
- Allows quilt_rails to find timing data without fishing through arbitrary headers
- Maybe helps us consider what a more generic metrics API looks like (how do we extend it for react-server?  How do consuming apps add their own metrics?)
- Embraces community standards, helping us interface with other libs / tools

GoodForOneFare: RE a generic metrics API.  We should be thinking of how to surface info like tracking [cancelled render loops](https://github.com/Shopify/quilt/blob/32f04a86d8ae0f585e96dda5160ffb152073587e/packages/react-effect/src/server.tsx#L41).

GoodForOneFare: RE a generic metrics API: it's weird that this does [a chunk of work that's identical to `requestLogger`](https://github.com/Shopify/quilt/blob/master/packages/react-server/src/logger/logger.ts#L37).  What if *another* middleware handled request timing, and Logger / Metrics middleware fished it out of the context?

PR #1000, opened 2019-09-13
Replace @shopify/javascript-utilities/dates with @shopify/dates

rexmac:
### Changed

- Replaced `@shopify/javascript-utilities/dates` functions with those from `@shopify/dates`.
- Removed `@shopify/javascript-utilities/dates` dependency.

### Fixed

- Fixed translation of weekday in `humanizeDate` for dates less than one week old.

## Type of change

- [x] @shopify/react-i18n Patch: Bug/ Documentation fix (non-breaking change which fixes an issue or adds documentation)

## Checklist

### DIFF
--- a/packages/react-i18n/src/i18n.ts
+++ b/packages/react-i18n/src/i18n.ts
@@ -1,5 +1,4 @@
 import {
-  getDateDiff,
   isLessThanOneHourAgo,
   isLessThanOneMinuteAgo,
   isLessThanOneWeekAgo,
@@ -7,9 +6,7 @@ import {
   isToday,
   isYesterday,
   TimeUnit,
-  Weekdays,
-} from '@shopify/javascript-utilities/dates';
-import {applyTimeZoneOffset} from '@shopify/dates';
+} from '@shopify/dates';
 import {memoize as memoizeFn} from '@shopify/function-enhancers';
 import {memoize} from '@shopify/decorators';
 import {languageFromLocale, regionFromLocale} from '@shopify/i18n';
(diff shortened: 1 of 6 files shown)


packages/react-i18n/src/test/i18n.test.ts
```diff
@@ -1114,7 +1114,7 @@ describe('I18n', () => {
           'humanize.weekday',
           {
             pseudotranslate: false,
-            replacements: {day: 'Friday', time: '11:00 a.m.'},
+            replacements: {day: 'Saturday', time: '11:00 a.m.'},
```
michenly: I dont quite understand what this is fixing, do you have an original issue that I can take a look?

rexmac: Reverting a prior fix: https://github.com/Shopify/quilt/pull/916#pullrequestreview-285701811, which changed it to "Friday" as for some reason that was needed to get the test to pass in that PR, but prior to that it was "Saturday" (and _it should be_ "Saturday") as that is the correct weekday for 2012-12-15T00:00:00-00:00 GMT in Sydney, Australia.

packages/react-i18n/src/i18n.ts
```diff
@@ -453,15 +449,12 @@ export class I18n {
...
-      const localTimeZone = Intl.DateTimeFormat().resolvedOptions().timeZone;
-
-      const dateInTimeZone =
-        timeZone === undefined
-          ? date
-          : applyTimeZoneOffset(date, timeZone, localTimeZone);
-
+      const weekday = this.formatDate(date, {
+        ...options,
+        weekday: 'long',
```
kyledurand: Nice, didn't know this was a thing 👍

PR #1002, opened 2019-09-16
🍪`react-cookie` package

cartogram:
## Description

~~This PR adds a `@shopify/react-cookie` library that gathers cookies from both the server and client. There are three main components to this library, 1) A universal-provider that _provides_ a set of initial cookies and instantiates 2) a manager. This manager provides common methods for getting/ setting cookies both internally, and in the `document`. 3) Finally, there are 2 common React hooks for operating on the cookies from within a React component.~~

#### Demo/ Tophat
You can view a [demo here](https://monosnap.com/file/YJuB4TUw6etygwhCj5A4DA38r1Dlmt) from [this branch](https://github.com/Shopify/web-rails-proving-ground/compare/cookies?expand=1) of @shopify/web-rails-proving-ground

If you want to tophat yourself, pull down this branch and also the [`cookies`]() branch from `web-rails-proving-ground`.

1. `dev up` in the `cookies` branch of `web-rails-proving-ground`
1. From the quilt repo run:
```
cp -R packages/react-cookie/src/ ../web-rails-proving-ground/packages/\@shopify/react-cookie
cp -R packages/react-network/src/ ../web-rails-proving-ground/packages/\@shopify/react-network
cp -R packages/react-server/src/ ../web-rails-proving-ground/packages/\@shopify/react-server
```
1. Then run `dev run` in the `web-rails-proving-ground` repo.


packages/react-cookie/src/hooks.ts
```diff
@@ -0,0 +1,56 @@
...
+export function useCookie(
+  key: string,
+): [
+  string | undefined,
+  (value?: string, options?: CookieSerializeOptions) => void
+] {
+  const manager = useContext(CookieContext);
+
+  if (!manager) {
+    throw new Error(NO_MANAGER_ERROR);
```
cartogram: Does it make sense to `throw` here?

packages/react-cookie/README.md
```diff
@@ -0,0 +1,85 @@
+# `@shopify/react-cookie`
```
cartogram: Just roughed this out for completeness, but will put some more ❤️  into this `Readme` either before or immediately after merging.

packages/react-cookie/src/tests/hooks.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+      const key = 'foo';
+      const value = 'bar';
+      const cookies = {[key]: value};
+
+      const wrapper = mount(<MockComponent cookie={key} />, {
+        manager: new CookieManager({...cookies}),
+      });
+
+      wrapper
+        .find(MockComponent)!
```
cartogram: This `find` returns `null` occasionally and is why CI is failing. I am not sure why yet.

packages/react-cookie/package.json
```diff
@@ -0,0 +1,41 @@
...
+  },
+  "bugs": {
+    "url": "https://github.com/shopify/quilt/issues"
+  },
+  "homepage": "https://github.com/Shopify/quilt/blob/master/packages/react-cookie/README.md",
+  "dependencies": {
+    "@shopify/react-effect": "^3.2.3",
+    "@shopify/react-hooks": "^1.2.1",
+    "@shopify/react-html": "^9.0.5",
+    "cookie": "^0.4.0",
```
cartogram: TODO: `@shopify/react-network` is missing here.

packages/react-cookie/src/hooks.ts
TheMallen: I think so?

packages/react-cookie/src/hooks.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  const [cookie, setCookieValue] = useState(
+    () => (manager ? manager.getCookie(key) : undefined),
+  );
+
+  const setCookie = (value?: string, options?: CookieSerializeOptions) => {
+    if (!manager) {
```
TheMallen: Manager should always be defined given that we are throwing before we construct this when it isn't.

This is actually a great usecase for the upcoming [assert signatures](https://httptoolkit.tech/blog/5-big-features-of-typescript-3.7/) feature, but till we have that I'd just `manager!` or cast it.

TheMallen: on a side note it feels kinda weird that this lives here and not in the manager itself.

packages/react-cookie/src/manager.ts
```diff
@@ -0,0 +1,65 @@
+import cookie from 'cookie';
+import {hasDocumentCookie} from './utilities';
+import {Cookies, CookieValue} from './types';
+
+export {CookieContext} from './context';
+
+export class CookieManager {
```
TheMallen: tbh I'm not super sure why we aren't using `universal-cookies` as our manager, does it clash with our expected APIs in some way? It's simple enough that it's not a huge deal for us to have our own but if there's no real reason not to we might as well use it.

cartogram:
@TheMallen on doing our own thing vs third-party: this library allows us to provide a cookie solution for applications devs that uses consistent patters as with the other quilt packages that also require similar client/server interaction (self-serializer, react-effect, network, etc…) . Going this route also requires no server-side middleware and everything is confined to only this package (and really one component) rather than requiring orchestrated changes across `@shopify/react-server`, `@shopify/sewing-kit`, etc. 

Less important, there is only one viable third-party solution out there (the `universal-cookie`/ `react-cookie` stack) and I wasn’t the biggest fan of the imperative change-listener way that those libraries operate. I wanted to explore a way that we could leverage our own libraries to provide something better.

That said, I get that additional code is a liability (even given this library is pretty tiny) and understand the hesitation to introduce this package.

packages/react-cookie/src/Cookie.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
```
GoodForOneFare: This follows the universal provider pattern so closely, would it make sense to call this `CookiesUniversalProvider`?

packages/react-cookie/src/Cookie.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+    return new CookieManager(initialCookies || cookies);
+  }).current;
+
+  return (
+    <>
+      <CookieContext.Provider value={manager}>
+        {children}
+      </CookieContext.Provider>
+      <Serialize data={() => manager.getCookies()} />
```
GoodForOneFare: I don't understand this 😕 Why wouldn't cookies be sent as response headers?

cartogram: The `useRequestHeader ` hook returns `undefined` on the client so without this, the server cookies are lost.

GoodForOneFare: I haven't worked much with cookies, but as I understand it, this should be possible:
- The server side of this generates `Set-Cookie` headers for each name/value pair
- The browser sees these headers, and populates `document.cookie` to reflect them
- The client side of this grabs `document.cookie`, and parses it into name/value pairs

Not an expert on this, so I could definitely be wrong.

TheMallen: The above seems right

packages/react-cookie/src/CookieUniversalProvider.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+    return new CookieManager(initialCookies || cookies);
+  }).current;
+
+  return (
+    <>
+      <CookieContext.Provider value={manager}>
+        {children}
+      </CookieContext.Provider>
+      <Serialize data={() => manager.getCookies()} />
```
TheMallen: as Gord mentioned, we shouldn't need to actually serialize the cookies since we can build the manager out of `document.cookie` on the client anyway.

packages/react-cookie/src/CookieUniversalProvider.tsx
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export function CookieUniversalProvider({children}: Props) {
+  const [initialCookies, Serialize] = useSerialized<Cookies>('cookies');
+  const serverCookieHeader = useRequestHeader('cookie');
+
+  const manager = useLazyRef(() => {
+    const cookies = hasDocumentCookie() ? document.cookie : serverCookieHeader;
+
+    return new CookieManager(initialCookies || cookies);
```
TheMallen: `new CookieManager(cookies)` should be fine here

packages/react-cookie/src/manager.ts
```diff
@@ -0,0 +1,65 @@
...
+    for (const [key, value] of this.cookies) {
+      cookies[key] = value;
+    }
+
+    return {
+      ...cookies,
+    };
+  }
+
+  setCookie(name: string, value: string | CookieValue) {
```
TheMallen: I think realistically on the server we want to set the cookie using a `networkManager` form `@shopify/react-network`,  which would probable entail

- depending on it here and passing it in from the component, or rolling this back into that library
- adding a `setCookie` property on the network manager
- adding logic to `applyToContext` that calls `ctx.cookies.set` (which will handle setting `set-cookie` as a user would expect)

packages/react-cookie/src/UniversalCookies.ts
```diff
@@ -0,0 +1,54 @@
+import cookie from 'cookie';
+import {Context} from 'koa'
+import {Cookie, CookieValue} from './types';
+
+export {CookieContext} from './context';
+
+export class UniversalCookies {
```
cartogram: open to better names here

packages/react-cookie/src/UniversalCookies.ts
```diff
@@ -0,0 +1,54 @@
...
+    const fullCookie = typeof value === 'string' ? {value} : value;
+
+    this.cookies.set(name, fullCookie);
+  }
+
+  removeCookie(name: string) {
+    this.cookies.delete(name);
+  }
+
+  applyToContext(ctx: Context) {
```
cartogram: embedding the server logic into the class

packages/react-cookie/src/context.ts
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+import {CookieManager} from './manager';
+import {UniversalCookies} from './UniversalCookies';
+
+export const CookieContext = React.createContext<
+  CookieManager | UniversalCookies | null | any
```
cartogram: todo: remove any

packages/react-cookie/README.md
```diff
@@ -0,0 +1,85 @@
...
+To use the hooks provided by this library you must first wrap your application tree in the `<CookieUniversalProvider />` component.
+
+```tsx
+// App.tsx
+
+import {CookieUniversalProvider} from '@shopify/react-cookie';
+import {SomeComponent} from './someComponent';
+
+function App() {
+  <Cookie>
```
TheMallen: `CookieUniversalProvider`

packages/react-cookie/src/CookieUniversalProvider.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export interface Props {
+  children?: React.ReactNode;
+}
+
+export function CookieUniversalProvider({children}: Props) {
+  const manager = useNetworkManager();
+
+  const cookieManager = useLazyRef(() => {
+    return !hasDocumentCookie() && manager
```
TheMallen: this seems slightly wrong to me since it could make a front-end focused cookie manager if the user forgot to wrap their tree in a network provider or it isn't there for some other weird reason. I'd rather we do something like

```tsx
if (hasDocumentCookie()) {
  return new CookieManager();
}

if (manager == null) {
  throw new error('You fool! You fell victim to one of the classic blunders! Never server render a react-html app without a network provider when death is on the line!');
}

return manager.cookies;
```

packages/react-cookie/src/context.ts
TheMallen: IMO we just want to have an `interface CookieManager {}` that lays out the API that both of these managers use.

packages/react-cookie/src/UniversalCookies.ts
TheMallen: This one only works on the server right? ServerCookieManager?

packages/react-cookie/src/UniversalCookies.ts
TheMallen: 🤔 seems weird to me, I'd expect it to be like `react-network` and expose a function separately (that way the manager isn't tied to koa, just the `applyToContext` function is)

```
import {ServerCookieManager} '@shopify/react-cookie'
import {applyToContext} from '@shopify/react-cookie/koa'

// in the middleware
applyToContext(ctx, cookies);
```

packages/react-cookie/src/tests/cookie.ts
```diff
@@ -0,0 +1,15 @@
+export default class Cookie {
+  mock(value: string) {
```
TheMallen: should this go live in `jest-dom-mocks`?

packages/react-network/src/manager.ts
```diff
@@ -1,12 +1,14 @@
 import {StatusCode, CspDirective, Header} from '@shopify/network';
 import {EffectKind} from '@shopify/react-effect';
+import {Cookie, UniversalCookies} from '@shopify/react-cookie';
```
TheMallen: isn't this a circular dependency? Should the server cookie manager live in `@shopify/react-network`?

packages/react-cookie/src/context.ts
cartogram: Yeah that works

packages/react-cookie/src/tests/cookie.ts
cartogram: yeah it should 👍

packages/react-network/src/manager.ts
cartogram: Yeah also agree with this

packages/react-cookie/src/tests/cookie.ts
cartogram: I'm going to do this in a follow up though. There are already a lot of packages involved in this PR and this is an easy thing to extract afterward.

cartogram:
@TheMallen another look when you get a chance 🙇 Still doing a final comb through all the Readme, but otherwise all changes are in.

packages/react-cookie/README.md
```diff
@@ -0,0 +1,132 @@
...
+
+function App() {
+  <CookieUniversalProvider>
+    // rest of your tree
+    <SomeComponent />
+  </CookieUniversalProvider>;
+}
+```
+
+You can then use the [hooks](#hooks) provided by this package. See [hooks](#hooks) below for more information.
```
TheMallen: I don't think you need the second sentence here since you already have a link.

packages/react-cookie/README.md
```diff
@@ -0,0 +1,132 @@
...
+
+```bash
+$ yarn add @shopify/react-cookie
+```
+
+## Usage
+
+### Server
+
+To extract cookies during the server-side render of your application, your component need to have access to the `NetworkManager` from `@shopify/react-network`. You can pass the initial server cookie value when the manager is instantiated within your server side middleware.
```
TheMallen: needs*

packages/react-cookie/src/server.ts
```diff
@@ -0,0 +1,12 @@
+import {Context} from 'koa';
```
TheMallen: I still think the way these two packages are broken up is strange. This function in particular is strange when the rest of the server cookie stuff is just part of react-network. Why can't this just be part of `react-network`'s `applyToContext`?

cartogram: It can be 👍 Slightly related, I am considering pulling the shared types and server manager into a more generic `@shopify/cookie` package that `react-network` and `react-cookie` would depend on.

packages/react-cookie/src/server.ts
```diff
@@ -0,0 +1,12 @@
...
+import {Cookie} from './types';
+
+export function applyCookieToContext(cookies: Cookie) {
+  return function(ctx: Context) {
+    Object.entries(cookies).forEach(([cookie, options]) => {
+      const {value, ...cookieOptions} = options;
+
+      ctx.cookies.set(cookie, value, cookieOptions as any);
```
GoodForOneFare: Could you add `none` in a PR to https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/cookies/index.d.ts#L91, please?

packages/react-network/src/ServerCookieManager.ts
```diff
@@ -0,0 +1,43 @@
+import cookie, {CookieSerializeOptions} from 'cookie';
```
GoodForOneFare: Should this class have a unit test?

packages/react-cookie/src/BrowserCookieManager.ts
```diff
@@ -0,0 +1,65 @@
+import cookie from 'cookie';
+import {hasDocumentCookie} from './utilities';
+import {CookieValue, Cookie} from './types';
+
+export {CookieContext} from './context';
+
+export class BrowserCookieManager {
```
GoodForOneFare: I guess this is consistent with the rest of Quilt, but I miss the lint rule for "this isn't really a class".

.vscode/settings.json
```diff
@@ -10,7 +10,6 @@
     "**/.git": true,
     "**/.svn": true,
     "**/.hg": true,
-    "**/node_modules": true,
```
GoodForOneFare: Please revert this change before merging.

packages/react-cookie/server.d.ts
```diff
@@ -0,0 +1 @@
+export * from './dist/server';
```
GoodForOneFare: I don't think you can export useful types from a js file.

packages/react-cookie/README.md
```diff
@@ -0,0 +1,132 @@
...
+      <p>the cookie is {cookie}</p>
+      <TextField value={cookie} onChange={handleChange} />
+    </>
+  );
+}
+```
+
+### Hooks
+
+#### `useCookies()`
```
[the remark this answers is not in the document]
cartogram: It was helpful for testing, but I can't imagine purpose for consumers so I am removing it for now. 👍

packages/react-cookie/src/BrowserCookieManager.ts
```diff
@@ -0,0 +1,65 @@
...
+  getCookie(name: string) {
+    if (!hasDocumentCookie()) {
```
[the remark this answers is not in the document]
cartogram: Good catch, was left in from a "universal" implementation of this manager.

packages/react-cookie/src/hooks.ts
```diff
@@ -0,0 +1,52 @@
...
+      manager.removeCookie(key);
+
+      return;
+    }
+
+    setCookieValue(value);
+    manager.setCookie(key, {value, ...options});
+  };
+
+  return [cookie, setCookie];
```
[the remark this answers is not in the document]
cartogram: This would break our lint rule https://github.com/Shopify/eslint-plugin-shopify/blob/master/docs/rules/react-hooks-strict-return.md. 

I opted for "setting" to null/undefined to remove the cookie, but I did not add documentation or tests for this feature. Those are both added now and I think provides a simple API for consumers. Do you think an explicit `removeCookie()` function is needed?

cartogram:
@TheMallen @GoodForOneFare All changes have been addressed. I also included:

- A `react-server` test that sending cookies with the request will be accessible via the client-side hooks.
- Added the UniversalCookieProvider to our default providers
- Tests for ServerCookieProvider 
- Tests for the react-network/server component
- More docs for all the changed things
- Utils for creating and clearing cookies in tests
- probably much more that I am forgetting

The only way I was able to tophat this successfully was by building each library (`react-server`, `react-cookie` and `react-network`) and copying them into web-rails-proving-ground`'s `node_modules. There is a branch there called `cookies` that has some application code for testing the hooks in this package.

Thanks again for your continual feedback and patience!

packages/react-cookie/src/BrowserCookieManager.ts
TheMallen: Yeah this could easily just be a constant object exported directly since it doesn't have any local state. Doesn't matter too much.

packages/react-cookie/src/hooks.ts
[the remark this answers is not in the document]
TheMallen: Personally I like the consistency with koa of doing `set('name', null)` as long as it's documented

packages/react-cookie/src/tests/CookieUniversalProvider.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+  afterEach(() => {
+    hasDocumentCookie.mockClear();
+  });
+
+  function MockApp() {
+    return null;
+  }
+
+  it('provides a universal cookie manager for the server', () => {
+    hasDocumentCookie.mockImplementation(false);
```
TheMallen: we might want to add a 'why' comment explaining why we have to mock this (so that we can test the server implementation despite jest always having a document.cookie defined)

packages/react-cookie/src/tests/hooks.test.tsx
```diff
@@ -0,0 +1,115 @@
...
+    });
+
+    it('removes a cookie', async () => {
+      const setValue = 'bar';
+
+      const wrapper = await mount(
+        <MockComponent cookie="foo" setValue={setValue} />,
+      );
+
+      wrapper.find('button')!.trigger('onClick');
```
TheMallen: this spacing is a bit weird IMO

packages/react-cookie/README.md
```diff
@@ -0,0 +1,110 @@
...
+    <>
+      <p>the cookie is {cookie}</p>
+      <TextField value={cookie} onChange={handleChange} />
+      <button type="button" onClick={removeCookie}>
+        Remove Cookie
+      </button>
+    </>
+  );
+}
+```
```
TheMallen: Can we also document the test utilities we export?

packages/react-cookie/src/hooks.ts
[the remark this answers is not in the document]
GoodForOneFare: If it makes sense to follow our rule, I guess it's fine.  I'm not 100% sure about how strict we about following it.

Is there any kind of type-hint we can provide like:
```
export const UNSET = '';
...
setCookie(name: string, value: string | CookieValue | UNSET)
```

Eww :/

Issue #971, opened 2019-09-06
[react-i18n] Add support for ordinals


rexmac:
## Overview

Add support for ordinals to `react-i18n`'s `translate` function.

[Inlt.PluralRules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/PluralRules#Using_options) supports a `{type: 'ordinal'}` option as a second parameter:

```js
var pr = new Intl.PluralRules('en-US', { type: 'ordinal' });

pr.select(0);
// → 'other'
pr.select(1);
// → 'one'
pr.select(2);
// → 'two'
pr.select(3);
// → 'few'
pr.select(4);
// → 'other'
pr.select(42);
// → 'two'
```

## Type

- [ ] New feature
- [x] Changes to existing features

## Motivation

> What inspired this feature request? What problems were you facing,
> or what else makes you think this should be in `quilt`?

Web needs the ability to translate ordinal strings for display in the UI. 

## Proposed usage

```js
// Assuming a dictionary like:
{
  "MyComponent": {
    "searchResult": {
      "one": "{count}st widget found",
      "two": "{count}nd widget found",
      "few": "{count}rd widget found",
      "other": "{count}th widget found"
    }
  }
}

i18n.translate('MyComponent.searchResult', {count: searchResults, ordinal: true});
```

Or perhaps `ordinal` is used rather than `count`?

```js
i18n.translate('MyComponent.searchResult', {ordinal: searchResults});
```

Or perhaps `ordinal` is an option?

```js
i18n.translate('MyComponent.searchResult', {ordinal: true}, {count: searchResults});
```

---

## Checklist

- [x] Please delete the labels section before submitting your issue
- [x] I have described this issue in a way that is actionable (if possible)

PR #1003, opened 2019-09-16
Add `i18n.ordinal()` method to `@shopify/react-i18n`

rexmac:
## Description

Fixes https://github.com/Shopify/quilt/issues/971.

Adds `I18n#ordinal(number)` method for translating integers into ordinals, e.g., 1st, 2nd, 3rd, etc.

Uses the [`{type: 'ordinal'}` option of Intl.PluralRules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/PluralRules#Using_options).

Given that ordinals are not context-specific (confirmed with #intl-services), they only need to be translated once and therefore translations are intended to be stored at the root level, e.g., `app/locales/en.json` in web, with keys of "one", "two", "few", and "other" under an "ordinal" key:

```json
{
  "ordinal": {
    "one": "{number}st",
    "two": "{number}nd",
    "few": "{number}rd",
    "other": "{number}th",
  }
}
```

### Example usage

Given the above "ordinal" translation entry and the following:

```json
{
  "myKey": "This was their {ordinal} visit."
}
```

The `ordinal` method can be combined with `translate` to include ordinal numbers in translated strings:

```js
const ordinal = i18n.ordinal(42);
// ordinal => "42nd"
const text = i18n.translate('myKey', {ordinal});
// text => "This was their 42nd visit."
```

## Type of change

## Checklist

### DIFF
--- a/packages/react-i18n/src/i18n.ts
+++ b/packages/react-i18n/src/i18n.ts
@@ -316,6 +316,14 @@ export class I18n {
     }).format(date);
   }
 
+  ordinal(amount: number) {
+    const {locale} = this;
+    const group = new Intl.PluralRules(locale, {type: 'ordinal'}).select(
+      amount,
+    );
+    return this.translate(group, {scope: 'ordinal'}, {amount});
+  }
+
   weekStartDay(argCountry?: I18n['defaultCountry']): Weekday {
     const country = argCountry || this.defaultCountry;
 
--- a/packages/react-i18n/src/i18n.ts
+++ b/packages/react-i18n/src/i18n.ts
@@ -78,6 +78,7 @@ export class I18n {
   readonly defaultTimezone?: string;
   readonly onError: NonNullable<I18nDetails['onError']>;
   readonly loading: boolean;
+  readonly ordinalPluralRules: Intl.PluralRules;
 
   get language() {
     return languageFromLocale(this.locale);
@@ -127,6 +128,7 @@ export class I18n {
     this.pseudolocalize = pseudolocalize;
     this.onError = onError || defaultOnError;
     this.loading = loading || false;
+    this.ordinalPluralRules = new Intl.PluralRules(locale, {type: 'ordinal'});
   }
 
   translate(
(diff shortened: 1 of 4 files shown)


packages/react-i18n/src/test/i18n.test.ts
```diff
@@ -229,6 +229,61 @@ describe('I18n', () => {
     });
   });
 
+  describe('#ordinal()', () => {
+    it('calls translate() utility wth ordinal scope and replacement amount', () => {
```
michaelmelanson:
```suggestion
    it('calls translate() utility with ordinal scope and replacement amount', () => {
```

packages/react-i18n/src/test/i18n.test.ts
```diff
@@ -229,6 +229,61 @@ describe('I18n', () => {
...
+      i18n.ordinal(42);
+      expect(translate).toHaveBeenCalledWith(
+        'two',
+        {scope: 'ordinal', replacements: {amount: 42}, pseudotranslate: false},
+        defaultTranslations,
+        i18n.locale,
+      );
+    });
+  });
+
```
michaelmelanson: Should there also be a test that it returns the expected values in each case, not just the interaction with `translate()`? There's nothing here that tests that it actually uses the value it gets back from `translate()`.

rexmac: Can't. [`translate()` is mocked](https://github.com/Shopify/quilt/blob/c38f6f7d4a9856a6a7880423362d92fc53b254b2/packages/react-i18n/src/test/i18n.test.ts#L12) so it just returns `undefined` unless we force a return value using `translate. mockReturnValue(....)`, but I don't see any value in that for these tests.

packages/react-i18n/src/i18n.ts
```diff
@@ -316,6 +316,14 @@ export class I18n {
     }).format(date);
   }
 
+  ordinal(amount: number) {
+    const {locale} = this;
+    const group = new Intl.PluralRules(locale, {type: 'ordinal'}).select(
```
tleunen: I'd create the object in the constructor to avoid creating it on every call.

rexmac: Ooh, great point!

rexmac: Done! 🎉

Issue #957, opened 2019-09-05
First request of a session always fails on `useQuery` calls


GoodForOneFare:
## Overview
A session's very first request always fails because:
- There is no CSRF token defined for the session
- We create a CSRF token and pass it along to the Node server
- The Node server passes the CSRF token along to GraphQLController
- GraphQLController's CSRF protection tries to validate the passed token against a (non-existent) persisted session token
  - It fails

When the user's first request is rejected, Rails passes back a session + csrf token, and everything works because:
- The session has a _persisted_ csrf token
- Browser requests now contain a session id + csrf token
- ... this csrf token is passed through to Node->GraphQLController with no validation issues

## Consuming repo
beta-buddy

TheMallen:
Fixed via our new reccs around CSRF

PR #1004, opened 2019-09-16
✨ quilt_rails CSRF strategy to allow UI server requests

GoodForOneFare:
## Description

Fixes `First request of a session always fails on useQuery calls` (issue #957).

The flow that it's trying to fix is:
1. User visits the site for the first time
2. User goes through oauth flow (session is persisted with only a `session_id` value)
3. UiController receives the first authed request
  3.1. UiController fetches the session
  3.2. No CSRF token is defined, so a new one is associated with the session in memory but *not* persisted
  3.3. UiController passes the session id + CSRF token on to Node
4. Node starts rendering the UI
  4.1. It serializes the CSRF token
  4.2. It sends GraphQL requests to (Rails) GraphQLController
5. GraphQLController receives a request containing a CSRF token
  5.1. Rails' CSRF handling finds the requests CSRF token, but fails to find an in-session equivalent (because UiController's session has not yet persisted)
  5.2. Rails' CSRF handler fails

In this version of the library:

`4.2.` becomes:
> 4.2. It sends GraphQL requests to (Rails) GraphQLController _and_ an `X-Shopify-Server-Side-Rendered` header

`5.2` becomes:
> 5.2. Rails CSRF handler fails, but defers to `TrustedUiServerCsrfStrategy`
> 5.3. The custom strategy finds the `X-Shopify-Server-Side-Rendered` header, and allows the query to use the session

## Can't malicious XSS scripts exploit this by setting `X-Shopify-Server-Side-Rendered` on their GraphQL requests?

Nope!  The Rails server will reject any request containing this header if it originates from a different domain.

### Verify that session request now works

* Check out the `csrf-fixes` branch in beta-buddy
* Start your local server
* Visit the local site
* Clear cookies for the site
* Refresh the page
* Note that a JSON exception isn't thrown 🎉 

### Verify that cross-origin requests are rejected

* In Chrome dev tools, right-click a `/graphql` request and `Copy as fetch`
* Edit the fetch command's `headers` hash to include `"x-shopify-server-side-rendered":"1"`
* In another browser tab, open dev tools and paste in the fetch command
* Note that the fetch is rejected with `422 - unprocessable entity`

## Checklist


gems/quilt_rails/lib/quilt_rails/trusted_ui_server_csrf_strategy.rb
```diff
@@ -0,0 +1,25 @@
...
+      null_session.handle_unverified_request
+    end
+
+    private
+
+    def node_server_side_render?
+      @controller.request.headers['x-shopify-server-side-rendered'] == '1'
+    end
+
+    def null_session
```
JackMc: 🤔 do we want to handle non-header requests via the Exception or NullSession strategy? Seems like currently we're using Exception but naming it null_session.

gems/quilt_rails/README.md
```diff
@@ -434,3 +434,28 @@ const app = createServer({
...
 export default app;
 ```
+
+### Fixing rejected CSRF tokens for new user sessions
+
+If a React component calls back to a Rails endpoint (e.g., `/graphql`), Rails may throw a `Can't verify CSRF token authenticity` exception. This stems from the Rails CSRF tokens not persisting until after the first `UiController` call ends.
+
+To fix this:
+
+- Add an `X-Shopify-Server-Side-Rendered: 1` header to all GraphQL requests
```
JackMc:
```suggestion
- Add an `X-Shopify-Server-Side-Rendered: 1` header to all server-side GraphQL requests
```

gems/quilt_rails/lib/quilt_rails/trusted_ui_server_csrf_strategy.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module Quilt
+  class TrustedUiServerCsrfStrategy
```
JackMc: Could we instead inherit from `NullSession` here and just call `super` in `handle_unverified_request`?

gems/quilt_rails/lib/quilt_rails/trusted_ui_server_csrf_strategy.rb
GoodForOneFare: Whoops.  Definitely `Exception`.  I'll update 👍

gems/quilt_rails/lib/quilt_rails/trusted_ui_server_csrf_strategy.rb
JackMc: If we're going with `Exception`, we could inherit from that instead.

GoodForOneFare: I feel like that's stretching OO for minor benefit.  It would create an, `is one thing and another thing` relationship, rather than an `is a`.

That said, I'm not a Rails developer.  If you feel that inheritance would be more idiomatic, I can be easily swayed 😄

GoodForOneFare:
> Interested in your thoughts on Shopify/web#18367 as it relates to this as well.

@JackMc `SameSite=Lax` looks great, but I think it needs at least one more Safari + iOS release to get 100% support for our baseline browsers (per [the caniuse](https://caniuse.com/#feat=same-site-cookie-attribute stats)).

JackMc:
Sounds good to me! Let's do this and SameSite=Lax then. The default for browsers when they see and don't understand it is to ignore it.

gems/quilt_rails/lib/quilt_rails/trusted_ui_server_csrf_strategy.rb
JackMc: Yeah, that sounds fine to me :D I wish it was a concern so we wouldn't have to instantiate another complete object, but I doubt it makes a substantial difference.

gems/quilt_rails/README.md
```diff
@@ -434,3 +434,28 @@ const app = createServer({
...
 export default app;
 ```
+
+### Fixing rejected CSRF tokens for new user sessions
+
+If a React component calls back to a Rails endpoint (e.g., `/graphql`), Rails may throw a `Can't verify CSRF token authenticity` exception. This stems from the Rails CSRF tokens not persisting until after the first `UiController` call ends.
+
+To fix this:
+
+- Add an `X-Shopify-Server-Side-Rendered: 1` header to all server-side GraphQL requests
```
TheMallen: Once we have the work @ismail-syed  is doing merged it will be relatively easy for us to add this header automagically I think. Until then this makes sense 👍

PR #1006, opened 2019-09-17
Copy templates to `dist` in `react-server-webpack-plugin`

cartogram:
## Description
Because the templates for generating the virtual entry point modules are never actually being imported/ used within this package, they never ended up in the `dist` folder during the build step. When using the package, an error that `no such file or directory....@shopify/react-server-webpack-plugin/dist/templates/client.ts' exists` occurs.

This PR solves the issues by ~~manually copying the files to the `dist` directory in script that runs after the build step~~ removing them from the `exclude` property of the tsconfig and using `@ts-ignore` comments to suppress TS errors.

- [X] `react-server-webpack-plugin` Patch: Bug/ Documentation fix (non-breaking change which fixes an issue or adds documentation)

## Checklist

cc @kirbyfike 

Fixes https://github.com/Shopify/quilt/issues/1009


packages/react-server-webpack-plugin/package.json
```diff
@@ -7,7 +7,8 @@
   "types": "dist/index.d.ts",
   "sideEffects": false,
   "scripts": {
-    "build": "tsc --p tsconfig.build.json",
+    "build": "tsc --p tsconfig.build.json && yarn run copy-templates",
```
ismail-syed: This feels a bit hacky to me and won't scale well when we migrate to TypeScript project references (see https://github.com/Shopify/quilt/pull/962). Instead, I think we should tweak our configs a bit.

* Add `"./src/templates/*"` to the `tsconfig.json#include`
* Remove `"**/*/templates/**/*""` from the `tsconfig.json#exclude`
* Add `// @ts-ignore` to where the build fails since these are template files. If we could opt the whole file out of typechecking that would be better imo.

TheMallen: I think this plan makes sense

cartogram:
@ismail-syed thanks! I just didn't need the include, seems like it works without it.

PR #1007, opened 2019-09-18
Handle invalid error path for submission errors in react-form

aishad:
## Description

Passing invalid error path on form submission was causing a JS error trace :

image

Issue found here: (https://github.com/Shopify/store/issues/9108)

In react-form, when propagating remote errors to its matching field,  it is possible that the return value of `get` can be `undefined` [here](https://github.com/Shopify/quilt/blob/27862acdc3b13f02dd1e0b4f2245e325955d9d6b/packages/react-form/src/utilities.ts#L61).  If an invalid field is passed, this causes `isField(undefined)` to throw an error:  

**Fix**
If `get` is undefined, we ignore the invalid field path when propagating through submission errors. Errors will still be returned to the component, but they will not be mapped to a specific field. 

## Type of change

## Checklist

### DIFF
--- a/packages/react-form/src/hooks/test/form.test.tsx
+++ b/packages/react-form/src/hooks/test/form.test.tsx
@@ -262,6 +262,37 @@ describe('useForm', () => {
       });
     });
 
+    it('ignores invalid field path when propagating remote submission errors', async () => {
+      const errors = [
+        {
+          field: ['invalid'],
+          message: 'The server hates your price',
+        },
+      ];
+
+      const promise = Promise.resolve(submitFail(errors));
+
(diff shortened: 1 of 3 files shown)


packages/react-form/src/hooks/test/form.test.tsx
```diff
@@ -262,6 +262,37 @@ describe('useForm', () => {
       });
     });
 
+    it('ignores invalid field path when propagating remote submission errors', async () => {
+      const errors = [
+        {
+          field: ['invalid'],
+          message: 'The server hates your price',
```
TheMallen: :(

Issue #1009, opened 2019-09-18
[react-server-webpack-plugin] missing templates


macdonaldr93:
## Overview

When creating a new app from scratch, the webpack tooling expects to be able to find the templates to start the server and client.

```
$ sewing-kit dev
{ Error: ENOENT: no such file or directory, open '/Users/ryanmacdonald/src/github.com/Shopify/sandbox/node_modules/@shopify/react-server-webpack-plugin/dist/templates/client.ts'
    at Object.openSync (fs.js:443:3)
    at Object.readFileSync (fs.js:343:35)
    at clientSource (/Users/ryanmacdonald/src/github.com/Shopify/sandbox/node_modules/@shopify/react-server-webpack-plugin/dist/react-server-webpack-plugin.js:69:29)
    at ReactServerPlugin.modules (/Users/ryanmacdonald/src/github.com/Shopify/sandbox/node_modules/@shopify/react-server-webpack-plugin/dist/react-server-webpack-plugin.js:40:29)
    at ReactServerPlugin.apply (/Users/ryanmacdonald/src/github.com/Shopify/sandbox/node_modules/@shopify/react-server-webpack-plugin/dist/react-server-webpack-plugin.js:31:28)
    at Object.webpack (/Users/ryanmacdonald/src/github.com/Shopify/sandbox/node_modules/@shopify/sewing-kit/node_modules/webpack/lib/webpack.js:47:13)
    at Client.<anonymous> (/Users/ryanmacdonald/src/github.com/Shopify/sandbox/node_modules/@shopify/sewing-kit/dist/src/tools/webpack/dev/client.js:45:47)
    at Generator.next (<anonymous>)
    at fulfilled (/Users/ryanmacdonald/src/github.com/Shopify/sandbox/node_modules/@shopify/sewing-kit/dist/src/tools/webpack/dev/client.js:4:58)
  errno: -2,
  syscall: 'open',
  code: 'ENOENT',
  path:
   '/Users/ryanmacdonald/src/github.com/Shopify/sandbox/node_modules/@shopify/react-server-webpack-plugin/dist/templates/client.ts' }
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

## Checklist

- [x] Please delete the labels section before submitting your issue
- [x] I have described this issue in a way that is actionable (if possible)

PR #1010, opened 2019-09-18
WIP [react-server-webpack-plugin] include templates in build

macdonaldr93:
## Description

Fixes https://github.com/Shopify/quilt/issues/1009

When including this package from npm, it's missing the template files. It looks like we were ignoring the templates in the build step. After including the templates again, I can see that the files are available in the `build` folder.

## Type of change

- [x] react-server-webpack-plugin Patch: Bug/ Documentation fix (non-breaking change which fixes an issue or adds documentation)

## Checklist


macdonaldr93:
This doesn't actually work yet because of type issues but I think this is the fix

macdonaldr93:
Duplicate https://github.com/Shopify/quilt/pull/1006

PR #1011, opened 2019-09-18
[react-i18n] Modify translation keys used by I18n#humanizeDate

rexmac:
## Description

Modifies the translation keys used by the `I18n#humanizeDate` method to more accurately reflect the conditions rather than the (expected) English values.

### Before

```json
  "humanize": {
    "date": "{date} at {time}",
    "minutes": "{count} minutes ago",
    "now": "Just now",
    "weekday": "{day} at {time}",
    "yesterday": "Yesterday at {time}"
  },
```

### After

```json
  "date": {
    "humanize": {
      "lessThanOneMinuteAgo": "Just now",
      "lessThanOneHourAgo": "{count} minutes ago",
      "yesterday": "Yesterday at {time}",
      "lessThanOneWeekAgo": "{weekday} at {time}",
      "lessThanOneYearAgo": "{date} at {time}"
    }
  },
```

## Type of change

- [x] `@shopify/react-i18n` Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

### DIFF
--- a/packages/react-i18n/migration-guide.md
+++ b/packages/react-i18n/migration-guide.md
@@ -1,0 +1,83 @@
+# Migration guide
+
+This is a concise summary of changes and recommendations around updating `@shopify/react-i18n` in consuming projects. For a more detailed list of changes, see [the changelog](./CHANGELOG.md).
+
+## Unreleased
+
+🛑Breaking change - The translation keys for dates formatted with `DateStyle.Humanize` have changed. Consumers will need to modify the translation keys as outlined below.
+
+1. The top-level "humanize" key is now nested under a "date" key.
+
+   **Before**
+
+   ```json
+   {
+     "humanize": {
+       ...
+     }
+   }
+   ```
+
+   **After**
+
+   ```json
+   {
+     "date": {
+       "humanize": {
+         ...
+       }
+     }
+   }
+   ```
+
+2. The `day` replacement parameter for the `lessThanOneWeekAgo` (formerly `weekday`) string has been renamed to `weekday`.
+
+   **Before**
+
+   ```json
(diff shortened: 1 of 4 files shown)


rexmac:
> Isn't there a en.json or something missing too? Or is that handled elsewhere?

Handled by the consumers of the package.

michaelmelanson:
Who uses this function now? How do we migrate them?

rexmac:
> Who uses this function now?

Other than web, I'm not sure.

> How do we migrate them?

Even though this is a small change (in terms of lines/bytes), it _does_ break backwards compatibility and therefore will require a major version bump (likely v2.0.0) when released. Any consumers will have to force an upgrade whenever they're ready.

After this ships, assuming no one beats me to it, I'll create a PR to upgrade the package in web, which will involve bumping the version constraint on the dependency in `package.json` and updating the modified translation keys in the few places that they're referenced in the web repo.

rexmac:
[Migration guide](https://github.com/Shopify/quilt/pull/1011/files#diff-7daffa3b0831e6a46853922651580b2b) added.

packages/react-i18n/migration-guide.md
```diff
@@ -0,0 +1,83 @@
...
+   "date": {
+     "humanize": {
+       "lessThanOneMinuteAgo": "Just now",
+       "lessThanOneHourAgo": "{count} minutes ago",
+       "yesterday": "Yesterday at {time}",
+       "lessThanOneWeekAgo": "{weekday} at {time}",
+       "lessThanOneYearAgo": "{date} at {time}"
+     }
+   },
+   ```
```
kyledurand: This looks great to me 👍

Issue #856, opened 2019-08-15
Re-work react-server tests (Does it add all the things like an update-able title, etc...)


TheMallen:
The tests is `react-server` are not exactly our style and also do not cover most of what the server does.

Are we able to write a test that actually proves that everything is working together instead of just "it created the markup we were expecting".

PR #1014, opened 2019-09-18
Better server tests

cartogram:
## Description

This PR is my initial stab at https://github.com/Shopify/quilt/issues/856. The issue is a little vague so I'm just trying to hammer out a general approach.

~~The idea is that we want to test that the App component tree that users will pass into `createServer` can actually do the things we say it can. To accomplish this I've created a `mountAppWithServer` utility for mounting an App component on a random port and returning a response object that can be used in the assertions. The utility keeps track of the mounted servers in an internal array and provides a `stopServers` function to kill them all.~~

**EDIT: Sept 19th**
The idea is to have a class that does our setup (in a `mount` function) and takes in a function that will start our server. We can then call `request` and get back the returned response `text` for our assertions. The class will also expose a `unmountAll()` to kill all the servers.

In the tests themselves, I am calling some primitive hook in the App component tree that would not work if the provider did not exist in the `createServer` markup.

### DIFF
--- a/packages/react-server/src/server/test/server.test.tsx
+++ b/packages/react-server/src/server/test/server.test.tsx
@@ -1,9 +1,11 @@
-import {Server} from 'http';
 import React from 'react';
-import request from 'supertest';
-import getPort from 'get-port';
-import {createServer} from '../server';
-import {mockMiddleware} from '../../test/utilities';
+import {useTitle} from '@shopify/react-html';
+import {useAcceptLanguage} from '@shopify/react-network';
+import {
+  mockMiddleware,
+  stopServers,
+  mountAppWithServer,
+} from '../../test/utilities';
 
 jest.mock('@shopify/sewing-kit-koa', () => ({
   middleware: jest.fn(() => mockMiddleware),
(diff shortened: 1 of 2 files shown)


packages/react-server/src/server/test/server.test.tsx
```diff
@@ -16,37 +18,36 @@ jest.mock('@shopify/sewing-kit-koa', () => ({
...
+
+    const response = await mountAppWithServer(<MockApp />);
 
     expect(response.text).toBe(
       `<!DOCTYPE html><html lang="en"><head><meta charSet="utf-8"/><meta http-equiv="X-UA-Compatible" content="IE=edge"/><meta name="referrer" content="never"/></head><body><div id="app"><div>markup</div></div></body></html>`,
     );
   });
+
+  it('supports updatable meta components', async () => {
+    const myTitle = 'Shopify Mock App';
```
TheMallen: These tests are really nice *except* that they kind of obscure what we're actually testing. It might seem puritanical, but I think it's important to have the unit under test itself called directly in the test.

I've been thinking about how to do these node end-to-endy tests nicer and I think what we really want is something more like:

```
const rack = new TestRack();

beforeEach(() => {
  rack.unmountAll()
})

it('supports updatable meta components', async () => {
  const wrapper = rack.mount((port) => 
    createServer({port, render: () => <SomeJSX />})
  ) 

  const {text} = await wrapper.request('/').response()
  expect(response).toContainText('<title>whatever</title>')
```

where `TestRack` looks something like

```
interface TestRack {
  mount((port: number, ip: string) => Server): TestServer
  unmountAll(): void
}

interface ServerWrapper {
  request(fetchOptions?: FetchOptions):  Response
}
```

It's possible that `ServerWrapper` just wants to be a supertest(server), or a custom extended version of it, but I think the above is nice enough for now.

packages/react-server/src/test/utilities.ts
```diff
@@ -1,5 +1,43 @@
...
+
+  const server = await createServer({
+    port,
+    ip,
+    render: () => app,
+  });
+
+  servers.push(server);
+
+  return request(url)
```
TheMallen: it's really weird to pull in supertest but then not expose it's API at all, we could just use `fetch` or something if we were going to work purely with a response object.

cartogram: yeah true 👍

cartogram: I don't have a reason currently to use supertest, so going to remove it. We can add it if/when it's needed down the road.

cartogram:
@TheMallen Thanks for the feedback, think I've got something closer to what you were thinking now.

packages/react-server/src/test/utilities.ts
```diff
@@ -1,6 +1,37 @@
+import {Server} from 'http';
 import {Context} from 'koa';
+import getPort from 'get-port';
+
 import {KoaNextFunction} from '../types';
 
+export class TestRack {
```
TheMallen: We could actually use this to do e2e style tests for all our node server related libraries. We don't need to pull it out into a package now necessarily but it's worth keeping in mind.

packages/react-server/src/test/utilities.ts
```diff
@@ -1,6 +1,37 @@
...
+  async mount(mountFunction: ({port: number, ip: string}) => Server) {
+    const port = await getPort();
+    const ip = 'http://localhost';
+    const server = mountFunction({port, ip});
+
+    this.servers.push(server);
+
+    return {
+      request: async () => {
+        const text = await fetch(`${ip}:${port}`).then(response =>
```
TheMallen: the one thing here is I'd maybe just return the whole response since I think it's unlikely all tests universally care about the text.

PR #1005, opened 2019-09-17
[Proof of concept] SSR GraphQL operations tracking

ismail-syed:
## Description

Fixes (issue #)

## Type of change

- [ ]  Patch: Bug/ Documentation fix (non-breaking change which fixes an issue or adds documentation)
- [ ]  Minor: New feature (non-breaking change which adds functionality)
- [ ]  Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

- [ ] I have added a changelog entry, prefixed by the type of change noted above

PR #1018, opened 2019-09-19
Add support for GraphQL operations tracking on server-side renders

ismail-syed:
## Description

Related to `Add Performance Metrics`

I tried to split this PR up into separate PRs but that seemed to be more trouble than good.
- Its easier to grok this solution in one diff than 4
- It's hard to teach Lerna about different lib changes without them in the branch. 

This PR adds GraphQL metric support in `@shopify/react-server` by introducing new features into our pre-existing libraries.  

### `@shopify/react-graphql`

This is where the core of the solution comes from. I'm adding a `createOperationDetailsLink` Apollo link to expose GraphQL operation details. Also adding a `totalGraphQLTime` utiltiy to determine the duration of those GraphQL operations. 

### `@shopify/react-network`

We need to eventually share the provided GraphQL operations from `createOperationDetailsLink` to our koa state so we can metric, yo. I've introduced `serverState` into the `NetworkManger` to enable this and updated `applyToContext` to add it to the koa context. 

### `@shopify/react-graphql-universal-provider`

- Update our API to create the `ApolloClient` instead of letting consumers create it. (Context: https://github.com/apollographql/apollo-client/issues/5347)
-  Add `createOperationDetailsLink` from `react-graphql` to `ApolloClient` and use it to track GraphQL operations using the NetworkManager. The use `NetworkManger`'s `serverState` APIs to store these operations in koa state.  

### `@shopify/react-server`

- Use the `@shopify/react-network` changes to get the GraphQL operations stored from the server-side render. Throw them into the GraphQL timing utilities added in `@shopify/react-graphql` to get the graphql operation durations. Then send them off in the response header

## 🎩 

* In `gems/quilt_rails/lib/quilt_rails/react_renderable.rb` make some changes so we can see the headers get sent
  * Change the `_response` to response in the `reverse_proxy` block
  * Add `puts "************* Headers: #{response.to_hash.inspect}"` in the block
* Check out the `poc-graphql-server-operations-2` branch in `web-rails-proving-ground` and run the app
* Hit a full page reload on `/customers` so we get an SSR page.   
     * Verify that the logs prepended with `************* Headers:  ` contains metrics for `graphql_time` and `react-rendering`

## Checklist


packages/react-graphql/package.json
```diff
@@ -36,6 +37,7 @@
     "react-apollo": ">=2.2.3 <3.0.0"
   },
   "devDependencies": {
+    "@shopify/apollo-link-test-utilities": "^0.0.1-beta.1",
```
ismail-syed: Did a quick ship to unblock myself, https://github.com/Shopify/quilt/pull/1019

packages/react-graphql-universal-provider/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+## 2.0.0 - 2019-09-13
+- 🛑Breaking change: `GraphQLUniversalProvider` expects a `createClientOptions` and will create ApolloClient using the options provided [#1018](https://github.com/Shopify/quilt/pull/1018)
```
ismail-syed: Update the readme docs to reflect this

packages/react-graphql-universal-provider/src/test/GraphQLUniversalProvider.test.tsx
```diff
@@ -29,55 +29,27 @@ jest.mock('@shopify/react-graphql', () => {
...
-      <GraphQLUniversalProvider createClient={() => client} />,
+      <GraphQLUniversalProvider createClientOptions={() => clientOptions} />,
     );
 
     expect(graphQL).toContainReactComponent(ApolloProvider, {
       client: expect.any(ApolloClient),
     });
   });
 
-  it('serializes the apollo apollo cache and re-uses it to hydrate the cache', async () => {
```
ismail-syed: This test called `client.extract()` manually which is no longer accessible. Is there any other way we can test this functionality still?

packages/react-graphql-universal-provider/src/GraphQLUniversalProvider.tsx
```diff
@@ -1,33 +1,64 @@
...
+  createSsrExtractableLink,
+  createOperationDetailsLink,
+  GraphQLOperationDetails,
+} from '@shopify/react-graphql';
 import {useLazyRef} from '@shopify/react-hooks';
 import {createApolloBridge} from '@shopify/react-effect-apollo';
+import {useNetworkManager} from '@shopify/react-network';
+
+export type GraphQLClientOptions = ApolloClientOptions<any>;
+export const GRAPHQL_OPERATIONS = Symbol('graphQLOperations');
```
ismail-syed: I considered having this in `@shopify/react-graphql` at one point. I ended up leaving it here since `GraphQLUniversalProvider` does the _handshaking_ between the `@shopify/react-graphql` & `@shopify/react-network`.

packages/react-graphql/src/links/operations-details-link.ts
```diff
@@ -0,0 +1,84 @@
...
+  end: number;
+  success: boolean;
+  errors?: {message: string; path?: string}[];
+  url?: string;
+  status?: number;
+  requestId?: string;
+}
+
+export interface Options {
+  requestIdHeader?: string;
```
ismail-syed: In web we used some custom header. Exposing this so the consumer can decide what request Id header they want to use.

ismail-syed: Actually, this header is simply `X-Request-ID`, should we just provide this header instead?

packages/react-graphql-universal-provider/src/test/GraphQLUniversalProvider.test.tsx
ismail-syed: cc/ @michenly

TheMallen: couldn't we do it the same way that `react-server` does? ie using `extract` from `react-effect/server`?

michenly: maybe we can try to get `client` from `ApolloProvider`'s prop for extraction?

packages/react-graphql-universal-provider/src/GraphQLUniversalProvider.tsx
```diff
@@ -1,33 +1,64 @@
...
-    const client = createClient();
-    client.link = link.concat(client.link);
+
+    if (clientOptions.link) {
+      clientOptions.link = link
+        .concat(
+          createOperationDetailsLink({
+            onOperation(operation) {
+              if (network) {
+                const newServerState = addOperationToServerState(
```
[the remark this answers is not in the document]
michenly: after which I wonder if we can use `useNetworkEffect` instead of `useNetworkManager`?
and if that is possible, maybe we can have `useGraphqlOperation` in `react-network` to encapsulate the network related things.

packages/react-graphql-universal-provider/src/GraphQLUniversalProvider.tsx
```diff
@@ -1,33 +1,64 @@
...
 import {createApolloBridge} from '@shopify/react-effect-apollo';
+import {useNetworkManager} from '@shopify/react-network';
+
+export type GraphQLClientOptions = ApolloClientOptions<any>;
+export const GRAPHQL_OPERATIONS = Symbol('graphQLOperations');
 
 interface Props {
   children?: React.ReactNode;
-  createClient(): import('apollo-client').ApolloClient<any>;
+  createClientOptions(): ApolloClientOptions<any>;
```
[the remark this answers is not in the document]
ismail-syed: It's to allow users to pass in functions to customize the returned options - [context](https://github.com/Shopify/quilt/pull/1005#discussion_r325369735).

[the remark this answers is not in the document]
michenly: > for the same reasons the previous api was createClient(): ApolloClient not client: ApolloClient.

😬 I dont know if I entirely understand why. I can only guess it is because of link or resolvers? Everything else seems pretty static.

packages/react-graphql-universal-provider/src/GraphQLUniversalProvider.tsx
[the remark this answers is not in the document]
ismail-syed: >  maybe we can have useGraphqlOperation in

In the end, we need to be able to use these graphql operations in a koa middleware. We can't call `useGraphqlOperation` in the koa middleware. Unless there's a pattern here that I'm not thinking of.

packages/react-graphql-universal-provider/src/test/GraphQLUniversalProvider.test.tsx
ismail-syed: Michelle's suggestion worked. Pulled client from the `ApolloProvider`'s props.

packages/react-graphql/src/links/operations-details-link.ts
ismail-syed: Added `Header.RequestId` to `@shopify/network` and consumed it here

packages/react-graphql-universal-provider/src/GraphQLUniversalProvider.tsx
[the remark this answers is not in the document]
michenly: no, hook have to be use in a react component. Can you remind me why this is have to be use in koa middleware and not in the App code?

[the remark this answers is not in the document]
ismail-syed: > why this is have to be use in koa middleware

Because we need to send off the [metrics](https://github.com/Shopify/quilt/pull/1018/files#diff-167e62803c79a1e1b024bbead47ca47bR36-R40) in  the [response header](https://github.com/Shopify/quilt/pull/1018/files#diff-167e62803c79a1e1b024bbead47ca47bR42-R43) to the rails server.

packages/react-graphql-universal-provider/src/GraphQLUniversalProvider.tsx
```diff
@@ -1,33 +1,64 @@
...
+    const clientOptions = createClientOptions();
     const link = createSsrExtractableLink();
-    const client = createClient();
-    client.link = link.concat(client.link);
+
+    if (clientOptions.link) {
+      clientOptions.link = link
+        .concat(
+          createOperationDetailsLink({
+            onOperation(operation) {
```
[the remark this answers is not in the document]
ismail-syed: There's not much implementation here to export. We're just appending operations to network state here. Consumers with multiple graphql endpoints doing custom things can call  `createOperationDetailsLink` themselves with their own `onOperation` callback.

packages/react-graphql-universal-provider/src/GraphQLUniversalProvider.tsx
```diff
@@ -1,33 +1,64 @@
...
 interface Props {
   children?: React.ReactNode;
-  createClient(): import('apollo-client').ApolloClient<any>;
+  createClientOptions(): ApolloClientOptions<any>;
 }
 
-export function GraphQLUniversalProvider({children, createClient}: Props) {
+export function GraphQLUniversalProvider({
+  children,
+  createClientOptions,
```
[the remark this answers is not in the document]
ismail-syed: Updated docs

ismail-syed:
I'm going to close this for now. As Chris mentioned, there's still a fair amount of work required to Add performance metrics properly to our _quilt framework_. Extracting more performance-related libraries out of web, better collection mechanics from the app, and better patterns to interact with rails to name a few.

Issue #1001, opened 2019-09-13
Prevent missing `@shopify/*` packages from going out to releases


cartogram:
## Overview

Packages found in `./packages` often import one another. For example, `@shopify/react-server-webpack-plugin` makes use of `@shopify/ast-utilities`. In these cases, the imported package must be listed in the importing package's `package.json`. This can be easy to forget as there are no build, development, or publishing error reported until the package is released and installed in a consuming project. The consuming project will get an error, such as `'@shopify/ast-utilities' modules not found`.

An easy temporary fix when this happens is to install the missing package in the consuming project, but the permanent fix is to add ’@shopify/ast-utilities' to `@shopify/react-server-webpack-plugin` package.json#dependencies. This will require a re-release, which can be time-consuming.

### 1. Lint

We can lint for missing packages and fail the lint step. A rule exists called `import/no-extraneous-packages` from `eslint-plugin-import`, but this rule is recognizing the @shopify namespace imports as “internal” (not “external”) causing them to be skipped over/ not reported as extraneous.

#### To see the problem:

- Hack the typescript config in node modules and turn on the rule. (I am not sure why we have it turned off at the moment, the comment says it has something to do with not supporting typescript, but I haven't experienced any issues with the rule https://github.com/Shopify/eslint-plugin-shopify/blob/master/lib/config/typescript.js#L41) 
- Remove the `@shopify/ast-utilities` package from `react-server-webpack-plugin` and `run yarn lint '*/**/react-server-webpack-plugin/**/**.ts’` and you should see 2 errors. This is wrong because there is no indication of the missing `ast-utilities` package we just deleted.
- Now run ```$ node --inspect-brk node_modules/eslint/bin/eslint.js '*/**/react-server-webpack-plugin/**/**.ts’``` from the quilt repo.
- Go to `chrome://inspect`  and adding a breakpoint In on line ~123 of the no-extraneous-import rule. 
- Change the import type condition to check for both external and internal, you should now see 3 errors instead of 2.

I haven't been able to figure out why this is happening. It could be a configuration problem (I tried a number of different settings) or could require a patch to the repo.

### 2. Run a script/test

This can be wrapped into a jest test, but something that runs in CI and checks to make sure that any imported packages are also listed in the package.json. 

A rough example below.

```tsx
const path = require('path');
const {readdirSync} = require('fs-extra');
const grep = require('simple-grep');
const uniq = require('lodash/uniq');

[...]

PR #1026, opened 2019-09-22
Restructure repo to report on missing dependencies

cartogram:
## Description

Fixes https://github.com/Shopify/quilt/issues/1001

The changes in this PR allow us to turn on the eslint rule [no-extraneous-dependencies](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-extraneous-dependencies.md). This is a big change so I'm breaking down the three major things happening here:

1) I moved all of the packages to `packages/@shopify` to allow the eslint rule to recognize the packages in this repository properly. More context in this issue https://github.com/Shopify/quilt/issues/1001

2) Adjusted our `.eslintrc` to automatically populate the overrides for each package in this repo to only look locally in that package for existing dependencies.

3) With the above 2 changes, it now successfully reporting errors and the final changes are me retro-actively applying the missing dependencies for the ~200 times we broke this rule, but got lucky in that it didn't break anything.

Additional things
- Point 3 is work in progress, I want to get a general approval before going through that additional effort.
- In the config for the rule I chose to `devDependencies: true, optionalDependencies: false, peerDependencies: true,` as it seemed like the best way to accommodate the missing `react`/`react-dom` dependencies across this repo.
- This PR seems a little over-the-top, but I think organizing the repo in this way would have been good idea anyways. Mat brought up the point of migrating to a `quilt` framework more easily which I agree with.
- Anything repercussions of this change that I am missing? For example, additional libraries being in conflict when packages are installed in other projects?


GoodForOneFare:
^ That aside, great job on figuring all of this out 👍  It's well on the way to solving a significant pain point for us and our consumers 🎉

cartogram:
Closing this PR. @lemonmade has a different solution in the works that doesn't require the massive restructure.

PR #1027, opened 2019-09-23
React i18n babel coverage

vsumner:
## Description
Since the upgrade to jest 24 the react i18n babel plugin has not been compatible with code coverage. This PR fixes that by adjusting how referencePath parents are checked.

https://github.com/Shopify/web/pull/13887

### Why did this break?
This is the code output after the babel transforms without `--coverage`
Screen Shot 2019-09-23 at 1 46 47 PM

This is the code output after the babel transforms with `--coverage`
Screen Shot 2019-09-23 at 1 42 48 PM

The issue is that `babel-plugin-istanbul` is running before `@shopify/react-i18n/babel` is applied.

This changes `parentPath` from `CallExpression` to `SequenceExpression` as you can see the code in the second screenshot. `parent` remains `CallExpression` and also contains the `arguments`. I adjusted this and 🎩 ed against `plus-web`.

## Type of change

- [X]  Patch: Bug/ Documentation fix (non-breaking change which fixes an issue or adds documentation)

## Checklist

### DIFF
--- a/packages/react-i18n/CHANGELOG.md
+++ b/packages/react-i18n/CHANGELOG.md
@@ -8,6 +8,10 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 <!-- ## [Unreleased] -->
 
+### Changed
+
+- Fixed babel plugin incompatiblity with jest code coverage
+
 ## [2.0.0] - 2019-09-19
 
 ### Changed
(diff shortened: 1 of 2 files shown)


GoodForOneFare:
@vsumner  Could you add a bit of context to the PR description about why the old code broke, and how this fixes it, please?

vsumner:
> @vsumner Could you add a bit of context to the PR description about why the old code broke, and how this fixes it, please?

Done. Had it open in another tab and forgot to save last night 👍

Issue #1033, opened 2019-09-24
sewing-kit.config.ts generated file is incompatible with default project 


ewalk153:
## Overview

When running `rails g sewing_kit:install`, the config file does export the module properly. The following error message is generated:
```
[error] Type errors encountered while parsing sewing-kit config:
    /Users/ericwalker/src/github.com/Shopify/quilt-rails-example/config/sewing-kit.config.ts (3,1): All files must be modules when the '--isolatedModules' flag is provided.
Error: Type errors encountered while parsing sewing-kit config:
    /Users/ericwalker/src/github.com/Shopify/quilt-rails-example/config/sewing-kit.config.ts (3,1): All files must be modules when the '--isolatedModules' flag is provided.
    at /Users/ericwalker/src/github.com/Shopify/quilt-rails-example/node_modules/@shopify/sewing-kit/dist/src/workspace/config.js:123:19
    at Generator.next (<anonymous>)
    at fulfilled (/Users/ericwalker/src/github.com/Shopify/quilt-rails-example/node_modules/@shopify/sewing-kit/dist/src/workspace/config.js:4:58)
```

The fix is to export with: `export default sewingKitConfig;`

GoodForOneFare:
Thanks for letting us know!  Fixed by #1033, releasing in `quilt_rails@1.8.0`.

PR #1038, opened 2019-09-24
`quilt:install` should fully bootstrap TypeScript / React

GoodForOneFare:
## Description

Fixes (issue #1033)

- Adds typescript as a dependency to stop compilation throwing exceptions about `App.tsx`
- Generates a `tsconfig.json` to stop sewing-kit complaining about `cannot build a typescript project without tsconfig.json`
- Use `export default` in `sewing-kit.config.ts` to compile without depending on CommonJS
  - This addresses the warnings about `--isolated-modules` in #1033 
- Add `react`, et al. as dependencies because `react-server` crashes hard without them

## Type of change

- [x] quilt_rails - Minor: New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/gems/quilt_rails/lib/generators/quilt/install_generator.rb
+++ b/gems/quilt_rails/lib/generators/quilt/install_generator.rb
@@ -8,7 +8,24 @@ module Quilt
 
     def install_js_dependencies
       say "Installing @shopify/react-server and @shopify/sewing-kit dependencies"
-      system("yarn add @shopify/sewing-kit @shopify/react-server") unless Rails.env.test?
+      system("yarn add "\
+        "@shopify/sewing-kit "\
+        "@shopify/react-server "\
+        "typescript "\
+        "react "\
+        "react-dom "\
+        "@types/react "\
+        "@types/react-dom") unless Rails.env.test?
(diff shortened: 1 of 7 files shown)


gems/quilt_rails/lib/generators/quilt/install_generator.rb
```diff
@@ -8,7 +8,24 @@ class InstallGenerator < Rails::Generators::Base
 
     def install_js_dependencies
       say "Installing @shopify/react-server and @shopify/sewing-kit dependencies"
-      system("yarn add @shopify/sewing-kit @shopify/react-server") unless Rails.env.test?
+      system("yarn add "\
```
cartogram: I think we might want to lock these to a version. What do you think?

michenly: definitely for react and typescript.

GoodForOneFare: Agreed, but stuff is busted, so it's a tomorrow problem: https://github.com/Shopify/quilt/issues/1040

PR #1039, opened 2019-09-24
Accept ApolloClient options instead of the client iteself

ismail-syed:
## Description

This PR updates the `GraphQLUniversalProvider` props to expect a `createClientOptions` function over a `createClient`. This prevents us from falling victim to this [issue](https://github.com/apollographql/apollo-client/issues/5347). 

This issue became apparent when working out https://github.com/Shopify/quilt/pull/1018. I've pulled it out from that PR. 

## Type of change

## Checklist

### DIFF
--- a/packages/react-graphql-universal-provider/src/GraphQLUniversalProvider.tsx
+++ b/packages/react-graphql-universal-provider/src/GraphQLUniversalProvider.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 
+import ApolloClient, {ApolloClientOptions} from 'apollo-client';
 import {useSerialized} from '@shopify/react-html';
 import {ApolloProvider, createSsrExtractableLink} from '@shopify/react-graphql';
 import {useLazyRef} from '@shopify/react-hooks';
@@ -7,10 +8,13 @@ import {createApolloBridge} from '@shopify/react-effect-apollo';
 
 interface Props {
   children?: React.ReactNode;
-  createClient(): import('apollo-client').ApolloClient<any>;
+  createClientOptions(): ApolloClientOptions<any>;
 }
 
-export function GraphQLUniversalProvider({children, createClient}: Props) {
+export function GraphQLUniversalProvider({
+  children,
+  createClientOptions,
+}: Props) {
   const [initialData, Serialize] = useSerialized<object | undefined>('apollo');
 
   const [client, link] = useLazyRef<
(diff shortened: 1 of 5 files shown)


packages/react-graphql-universal-provider/src/GraphQLUniversalProvider.tsx
```diff
@@ -1,16 +1,20 @@
 import React from 'react';
 
+import ApolloClient, {ApolloClientOptions} from 'apollo-client';
```
michenly: Please add `apollo-client` to the `dependencies` list. It is sadly missing from before...><...

packages/react-graphql-universal-provider/src/GraphQLUniversalProvider.tsx
```diff
@@ -19,15 +23,20 @@ export function GraphQLUniversalProvider({children, createClient}: Props) {
       ReturnType<typeof createSsrExtractableLink>
     ]
   >(() => {
+    const clientOptions = createClientOptions();
     const link = createSsrExtractableLink();
-    const client = createClient();
-    client.link = link.concat(client.link);
+
+    if (clientOptions.link) {
+      clientOptions.link = link.concat(clientOptions.link);
```
lemonmade: You shouldn't mutate their original object, and shouldn't you still assign `clientOptions.link` to `link` if they didn't have an explicit one?

ismail-syed: Don't we want to provide the `createSsrExtractableLink` link as we were doing before? 

> shouldn't you still assign clientOptions.link to link if they didn't have an explicit one?

Will fix

ismail-syed: This should be `if (!clientOptions.link)` then, right?

lemonmade: no, like, if they have no link, add one. If they have one, augment it.

packages/react-graphql-universal-provider/src/test/GraphQLUniversalProvider.test.tsx
```diff
@@ -64,22 +72,36 @@ describe('<GraphQLUniversalProvider />', () => {
...
+      cache: new InMemoryCache(),
+    };
+
+    const client = mount(
+      <GraphQLUniversalProvider createClientOptions={() => clientOptions} />,
+    )
+      .find(ApolloProvider)!
+      .prop('client');
+
+    expect(client.link instanceof SsrExtractableLink).toBe(true);
```
lemonmade: You can do nicer assertions than this: expect(client).toContainReactComponent(ApolloProvider, {client: expect.any(SarExtractableLink)})

ismail-syed: That doesn't work since `client` expects an `ApolloClient` constructor.  Is there a way to perform nested assertions. Something like this (which doesn't work either) 

```
expect(client).toContainReactComponent(ApolloProvider, {
  client: {
      link: expect.any(SarExtractableLink),
  }
)
```

lemonmade: sorry it would be `expect.objectContaining({link: expect.any(SsrExtractableLink)})`

ismail-syed: Going to merge, for now. Will follow up if I get it the cleaner assertion working

Issue #958, opened 2019-09-05
Some Node exceptions aren't logged


GoodForOneFare:
## Overview
In production, beta-buddy the first `useQuery` call for a session fails (#957), but no Node exceptions appear in our production logs.  There should be _something_ about the call out to GraphQLController failing here.

## Consuming repo

beta-buddy

ismail-syed:
Given that we've disabled SSR in beta-buddy, would this be re-producible in web-rails-proving-ground?

GoodForOneFare:
It's not specific to `useQuery`.  To test:
- Create a `server.ts` file that intentionally throws an exception
  - Try to find the exception logs in Splunk
- Revert the above change, and throw an unchecked exception from a component
  - Try to find the exception logs in Splunk

Note that Mat has a PR open in sewing_kit to enable more logging, so you might want to test with that integrated.

GoodForOneFare:
And note that this is behaviour that I observed in passing almost a month ago.  Things may have changed since then, so don't take my word as gospel ¯\\_(ツ)_/¯

ismail-syed:
> Create a server.ts file that intentionally throws an exception

* Working on this one right now. As of now, not seeing any logs in splunk from the server. [test changeset](https://github.com/Shopify/web-rails-proving-ground/commit/2c9f2558ea4cf1bee6c5067d6d76dd964276b297)

> Revert the above change, and throw an unchecked exception from a component

Splunk logs are readable with https://github.com/Shopify/quilt/pull/978

ismail-syed:
> As of now, not seeing any logs in splunk from the serve

Found the logs! Though the server-level exception logs from `server.ts` are chopped up like what was mentioned in https://github.com/Shopify/quilt/issues/959

ismail-syed:
> Create a server.ts file that intentionally throws an exception

We spoke about this yesterday in the pod. If we throw an exception in a server file outside the middlewares (which now captures readbale logging on splunk) it's almost certain that our server will fail to start. When our server fails to start, [shipit will fail on deploy](https://shipit.shopify.io/shopify/web-rails-proving-ground/production/deploys/839899). This is enough for us to write off this concern. 

That said, we still want some reasonable feedback in development. In development, our logs are already visible by default now in `@shopify/react-server` - See https://github.com/Shopify/quilt/pull/978. 

@GoodForOneFare, can we close this off or do you want to verify with `useQuery` in beta buddy first?

GoodForOneFare:
[...]

PR #1050, opened 2019-09-26
Log errors on `uncaughtException` `unhandledRejection` events

ismail-syed:
## Description
See title

Fixes https://github.com/Shopify/quilt/issues/958

I did a 🎩 in [`web-rails-proving-ground`](https://github.com/Shopify/web-rails-proving-ground/pull/164) with this change along with a change to generate a [exception](https://github.com/Shopify/quilt/pull/react-server-uncaught-server-exceptions-test-release).

Shipit deploy [logs](https://shipit.shopify.io/shopify/web-rails-proving-ground/production/deploys/849980) contain more meaningful logs on why the node server failed on startup.

```
[FATAL][2019-09-26 14:07:54 +0000]	      $ node build/server/main.js
[FATAL][2019-09-26 14:07:54 +0000]	      React Server failed to start.
[FATAL][2019-09-26 14:07:54 +0000]	      Error: Test error. I am an uncaught exception thrown in the server that shows in shipit logs
[FATAL][2019-09-26 14:07:54 +0000]	      at breakThings (/app/build/server/webpack:/app/ui/server.js:18:15)
[FATAL][2019-09-26 14:07:54 +0000]	      at Module.breakThings (/app/build/server/webpack:/app/ui/server.js:34:5)
```  

## Checklist

### DIFF
--- a/packages/react-server-webpack-plugin/src/react-server-webpack-plugin.ts
+++ b/packages/react-server-webpack-plugin/src/react-server-webpack-plugin.ts
@@ -81,6 +81,13 @@ function serverSource(options: Options) {
     import React from 'react';
     import {createServer} from '@shopify/react-server';
     import App from 'index';
+    process.on('uncaughtException', logError);
+    process.on('unhandledRejection', logError);
+    function logError(error) {
+      const errorLog = \`\${error.stack || error.message || 'No stack trace was present'}\`;
+      console.log(\`React Server failed to start.\n\${errorLog}\`);
+      process.exit(1);
+    }
     const render = (ctx) =>
     React.createElement(App, {
       server: true,
--- a/packages/react-server-webpack-plugin/CHANGELOG.md
+++ b/packages/react-server-webpack-plugin/CHANGELOG.md
@@ -7,6 +7,10 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## Unreleased -->
 
+## [2.2.0] - 2019-09-26
+
+- Log errors on `uncaughtException` `unhandledRejection` events [#1006](https://github.com/Shopify/quilt/pull/1006)
+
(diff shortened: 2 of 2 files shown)


packages/react-server-webpack-plugin/src/react-server-webpack-plugin.ts
```diff
@@ -81,6 +81,13 @@ function serverSource(options: Options) {
     import React from 'react';
     import {createServer} from '@shopify/react-server';
     import App from 'index';
+    process.on('uncaughtException', logError);
+    process.on('unhandledRejection', logError);
+    function logError(error) {
+      const errorLog = \`\${error.stack || error.message || 'No stack trace was present'}\`;
+      console.log(\`React Server failed to start.\n\${errorLog}\`);
```
GoodForOneFare: How confident are you about this message only being thrown during startup?  Does koa catch 100% of rejected promises?

ismail-syed: I'm pretty confident about this change. I tried creating some e2e tests which would involve testing dev startup which gets too messy with our e2e testing approach. Unless there's an intuitive way for us to test dev startups that I'm not aware of. 


> Does koa catch 100% of rejected promises

No, not natively to koa. However, our [logger](https://github.com/Shopify/quilt/blob/master/packages/react-server/src/logger/logger.ts#L88-L91) will handles rejected promises within koa.

GoodForOneFare:
I'm a little confused that this change didn't happen in [`templates/server.ts`](https://github.com/Shopify/quilt/blob/master/packages/react-server-webpack-plugin/src/templates/server.ts).  Are those templates not used, yet?

ismail-syed:
> Are those templates not used, yet?

This drove me crazy for longer that I'd like 😅 ... Turns out [they're not](https://github.com/Shopify/quilt/pull/1021).

PR #1052, opened 2019-09-26
adds version to webhook package api call

katiedavis:
## Description
Adds the ability to pass a version of the API within webhook registration. 
@TheMallen would this be considered a breaking change?
## Type of change

- [x] koa-shopify-webhooks: Major change

## Checklist

### DIFF
--- a/packages/koa-shopify-webhooks/src/register.ts
+++ b/packages/koa-shopify-webhooks/src/register.ts
@@ -1,11 +1,19 @@
 import {Method, Header} from '@shopify/network';
 import {WebhookHeader, Topic} from './types';
 
+export enum ApiVersion {
+  April19 = '2019-04',
+  July19 = '2019-07',
+  Unstable = 'unstable',
+  Unversioned = 'unversioned',
+}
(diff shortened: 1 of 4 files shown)


packages/koa-shopify-webhooks/src/register.ts
```diff
@@ -1,27 +1,39 @@
...
+  Unstable = 'unstable',
+  Unversioned = 'unversioned',
+}
+
 export interface Options {
   topic: Topic;
   address: string;
   shop: string;
   accessToken: string;
+  apiVersion: ApiVersion;
```
GoodForOneFare: Since I rudely didn't respond to this in Slack... yes, I think this is breaking.  Unless we default to the old behaviour.  

My preference would be to bite the bullet here because -- in my limited understanding -- it makes sense to always specify a version.

katiedavis: I totally agree. Partners have been having issues with the old behaviour and certain topics anyway.

PR #1064, opened 2019-09-27
Fix no-extraneous-dependencies: Take 2

### LINKED ISSUE #1001: Prevent missing `@shopify/*` packages from going out to releases

cartogram:
## Description

Fixes #1001
Follow up from https://github.com/Shopify/quilt/pull/1026

This PR is split into [2 commits](https://github.com/Shopify/quilt/pull/1064/commits). The [first](https://github.com/Shopify/quilt/pull/1064/commits/11345db3fccbe75a9c64f34d6087c5064071e327) contains only the changes required to make the `import/no-extraneous-dependencies` rule work for this repo, the [second](https://github.com/Shopify/quilt/pull/1064/commits/af001140efeb3953c4018fd0ecb4335a066792fa) is all the lint fixes required to pass CI.

The fix here was to use the `import/resolver` setting to override the typescript resolver from `plugin:shopify/typescript`. The typescript resolver was problematic because it resolves modules to their source location, causing the rule think of them as "internal" (not extraneous) modules. Creating a new configuration and extending it before extending the `plugin:shopify/typescript` tells the rule to use the node resolver, which looks for these packages in `node_modules` instead and properly sees them as "external" to the eslint rule.

The downside to having this rule enabled is that it (correctly) enforces all our `@shopify/*` dependencies to be listed when they are used and thus increases the dependency web and number of "patch" dependency bumps needed (releasing a new version of dependency `x` causes all related dependencies that use `x`  to need a new version). The largest offender of this was `@shopify/react-testing`, which is used in all of the `@shopify/react-*` dependencies. 

The fix for this was to add an override for test files that disables this rule. We don't care if `@shopify/react-testing` is listed or not since its really just needed for development and will never cause a missing dependency error when installed in a consuming package, which is the only thing we _really_ care about here.


packages/graphql-persisted/package.json
```diff
@@ -29,7 +29,8 @@
     "koa-compose": ">=3.0.0 <4.0.0"
   },
   "devDependencies": {
-    "@shopify/sewing-kit-koa": "^6.1.2"
+    "@shopify/sewing-kit-koa": "^6.1.2",
+    "@types/koa": "^2.0.45"
```
cartogram: This library, `@shopify/graphql-persisted`, uses the `Context` from `koa` and would require the consumer to have `koa` installed in their project. There were a number of packages that met this criteria, and I use the following method to resolve them: 1) Add the packages the host is required to have in the `peerDependencies` with a pretty loose version range. and 2) add the types packages used to the `devDependencies`.

Why? I added the `peerDependency` because it is needed yet unused in this library. I added a loose version range to reduce the chances of 2 version of the package been installed as a result of this library.

I added the types in `devDependencies` because they should be there in development, but are not exported in this library. However, there is a middleware `persistedOperationAssociationMiddleware` that uses the `Context` type in one of its arguments and _that_ function _is_ being exported. For this reason, I wonder if the types should be moved to dependencies. If anyone can clarify this, please comment.

.eslintrc.js
```diff
@@ -0,0 +1,52 @@
...
+      files: [
+        '**/*/test/**/*.ts',
+        '**/*/test/**/*.tsx',
+        '**/*/tests/*.ts',
+        '**/*/tests/*.tsx',
+        '*.test.ts',
+        '*.test.tsx',
+      ],
+      rules: {
+        'shopify/jsx-no-hardcoded-content': 'off',
```
cartogram: Here we turn this rule off to prevent requiring packages such as `@shopify/react-testing` in ever react-based package.

packages/jest-koa-mocks/package.json
```diff
@@ -23,13 +23,13 @@
   },
   "homepage": "https://github.com/Shopify/quilt/blob/master/packages/jest-koa-mocks/README.md",
   "dependencies": {
-    "koa": "^2.5.0",
+    "koa": "^2.8.2",
```
cartogram: Bringing our `koa` and `@types/koa` to be more consistent across packages in order to remove duplicate versions.

packages/graphql-persisted/package.json
```diff
@@ -24,6 +24,7 @@
   "dependencies": {
     "@types/koa-bodyparser": "*",
     "@types/koa-compose": "*",
+    "@types/koa": "^2.0.50",
```
cartogram: This library uses the `Context` from `koa` and would require the consumer to have `koa` installed in their project. There were a number of packages that met this criteria, and I use the following method to resolve them: 1) Add the packages the host is required to have in the `peerDependencies` with a pretty loose version range, and 2) add the types packages used to the dependencies.

Why? I added the peerDependency because it is needed (yet unused) in this library. You can't really use a `koa` middleware as intended without `koa`, but you also don't need `koa` to write them. They are, more or less, just functions. I added a loose version range to reduce the chance 2 versions of the package will be installed, in this case `koa`, as a result of this library.

I added the types in dependencies because there is a middleware, `persistedOperationAssociationMiddleware`, that uses the `Context` type in one of its arguments and that function is being exported. Even though the use of the type is tangental, we are not exporting the type itself or expecting consumers will use it, they will use the function that uses the type. For this reason, it needed to be in dependencies.

packages/react-html/package.json
```diff
@@ -29,7 +29,8 @@
     "@shopify/useful-types": "^2.0.1",
     "@types/multistream": "^2.1.1",
     "multistream": "^2.1.1",
-    "tslib": "^1.9.3"
+    "tslib": "^1.9.3",
+    "serialize-javascript": "^2.1.0"
```
cartogram: This was a strange one to be missing.

packages/graphql-persisted/package.json
lemonmade: Good rationale. Is this version range too restrictive, though?

packages/react-i18n/src/i18n.ts
```diff
@@ -1,3 +1,6 @@
+import {memoize as memoizeFn} from '@shopify/function-enhancers';
+import {memoize} from '@shopify/decorators';
+import {languageFromLocale, regionFromLocale} from '@shopify/i18n';
```
cartogram: With the resolver changes and re-classification of these libraries (from "internal" to "external"), they are being rearranged as a result of `import/order`.

packages/react-router/src/components/Link/Link.tsx
```diff
@@ -1,4 +1,4 @@
-import * as React from 'react';
+import React from 'react';
```
cartogram: This is unrelated to the purpose of this PR, just cleaning up while in here.

packages/react-server/src/logger/logger.ts
```diff
@@ -3,7 +3,10 @@ import {Context, Request} from 'koa';
 import chalk from 'chalk';
 import {KoaNextFunction} from '../types';
 
-export const LOGGER = Symbol('logger');
+// Have to cast as `any` or typescript will complain
+// about using a "unique symbol as an index type".
+// https://github.com/microsoft/TypeScript/issues/1863
+export const LOGGER: string = Symbol('logger') as any;
```
cartogram: Open to other ways around this. It is a pattern we use a lot so unfortunate that we have to cast.

packages/koa-shopify-auth/package.json
```diff
@@ -32,10 +32,10 @@
   "devDependencies": {
     "@shopify/jest-dom-mocks": "^2.8.0",
     "@shopify/jest-koa-mocks": "^2.1.5",
-    "@types/koa": "^2.0.44",
+    "@types/koa": "^2.0.50",
     "@types/koa-compose": "*",
     "@types/safe-compare": "^1.1.0",
-    "koa": "^2.5.0"
+    "koa": "^2.8.2"
```
lemonmade: Just a heads up that you don't generally need to update dependencies in this way, using `yarn-deduplicate` can get you down to a single set of installed dependencies without forcing more restrictive version ranges

packages/react-cookie/src/CookieUniversalProvider.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
-import {useNetworkManager} from '@shopify/react-network';
 import {useLazyRef} from '@shopify/react-hooks';
+import {useNetworkManager} from '@shopify/react-network';
```
lemonmade: Why does this kind of re-order need to happen? Aren't they both external?

packages/react-html/package.json
lemonmade: Why?

packages/react-i18n/package.json
```diff
@@ -34,6 +34,8 @@
     "intl-pluralrules": "^0.2.1"
   },
   "dependencies": {
+    "@babel/template": "^7.6.0",
+    "@babel/traverse": "^7.6.2",
```
lemonmade: Hmm, I'm not sure if these are strictly dependencies. Aren't they just used for types? Babel should generally be a peer/ optional dependency IMO

packages/sewing-kit-koa/src/middleware.ts
```diff
@@ -17,7 +17,10 @@ export interface Options {
   manifestPath?: string;
 }
 
-const ASSETS = Symbol('assets');
+// Have to cast as `any` or typescript will complain
+// about using a "unique symbol as an index type".
+// https://github.com/microsoft/TypeScript/issues/1863
```
lemonmade: I don't understand, what changed that caused this?

packages/react-html/package.json
cartogram: This library uses it heavily so I'm surprised the library works without it listed. I think its because `react-serialize` has it.

packages/sewing-kit-koa/src/middleware.ts
cartogram: @lemonmade the change from `types/koa` from `2.0.49 => 2.0.50`. My guess is this change https://github.com/DefinitelyTyped/DefinitelyTyped/commit/80c6d507792b9706617ca27e4fa5ec9e314db8a5#diff-7605d695f1da86ef8f6a929780113155R724

cartogram: `State` is no longer `any`

lemonmade: Can we cast `ctx.state` as `any` then? I don't like a type that is a straight up lie when we could switch to looser typing in one spot instead.

packages/react-cookie/src/CookieUniversalProvider.tsx
cartogram: Yeah good question, this looks like a mistake.

packages/jest-koa-mocks/src/create-mock-context/create-mock-context.ts
```diff
@@ -109,6 +109,6 @@ export default function createContext<
 
   // ctx.request.body is a common enough custom property for middleware to add that it's handy to just support it by default
   context.request.body = requestBody;
-  context.request.rawBody = rawBody;
+  context.request.rawBody = rawBody || '';
```
jgodson: could we set this as a default argument on line 53 instead? (like `url` is)

BPScott:
I wonder if there's some upstream PR we can make to https://github.com/alexgorbatchev/eslint-import-resolver-typescript to allow support for our monorepo use case

.eslintrc.js
```diff
@@ -0,0 +1,52 @@
...
+          },
+        ],
+      },
+    },
+    {
+      files: [
+        '**/*/test/**/*.ts',
+        '**/*/test/**/*.tsx',
+        '**/*/tests/*.ts',
+        '**/*/tests/*.tsx',
```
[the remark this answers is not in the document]
cartogram: Moving 2 files (example: `react-i18n/src/e2e/translations.test.tsx` => `react-i18n/src/test/e2e/translations.test.tsx`) gets us just the 4 patterns. The test/tests thing can be resolved in a follow-up.

config/eslint.js
```diff
@@ -0,0 +1,13 @@
+// We need to specify the import/resolver setting in an extendable
```
BPScott: I'm surprised this works - rules later in the config overrode earlier ones - it's how you can enable a rule in a config, then disable it in your local eslint config. I would have thought settings would work in the same way.

lemonmade: I think the import plugin iterates over the setting object, which gets merged together by ESLint's config. However, it merges in the order of plugin addition, and then iterates over the object (iteration happens in insertion order), so resolvers added by earlier plugins are actually handled first.

.eslintrc.js
```diff
@@ -0,0 +1,50 @@
...
+    {
+      files: [
+        '**/test/**/*.ts',
+        '**/test/**/*.tsx',
+        '**/tests/**/*.ts',
+        '**/tests/**/*.tsx',
+      ],
+      rules: {
+        'shopify/jsx-no-hardcoded-content': 'off',
+        'import/no-extraneous-dependencies': 'off',
```
lemonmade: Can we leave a note for why we turn this off (we should do this always but alas)

packages/address/src/tests/AddressFormatter.test.ts
```diff
@@ -1,5 +1,5 @@
-import {fetch} from '@shopify/jest-dom-mocks';
 import {Address} from '@shopify/address-consts';
+import {fetch} from '@shopify/jest-dom-mocks';
```
lemonmade: Did you explain why this would need to go below `@shopify/address-consts`?

packages/graphql-persisted/package.json
```diff
@@ -23,10 +23,11 @@
   "homepage": "https://github.com/Shopify/quilt/blob/master/packages/graphql-persisted/README.md",
   "dependencies": {
     "@types/koa-bodyparser": "*",
-    "@types/koa-compose": "*",
+    "@types/koa-compose": "^3.2.2",
+    "@types/koa": "^2.0.0",
     "apollo-link": ">=1.0.0 <2.0.0",
     "koa-bodyparser": ">=4.0.0 <5.0.0",
-    "koa-compose": ">=3.0.0 <4.0.0"
+    "koa-compose": "^4.1.0"
```
lemonmade: Do we need to restrict the version range on this?

packages/graphql-persisted/package.json
```diff
@@ -23,10 +23,11 @@
   "homepage": "https://github.com/Shopify/quilt/blob/master/packages/graphql-persisted/README.md",
   "dependencies": {
     "@types/koa-bodyparser": "*",
-    "@types/koa-compose": "*",
+    "@types/koa-compose": "^3.2.2",
```
lemonmade: Why did this change?

packages/jest-dom-mocks/package.json
```diff
@@ -30,7 +30,8 @@
     "fetch-mock": "^6.3.0",
     "lolex": "^2.7.5",
     "promise": "^8.0.3",
-    "tslib": "^1.9.3"
+    "tslib": "^1.9.3",
+    "lodash": "^4.17.15"
```
lemonmade: Can this be looser?

packages/koa-shopify-auth/package.json
```diff
@@ -24,16 +24,16 @@
   "homepage": "https://github.com/Shopify/quilt/blob/master/packages/koa-shopify-auth/README.md",
   "dependencies": {
     "@shopify/network": "^1.4.2",
-    "koa-compose": ">=3.0.0 <4.0.0",
+    "koa-compose": "^4.1.0",
```
lemonmade: Same question about these dependencies

packages/react-app-bridge-universal-provider/package.json
```diff
@@ -32,7 +32,6 @@
   },
   "devDependencies": {
     "@shopify/react-effect": "^3.2.3",
```
lemonmade: 🔥 this one too?

packages/react-effect-apollo/package.json
```diff
@@ -24,7 +24,8 @@
   },
   "homepage": "https://github.com/Shopify/quilt/blob/master/packages/react-effect-apollo/README.md",
   "dependencies": {
-    "@shopify/react-effect": "^3.2.3"
+    "@shopify/react-effect": "^3.2.3",
+    "prop-types": "^15.7.2"
```
lemonmade: Should be able to delete all of these because I deprecated the package

packages/react-server/package.json
```diff
@@ -33,14 +33,18 @@
...
     "@shopify/react-network": "^3.3.1",
     "@shopify/sewing-kit-koa": "^6.1.2",
+    "@shopify/useful-types": "^1.3.0",
     "chalk": "^2.4.2",
     "koa": "^2.5.0",
     "koa-compose": "^4.1.0",
-    "koa-mount": "^4.0.0"
+    "koa-mount": "^4.0.0",
+    "fs-extra": "^8.1.0",
+    "isomorphic-fetch": "2.2.1"
```
lemonmade: Is this actually used for it directly?

packages/react-server/package.json
```diff
@@ -33,14 +33,18 @@
     "@shopify/react-hydrate": "^1.1.4",
     "@shopify/react-network": "^3.3.1",
     "@shopify/sewing-kit-koa": "^6.1.2",
+    "@shopify/useful-types": "^1.3.0",
     "chalk": "^2.4.2",
     "koa": "^2.5.0",
     "koa-compose": "^4.1.0",
-    "koa-mount": "^4.0.0"
+    "koa-mount": "^4.0.0",
+    "fs-extra": "^8.1.0",
```
lemonmade: What is this used for?

cartogram: It is used for checking if there is a Gemfile, when getting the manifest file.

```

function getManifestPath(root: string) {
  const gemFileExists = pathExistsSync(join(root, 'Gemfile'));
  if (!gemFileExists) {
    return;
  }

  // eslint-disable-next-line no-process-env
  return process.env.NODE_ENV === 'development'
    ? `tmp/sewing-kit/sewing-kit-manifest.json`
    : `public/bundles/sewing-kit-manifest.json`;
}
```

packages/react-server/package.json
cartogram: It isn't no.

cartogram: Might make more sense as a peer dependency.

packages/react-server/package.json
BPScott: Looks like it's an alias for the built in fs.existsSync.

https://github.com/jprichardson/node-fs-extra/blob/master/lib/path-exists/index.js#L11

packages/koa-performance/package.json
```diff
@@ -27,14 +27,15 @@
     "@shopify/network": "^1.4.2",
     "@shopify/performance": "^1.2.0",
     "@shopify/statsd": "^1.1.0",
+		"@types/koa": "^2.0.0",
```
lemonmade: Weird spacing

PR #1065, opened 2019-09-27
Memoize numberFormatter and pluralRules in translate utility

rexmac:
## Description

Memoizes some usages of `Intl.NumberFormat` and `Intl.PluralRules` to significantly reduce the number of new objects being created. 

- Created `memoizedPluralRules` in `utilities/translate.ts`
- Removed creation of `Intl.PluralRules` object from `I18n` constructor (`this.ordinalPluralRules`)
   - This object only existed for memoization purposes
   - The `ordinal` method now uses `memoizedPluralRules` instead
- Moved `memoizedNumberFormatter` from `i18n.ts` to `utilities/translate.ts` so that it can be used by other functions in that file and exported for use by `I18n`
- _Note: This has no effect for SSR because [memoization is disabled](https://github.com/Shopify/quilt/tree/master/packages/function-enhancers#memoize):_
   > Know that memoization will be skipped on server process

## Type of change

- [x] `@shopify/react-i18n` Patch: Bug/ Documentation fix (non-breaking change which fixes an issue or adds documentation)

## Checklist

### DIFF
--- a/packages/react-i18n/CHANGELOG.md
+++ b/packages/react-i18n/CHANGELOG.md
@@ -6,7 +6,16 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ---
 
-<!-- ## [Unreleased] -->
+## [Unreleased]
+
+### Added
+
+- Added `memoizedPluralRules` utility function ([#1065](https://github.com/Shopify/quilt/pull/1065)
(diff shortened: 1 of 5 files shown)


rexmac:
I'm thinking this should actually be a _minor_ version bump. Is exporting new utility methods "added functionality"?

PR #1066, opened 2019-09-29
Add BILLING Webhook Topics

gil--:
~Add required GDPR webhook topics for public apps as outlined at https://help.shopify.com/en/api/guides/gdpr-resources~

~- customers/redact -- Requests deletion of customer data.~
~- shop/redact -- Requests deletion of shop data.~
~- customers/data_request -- Requests to view stored customer data.~

- Add billing webhooks

## Description

- ~Currently the `@shopify/koa-shopify-webhooks` library does not support the 3 GDPR webhook topics outlined on https://help.shopify.com/en/api/guides/gdpr-resources~
- Newer billing webhook topics are also not supported on https://help.shopify.com/en/api/guides/billing-api#webhooks-for-billing

> Note: the webhook topics on [that page](https://help.shopify.com/en/api/guides/billing-api#webhooks-for-billing) are incorrect. They're listed as `APP_PURCHASE_ONE_TIME_UPDATE` instead of `APP_PURCHASES_ONE_TIME_UPDATE ` and `APP_SUBSCRIPTION_UPDATE` instead of `APP_SUBSCRIPTIONS_UPDATE`

## Type of change

- [X] @shopify/koa-shopify-webhooks Patch: Adds 2 new webhook topics for Billing. Existing functionality unchanged.

## Checklist


gil--:
@lemonmade got it. I'll remove for this PR the GDPR topics. My main reasoning for adding was that I'm using koa-shopify-webhooks and wanted to validate incoming webhooks with the middleware but they're definitely handled differently since we don't subscribe and instead specify the callback/url in the app settings.

tolgap:
Hey guys, I happen to exactly need one of these extra webhook topics. Any chances of getting this merged soon?

lemonmade:
Sorry for the delay, this will be in version `2.3.0`, which should be released in a few minutes. Just a note, though: even though the types restricted what can go here, I hope people are comfortable temporarily "fudging" the types a little in order to get your work done. Something like this would have worked, and it's something we often do internally at Shopify when we are stuck waiting:

```ts
{topic: 'APP_PURCHASES_ONE_TIME_UPDATE' as any}
```

but, correct types are obviously better, so thanks again for the PR @gil--!

Issue #1067, opened 2019-09-30
Unify statsd-client into one package `@shopify/statsd-client`


michenly:
## Overview

There are 3 implementations of statsd-client across quilt & web.
1. https://github.com/Shopify/web/blob/master/packages/%40shopify/koa-performance/middleware.ts#L66
2. https://github.com/Shopify/quilt/blob/master/packages/koa-metrics/src/Metrics.ts#L44
3. https://github.com/Shopify/web/blob/master/packages/%40shopify/statsd/client.ts#L10

Lets unify it into one.

PR #1068, opened 2019-09-30
✨ start a new package

michenly:
## Description

Fixes (issue #) https://github.com/Shopify/quilt/issues/1067

Unifying these 3
1. https://github.com/Shopify/web/blob/master/packages/%40shopify/koa-performance/middleware.ts#L66
2. https://github.com/Shopify/quilt/blob/master/packages/koa-metrics/src/Metrics.ts#L44
3. https://github.com/Shopify/web/blob/master/packages/%40shopify/statsd/client.ts#L10

## 🎩 
1. Using this companion branch in web
https://github.com/Shopify/web/pull/18952
2. start by creating folder for the new package in `web/node_modules/@shopify/statsd` and copy the `package.json` over
3. `yarn tophat statsd ../web` in quilt

## Todo

- [x] Bring in all the `snakeCase ` customization from https://github.com/Shopify/web/blob/master/packages/%40shopify/statsd/client.ts
- [x] Unify with https://github.com/Shopify/quilt/blob/master/packages/koa-metrics/src/Metrics.ts#L91
- [x] add test to the client
- [x] add documentation to the package


packages/koa-metrics/package.json
```diff
@@ -23,6 +23,7 @@
   },
   "homepage": "https://github.com/shopify/quilt/blob/master/packages/koa-metrics/README.md",
   "dependencies": {
+    "@shopify/statsd": "^1.0.0",
```
michenly: can I do this yet? I remember us discussing a fix recently. 
Or would I need to wait till `@shopify/statsd` is actually publish.

packages/koa-metrics/src/Metrics.ts
```diff
@@ -1,104 +0,0 @@
...
-      host,
-      port,
-      prefix,
-      globalTags,
-    });
-    this.client = this.rootClient;
-  }
-
-  public distribution(name: string, value: number, tags?: TagsMap) {
-    this.log(MetricType.Distribution, name, value, tags);
```
michenly: this is the only line I did not carry over

michenly:
Still adding test & document. But the implementation can be review + 🎩 now!

packages/koa-metrics/package.json
michenly: ya I can't...grr...

README.md
```diff
@@ -96,10 +97,10 @@ Check out our [Contributing Guide](./.github/CONTRIBUTING.md)
 
 ## Questions?
 
-For Shopifolk, you can reach out to us in Slack in the `#web-foundation-tech` channel. For external inquiries, we welcome bug reports, enhancements, and feature requests via Github issues.
+For Shopifolk, you can reach out to us in Slack in `#web-foundation-tech` channel. For external inquiries, we welcome bug reports, enhancements, and feature requests via Github issues.
```
TheMallen: Either remove `channel` or add back the `the` here please :)

packages/statsd/src/client.ts
```diff
@@ -0,0 +1,127 @@
...
+}
+
+export interface Options extends ClientOptions {
+  logger?: Logger;
+  snakeCase?: boolean;
+}
+
+export class StatsDClient {
+  private statsd: StatsD;
+  private logger: Logger | null = null;
```
TheMallen: can we default this to a dirt simple object that just exposes console.log when there is no logger provided?

packages/statsd/src/client.ts
```diff
@@ -0,0 +1,127 @@
...
+  private options: Options;
+
+  constructor({logger, ...options}: Options) {
+    this.logger = logger || null;
+    this.options = {
+      ...options,
+      errorHandler: options.errorHandler
+        ? options.errorHandler
+        : this.handleError.bind(this),
```
TheMallen: Can we just make it an error function and not bother with this bind?

packages/statsd/src/client.ts
```diff
@@ -0,0 +1,127 @@
...
+  ) {
+    return (error?: Error) => {
+      if (error) {
+        if (this.options.errorHandler) {
+          this.options.errorHandler(error);
+        }
+        reject(error);
+      }
+
+      // Failing to send the metrics is not something we need to propagate
```
TheMallen: This comment is a bit of a lie, as far as I can tell from the way this behaves / the code in general if we get here it is because there was no error (therefore sending stats succeeded). We can just🔥 this comment.

packages/statsd/src/client.ts
```diff
@@ -0,0 +1,127 @@
...
+    if (this.logger == null) {
+      return;
+    }
+
+    this.logger.log(
+      `Error occurred in the StatsD client:\n${error.stack || error.message}`,
+    );
+  }
+
+  private createErrorCallback<T>(
```
TheMallen: `errorCallback` is a bit misleading here, since the returned function will get called on every metric regardless of pass or fail

packages/statsd/src/test/client.ts
```diff
@@ -0,0 +1,3 @@
+describe('statsd', () => {
+  it.todo('works');
```
TheMallen: I'd like to see some tests here, not sure exactly how we could make them meaningful. Looking at `hot-shots` itself they setup a weird local server to see that it receives the expected requests. :/

I'm ok with them coming in as a subsequent PR at worst.

packages/statsd/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+An opinionated StatsD client for Shopify Node.js server and other StatsD utilities.
+
+## Installation
+
+```bash
+$ yarn add @shopify/statsd
+```
+
+## Usage
```
TheMallen: We should add these before merging :)

packages/statsd/src/test/client.test.ts
```diff
@@ -0,0 +1,262 @@
...
+      const error = new Error('Something went wrong!');
+      statsDMock.close.mockImplementation(callback => {
+        callback(error);
+      });
+
+      statsDClient.close();
+      expect(errorHandler).toHaveBeenCalledWith(error);
+    });
+
+    it("calls logger's log on the promise when the client returns an error and there is no errorHandler in options", () => {
```
TheMallen: Personally I'd just🔥 the tests that test what we log since we generally don't test logs in other libraries.

packages/statsd/src/client.ts
```diff
@@ -0,0 +1,125 @@
...
+    });
+  }
+
+  addGlobalTags(globalTags: Tags) {
+    this.statsd = this.statsd.childClient({
+      globalTags,
+    });
+  }
+
+  initTimer(): Timer {
```
TheMallen: I'm not super sure this makes sense as part of this class tbh, it seems unrelated to the rest, and doesn't share any state.

michenly: good call, 🔥ing

PR #1070, opened 2019-09-30
adding asChoiceField utility to react-form and react-form-state

### LINKED ISSUE #673: FormState Typescript issues when using Polaris checkbox
## Overview
Typescript complains when I use `FormState` boolean field and the spread operator on a Polaris Checkbox.

I saw some solutions in other repos like using an utility `asChoiceField`
```
<Checkbox
      id={`${Id.Root}${Id.AllowOutOfStockPurchases}`}
      label={i18n.translate('InventoryCard.inventoryPolicyLabel')}
      {...asChoiceField(allowOutOfStockPurchases)}
    />
```
or changing the type when using the spread operator:
```
 <Checkbox
                    label={i18n.translate("tos_checkbox")}
                    checked={tosCheckbox.value}
                    {...tosCheckbox as FieldDescriptor<any>}
  />
```
1. What is the problem with the spread operator here? could someone explain me a bit?
2. Will choicefield be supported at some point as mentioned here https://github.com/Shopify/web/pull/5581 ?

## Consuming repo

https://github.com/Shopify/shopify-products-web

## Checklist

patsissons:
## Description

Fixes #673
Related to https://github.com/Shopify/web/pull/18953

Migrating the `asChoiceField` utility function that has been living in `web` for a while now to quilt, which simplifies spreading the field into a `Checkbox` or `RadioButton.

I have simplified these two utility functions a bit by just aliasing `value` to `checked` instead of performing a `Boolean(value)` cast (which seemed redundant).

this new utility is used by spreading its result into the `Checkbox` or `RadioButton`.

For `react-form` we also now have a simplified `useChoiceField` hook that will automatically perform the conversion which simplifies the API for consumers that only use the choice field for choice components.

```tsx
const field = useField(false);
const choiceField = useChoiceField(false);

return (
  <>
    <Checkbox label="Enabled" {...asChoiceField(field)} />
    <Checkbox label="Enabled" {...choiceField} />
  </>
);
```

## Type of change

- [x] `react-form` and `react-form-state` Minor: New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/packages/react-form/src/hooks/field/field.ts
+++ b/packages/react-form/src/hooks/field/field.ts
@@ -195,6 +195,45 @@ export function useField<Value = string>(
   return field as Field<Value>;
 }
 
+/**
+ * Converts a standard `Field<boolean>` into a `ChoiceField` that is compatible
+ * with `<Checkbox />` and `<RadioButton />` components in `@shopify/polaris`.
+ *
+ * For fields that are used by both a choice components and other components, it
+ * can be beneficial to retain the original `Field<boolean>` shape and convert
+ * the field on the fly for the choice component.
+ *
+ * ```typescript
+ * const enabled = useField(false);
+ * return (<Checkbox label="Enabled" {...asChoiceField(enabled)} />);
+ * ```
+ */
+export function asChoiceField({value: checked, ...fieldData}: Field<boolean>) {
+  return {
+    checked,
+    ...fieldData,
+  };
+}
+
+export type ChoiceField = ReturnType<typeof asChoiceField>;
+
+/**
+ * A simplification to `useField` that returns a `ChoiceField` by automatically
+ * converting the field using `asChoiceField` for direct use in choice
+ * components.
+ *
+ * ```typescript
+ * const enabled = useChoiceField(false);
+ * return (<Checkbox label="Enabled" {...enabled} />);
+ * ```
+ */
+export function useChoiceField(
+  input: FieldConfig<boolean> | boolean,
+  dependencies: unknown[] = [],
+) {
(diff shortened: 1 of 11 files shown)


patsissons:
do we need any documentation changes for this new utility?

packages/react-form/src/utilities.ts
```diff
@@ -97,3 +97,14 @@ export function validateAll(fieldBag: {[key: string]: FieldOutput<any>}) {
 }
 
 export function noop() {}
+
+/**
+ * Transforms a boolean Field object to work with checkboxes and radios.
+ * @param field
+ */
+export function asChoiceField({value: checked, ...fieldData}: Field<boolean>) {
```
TheMallen: could we also expose this as `useChoiceField` where we create the field inline?

patsissons: oooooooo great idea!

packages/react-form/src/hooks/field/field.ts
```diff
@@ -195,6 +195,45 @@ export function useField<Value = string>(
...
+ * A simplification to `useField` that returns a `ChoiceField` by automatically
+ * converting the field using `asChoiceField` for direct use in choice
+ * components.
+ *
+ * ```typescript
+ * const enabled = useChoiceField(false);
+ * return (<Checkbox label="Enabled" {...enabled} />);
+ * ```
+ */
+export function useChoiceField(
```
patsissons: @TheMallen i moved the functions straight into this module. IMO it feels more natural to put them here.

patsissons:
@TheMallen should i put anything in the `README.md` too?

PR #1071, opened 2019-09-30
Update Router readme

cartogram:
## Description

This PR adds some documentation for the `Router` component in `@shopify/react-router`.

### DIFF
--- a/packages/react-router/README.md
+++ b/packages/react-router/README.md
@@ -17,7 +17,22 @@ $ yarn add @shopify/react-router
 
 ### `<Router />`
 
-_TBD_
+Rendering the `Router` component at the top-level of your application will create the router that is provided to the rest of the React tree. It takes a single optional prop, `location`, that represents the current location in the server-side render of the application. This is not used or required in the client-side render of your application and can be `undefined` in that environment.
+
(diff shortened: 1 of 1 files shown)


packages/react-router/README.md
```diff
@@ -17,7 +17,23 @@ $ yarn add @shopify/react-router
 
 ### `<Router />`
 
-_TBD_
+Rendering the `Router` component at the top-level of your application will create the router that is provided to the rest of the React tree. It takes a single prop, `location`, that represents the current location to be used in the initial server-side render of the application.
```
michenly:
```suggestion
Rendering the `Router` component at the top-level of your application will create the router that is provided to the rest of the React tree. It takes an optional single prop, `location`, that can be left `undefined` during client side rendering, but need to be passed in to represents the current location in the initial server-side render of the application.
```

packages/react-router/README.md
```diff
@@ -17,7 +17,23 @@ $ yarn add @shopify/react-router
...
+
+```
```
michenly:
```suggestion
```tsx
```

packages/react-router/README.md
```diff
@@ -17,7 +17,23 @@ $ yarn add @shopify/react-router
...
+import React from 'react';
+import {Router} from '@shopify/react-router';
+
+export function App({location}) {
```
michenly:
```suggestion
export function App({location}: {location?: string}) {
```

PR #1072, opened 2019-09-30
remove ensyme from a few packages

cartogram:
## Description

This PR removes some lingering `enzyme` from a the following packages:

`react-async`
`react-serialize`
`react-compose`
`react-form-state`

I happened to be in here fixing some extraneous dependency violations for `enzyme` and found it easy enough to just remove it.

### DIFF
--- a/packages/react-compose/package.json
+++ b/packages/react-compose/package.json
@@ -30,8 +30,5 @@
   "peerDependencies": {
     "react": "^16.3.2"
   },
-  "devDependencies": {
-    "@shopify/react-testing": "^1.7.3"
-  },
   "sideEffects": false
 }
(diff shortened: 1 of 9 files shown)


packages/react-compose/package.json
```diff
@@ -31,7 +31,7 @@
     "react": "^16.3.2"
   },
   "devDependencies": {
-    "@types/enzyme": "^3.1.10"
+    "@shopify/react-testing": "^1.7.3"
```
lemonmade: I'd prefer we not declare this dependency if possible, I think it means react-testing bumps unnecessarily bump all these other libraries too

packages/react-form-state/package.json
```diff
@@ -32,6 +32,7 @@
   },
   "devDependencies": {
     "@shopify/enzyme-utilities": "^2.1.4",
+    "@shopify/useful-types": "^1.3.0",
     "@types/enzyme": "^3.1.10",
```
lemonmade: Don't need to remove this?

PR #1080, opened 2019-10-02
add october first versions

katiedavis:
## Description

Adds October1st API version to packages.

## Type of change

- [ ] koa-shopify-graphql-proxy: Patch
- [ ] koa-shopify-webhooks: Patch

## Checklist


packages/koa-shopify-webhooks/src/register.ts
```diff
@@ -4,6 +4,7 @@ import {WebhookHeader, Topic} from './types';
 export enum ApiVersion {
   April19 = '2019-04',
   July19 = '2019-07',
+  October1 = '2019-10',
```
TheMallen: wait should this be `October19` for both of these to match the others?

katiedavis: Wow yes. This is why you don't do any work at 6 am 🙃I'll update.

katiedavis:
> Looks good.
> 
> We might want to throw the `ApiVersion` and any other similar constants that are shared between our shopify-api specific libraries into like a `@shopify/shopify-constants` or something down the line 🤔

Agreed! I'll merge this for now and add that as a todo

PR #1082, opened 2019-10-02
✨ add the browser package

michenly:
Pull out the browser package from web
https://github.com/Shopify/web/tree/master/packages/%40shopify/browser

- [x] Add more test
- [x] Add Docs

### DIFF
--- a/packages/browser/src/browser.ts
+++ b/packages/browser/src/browser.ts
@@ -1,0 +1,115 @@
+import {UAParser} from 'ua-parser-js';
+
+export interface Options {
+  userAgent: string;
+  supported?: boolean;
+  bypass?: boolean;
+}
+
+export class Browser {
+  userAgent: string;
+  supported: boolean;
+  private ua: UAParser;
+
+  get name() {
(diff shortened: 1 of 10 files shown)


TheMallen:
@michenly I updated this branch with some docs and more tests, could you take a look? :)

packages/browser/src/browser.ts
```diff
@@ -0,0 +1,115 @@
...
+
+  get isMobile() {
+    return this.ua.getDevice().type === 'mobile';
+  }
+
+  get isDesktop() {
+    return !this.isMobile;
+  }
+
+  get isShopifyMobile() {
```
TheMallen: Switched `isNativeApp` to the more clearly-shopify-specific `isShopifyMobile`. Left it out of the docs because it is something only we really need. We could remove it entirely and just have our own little checks in web / other places that consume this I guess, not sure that's great though.

TheMallen:
Closing this PR for now as we can replicate the koa middleware's functionality by just passing the `user-agent` and letting consumers use their own parsing logic.

PR #1083, opened 2019-10-02
✨ React performance

TheMallen:
## Description
This PR adds `@shopify/react-performance`,  a library built on top of `@shopify/performance`, and pulled as-is out of the updated version of these components from https://github.com/Shopify/web/pull/18972.

This ends up being a pretty big diff but the actual runtime code is all currently running in production in web, so all you should need to review are the tests and docs, as well as the tweaks to our mocks to make them work.

### Changes of note from the version in web
- the tests are totally new
- had to add a `navigator.connection` mock to `jest-dom-mocks` to facilitate the tests
- the docs are new, incomplete
- added `mark` to `@shopify/performance`'s `Performance` class since it seemed weird to randomly call out to the window primitive after checking `supportsMarks`, and this makes things easier to test anyway

## Checklist

### DIFF
--- a/packages/jest-dom-mocks/src/connection.ts
+++ b/packages/jest-dom-mocks/src/connection.ts
@@ -1,0 +1,55 @@
+import {set} from './utilities';
+
+export interface NavigatorWithConnection extends Navigator {
+  connection: {
+    downlink: number;
+    effectiveType: string;
+    onchange: null | Function;
+    rtt: number;
+    saveData: boolean;
+  };
+}
+
+export class Connection {
+  private isUsingMockConnection = false;
+  private originalConnection?: NavigatorWithConnection['connection'];
+
+  mock(timing: Partial<NavigatorWithConnection['connection']> = {}) {
+    const globalNavigator = navigator as NavigatorWithConnection;
+    if (this.isUsingMockConnection) {
+      throw new Error(
+        'You tried to mock navigator.connection when it was already mocked.',
+      );
+    }
+
+    this.originalConnection = globalNavigator.connection;
+
+    const mockConnection: NavigatorWithConnection['connection'] = {
+      downlink: 0,
+      effectiveType: '3g',
+      onchange: null,
+      rtt: 100,
+      saveData: false,
+      ...timing,
+    };
+
+    set(globalNavigator, 'connection', mockConnection);
+    this.isUsingMockConnection = true;
+  }
+
+  restore() {
+    const globalNavigator = navigator as NavigatorWithConnection;
+    if (!this.isUsingMockConnection) {
+      throw new Error(
+        'You tried to restore navigator.connection when it was already restored.',
+      );
+    }
+
+    set(globalNavigator, 'connection', this.originalConnection);
+    this.isUsingMockConnection = false;
+  }
+
+  isMocked() {
+    return this.isUsingMockConnection;
+  }
+}
--- a/packages/react-performance/README.md
+++ b/packages/react-performance/README.md
@@ -1,0 +1,261 @@
+# `@shopify/react-performance`
+
+[![Build Status](https://travis-ci.org/Shopify/quilt.svg?branch=master)](https://travis-ci.org/Shopify/quilt)
+[![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](LICENSE.md) [![npm version](https://badge.fury.io/js/%40shopify%2Freact-performance.svg)](https://badge.fury.io/js/%40shopify%2Freact-performance.svg) [![npm bundle size (minified + gzip)](https://img.shields.io/bundlephobia/minzip/@shopify/react-performance.svg)](https://img.shields.io/bundlephobia/minzip/@shopify/react-performance.svg)
+
+Primitives to measure your React application's performance using `@shopify/performance`
+
+## Installation
+
+```bash
+$ yarn add @shopify/react-performance
+```
+
+## Quick-start
+
+### Basic
+
+The most basic way to use the tools in this package is to record information and display it locally to the user. In practice you usually only want to do this in development so that developers can easily see performance information.
+
+#### Add the context provider
+
+Before we can use any of the components or hooks in the package we must wrap our app tree with the `PerformanceContext.Provider` component.
+
+```tsx
+// App.tsx
+import React from 'react';
+import {Performance, PerformanceContext} from '@shopify/react-performance';
+
+// in a Server-Side-Rendering enabled app you will likely only want to instantiate this if `document` is defined.
+const performance = new Performance();
+
+function App() {
+  return (
+    <PerformanceContext.Provider value={performance}>
+      {/* The rest of your app */}
+    </PerformanceContext.Provider>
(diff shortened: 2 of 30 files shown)


packages/jest-dom-mocks/src/connection.ts
```diff
@@ -0,0 +1,55 @@
+import set from 'lodash/set';
```
michenly: `lodash/set` should be included in `package.json`

TheMallen: oh weird, it's used in other ones but isn't there. I'll add it

TheMallen: Actually I'm going to try just removing it instead

michenly: I am not surprise actually, @cartogram told me there is actually A LOT of missing dependencies in quilt when he was working on an automatic way of checking this.

so I am trying to be more carful recently.

packages/react-performance/README.md
```diff
@@ -0,0 +1,261 @@
...
+> This section is a work in progress. It will be filled out as helper libraries for marshalling data on the server are added to Quilt.
+
+## API
+
+Most APIs provided by this package are available as both hooks (ie. `useNavigationListener`), as well as components (ie. `<NavigationListener />`).
+
+This package also re-exports all of the API of [`@shopify/performance`](https://www.npmjs.com/package/@shopify/performance).
+
+### Hooks
+
```
michenly: Maybe it will be good to add some explanation on what life cycle events are before starting.

We can copy those from web's dev UI.

Screen Shot 2019-10-02 at 4 14 44 PM

packages/react-performance/README.md
```diff
@@ -0,0 +1,261 @@
...
+    console.log(event);
+  });
+
+  return <div>something</div>;
+}
+```
+
+#### useNavigationListener
+
+A custom hook which takes in a callback to invoke whenever a navigation takes place, whether it is a full-page load (such as the result of `location.assign`) or an incrmental load (such as the result of `router.push` in a ReactRouter app).
```
michenly: ❤️ really awesome explanations.

packages/react-performance/README.md
```diff
@@ -0,0 +1,261 @@
...
+### Hooks
+
+#### useLifecycleEventListener
+
+A custom hook which takes in callback to invoke whenever the `Performance` context object emits a lifecycleEvent.
+
+```tsx
+import {useLifecycleEventListener} from '@shopify/react-performance';
+
+function SomeComponent() {
```
michenly: I would almost just skip this example all together. It does not feel very useful till it is compose together later with more meaningful example that show case where this should go in an app and the reason why.

packages/react-performance/README.md
```diff
@@ -0,0 +1,261 @@
...
+}
+```
+
+#### usePerformanceReport
+
+> Todo
+
+### Components
+
+#### LifecycleEventListener
```
michenly: Maybe the doc instead of showing each individual API, should explain each concept first and then mention you can use it as a component or a hook.

Possible outline:

### Lifecycle Event Listener
Emit an `onEvent` callback whenever a lifecycle event occur.
The list of  lifecycle event is as follow:
1. `TimeToFirstByteEvent`: The time until the server sends the first part of response. [Learn more](https://developers.google.com/web/tools/chrome-devtools/network/understanding-resource-timing)
2.......

Should be place near the top of the react tree.
Only need to be include once in an application.
Can use component or hook.
See `Performance Report` below for how to compose it together.

packages/react-performance/src/test/LifecycleEventListener.test.tsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
```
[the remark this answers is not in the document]
TheMallen: I'm not sure I agree, I think generally we have avoided mocking hooks in tests.  Including `context` here is nice because it makes the test closer to how you would actually expect to use the component. Tests as documentation is very useful IMO.

packages/react-performance/src/test/LifecycleEventListener.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+    mount(
+      <PerformanceContext.Provider value={performance}>
+        <LifecycleEventListener onEvent={listener} />
+      </PerformanceContext.Provider>,
+    );
+
+    expect(performance.on).toHaveBeenCalledWith('lifecycleEvent', listener);
+  });
+
+  it('cleans up the event listener when unmounted', () => {
```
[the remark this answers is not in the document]
TheMallen: I could see also testing `usePerformanceEffect` setting up cleanup correctly, but I think it's important to call out that `LifecycleEventListener` and friends clean up their listeners. The fact that it uses `usePerformanceEffect` under the hood is an implementation detail to me.

packages/react-performance/README.md
TheMallen: I avoided spelling out some of the details of what `@shopify/react-performance`'s events are since ultimately this library is just a react-based abstraction on top of that, but looking at that library it seems like it's docs are pretty lacking. Probably we'll want to make those a lot better in a subsequent PR and then add more link-outs to those docs from this one.

packages/react-performance/README.md
TheMallen: I think we definitely need to add docs about what exact lifecycle events there are, but probably they belong to `@shopify/performance` and then `@shopify/react-performance` should link out. 

I will definitely do a subsequent PR to flesh out the docs in the base library though 👍

packages/react-performance/README.md
TheMallen: hmm, I'm not sure what the best choice is here. I like having something you can see that shows the API, and I don't think going into how everything fits together makes sense in the API section. I do want to add more to the first section (quick start) that talks about how to hook things up to server-side reporting in the future though. :/

packages/react-performance/src/test/LifecycleEventListener.test.tsx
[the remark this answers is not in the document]
michenly: I see what you mean by avoided mocking hooks, but this still feel really integrated. 


I try to imagine what I would do if `usePerformanceEffect` and `PerformanceContext` both came from a library, and I've come around on including `PerformanceContext.Provider`, those are similar to what we do with `mountWithAppContext`.

Testing clean up still feels like something that belongs to `usePerformanceEffect`'s test thou.

packages/react-performance/README.md
michenly: > I like having something you can see that shows the API

me too! 
however in this case some foresight of where it should be place is not as clear, and the example code so far feel like if you know react you can get that far.

if we would to keep the example, I would like to see a more involved `onEvent` in life cases.

packages/react-performance/src/test/LifecycleEventListener.test.tsx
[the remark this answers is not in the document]
TheMallen: updated it based on our convo :)

PR #1084, opened 2019-10-02
Introduce typings for web workers library

ismail-syed:
## Description

Fixes (issue #)

## Type of change

## Checklist


packages/web-workers/src/types.d.ts
```diff
@@ -0,0 +1,17 @@
...
+    : TypeReturned extends Promise<any> // check to prevent returning Promise<Promise<thing>>> in async function
+      ? (...args: Args) => TypeReturned
+      : (...args: Args) => Promise<TypeReturned>
+  : T extends Array<infer ArrayElement>
+    ? ArrayOfPromises<ArrayElement>
+    : {[K in keyof T]: Promisify<T[K]>};
+
+interface ArrayOfPromises<T> extends Array<Promisify<T>> {}
+
+export type WorkerInput<T> = Promisify<T>;
```
lemonmade: `WorkerInput` is not the same as `Promisify`. A worker **can** accept `() => string | Promise<string>`, so we should turn `() => string` into that instead of forcing a promise. A worker **can't** return a `() => string`, because that is returned over the bridge, which means it will only ever return a promise. So, `() => string` => `() => Promise<string>`.

packages/web-workers/src/types.d.ts
```diff
@@ -0,0 +1,17 @@
+export function createWorker<T>(load: () => Promise<T>): () => Promisify<T>;
+
+type Arguments<T> = T extends (...arg: infer U) => any ? U : any;
+
+export type Promisify<T> = T extends (...args: infer Args) => infer TypeReturned
+  ? TypeReturned extends () => infer Result
+    ? (...args: Args) => () => Promise<Result>
+    : TypeReturned extends Promise<any> // check to prevent returning Promise<Promise<thing>>> in async function
+      ? (...args: Args) => TypeReturned
+      : (...args: Args) => Promise<TypeReturned>
```
lemonmade: I don't think I understand this entire block (from `TypeReturned extends () => ...` down). Shouldn't this type just be applied recursively on the return type (so `() => {foo(): string}` becomes `Promise<() => Promise<{foo(): Promise<string>}>>`)

ismail-syed: This is because of the return types on promisey/async function cases. 

The snippet below is typed as `(() => string | Promise<string>): string | Promise<string>`

```
export function functionArgStringReturn(
  inputFunction: WorkerInput<() => string>,
) {
  const result = inputFunction();
  return typeof result === 'string'
    ? `Hello ${result}`
    : result.then(final => `Hello ${final}`);
}
```

The return type is `string | Promise<string>`. When we augment this type with `Promisify` it gets augmented to `Promise<string> | Promise<Promise<string>>`. This conditional type checks against this `Promise<Promise<string>>` case because we don't want to return the api as `Promise<Promise<string>>`

PR #1087, opened 2019-10-02
Support modern react-apollo

lemonmade:
## Description

This PR fixes support for our GraphQL libraries and newer versions of Apollo. It does so by:

- Removing some version constraints on Apollo and upgrading our development-time version to 3.x (this required some type changes, might make this version of the library type-incompatible with 2.x of Apollo client but should work fine at runtime).
- Deprecate `react-effect-apollo` and remove it from the universal GraphQL serializer. This library relies on internal details no longer present in 3.x, and is not needed with the more fool-proof solution we employ in that library (we mess with the client so we see all GraphQL queries, regardless of where they happen or what API they use).
- Minor documentation improvements

cc/ @amooz


yarn.lock
```diff
@@ -11243,6 +11290,11 @@ tsconfig-paths@^3.6.0:
     minimist "^1.2.0"
     strip-bom "^3.0.0"
 
+tslib@^1.10.0:
```
cartogram: looks like multiple `tslib`s here from the `@apollo/*` dependencies

packages/react-graphql/package.json
```diff
@@ -33,7 +33,7 @@
     "apollo-client": "^2.6.0",
     "apollo-link-http": "^1.5.15",
     "graphql-typed": "^0.4.0",
-    "react-apollo": ">=2.2.3 <3.0.0"
+    "react-apollo": ">=2.2.3 <4.0.0"
```
cartogram: Is the bottom end of this range still supported? (as in, not `>=3.0.0 <4.0.0`)

lemonmade: The library itself is still supported, types are maybe a bit wonky but I can't select a more restrictive dependency for the types themselves :/

PR #1088, opened 2019-10-02
adding enzyme-like .debug() feature to Node instances

patsissons:
## Description

Fixes (issue #) _do we have an issue for this???_

Similar to how [enzyme handles `.debug()`](https://github.com/airbnb/enzyme/blob/master/packages/enzyme/src/Debug.js), we can now emit mounted structure (and sub-structure) using `wrapper.debug()` or `wrapper.find(Foo)!.debug()`.

increase (or decrease) verbosity of the props object expansion by changing the `verbosity` option (default is `1`, `wrapper.debug(3)`). limit how deep to walk the graph by changing the `depth` option (default is undefined, walk the entire graph). Use either position parameters (`wrapper.debug(verbosity, depth)`) or a single options object (`wrapper.debug({depth, verbosity})`)

Test it out in your project using the debug console, `wrapper.findWhere((type) => type && type.name === 'Breadcrumbs')!.debug(5)`

```tsx
<Header breadcrumbs={[{"content": "Settings", "url": "/admin/settings"}]} title="Capital" separator polaris={{"appBridge": undefined, "intl": {"translate": [Function anonymous], "translation": [Object]}, "link": undefined, "scrollLockManager": {"locked": false, "scrollLocks": 0}, "stickyManager": {"handleResize": [Function debounced], "handleScroll": [Function debounced], "stickyItems": [Array], "stuckItems": [Array], "topBarOffset": 0}, "theme": {"logo": null}}}>
  <div>
    <div>
      <div>
        <Breadcrumbs breadcrumbs={[{"content": "Settings", "url": "/admin/settings"}]}>
          <nav role="navigation">
            <Component url="/admin/settings" onMouseUp={[Function handleMouseUpByBlurring]}>
              <Component url="/admin/settings" onMouseUp={[Function handleMouseUpByBlurring]}>
                <a onMouseUp={[Function handleMouseUpByBlurring]} href="/admin/settings">
                  <span>
                    <Icon source={[Function SvgChevronLeftMinor]}>
                      <span>
                        <SvgChevronLeftMinor focusable="false">
                          <svg viewBox="0 0 20 20" focusable="false" />
                        </SvgChevronLeftMinor>
                      </span>
                    </Icon>
                  </span>
                  <span />
                </a>
              </Component>
            </Component>
          </nav>
        </Breadcrumbs>
      </div>
    </div>
    <div>
      <div>
        <Title title="Foo">
          <div>
            <div>
              <div>
                <DisplayText size="large" element="h1">
                  <h1 />
                </DisplayText>
              </div>
            </div>
          </div>
        </Title>
      </div>
    </div>
    <EventListener event="resize" handler={[Function debounced]} passive />
  </div>
</Header>
```

I'm open to changing the name from `.debug()`, but maybe we want to keep it the same (or at least an alias to it) for posterity?

## Type of change

## Checklist

### DIFF
--- a/packages/react-testing/README.md
+++ b/packages/react-testing/README.md
@@ -11,6 +11,7 @@ A library for testing React components according to [Shopify conventions](https:
 1. [Usage](#usage)
    1. [`Root`](#root)
    1. [`Element`](#element)
+   1. [`debug()`](#debug)
    1. [`mount()`](#mount)
    1. [`createMount()`](#createMount)
    1. [`destroyAll()`](#destroyAll)
@@ -292,6 +293,48 @@ const myComponent = mount(
 );
 myComponent.triggerKeypath('action.onAction');
 expect(spy).toHaveBeenCalled();
+```
+
+### <a name="debug"></a> `debug(options?: {allProps?: boolean, depth?: number, verbosity?: number}): string`
+
+Returns a text representation of either the root node, or any element within the mounted graph. `debug()` output can be tweaked using the `options` parameter.
+
+- `allProps` overrides the default props filtering behaviour and instead includes all props in the output, by default `className`, `aria-*`, and `data-*` props are omitted.
+- `depth` defines the number of children printed, by default all children are printed.
+- `verbosity` defines the level of expansion that non-scalar props experience, the default value of `1` will expand objects one level deep
+
+Typical usage should not require providing any options as the default `verbosity` and `depth` should be appropriate for the majority of inspections.
+
+```tsx
+function ObjectText({data}: {data: {}}) {
+  return <span>{JSON.stringify(data)}</span>;
+}
+
+function Container({children}: PropsWithChildren<{}>) {
+  return children;
+}
+
+function MyComponent({onClick}: {onClick(id: string): void}) {
+  return (
+    <Container>
+      <button type="button" onClick={() => onClick(String(Math.random()))}>
+        <ObjectText data={{a: {very: {deep: {data: {object: 'with text'}}}}}} />
+      </button>
+    </Container>
+  );
+}
+
+const wrapper = mount(<MyComponent />);
+// print the whole structure with one level of prop verbosity
+console.log(wrapper.debug());
+// print only the Container and button without any other children
+console.log(wrapper.find(Container)!.debug({depth: 1}));
+// find button by name and print all children with verbose props details
+console.log(
+  wrapper
+    .findWhere(type => type && type.name === 'button')!
+    .debug({verbosity: 9}),
+);
 ```
 
 ### <a name="mount"></a> `mount(element: React.ReactElement<any>)`
--- a/packages/react-testing/src/element.ts
+++ b/packages/react-testing/src/element.ts
@@ -3,6 +3,7 @@ import {
   Arguments,
   MaybeFunctionReturnType as ReturnType,
 } from '@shopify/useful-types';
+import {nodeName, toReactString} from './toReactString';
 import {
   Tag,
   Node,
@@ -10,6 +11,7 @@ import {
   FunctionKeys,
   DeepPartialArguments,
   PropsFor,
+  DebugOptions,
 } from './types';
 
 type Root = import('./root').Root<any>;
@@ -220,17 +222,12 @@ export class Element<Props> implements Node<Props> {
     });
   }
 
+  debug(options?: DebugOptions) {
+    return toReactString(this, options);
+  }
+
   toString() {
-    const {type} = this;
-
-    if (type == null) {
-      return '<Element />';
-    }
-
-    const name =
-      typeof type === 'string' ? type : type.displayName || type.name;
-
-    return `<${name} />`;
+    return `<${nodeName(this)} />`;
   }
 }
 
--- a/packages/react-testing/src/types.ts
+++ b/packages/react-testing/src/types.ts
@@ -112,5 +112,12 @@ export interface Node<Props> {
   ): MaybeFunctionReturnType<NonNullable<Props[K]>>;
   triggerKeypath<T = unknown>(keypath: string, ...args: unknown[]): T;
 
+  debug(options?: DebugOptions): string;
   toString(): string;
+}
+
+export interface DebugOptions {
+  allProps?: boolean;
+  depth?: number;
+  verbosity?: number;
 }
--- a/packages/react-testing/CHANGELOG.md
+++ b/packages/react-testing/CHANGELOG.md
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## [Unreleased] -->
+## [Unreleased]
+
+### Added
+
+- new `debug()` function added to `Root` and `Element` to inspect mounted structure ([#1088](https://github.com/Shopify/quilt/pull/1088))
 
 ## [1.7.10] - 2019-09-30
 
--- a/packages/react-testing/src/index.ts
+++ b/packages/react-testing/src/index.ts
@@ -3,3 +3,4 @@ export {Element} from './element';
 export {Node} from './types';
 export * from './mount';
 export * from './destroy';
+export * from './toReactString';
--- a/packages/react-testing/src/root.tsx
+++ b/packages/react-testing/src/root.tsx
@@ -17,6 +17,7 @@ import {
   FunctionKeys,
   DeepPartialArguments,
   PropsFor,
+  DebugOptions,
 } from './types';
 
 // Manually casting `act()` until @types/react is updated to include
@@ -241,6 +242,11 @@ export class Root<Props> implements Node<Props> {
   forceUpdate() {
     this.ensureRoot();
     this.act(() => this.wrapper!.forceUpdate());
+  }
+
+  debug(options?: DebugOptions) {
+    this.ensureRoot();
+    return this.root!.debug(options);
   }
 
   toString() {
--- a/packages/react-testing/src/toReactString.ts
+++ b/packages/react-testing/src/toReactString.ts
@@ -1,0 +1,112 @@
+import {stringify} from 'jest-matcher-utils';
+import {DebugOptions, Node} from './types';
+
+export function toReactString<Props>(
+  node: Node<Props>,
+  options: DebugOptions = {},
+  level = 0,
+) {
+  // if this is an array node then print all children at the current level
+  if (!node.type && node.children.length > 0) {
+    return node.children
+      .map(child => toReactString(child, options, level))
+      .join('\n');
+  }
+
+  const name = nodeName(node);
+  const indent = '  '.repeat(level);
+  const props = Object.keys(node.props)
+    // we always filter out children no matter what, but unless allProps option
+    // is present we will also filter out insigificant props
+    .filter(
+      key =>
+        options.allProps
+          ? key !== 'children'
+          : !/^(children|className)$|^(aria|data)-/.test(key),
+    )
+    .reduce(
+      (list, key) => {
+        if (!key) {
+          return list;
+        }
+
+        const value = node.props[key];
+
+        if (value === undefined && !options.allProps) {
+          return list;
+        }
+
+        list.push(toPropString(key, value, options.verbosity));
+
+        return list;
+      },
+      [] as string[],
+    );
+  const hasChildren = node.children.length > 0 && !['svg'].includes(name);
+  const open = `${indent}<${name}${
+    props.length > 0 ? ` ${props.join(' ')}` : ''
+  }${hasChildren ? '>' : ' />'}`;
+
+  if (!hasChildren) {
+    return open;
+  }
+
+  const close = `${indent}</${name}>`;
+
+  if (options.depth != null && level >= options.depth) {
+    return [
+      open,
+      `${indent}  {/* <${node.children.length} child${
+        node.children.length === 1 ? '' : 'ren'
+      }... /> */}`,
+      close,
+    ].join('\n');
+  }
+
+  return [
+    open,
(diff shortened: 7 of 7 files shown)


packages/react-testing/src/types.ts
```diff
@@ -112,5 +112,6 @@ export interface Node<Props> {
   ): MaybeFunctionReturnType<NonNullable<Props[K]>>;
   triggerKeypath<T = unknown>(keypath: string, ...args: unknown[]): T;
 
+  debug(maxDepth?: number): string;
```
patsissons: could also be handy to have an `omitChildren` flag.

patsissons: also `maxDepth` is a poor choice for a name, it really represents prop object expansion depth. perhaps a better approach is `{propsDepth?: number; childDepth?: number}`?

BPScott: I'm not sure when I'd want those two to be different values, or if i did that I'd remember that as an option. A single value for both feels more intuitive.

patsissons: i can see some legit use cases.
> I want to see max detail for all props on a single component with no children
`wrapper.findWhere(({type}) => type && type.name === 'Header').debug({verbosity: 5, depth: 0})`
```tsx
<Header
  breadcrumbs={[{content: 'Settings', url: '/admin/settings'}]}
  title="Capital"
  separator
  polaris={{
    appBridge: undefined,
    intl: {
      translate: ['Function anonymous'],
      translation: {
        Polaris: {
          ActionMenu: {RollupActions: {rollupButton: 'Actions'}},
          Autocomplete: {spinnerAccessibilityLabel: 'Loading'},
          Avatar: {
            label: 'Avatar',
            labelWithInitials: 'Avatar with initials {initials}',
          },
          Badge: {
            PROGRESS_LABELS: {
              complete: 'Complete',
              incomplete: 'Incomplete',
              partiallyComplete: 'Partially complete',
            },
            STATUS_LABELS: {
              attention: 'Attention',
              info: 'Info',
              new: 'New',
              success: 'Success',
              warning: 'Warning',
            },
          },
          Button: {spinnerAccessibilityLabel: 'Loading'},
          Common: {
            cancel: 'Cancel',
            checkbox: 'checkbox',
            clear: 'Clear',
            close: 'Close',
            more: 'More',
            newWindowAccessibilityHint: '(opens a new window)',
            submit: 'Submit',
            undo: 'Undo',
          },
          ContextualSaveBar: {discard: 'Discard', save: 'Save'},
          DataTable: {
            navAccessibilityLabel:
              'Scroll table {direction} one column',
            sortAccessibilityLabel: 'sort {direction} by',
            totalsRowHeading: 'Totals',
          },
          DatePicker: {
            daysAbbreviated: {
              friday: 'Fr',
              monday: 'Mo',
              saturday: 'Sa',
              sunday: 'Su',
              thursday: 'Th',
              tuesday: 'Tu',
              wednesday: 'We',
            },
            months: {
              april: 'April',
              august: 'August',
              december: 'December',
              february: 'February',
              january: 'January',
              july: 'July',
              june: 'June',
              march: 'March',
              may: 'May',
              november: 'November',
              october: 'October',
              september: 'September',
            },
            nextMonth: 'Show next month, {nextMonth} {nextYear}',
            previousMonth:
              'Show previous month, {previousMonthName} {showPreviousYear}',
            today: 'Today ',
          },
          DiscardConfirmationModal: {
            message:
              'If you discard changes, you’ll delete any edits you made since you last saved.',
            primaryAction: 'Discard changes',
            secondaryAction: 'Continue editing',
            title: 'Discard all unsaved changes',
          },
          DropZone: {
            FileUpload: {
              actionHintFile: 'or drop files to upload',
              actionHintImage: 'or drop images to upload',
              actionTitleFile: 'Add file',
              actionTitleImage: 'Add image',
              label: 'Upload file',
            },
            errorOverlayTextFile: 'File type is not valid',
            errorOverlayTextImage: 'Image type is not valid',
            overlayTextFile: 'Drop file to upload',
            overlayTextImage: 'Drop image to upload',
          },
          EmptySearchResult: {altText: 'Empty search results'},
          Filters: {
            cancel: 'Cancel',
            clear: 'Clear',
            clearAllFilters: 'Clear all filters',
            clearLabel: 'Clear {filterName}',
            done: 'Done',
            filter: 'Filter {resourceName}',
            moreFilters: 'More filters',
            noFiltersApplied: 'No filters applied',
          },
          Frame: {
            Navigation: {
              closeMobileNavigationLabel: 'Close navigation',
            },
            skipToContent: 'Skip to content',
          },
          Icon: {
            backdropWarning:
              'The {color} icon doesn’t accept backdrops. The icon colors that have backdrops are: {colorsWithBackDrops}',
          },
          Modal: {
            iFrameTitle: 'body markup',
            modalWarning:
              'These required properties are missing from Modal: {missingProps}',
          },
          Pagination: {
            next: 'Next',
            pagination: 'Pagination',
            previous: 'Previous',
          },
          ProgressBar: {
            exceedWarningMessage:
              'Values passed to the progress prop shouldn’t exceed 100. Setting {progress} to 100.',
            negativeWarningMessage:
              'Values passed to the progress prop shouldn’t be negative. Resetting {progress} to 0.',
          },
          ResourceList: {
            BulkActions: {
              actionsActivatorLabel: 'Actions',
              moreActionsActivatorLabel: 'More actions',
              warningMessage:
                'To provide a better user experience. There should only be a maximum of {maxPromotedActions} promoted actions.',
            },
            DateSelector: {
              FilterLabelForValue: {
                coming_month: 'next month',
                coming_quarter: 'in the next 3 months',
                coming_week: 'next week',
                coming_year: 'in the next year',
                on_or_after: 'after {date}',
                on_or_before: 'before {date}',
                past_month: 'in the last month',
                past_quarter: 'in the last 3 months',
                past_week: 'in the last week',
                past_year: 'in the last year',
              },
              SelectOptions: {
                ComingMonth: 'next month',
                ComingQuarter: 'in the next 3 months',
                ComingWeek: 'next week',
                ComingYear: 'in the next year',
                OnOrAfter: 'on or after',
                OnOrBefore: 'on or before',
                PastMonth: 'in the last month',
                PastQuarter: 'in the last 3 months',
                PastWeek: 'in the last week',
                PastYear: 'in the last year',
              },
              dateFilterLabel: 'Select a value',
              dateValueError: 'Match YYYY-MM-DD format',
              dateValueLabel: 'Date',
              dateValuePlaceholder: 'YYYY-MM-DD',
            },
            FilterControl: {
              textFieldLabel: 'Search {resourceNamePlural}',
            },
            FilterCreator: {
              addFilterButtonLabel: 'Add filter',
              filterButtonLabel: 'Filter',
              selectFilterKeyPlaceholder: 'Select a filter…',
              showAllWhere: 'Show all {resourceNamePlural} where:',
            },
            FilterValueSelector: {
              selectFilterValuePlaceholder: 'Select a filter…',
            },
            Item: {
              actionsDropdown: 'Actions dropdown',
              actionsDropdownLabel:
                'Actions for {accessibilityLabel}',
              viewItem: 'View details for {itemName}',
            },
            a11yCheckboxDeselectAllMultiple:
              'Deselect all {itemsLength} {resourceNamePlural}',
            a11yCheckboxDeselectAllSingle:
              'Deselect {resourceNameSingular}',
            a11yCheckboxSelectAllMultiple:
              'Select all {itemsLength} {resourceNamePlural}',
            a11yCheckboxSelectAllSingle:
              'Select {resourceNameSingular}',
            allItemsSelected:
              'All {itemsLength}+ {resourceNamePlural} in your store are selected.',
            ariaLivePlural: '{itemsLength} items',
            ariaLiveSingular: '{itemsLength} item',
            defaultItemPlural: 'items',
            defaultItemSingular: 'item',
            emptySearchResultDescription:
              'Try changing the filters or search term',
            emptySearchResultTitle: 'No {resourceNamePlural} found',
            loading: 'Loading {resource}',
            selectAllItems:
              'Select all {itemsLength}+ {resourceNamePlural} in your store',
            selectButtonText: 'Select',
            selected: '{selectedItemsCount} selected',
            showing: 'Showing {itemsCount} {resource}',
            sortingLabel: 'Sort by',
          },
          SkeletonPage: {loadingLabel: 'Page loading'},
          Spinner: {
            warningMessage:
              'The color {color} is not meant to be used on {size} spinners. The colors available on large spinners are: {colors}',
          },
          Tabs: {toggleTabsLabel: 'More tabs'},
          Tag: {ariaLabel: 'Remove {children}'},
          TextField: {
            characterCount: '{count} characters',
            characterCountWithMaxLength:
              '{count} of {limit} characters used',
          },
          TopBar: {
            SearchField: {
              clearButtonLabel: 'Clear',
              search: 'Search',
            },
            toggleMenuLabel: 'Toggle menu',
          },
        },
      },
    },
    link: undefined,
    scrollLockManager: {locked: false, scrollLocks: 0},
    stickyManager: {
      handleResize: ['Function debounced'],
      handleScroll: ['Function debounced'],
      stickyItems: [],
      stuckItems: [],
      topBarOffset: 0,
    },
    theme: {logo: null},
  }}
>
  {/* <1 child... /> */}
</Header>
```

BPScott: Ah that makes more sense, thanks! I think i'd go for a 2 args approach with some sensible-ish rather than an object. Seeing children `.debug(childrenMaxDepth = 5, propsMaxDepth = 1)`.


It's at this point that I mutter something something Polaris being on a warpath to kill our WithAppProvider component (the only HoC remaining in v4), and [we're down to <15 usages](https://github.com/Shopify/polaris-react/issues/1995), but that won't help the reset of web being very HoC happy.

packages/react-testing/src/element.ts
```diff
@@ -220,18 +222,120 @@ export class Element<Props> implements Node<Props> {
     });
   }
 
+  debug(verbosity?: number, depth?: number): string;
+  debug(options?: DebugOptions): string;
+  debug(options: number | DebugOptions = {}, depth?: number) {
```
lemonmade: Why have two options here?

packages/react-testing/src/element.ts
```diff
@@ -220,18 +222,120 @@ export class Element<Props> implements Node<Props> {
...
+    );
+  }
+
   toString() {
-    const {type} = this;
+    return `<${elementName(this)} />`;
+  }
+}
+
+function printElement(element: Element<any>, options: DebugOptions, level = 0) {
```
lemonmade: Is there no existing serializer for a react element we could use?

packages/react-testing/README.md
```diff
@@ -294,6 +295,34 @@ myComponent.triggerKeypath('action.onAction');
...
+  );
+}
+
+const wrapper = mount(<MyComponent />);
+// return the whole structure with one level of prop verbosity
+wrapper.debug();
+// return only the div and button without any other children
+wrapper.find(div)!.debug({depth: 1});
+// find by name and return all children with verbose props details
+wrapper.findWhere(type => type && type.name === 'button')!.debug(9);
```
lemonmade: These examples are kind of weird because they just produce strings but don’t log them or anything :/

packages/react-testing/README.md
```diff
@@ -294,6 +295,34 @@ myComponent.triggerKeypath('action.onAction');
 expect(spy).toHaveBeenCalled();
 ```
 
+### <a name="debug"></a> `debug(): string`
+
+Returns a text representation of either the root node, or any element within the mounted graph. `debug()` has two forms, `debug(verbosity?: number = 1, depth?: number)` for simple usage and `debug({verbosity?: number, depth?: number})` for more direct output control. `verbosity` defines the level of expansion that non-scalar props experience, the default value of `1` will expand objects for example one level deep. `depth` defines the number of children printed, by default all children are printed. Typical usage should not require any parameters as the default `verbosity` and `depth` should be appropriate for the majority of inspections.
```
cartogram: I guess most people will just call `debug()` and use the defaults as you say, but I found this pretty hard to parse with such similar arguments (It also doesn't help that the arguments can be passed in a few different ways). Do you think it might be worth simplifying this whole thing?

packages/react-testing/src/element.ts
patsissons: /shrug i wanted to avoid having to write `debug(undefined, 0)`

packages/react-testing/README.md
patsissons: well i guess the intent is that you run this in the debug execution console, so the result is automatically printed. i could wrap with `console.log` for more clarity

packages/react-testing/README.md
patsissons: maybe just the options object only then?

cartogram: I think thats the safe bet, yeah

packages/react-testing/src/element.ts
lemonmade: `debug({depth: 0})` is literally fewer characters than that and is not inscrutable...

packages/react-testing/src/element.ts
patsissons: if there is i didn't encounter one

packages/react-testing/src/element.ts
patsissons: i'll 🔥 the positional params

packages/react-testing/src/element.ts
patsissons: looks like are a few options:

- [`react-element-to-jsx-string`](https://github.com/algolia/react-element-to-jsx-string)
- [`pretty-format`](https://github.com/facebook/jest/tree/master/packages/pretty-format) from `jest`
- `React.renderToStaticMarkup(...)`

any of these could work, but really there isn't much code here so i'm not sure how interested i am in pulling in another dependency. the `renderToStaticMarkup` is probably a bit basic, `pretty-format` could be reasonable though if it's already present in the package dependences?

patsissons: i tried out `pretty-format` and it was a no go, we would need to change how we are rendering nodes which isn't really in the scope of this feature. The other options i don't feel make sense to attempt.

PR #1092, opened 2019-10-04
Initial @shopify/web-worker package

lemonmade:
This PR adds the initial set of pieces for the `@shopify/web-worker` package. This includes:

1. The actual API app code uses
1. A babel plugin that changes `createWorker(() => import())` into a call to the Webpack loader
1. A webpack plugin that does some things I couldn't figure out how to do in the loader alone (using `webpack-virtual-modules` for creating magic entry points for the workers)
1. A webpack loader that turns a require for a worker into a child compilation that outputs its assets, and returns the results of the build (the entry)

It also includes two E2E tests, which feel appropriate here since the individual parts of this are kind of not that relevant.

More tests/ docs to come, and this isn't the final implementation, just a base we can start to work off of.


packages/web-worker/src/test/e2e.test.ts
```diff
@@ -0,0 +1,138 @@
+/**
+ * @jest-environment node
```
ismail-syed: TIL

package.json
```diff
@@ -61,6 +64,7 @@
     "lerna": "^2.9.0",
     "plop": "^2.0.0",
     "prettier": "^1.14.0",
+    "puppeteer": "^1.20.0",
```
ismail-syed: Surpised we weren't already using this in quilt. Nice

packages/web-worker/src/webpack-parts/loader.ts
```diff
@@ -0,0 +1,124 @@
...
+  const callback = this.async();
+
+  const {
+    context,
+    resourcePath,
+    _compiler: compiler,
+    _compilation: compilation,
+  } = this;
+
+  // if (compiler.options.output.globalObject !== 'self') {
```
ismail-syed: 🔥 ?

packages/web-worker/src/test/e2e.test.ts
```diff
@@ -0,0 +1,138 @@
...
+ */
+
+import * as path from 'path';
+import {WebWorkerPlugin} from '../webpack-parts';
+import {withContext, Context, runWebpack as runWebpackBase} from './utilities';
```
ismail-syed: Some really nice testing utilities here! Maybe this could be its own package one-day.

packages/web-worker/src/index.ts
```diff
@@ -0,0 +1,2 @@
+export {expose} from './worker';
+export {createWorker} from './parent';
```
ismail-syed: Is `parent` the right term to use here. In an application, the UI thread spawns a worker thread so the application thread would be the parent. Unless this is a different context that is. 

Maybe just call it `create-worker.ts`?

packages/web-worker/src/webpack-parts/loader.ts
```diff
@@ -0,0 +1,124 @@
+import * as path from 'path';
+
+import SingleEntryPlugin from 'webpack/lib/SingleEntryPlugin';
+import WebWorkerTemplatePlugin from 'webpack/lib/webworker/WebWorkerTemplatePlugin';
```
ismail-syed: Ouu, this is neat.

packages/web-worker/src/webpack-parts/loader.ts
```diff
@@ -0,0 +1,124 @@
...
+import FetchCompileWasmTemplatePlugin from 'webpack/lib/web/FetchCompileWasmTemplatePlugin';
```
ismail-syed: This is interesting, what this wasm template for?

packages/web-worker/src/webpack-parts/loader.ts
```diff
@@ -0,0 +1,124 @@
...
+      return callback!(
+        null,
+        `export default __webpack_public_path__ + ${JSON.stringify(entry)};`,
+      );
+    },
+  );
+}
+
+function addWorkerSubExtension(file: string) {
+  return file.replace(/\.([a-z]+)$/i, '.worker.$1');
```
ismail-syed: It be nice if `WebWorkerTemplatePlugin` handled this for us.

packages/web-worker/tsconfig.build.json
```diff
@@ -0,0 +1,15 @@
+{
+  "extends": "../../tsconfig.json",
+  "compileOnSave": false,
+  "compilerOptions": {
+    "outDir": "./dist",
+    "baseUrl": "./src",
+    "rootDir": "./src"
+  },
+  "include": [
+    "../../config/typescript/**/*",
```
ismail-syed: Are we pulling in anything from here? If not we can 🔥

packages/web-worker/src/parent.ts
```diff
@@ -0,0 +1,51 @@
...
+          },
+        },
+      ) as any;
+    }
+
+    const workerScript = URL.createObjectURL(
+      new Blob([`importScripts(${JSON.stringify(script)})`]),
+    );
+
+    const worker = new Worker(workerScript);
```
ismail-syed: Do we have any ideas on how/when we'll want to `terminate` the worker?

lemonmade: I think it will just be part of the API for `createWorker()`? Not positive yet, but I think it's mostly up the consumer to terminate.

packages/web-worker/tsconfig.build.json
lemonmade: It's part of the template, I feel like it would be weird to remove

packages/web-worker/src/webpack-parts/loader.ts
lemonmade: It would be kind of weird for it to do that IMO, Webpack is not very opinionated about the filename and this is purely a "pretty filename" change.

packages/web-worker/src/webpack-parts/loader.ts
lemonmade: This was from the https://github.com/GoogleChromeLabs/worker-plugin plugin, where I got a lot of the implementation from. It just allows you to import `.wasm` files IIRC, which is a good approach for things in workers.

packages/web-worker/src/index.ts
lemonmade: Hmm, I feel like it's appropriate, because `createWorker` is for the application, which is the "parent" in the relationship. In the end it doesn't really matter much IMO because this is just so that they are in separate files for tree shaking.

packages/web-worker/src/webpack-parts/loader.ts
lemonmade: I left this in as a reminder to myself that I still have to figure out the right approach for these scripts on the server render :P

Issue #1044, opened 2019-09-25
Add basic API for creating a worker from an async import


lemonmade:
**Note: this task is just about getting the typings right, not making the whole thing actually work**

The API we want to expose for a worker is basically as close as possible to just using an async imported module. That is, its available methods should be able to be called directly with proper typings on arguments. The only way I can think to do this is to pass an actual async import of the module to a function:

```ts
// foo.ts
export function foo(count: number) {
  return `Count is ${count}`;
}

// bar.ts
const createWorker = workerize(() => import('./foo'));
const worker = await createWorker(); // need to wait for Worker to initialize

const result = await worker.foo(42); // all methods will need to be async because they go to a worker
```

As noted above, all method signatures on the resulting worker will need to be converted to be return a promise:

```tsx
// worker...
function foo(): string {}

const result: Promise<string> = worker.foo();
```

Similarly, any argument that gets passed to the worker will need to have any function portions be converted to always return promises, and we should provide a helper type to make that work:

```ts
// in worker...
// this is not possible, because the only argument that can be passed 
function foo(arg: {bar(): string}) {
  return `Message: ${arg.bar()}`;
}

// A helper type that knows to turn all functions into async function will reflect how the final implementation can work:
async function foo(arg: WorkerInput<{bar(): string}>) {
  return `Message: ${await arg.bar()}`;
}
```

The reverse operation can happen at the call site; if the worker accepts a function that returns `Promise<string>`, the person calling the worker can provide a function that returns a `string` directly, because once it is communicated over the worker, it will be in promise form.

PR #1094, opened 2019-10-04
Add type safe APIs

ismail-syed:
## Description

Fixes https://github.com/Shopify/quilt/issues/1044

* Add and use a `PromisifyModule` type in `createWorker` to return a type-safe API for the created web worker api. 
* Exposes a `SafeWorkerArgument` type to be used by consumers to ensure type safe inputs as mentioned in the [docs]https://github.com/Shopify/quilt/tree/master/packages/web-worker#worker)

**`createWorker` with a _promisified_ module:**

Screen Shot 2019-10-04 at 4 03 13 PM

### DIFF
--- a/packages/web-worker/src/parent.ts
+++ b/packages/web-worker/src/parent.ts
@@ -1,4 +1,8 @@
-export function createWorker<T>(script: () => Promise<T>): () => T {
+import {PromisifyModule} from './types';
+
+export function createWorker<T>(
+  script: () => Promise<T>,
+): () => PromisifyModule<T> {
   return function create() {
     if (typeof Worker === 'undefined') {
       return new Proxy(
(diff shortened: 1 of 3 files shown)


packages/web-worker/src/types.ts
```diff
@@ -0,0 +1,31 @@
+export type PromisifyModule<T> = {[K in keyof T]: PromisifyExport<T[K]>};
```
ismail-syed: I personally found [these _tests_](https://github.com/Shopify/quilt/pull/1084/files#diff-b4a026640b4aa57e3183d2c501452039R38) very helpful. @lemonmade, not sure if you want to keep them or not.

packages/web-worker/src/parent.ts
```diff
@@ -1,4 +1,8 @@
-export function createWorker<T>(script: () => Promise<T>): () => T {
+import {PromisifyModule} from './types';
+
+export function createWorker<T>(
+  script: () => Promise<T>,
```
lemonmade: script is a weird name here

PR #1096, opened 2019-10-04
Actually support custom plugins for worker

lemonmade:
## Description

Fixes (issue #)

## Type of change

## Checklist

### DIFF
--- a/packages/react-server-webpack-plugin/package.json
+++ b/packages/react-server-webpack-plugin/package.json
@@ -30,7 +30,7 @@
   "devDependencies": {
     "@shopify/react-server": "^0.7.3",
     "@types/webpack-virtual-modules": "^0.1.0",
-    "webpack": "^4.39.1"
+    "webpack": "^4.0.0"
   },
   "files": [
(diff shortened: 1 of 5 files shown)


packages/react-server-webpack-plugin/package.json
```diff
@@ -30,7 +30,7 @@
   "devDependencies": {
     "@shopify/react-server": "^0.7.3",
     "@types/webpack-virtual-modules": "^0.1.0",
-    "webpack": "^4.39.1"
+    "webpack": "^4.0.0"
```
ismail-syed: These both mean the same thing. Is this just a style thing?

ismail-syed: Or for peer-deps to resolve to at least `4.0.0`

lemonmade: Yeah, they are different because version 4.0–4.38 satisfies this but not the old range. In dev dependencies it doesn't matter much but in general having looser version restrictions for library code is desirable

Issue #1101, opened 2019-10-08
Include `Performance` instance creation boilerplate in `@shopify/react-performance`


TheMallen:
Something like the following is necessary in every use of `@shopify/react-performance`.

```tsx
import React from 'react';
import {useLazyRef} from '@shopify/react-hooks';

import {
  PerformanceContext,
  Performance as PerformanceManager,
  NavigationListener,
} from '@shopify/react-performance';

export function PerformanceProvider({children, isServer}: {children: React.ReactNode, isServer: boolean}) {
  const performance = useLazyRef(() => {
    if (isServer) {
      return undefined;
    }
    return new PerformanceManager();
  });

  return (
    <PerformanceContext.Provider value={performance.current}>
      {children}
    </PerformanceContext.Provider>
  );
}

```

We should just provide an enhanced standalone `PerformanceProvider` component that handles this boilerplate for the user.

PR #1102, opened 2019-10-08
✨ have a provider that create performance manager by default

michenly:
## Description

Fixes (issue #) https://github.com/Shopify/quilt/issues/1101

## Type of change

## Checklist


michenly:
Can't think of options we would add options we would add to the performance object in the future so this make sense!

michenly:
Done.
This pattern honestly feels kinda funny. But let's all look at it together.

packages/react-performance/README.md
```diff
@@ -17,30 +17,9 @@ $ yarn add @shopify/react-performance
 
 The most basic way to use the tools in this package is to record information and display it locally to the user. In practice you usually only want to do this in development so that developers can easily see performance information.
 
-#### Add the context provider
-
-Before we can use any of the components or hooks in the package we must wrap our app tree with the `PerformanceContext.Provider` component.
```
TheMallen: Oh weird we don't need this at all anymore?

michenly: Ya we don't because performance object is created by default.
This *feels* weired but work pretty nicely when I :tophat:

packages/react-performance/src/context.ts
```diff
@@ -1,6 +1,8 @@
 import React from 'react';
 import {Performance} from '@shopify/performance';
 
-export const PerformanceContext = React.createContext<Performance | undefined>(
-  undefined,
+import {isServer} from './utilities';
+
+export const PerformanceContext = React.createContext<Performance | null>(
+  isServer() ? null : new Performance(),
```
lemonmade: personally I prefer not having utilities for this kind of thing, because it's something we may want to some day optimize in build, and it's very hard to do that when programatic APIs wrap the statically-analyzable bits.

packages/react-performance/src/test/performance-effect-server.test.tsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
```
lemonmade: this test will probably fail eventually, if JSDom ever adds any API for `performance`. It's best to set the `jest-env` here to ensure you get a clean, `performance`-free place to run the test. Separately, not a huge fan of the name of the test not matching the name of the file it tests.

packages/react-performance/src/test/performance-effect-server.test.tsx
```diff
@@ -0,0 +1,37 @@
...
+jest.mock('@shopify/performance', () => {
+  return {
+    ...require.requireActual('@shopify/performance'),
+    Performance: jest.fn(),
+  };
+});
+
+jest.mock('../utilities', () => ({
+  isServer: () => true,
+}));
```
lemonmade: If this was the only reason for the function, you can definitely avoid this by setting `jest-env` :)

packages/react-performance/src/test/performance-effect-server.test.tsx
```diff
@@ -0,0 +1,37 @@
...
+  PerformanceEffectCallback,
+  PerformanceContext,
+} from '..';
+
+jest.mock('@shopify/performance', () => {
+  return {
+    ...require.requireActual('@shopify/performance'),
+    Performance: jest.fn(),
+  };
+});
```
lemonmade: Do you need these mocks?

packages/react-performance/src/test/performance-effect-server.test.tsx
michenly: it is the only reason 😅 
I tried doing it otherwise with environment setting for the whole file then `react-testing/mount` would fail because in the test wrapper uses `document`

packages/react-performance/src/test/performance-effect-server.test.tsx
michenly: I doooo, calling `new Performance ()` would fail otherwise on withTiming line where the warp method is not getting a valid object.

packages/react-performance/src/test/performance-effect-server.test.tsx
michenly: IMO its either we do a strange workaround like this or skip the test all together

lemonmade: Yeah we can't use `react-testing`, but it doesn't really make sense to do so anyways, because we are actually testing server behaviour, and react-testing is really react-dom-testing. I think it would be appropriate to use react's server API here (`renderToString`) instead, should still be able to keep the rest of the change relatively unchanged.

PR #1106, opened 2019-10-10
Return a properly typed web worker api

ismail-syed:
## Description

Returns a properly typed module for a web worker that we communicate with via our new bridge,  https://github.com/Shopify/quilt/pull/1104.

Tested in web-proving-ground. The worker types get augmented as expected. 

Screen Shot 2019-10-10 at 12 26 41 PM

### DIFF
--- a/packages/web-worker/src/create.ts
+++ b/packages/web-worker/src/create.ts
@@ -1,5 +1,4 @@
 import {createEndpoint, Endpoint} from './endpoint';
-import {PromisifyExport} from './types';
 
 const workerEndpointCache = new WeakMap<Endpoint<any>['call'], Endpoint<any>>();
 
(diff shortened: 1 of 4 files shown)


packages/web-worker/src/create.ts
```diff
@@ -14,9 +15,9 @@ export function getEndpoint(caller: Endpoint<any>['call']) {
   return workerEndpointCache.get(caller);
 }
 
-export function createWorker<T extends {[key: string]: () => Promise<any>}>(
+export function createWorker<T extends {[K in keyof T]: T[K]}>(
```
lemonmade: I don't think the `extends` here is doing anything, should just remove

packages/web-worker/src/create.ts
```diff
@@ -14,9 +15,9 @@ export function getEndpoint(caller: Endpoint<any>['call']) {
...
   script: () => Promise<T>,
-) {
+): () => {[K in keyof T]: PromisifyExport<T[K]>} {
```
lemonmade: Hmm, I'd like to try and just rely on the inferred type of `create()` if possible. Can `create` just return `Endpoint<T>['call']` still, and we update `Endpoint` to `PromisifyModule<T>`?

ismail-syed: This was actually how I approached it first, by having `Endpoint['call']` infer as `PromisifyModule<T>`. When I tested it out in web-proving-ground, I got `Property '[some_function_name]' does not exist on type 'PromisifyModule<[some_module_path]>`

ismail-syed:
Got a verbal go-ahead from Chris, merging.

PR #1107, opened 2019-10-11
cacheEffectiveness now uses duration=0 as a cache hit

GoodForOneFare:
## Description

Now that CDN metrics are available, it looks like `duration === 0` is a better indicator of cache hits:

Chrome development tools screencap showing a navigation timing event containing a 0 duration, and non-0 byte count

Longer term, I'm tempted to just log bytes cached and let Splunk math on `cached` vs `totalDownloadSize` do the rest.

## Type of change
You could argue that this is a breaking change.  IMO, this is just fixing behaviour that maybe never worked, so minor.

- [x] `@shopify/performance` Patch: Bug/ Documentation fix

## Checklist

### DIFF
--- a/packages/performance/CHANGELOG.md
+++ b/packages/performance/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
+- `cacheEffectiveness` now assumes duration=0 is a cache hit [[#1107](https://github.com/Shopify/quilt/pull/1107)]
+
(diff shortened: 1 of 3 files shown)


packages/performance/CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
+- `cacheEffectiveness` now assumes duration=0 is a cache hit [[#1107](https://github.com/Shopify/quilt/pull/1107)]
```
GoodForOneFare: As noted in the PR description, this could be a major change, but I think it's more of a "fixing something that maybe never worked" thing.

lemonmade: Definitely a patch fix IMO

Issue #1108, opened 2019-10-11
Make `assetPrefix` option configurable in `sewing-kit-koa`


tgxworld:
## Overview

In order to [improve the duration/reliability of load time](https://vault.shopify.io/projects/9226) for users connecting from within China, we've setup `cdn.shopify.cn` which can resolve to our Tencent CDN so as to reduce the number of requests that has to cross the firewall. Since we now have `cdn.shopify.com` and `cdn.shopify.cn`, the middleware exported by `sewing-kit-koa` needs to allow the `assetPrefix` variable to be configurable for each request. I'm thinking we could add a `assetPrefixCallback` option when creating the middleware which will then be called during each request.

## Type

- [ ] New feature
- [X] Changes to existing features

## Motivation

> What inspired this feature request? What problems were you facing,
> or what else makes you think this should be in `quilt`?

See overview.

## Checklist

- [X] Please delete the labels section before submitting your issue
- [X] I have described this issue in a way that is actionable (if possible)

tgxworld:
Closing this for now as we're taking a difference approach which limits the logic to `shopify/web` itself.

PR #1109, opened 2019-10-11
Add `assetPrefixCallback` to sewing-kit-koa middleware option

tgxworld:
## Description

Fixes https://github.com/Shopify/quilt/issues/1108

## Type of change
Minor: New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/packages/sewing-kit-koa/src/middleware.ts
+++ b/packages/sewing-kit-koa/src/middleware.ts
@@ -12,8 +12,7 @@ import Assets, {Asset} from './assets';
 export {Assets, Asset};
 
 export interface Options {
-  assetPrefix?: string;
-  assetPrefixCallback?: Function;
+  assetPrefix?: string | ((ctx: Context) => string);
   serveAssets?: boolean;
   manifestPath?: string;
 }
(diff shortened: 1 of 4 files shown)


packages/sewing-kit-koa/src/middleware.ts
```diff
@@ -13,6 +13,7 @@ export {Assets, Asset};
 
 export interface Options {
   assetPrefix?: string;
+  assetPrefixCallback?: Function;
```
lemonmade: instead of this, I think we'd usually prefer to augment the existing API. I think this would work OK if you made `assetPrefix` be `assetPrefix?: string | (ctx: Context) => string;`

lemonmade: This helps avoid the awkward "but what does this mean" situation where both `assetPrefix` and `assetPrefixCallback` are specified.

packages/sewing-kit-koa/src/middleware.ts
```diff
@@ -30,11 +31,20 @@ export function setAssets(ctx: Context, assets: Assets) {
...
+
+    if (!serveAssets && assetPrefixCallback) {
+      currentAssetPrefix = assetPrefixCallback(ctx);
+    } else {
+      currentAssetPrefix = assetPrefx;
+    }
+
     const assets = new Assets({
-      assetPrefix,
+      currentAssetPrefix,
```
lemonmade: I don't see the matching change to `Assets`'s constructor arguments? I kind of feel like I preferred the old name too /shrug

packages/sewing-kit-koa/src/tests/middleware.test.ts
```diff
@@ -48,13 +45,32 @@ describe('middleware', () => {
...
+    })(context, () => Promise.resolve());
+
+    expect(getAssets(context)).toHaveProperty('assetPrefix', customAssetPrefix);
+
+    await middleware({
+      assetPrefixCallback: ctx => {},
+    })(context, () => Promise.resolve());
+
+    expect(getAssets(context)).toHaveProperty('assetPrefix', webpackAssetUrl);
+  });
```
lemonmade: I think we probably want a test that verifies that `assetPrefix` is called with `context` (we don't currently use that feature anywhere in the tests, so it's not technically verified)

packages/sewing-kit-koa/src/middleware.ts
tgxworld: I thought about having the param accept either a string or function but thought it would be weird for a param to be either of two different types. With this approach though, doesn't it mean we would have to do some sort of `typeof` check in order to know how to handle the param?

packages/sewing-kit-koa/src/middleware.ts
tgxworld: Ah 😿 I had to change this because lint was complaining about reassigning the `assetPrefix` param. Oops just realised this is a mistake as well.

tgxworld:
@lemonmade I've updated the PR after your review :) For some reason, the CLA build is not passing even though I've already signed in.

packages/sewing-kit-koa/src/middleware.ts
lemonmade: Yeah, you’d use typeof to pick the right type at runtime. In general we opt for this because we try to keep our APIs fairly clean, which generally means one option for a thing (but different choices for what to pass when the API needs more customizability) rather than combinations of multiple options.

tgxworld:
Closing this for now as we're taking a difference approach which limits the logic to `shopify/web` itself.

PR #1111, opened 2019-10-11
Add api to terminate the web worker endpoint

ismail-syed:
## Description

This PR enables consumers to be able to terminate web workers created from our `@shopify/web-worker` library. 

Worth noting. We talked about `self.close()` being available from the worker scope. Looking at the latest [`WorkerGlobalScope.close()` docs](https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope/close) , this is a deprecated API


packages/web-worker/src/index.ts
```diff
@@ -1,3 +1,4 @@
-export {createWorker} from './create';
+export {createWorker, terminate} from './create';
 export {retain, release} from './memory';
 export {SafeWorkerArgument} from './types';
+export {expose} from './worker';
```
ismail-syed: I noticed this wasn't exposed, should it be?

packages/web-worker/src/endpoint.ts
```diff
@@ -146,6 +154,14 @@ export function createEndpoint<T>(
       makeCallable(port, () => newValue);
       functionStore.set(newValue, [id, port]);
     },
+    terminate() {
```
ismail-syed: We talked about `self.close()` being available from the worker scope. Looking at the latest [`WorkerGlobalScope.close()` docs](https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope/close) , this is a deprecated API .

packages/web-worker/src/endpoint.ts
```diff
@@ -101,12 +109,36 @@ export function createEndpoint<T>(
...
+
+    if (TERMINATE in data) {
+      if (messageEndpoint.terminate) {
+        terminated = true;
+
+        // This should terminate the worker and continue but it hangs and tests fails waiting on the completion of this
+        // In the browser it seems like this silently fails, no errors or relevant logs.
+        // Subsequent worker calls don't throw errors either, they just silently fail
+
+        // messageEndpoint.terminate();
```
ismail-syed: why this hang and silently fail :/

packages/web-worker/src/endpoint.ts
```diff
@@ -86,12 +92,14 @@ export interface Endpoint<T> {
   expose(api: {[key: string]: Function | undefined}): void;
   revoke(value: Function): void;
   exchange(value: Function, newValue: Function): void;
+  terminateEndpoint(): void;
```
ismail-syed: Named this to `terminateEndpoint` instead of `terminate` to avoid confusion between terminate calls

packages/web-worker/README.md
```diff
@@ -65,6 +65,22 @@ const result = await worker.default(
 ); // 'Hello, Tobi'
 ```
 
+When the web worker is no longer required, you can terminate the worker using `terminate()`. This will immedietly terminate any ongoing operations. If an attempt is made to make calls to a terminated worker, an error will be thrown.
+
+Note: A worker can only be terminated from the main thread. A worker can not terminate itself from within the worker module.
+
+```
```
lemonmade: Missing language type

packages/web-worker/src/endpoint.ts
lemonmade: Rename it somewhere else then, API should have the most obvious name which is definitely `terminate()` here

packages/web-worker/src/endpoint.ts
```diff
@@ -146,6 +155,13 @@ export function createEndpoint<T>(
       makeCallable(port, () => newValue);
       functionStore.set(newValue, [id, port]);
     },
+    terminateEndpoint() {
+      if (messageEndpoint.terminate) {
+        [functionStore, functionProxies, activeApi].forEach(map => map.clear());
+        messageEndpoint.terminate();
+        terminated = true;
```
lemonmade: We should definitely be doing this and the clearing of all the stores whether the `messageEndoint` has `terminate` or not.

lemonmade: Also, I still think the main thread should be notified when `terminate` is called in the worker. At the very least, it lets the main thread release all of its function store/ proxies/ activeApi/ etc

packages/web-worker/src/test/e2e.test.ts
```diff
@@ -520,6 +520,59 @@ describe('web-worker', () => {
       expect(page.workers()).toHaveLength(0);
     });
   });
+
+  it('throws an error when calling a function on a terminated worker', async () => {
```
lemonmade: The other necessary test here is that calling it from the main thread actually releases all relevant memory and that `page.workers()` is empty

packages/web-worker/src/endpoint.ts
ismail-syed: Sounds fair. Important to note that calling `terminate` from worker doesn't truely terminate the worker as [mentioned earlier](https://github.com/Shopify/quilt/pull/1111#discussion_r335038322). It just clears its maps and sets `terminated` to true, preventing subsequent worker calls. 

Updated PR with tests.

packages/web-worker/src/test/e2e.test.ts
ismail-syed: Added a [test](https://github.com/Shopify/quilt/pull/1111/files#diff-734ac9de6755db022949065ed7bfa5efR609-R686) to verify that `page.workers()` and all the relevant memory is released.

packages/web-worker/README.md
```diff
@@ -65,6 +65,22 @@ const result = await worker.default(
 ); // 'Hello, Tobi'
 ```
 
+When the web worker is no longer required, you can terminate the worker using `terminate()`. This will immedietly terminate any ongoing operations. If an attempt is made to make calls to a terminated worker, an error will be thrown.
```
lemonmade:
```suggestion
When the web worker is no longer required, you can terminate the worker using `terminate()`. This will immediately terminate any ongoing operations. If an attempt is made to make calls to a terminated worker, an error will be thrown.
```

packages/web-worker/README.md
```diff
@@ -65,6 +65,22 @@ const result = await worker.default(
...
+Note: A worker can only be terminated from the main thread. A worker can not terminate itself from within the worker module.
+
+```tsx
+import {createWorker, terminate} from '@shopify/web-worker;
+
+const makeWorker = createWorker(() => import('./worker'));
+const worker = makeWorker();
+
+// Assume `stuff` is some large payload that needs to be processed
+const result = await worker.someExpensiveOperation(stuff);
```
lemonmade: Does it actually matter what the intermediate code is? May as well just replace this with `// ...` IMO

packages/web-worker/src/create.ts
```diff
@@ -10,6 +10,15 @@ export function expose(
   return endpoint && endpoint.expose(api);
 }
 
+export function terminate(caller: Endpoint<any>['call']) {
+  const endpoint = getEndpoint(caller);
+  if (endpoint && endpoint.terminate) {
```
lemonmade: `endpoint.terminate` is always defined

packages/web-worker/src/index.ts
lemonmade: No, the intention is just to leave it where it is, it's available from a pseudo-private `@shopify/web-worker/worker` entry that is used internally by the tools.

lemonmade: However, we should be exporting the function of the same name from `./create`

packages/web-worker/src/test/e2e.test.ts
```diff
@@ -520,6 +520,231 @@ describe('web-worker', () => {
...
+        `
+        export function greet() {
+          return 'Hi, friend!';
+        }
+      `,
+      );
+
+      await runWebpack(context);
+
+      const page = await browser.go();
```
lemonmade: Personally I'd rather see an action that terminates (so we can show before there was a worker, after there was not)

packages/web-worker/src/test/e2e.test.ts
```diff
@@ -520,6 +520,231 @@ describe('web-worker', () => {
...
+    await withContext('error-on-terminated-worker-calls', async context => {
+      const {workspace, browser} = context;
+
+      await workspace.write(
+        mainFile,
+        `
+          import {createWorker, terminate} from '@shopify/web-worker';
+          self.worker = createWorker(() => import('./worker'))();
+          (async () => {
+            await terminate(self.worker);
```
lemonmade: terminate is not async

packages/web-worker/src/test/e2e.test.ts
```diff
@@ -520,6 +520,231 @@ describe('web-worker', () => {
...
+        `
+          import {createWorker, terminate} from '@shopify/web-worker';
+
+          self.prepare = () => {
+            start();
+
+            self.func = () => {};
+            self.worker = createWorker(() => import('./worker'))();
+
+            self.WorkerTestClass = class WorkerTestClass {constructor(id) {this.id = id;}}
```
lemonmade: Follow the more modern examples for setting up this style of test, this is using a slightly sloppier setup I had earlier I think?

packages/web-worker/src/test/e2e.test.ts
```diff
@@ -520,6 +520,231 @@ describe('web-worker', () => {
...
+            document.body.appendChild(element);
+          })();
+        `,
+        );
+
+        await workspace.write(
+          workerFile,
+          `
+          export async function terminateAttemptFromWorker(){
+            await self.endpoint.terminate();
```
lemonmade: Not async

packages/web-worker/src/test/e2e.test.ts
```diff
@@ -520,6 +520,231 @@ describe('web-worker', () => {
...
+
+      await page.evaluate(() => (self as any).run());
+      await page.waitForSelector(`#${testId}`);
+
+      expect(await getTestClassInstanceCount(page)).toBe(0);
+      expect(page.workers()).toHaveLength(0);
+    });
+  });
+
+  it('throws an error when calling a function on a terminated worker from the worker thread', async () => {
```
lemonmade: I don't think this quite describes what the test is checking

packages/web-worker/src/test/e2e.test.ts
ismail-syed: Sorry, what _more modern examples_ are you referring to?

lemonmade: See the other memory management tests, they do their setup more succinctly

ismail-syed: Updated

PR #1118, opened 2019-10-15
Adjust types on react-csrf to match usage

vsumner:
## Description
This PR adjust the types of `@shopify/react-csrf` to match it's usage. 

If the token is `csrfToken?: string` the type will be `string | undefined`. 

## Type of change

- Type change in `@shopify/react-csrf` 
- `@shopify/react-csrf-universal-provider` depends on it so will need to be bumped.

- [X]  Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist


vsumner:
Need to rethink this.

PR #1119, opened 2019-10-15
♻️ have `withTiming` return early if `performance.timing` does not ex…

michenly:
…ist.

## Description

The performance mock currently return an empty object which mean calling `new Performance` will run into error during test. 

I feel like I am not getting to the root of the issue with this solution. 

### Musing 1:
I think the performance mock come with [jsdom](https://github.com/jsdom/jsdom/pull/2094), which only mock a portion of the API in `performance` but this does not explain why the mock is an empty object.

### Musing 2:
we are using [`performance.timing`](https://github.com/Shopify/quilt/blob/add-check-to-performance/packages/performance/src/utilities.ts#L90), but checking for [`PerformanceNavigationTiming`](https://github.com/Shopify/quilt/blob/add-check-to-performance/packages/performance/src/performance.ts#L33) before use. Should we check for `PerformanceTiming` as well?

### Musing 3:
`PerformanceTiming` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceTiming) 😂, should the code be updated to use [`performance.navigation`](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceNavigationTiming) instead? Although the official types have `timing` and `navigation` type listed as [deprecated](https://github.com/microsoft/TypeScript/blob/master/lib/lib.dom.d.ts#L11669-L11674)

## Type of change

- [x] @Shopify/react-performance - Major
- [x] @Shopify/performance - Patch

## Checklist

### DIFF
--- a/packages/performance/CHANGELOG.md
+++ b/packages/performance/CHANGELOG.md
@@ -7,6 +7,10 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## Unreleased -->
 
+### Fixed
+
+- `Performance` object constructor will now check if `PerformanceTiming` is supported. [[#1119](https://github.com/Shopify/quilt/pull/1119)]
+
 ## [1.2.1] - 2019-10-11
 
 ### Fixed
--- a/packages/performance/src/performance.ts
+++ b/packages/performance/src/performance.ts
@@ -31,6 +31,7 @@ export class Performance {
   readonly supportsObserver = hasGlobal('PerformanceObserver');
   readonly supportsMarks = hasGlobal('PerformanceMark');
   readonly supportsNavigationEntries = hasGlobal('PerformanceNavigationTiming');
+  readonly supportsTimingEntries = hasGlobal('PerformanceTiming');
   readonly supportsLongtaskEntries = hasGlobal('PerformanceLongTaskTiming');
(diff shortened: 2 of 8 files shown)


packages/react-performance/src/context.ts
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import {Performance} from '@shopify/performance';
 
-export const PerformanceContext = React.createContext<Performance | null>(
-  typeof window === 'undefined' ? null : new Performance(),
```
michenly: cc @vsumner another place I kept going back and forth about using null or undefined.

packages/performance/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## Unreleased -->
 
+### Fixed
+
+- `withTiming` used in `Performance` object constructor will now check if `performance.timing` exist. This is mostly to reduce error encounter during test. [[#1119](https://github.com/Shopify/quilt/pull/1119)]
```
GoodForOneFare:
```suggestion
- `withTiming` used in `Performance` object constructor will now check if `performance.timing` exists. This is mostly to reduce errors encounter during tests [[#1119](https://github.com/Shopify/quilt/pull/1119)]
```

michenly:
actually changed the way I solved this. It feels less hacky now.

PR #1120, opened 2019-10-15
Adjust types on react-csrf and react-universal-provider to match usage

vsumner:
## Description
This PR adjust the types of `@shopify/react-csrf` to match it's usage. 

If the token is `csrfToken?: string` the type will be `string | undefined`. 

## Type of change
- [X] @Shopify/react-csrf  - Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] @shopify/react-csrf-universal-provider  - Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] @shopify/react-universal-provider  - Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

### DIFF
--- a/packages/react-universal-provider/src/create-universal-provider.tsx
+++ b/packages/react-universal-provider/src/create-universal-provider.tsx
@@ -8,13 +8,13 @@ export interface UniversalProviderProps<Value> {
 
 export function createUniversalProvider<Value>(
   id: string,
-  Context: React.Context<Value | null>,
+  Context: React.Context<Value>,
 ) {
   const UniversalProvider = React.memo(
     ({value: explicitValue, children}: UniversalProviderProps<Value>) => {
       const [value = explicitValue, Serialize] = useSerialized<Value>(id);
 
-      if (!value) {
+      if (value == null) {
         throw new Error(
           `You must provide a ${id} value, or have one previously serialized.`,
         );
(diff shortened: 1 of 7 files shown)


packages/react-csrf/src/hooks.ts
```diff
@@ -4,7 +4,7 @@ import {CsrfTokenContext} from './context';
 export function useCsrfToken() {
   const csrf = useContext(CsrfTokenContext);
 
-  if (csrf == null) {
```
michenly: we can probably keep this. It will catch both null & undefined.

packages/react-universal-provider/src/create-universal-provider.tsx
```diff
@@ -8,13 +8,13 @@ export interface UniversalProviderProps<Value> {
 
 export function createUniversalProvider<Value>(
   id: string,
-  Context: React.Context<Value | null>,
+  Context: React.Context<Value | undefined>,
```
michenly: I am thinking maybe this should be `Context: React.Context<Value>`, let the consumer decided what this should be instead.

packages/react-universal-provider/src/create-universal-provider.tsx
```diff
@@ -8,13 +8,13 @@ export interface UniversalProviderProps<Value> {
...
 ) {
   const UniversalProvider = React.memo(
     ({value: explicitValue, children}: UniversalProviderProps<Value>) => {
       const [value = explicitValue, Serialize] = useSerialized<Value>(id);
 
-      if (!value) {
+      if (value === undefined) {
```
michenly: I been wonder if this is valuable to have at all. If a consumer use `UniversalProvider` with null value, would that ever be on purpose?

vsumner: I switched this to `value == null` instead. I was concerned `value` could be a `boolean`.

PR #1121, opened 2019-10-15
:sparkles: add performance module to quilt_rails

TheMallen:
## Description
This PR introduces the `Rails` implementation of `@shopify/koa-performance` as a module within `quilt_rails` called `Quilt::Performance`. Most of the details of how this works can be found in the tests and the documentation.

See [the demo branch](https://github.com/Shopify/web-rails-proving-ground/compare/demo-quilt-rails-perf?expand=1) on `web-rails-proving-ground` for a working example.


gems/quilt_rails/app/controllers/quilt/performance_report_controller.rb
```diff
@@ -0,0 +1,26 @@
+module Quilt
+  class PerformanceReportController < ActionController::Base
+    protect_from_forgery except: :index
```
TheMallen: The bot is mad at me, but we don't really care about forgery protection here since it does not actually access data at all. I tried using `null_session` here but it was still yelling at me about csrf tokens, so this seemed fine.

gems/quilt_rails/test/quilt_rails/performance_measurable_test.rb
```diff
@@ -0,0 +1,401 @@
...
+module Quilt
+  class PerformanceReportableTest < Minitest::Test
+    include Quilt::Performance::Reportable
+
+    def setup
+      @params = ActionController::Parameters.new({})
+    end
+
+    def test_process_report_sends_a_distribution_for_events
+      @params = ActionController::Parameters.new({
```
TheMallen: thought about making a fancy factory here like in the sewing-kit tests but felt like overkill for this atm

gems/quilt_rails/README.md
```diff
@@ -215,18 +217,205 @@ class ReactController < ApplicationController
...
+The `Quilt::Performance::Reportable` mixin is intended to be used in Rails controllers, and provides only the `process_report` method. This method handles parsing an incoming report from [@shopify/react-performance's](https://www.npmjs.com/package/@shopify/react-performance) `<PerformanceReport />` component (or a custom report in the same format) and sending it to your application's StatsD endpoint as `distribution`s using [`StatsD-Instrument`](https://rubygems.org/gems/statsd-instrument).
+
+> **Note** `Quilt::Performance::Reportable` is usable even if you are using purely client-side React through `sewing_kit_script_tag`.
+
+```ruby
+class PerformanceController < ApplicationController
+  include Quilt::Performance::Reportable
+
+  def index
+    process_report
```
TheMallen: Waffled on this name a lot, what this does in full is:
- parse the `params` sent to the `PerformanceReportController` into a `Report` instance
- call `Quilt::Performance::Client.send!` on the constructed instance

So maybe like `parse_and_send_report` but that sounds weird :/

gems/quilt_rails/README.md
```diff
@@ -215,18 +217,205 @@ class ReactController < ApplicationController
 end
 ```
 
+### Performance
+
+#### Reportable
+
+The `Quilt::Performance::Reportable` mixin is intended to be used in Rails controllers, and provides only the `process_report` method. This method handles parsing an incoming report from [@shopify/react-performance's](https://www.npmjs.com/package/@shopify/react-performance) `<PerformanceReport />` component (or a custom report in the same format) and sending it to your application's StatsD endpoint as `distribution`s using [`StatsD-Instrument`](https://rubygems.org/gems/statsd-instrument).
+
+> **Note** `Quilt::Performance::Reportable` is usable even if you are using purely client-side React through `sewing_kit_script_tag`.
```
lemonmade: I think this would be more clear if you say you **don't** need to be using any of the react-server stuff. It reads a little weird to me right now

gems/quilt_rails/README.md
```diff
@@ -215,18 +217,205 @@ class ReactController < ApplicationController
...
+class PerformanceController < ApplicationController
+  include Quilt::Performance::Reportable
+
+  def index
+    process_report
+  end
+end
+```
+
+The params sent to the controller are expected to be of type `application/json`. Given the following example JSON sent by `@shopify/react-performance`,
```
lemonmade: Should they be doing any annotations to their controller actions to require JSON?

gems/quilt_rails/README.md
```diff
@@ -215,18 +217,205 @@ class ReactController < ApplicationController
...
+          },
+        ],
+        "result": 0
+      },
+      "metadata": {
+        "index": 0,
+        "supportsDetailedTime": true,
+        "supportsDetailedEvents": true
+      }
+  ],
```
lemonmade: Missing a closing brace above this

gems/quilt_rails/README.md
```diff
@@ -215,18 +217,205 @@ class ReactController < ApplicationController
...
+```ruby
+StatsD.distribution('time_to_first_byte', 2, ['browser_connection_type:3g'])
+StatsD.distribution('time_to_first_byte', 2, ['browser_connection_type:3g'])
+StatsD.distribution('navigation_complete', 23924, ['browser_connection_type:3g'])
+StatsD.distribution('navigation_usable', 23924, ['browser_connection_type:3g'])
+```
+
+##### Customizing `process_report` with a block
+
+The behaviour of `process_report` can be customized by manipulating the `Quilt::Performance::Client` instance yielded into it's implicit block parameter.
```
lemonmade: `it's` => `its`

gems/quilt_rails/README.md
```diff
@@ -215,18 +217,205 @@ class ReactController < ApplicationController
...
+
+The `on_distribution` method takes a block which is run for each distribution (including custom ones) sent during `process_report`. It can be used to easily add logging or other side-effects to your measurements.
+
+```ruby
+client.on_distribution do |metric_name, value, tags|
+  Rails.logger.debug "#{metric_name}: #{value}, tags: #{tags}"
+end
+```
+
+##### Client#on_distribution
```
lemonmade: Duplicated?

gems/quilt_rails/README.md
```diff
@@ -215,18 +217,205 @@ class ReactController < ApplicationController
...
+
+```ruby
+client.on_report do |report, tags|
+  client.distribution('some_metric', some_computation(report), tags)
+end
+```
+
+##### Client#on_navigation
+
+The `on_report` method takes a block which is run once per navigation in a given `Report`. It can be used to perform side-effects based on a given `Navigation`, including sending custom `distribution`s.
```
lemonmade: `on_report` => `on_navigation`

gems/quilt_rails/README.md
```diff
@@ -215,18 +217,205 @@ class ReactController < ApplicationController
...
+The `customize_navigation_tags` method allows custom tags to be computed for each individual navigation described by the POST body, and included on the `distribution` calls for that navigation.
+
+```ruby
+client.customize_navigation_tags do |navigation, tags|
+  # target will be included only for the distributions about this navigation
+  tags[:navigation_target] = navigation.target
+end
+```
+
+#### Report
```
lemonmade: It's kind of weird to me to have `Report` be a meaningful thing since it represents a collection of navigations/ events of unknown size/ distribution :/

gems/quilt_rails/README.md
```diff
@@ -215,18 +217,205 @@ class ReactController < ApplicationController
...
+
+[`Quilt::Performance::Report`](/lib/quilt_rails/performance/report.rb#L3) describes timing, size information, and metadata about a series of events and navigations on the front-end of a web application, as recorded by `@shopify/performance` and parsed from an `ActionController:Parameters` instance.
+
+##### Report#pathname
+
+The initial path (eg. `/foo`) of the request which generated the report.
```
lemonmade: Does this even make sense if it contains multiple navigations?

gems/quilt_rails/README.md
```diff
@@ -215,18 +217,205 @@ class ReactController < ApplicationController
...
+
 ### Engine
 
-`Quilt::Engine` provides a preconfigured controller which consumes `ReactRenderable` and provides an index route which uses it.
+`Quilt::Engine` provides:
+
+- a preconfigured `UiController` which consumes `ReactRenderable`
+- a preconfigured `PerformanceController` which consumes `Performance::Reportable`
+- a `/performance` route mapped to `PerformanceController#index`
+- a catch-all index route mapped to the `UiController#index`
```
lemonmade: Is there any way we can give them only the performance bit, or would that require them to manually use the mixin?

gems/quilt_rails/app/controllers/quilt/performance_report_controller.rb
lemonmade: Might be something to pass by security?

gems/quilt_rails/app/controllers/quilt/performance_report_controller.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module Quilt
+  class PerformanceReportController < ActionController::Base
+    protect_from_forgery except: :index
+    include Quilt::Performance::Reportable
+
+    def index
+      process_report do |client|
```
lemonmade: Why is there any customization here? Wouldn't a consumer expect that anything the default controller will do is all encompassed by the mixin? I can't say I understand how you decided what would be here versus in `process_report`.

gems/quilt_rails/config/routes.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 Quilt::Engine.routes.draw do
+  get '/performance_report', to: 'performance_report#index'
```
lemonmade: I think this name is different than the docs?

gems/quilt_rails/lib/quilt_rails/performance/connection.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+module Quilt
+  module Performance
+    class Connection
```
lemonmade: In general, when they log out these model-style classes, do they present useful information? I don't know how the default logging of Rails class instances work.

gems/quilt_rails/app/controllers/quilt/performance_report_controller.rb
TheMallen: Yeah that makes sense, I made the defaults in `process_report` the same as in koa but I added here behaviour that made sense to me as defaults, probably should just throw the bits that are most obvious into the mixin itself and just do the call + response rendering here.

gems/quilt_rails/README.md
TheMallen: manually use the mixin

gems/quilt_rails/lib/quilt_rails/performance/connection.rb
TheMallen: It just says the class name and a uuid type thing. I think there's like a toString / inspect type method you can implement to show more details, I'll look into it.

gems/quilt_rails/app/controllers/quilt/performance_report_controller.rb
TheMallen: This is a POST endpoint, it parses a JSON body.

gems/quilt_rails/README.md
TheMallen: `Report` represents a specifically a report sent from the client. We even call the client-side component `<PerformanceReport />`. I'm not sure what else we'd name the class?

lemonmade: Well `PerformanceReport` is a long-lived thing that sends all events to the server. I am just talking about how we interact with that on the server, and what I am saying in that instance is that anything that represents the "report" as being significant is weird (because it is literally just a throttled grouping of random events from the client).

gems/quilt_rails/README.md
TheMallen: That's a good question, probably not. Should we delete our usage of `pathname` from web and koa-metrics?

gems/quilt_rails/README.md
TheMallen: I'm not sure what you mean by significant. This literally says:

> describes timing, size information, and metadata about a series of events and navigations on the front-end of a web application

This thing, whatever it is called, is what the user of the library will need to understand the shape of to be able to write custom metrics code using `on_report`. If we called it `on_metrics_body` and named this `metrics_body` like we did in `koa-performance` I think that actually makes things less clear.

It will always be "significant" in the sense that it is part of the API consumers use to customize handling of the incoming requests. This is the same as it is in `koa-metrics` but with a name that feels more accurate / more obviously connected to the component that sends it on the react side.

lemonmade: But why do they do anything on a report? Sure, they can add stuff on navigations, and they can add stuff on events, but a report is basically just a big array. Why does that need to be exposed to the consumer? Put another way, if the `PerformanceReport` were not batching (it was just sending events as it gets them), would this construct make any sense?

gems/quilt_rails/README.md
TheMallen: Thinking about it more, it makes sense specifically as the pathname that the top level events (such as TTFB) refer to. `Navigation`s are just client-side `History` style navigations. There's no way for batching to even happen across multiple server side requests.

gems/quilt_rails/README.md
TheMallen: Report also contains other top-level info though, such as extra metadata about the browser and it's internet connection. If you want to generate tags off of those we'd either need to thread that through to the navigation (which I guess is reasonable) or expose the report itself.

TheMallen: 🤔 I'm going to explore just passing the `connection` down to `navigation` and see how that feels

TheMallen: tldr: it feels awful

gems/quilt_rails/README.md
```diff
@@ -404,3 +565,77 @@ class GraphqlController < ApplicationController
...
+```
+
+4. If your application is not using `Quilt::Engine`, you will need to manually configure the server-side portion of performance tracking. Add a `PerformanceController` and the corresponding routes to your Rails app.
+
+```ruby
+# app/controllers/performance_report_controller.rb
+
+class PerformanceReportController < ActionController::Base
+  include Quilt::Performance::Reportable
+  protect_from_forgery with: :null_session
```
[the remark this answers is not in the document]
TheMallen: `null_session` makes the most sense here, not sure why it wasn't working when I tried it before but it seems good now :)

gems/quilt_rails/lib/quilt_rails/performance/client.rb
```diff
@@ -0,0 +1,86 @@
...
+      private
+
+      def process_report!
+        report_events
+        report_navigations
+      end
+
+      def report_events
+        @report.events.each do |event|
+          event_tags = @base_tags.dup
```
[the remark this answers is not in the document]
TheMallen: It comes from the request, so constant seems wrong as it would change each invocation.

Probably it should be `.freeze`d though

gems/quilt_rails/app/controllers/quilt/performance_report_controller.rb
TheMallen: @clayton-shopify could you take another look quick? I got `with: :null_session` working now

gems/quilt_rails/lib/quilt_rails/performance/client.rb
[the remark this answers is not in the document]
lemonmade: Why is constant wrong if you are just duplicating it?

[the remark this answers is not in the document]
TheMallen: it gets computed at instance construction not at the module level, I don't get how it would be a constant.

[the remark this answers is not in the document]
lemonmade: I dum

PR #1122, opened 2019-10-15
Add @shopify/react-web-worker

ismail-syed:
Pretty small package here. 

* Provides a `useWorker` hook
* Re-exports all of `@shopify/web-worker`


packages/react-web-worker/src/hooks.ts
```diff
@@ -0,0 +1,19 @@
...
+
+import {useLazyRef} from '@shopify/react-hooks';
+
+export function useWorker<T>(creator: () => T) {
+  const {current: worker} = useLazyRef(creator);
+
+  useEffect(
+    () => {
+      return () => {
+        // terminate(worker);
```
ismail-syed: Will add in once #1111 is merged.
I'm ok with shipping this independently from #1111

packages/react-web-worker/src/test/hooks.test.tsx
```diff
@@ -0,0 +1,32 @@
...
+  return () => {
+    return {
+      hello(friend: string) {
+        return `Hello, ${friend}`;
+      },
+    };
+  };
+}
+describe('useWorker', () => {
+  function MockComponent({create = mockCreate(), children}) {
```
lemonmade: We should need to provide a type here (IIRC, TS was complaining in the editor about this prop). Yes, you will find that quilt's CI does not check types for tests, which is fairly shocking, I agree :P

packages/react-web-worker/README.md
```diff
@@ -0,0 +1,53 @@
...
+```tsx
+import React, {useEffect} from 'react';
+import {Page} from '@shopify/polaris';
+import {createWorker, useWorker} from '@shopify/react-web-worker';
+
+// assume ./worker.ts contains
+// export function hello(name) {
+//  return `Hello, ${name}`;
+// }
+const create = createWorker(() => import('./worker'));
```
lemonmade: Newline above

packages/react-web-worker/README.md
```diff
@@ -0,0 +1,53 @@
...
+      })();
+    },
+    [worker],
+  );
+
+  return <Page title="Home"> {message} </Page>;
+}
+```
+
+### `@shopify/web-worker`
```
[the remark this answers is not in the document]
ismail-syed: Moving to the top

Issue #1123, opened 2019-10-16
Prettier to ^1.18.0 conflicts with Typescript 3.2.1


cartogram:
**TLDR: Updating Prettier to the latest minor version requires us to be on Typescript 3.3 or later.**

When attempting to update to the latest minor version of Prettier, I encountered an issue where the formatted TypeScript output added a trailing comma to `Tuples`. 

For example

```ts
// Before
export function useSomeHook(): [
  string,
  string
] { /*...*/}

// After 
export function useSomeHook(): [
  string,
  string, <-- Trailing comma 
] { /*...*/}
```

This is a problem because our version of Typescript does not support trailing commas on Tuples. It looks like they were were added in Typescript 3.3 via this [PR](https://github.com/Microsoft/TypeScript/issues/28893).

Work has already been started in [this PR to upgrade Typescript](https://github.com/Shopify/quilt/pull/962/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R71). When that is shipped, we can upgrade to the latest prettier.

PR #1110, opened 2019-10-11
Update eslint-plugin-shopify v31.0.0

cartogram:
## Description

This PR bring this project to the latest `eslint` and `eslint-plugin-shopify`. I've left comments inline for a few obscure new rules, but happy to clarify any additional changes if anything unclear.

PR #1124, opened 2019-10-16
Update Prettier

cartogram:
## Description

This PR updates prettier to version `1.17.0`. Prettier was originally installed in this PR https://github.com/Shopify/quilt/pull/203, and hasn't changed in the ~14 months since. 

When updating our eslint dependencies ([in this PR](https://github.com/Shopify/eslint-plugin-shopify/pull/433)), the updates to `eslint-config-prettier`/`eslint-plugin-prettier` modifies our prettier version. 

Rather than keep the prettier changes for [when we update our eslint/eslint-plugin-shopify versions](https://github.com/Shopify/quilt/pull/1110) (which are distracting and formatting only), I've pulled them out here. I cannot bump this repo right to the latest version (`1.18.2`) because of an incompatibility with our version of Typescript ([documented here](https://github.com/Shopify/quilt/issues/1123)).

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -2,5 +2,6 @@
   "arrowParens": "avoid",
   "singleQuote": true,
   "bracketSpacing": false,
-  "trailingComma": "all"
+  "trailingComma": "all",
+  "quoteProps": "as-needed"
 }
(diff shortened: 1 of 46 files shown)


cartogram:
@ismail-syed yeah, we agree. I'm going to do the `eslint-plugin-shopify` prettier update, but in this PR I am bringing us as close as we can without needing to bump TS as well. Does that make sense?

ismail-syed:
**facepalm** I just had the biggest brain fart. For some reason, I thought this was a PR for `eslint-plugin-shopify`. 😅

PR #1131, opened 2019-10-21
Warn developers about quilt_rails' performance issues

GoodForOneFare:
## Description
Warn consumers about quilt_rails' limitations.

## Checklist


gems/quilt_rails/README.md
```diff
@@ -1,3 +1,11 @@
+# Alpha functionality - do not use in high-traffic production applications
```
lemonmade: Is there anything we can link to if they want more details?

GoodForOneFare: Added links to the "what the problem is" issue, and the "future architecture discussion" issue.

Also added a snipped about altering existing apps to skip server-side queries.

PR #1137, opened 2019-10-23
Enhance documentation for using performance libraries together

TheMallen:
This PR takes a stab at filling out some of the unfinished documentation for our performance tracking libraries, and clarifying how to use them together with stronger in depth examples.

### DIFF
--- a/packages/koa-performance/README.md
+++ b/packages/koa-performance/README.md
@@ -3,21 +3,38 @@
 [![Build Status](https://travis-ci.org/Shopify/quilt.svg?branch=master)](https://travis-ci.org/Shopify/quilt)
 [![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](LICENSE.md) [![npm version](https://badge.fury.io/js/%40shopify%2Fkoa-performance.svg)](https://badge.fury.io/js/%40shopify%2Fkoa-performance.svg) [![npm bundle size (minified + gzip)](https://img.shields.io/bundlephobia/minzip/@shopify/koa-performance.svg)](https://img.shields.io/bundlephobia/minzip/@shopify/koa-performance.svg)
 
-A middleware which makes it easy to send metrics from your front end application and forward them to a stats-d server of your choice.
+A middleware which makes it easy to send metrics from your front end application and forward them to a StatsD server of your choice.
 
 Best used with [`@shopify/performance`](https://www.npmjs.com/package/@shopify/performance) and/or [`@shopify/react-performance`](https://www.npmjs.com/package/@shopify/react-performance).
 
-## Installation
+## Table of Contents
+
+- [Quick-start](#quick-start)
+
+  - [Install the package](install-the-package)
+  - [Add the middleware](add-the-middleware)
+  - [Verify with CURL](verify-with-curl)
+  - [Send data from the frontend](send-data-from-the-frontend)
+  - [Process report data and forward it to StatsD](process-report-data-and-forward-it-to-statsd)
+
+- [API](#api)
+  - [Hooks](#hooks)
+  - [Components](#components)
+  - [Other](#other)
+
+## Quick-start
+
+### Install the package
+
+First we will need to install the npm package.
 
 ```bash
 $ yarn add @shopify/koa-performance
 ```
 
-## Usage
+### Add the middleware
 
-### Basic server setup
-
-First add the middleware to your Koa app. We recommend using `koa-mount` or `koa-router` to restrict it to a specific endpoint.
+Next we import the `clientPerformanceMetrics` factory into our server, use it to create a middleware to collect performance data, and mount it. We use something `koa-mount` to restrict it to a specific endpoint. If your application uses `koa-router`, you can use that instead.
 
 ```tsx
 // server.ts
@@ -25,24 +42,32 @@ import Koa from 'koa';
 import mount from 'koa-mount';
 import {clientPerformanceMetrics} from '@shopify/koa-performance';
 
+// create our Koa instance for the server
 const app = new Koa();
 
-app.use(mount('/client-metrics', clientPerformanceMetrics({
-  prefix: 'MyApp.',
-  // when development is true the middleware will skip sending metrics
-  development: process.env.NODE_ENV === 'development',
-  // the host of the statsd server you want to send metrics to
-  statsdHost: 'localhost',
-  // the port of the statsd server you want to send metrics to
-  statsdPort: 3000,
-})));
+app.use(
+  mount(
+    '/client-metrics',
+    clientPerformanceMetrics({,
+      // the prefix for metrics sent to StatsD
+      prefix: 'ExampleCode.',
+      // the host of the statsd server you want to send metrics to
+      statsdHost: 'YOUR STATSD HOST HERE'
+      // the port of the statsd server you want to send metrics to
+      statsdPort: 3000,
+    }),
+  ),
+);
+
+// other middleware for your app
+// ...
 
 app.listen(3000, () => {
   console.log('listening on port 3000');
-})
+});
 ```
 
-Now the app will respond to requests to `/client-metrics`. The middleware returned from `clientPerformanceMetrics` expects to receive JSON POST requests meeting the following interface:
+Now the app will respond to requests to `/performance-report`. The middleware returned from `clientPerformanceMetrics` expects to receive JSON POST requests meeting the following interface:
 
 ```tsx
 interface Metrics {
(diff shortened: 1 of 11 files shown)


packages/react-performance/README.md
```diff
@@ -50,36 +86,131 @@ export function LastNavigationDetails() {
...
+```
+
+With these changes we will see separate timing information for `usable` and `complete` in our `usePerformanceReport` output.
 
 ## API
 
 Most APIs provided by this package are available as both hooks (ie. `useNavigationListener`), as well as components (ie. `<NavigationListener />`).
 
-This package also re-exports all of the API of [`@shopify/performance`](https://www.npmjs.com/package/@shopify/performance).
+This package also re-exports all of the API from [`@shopify/performance`](https://www.npmjs.com/package/@shopify/performance).
```
michenly: aww thats what you are going for before. Both work really, but to keep consistent I will suggest the following.

```suggestion
This package also re-exports all of the API from [`@shopify/performance`](../performance/README.md).
```

packages/react-performance/README.md
```diff
@@ -17,7 +17,36 @@ $ yarn add @shopify/react-performance
 
 The most basic way to use the tools in this package is to record information and display it locally to the user. In practice you usually only want to do this in development so that developers can easily see performance information.
 
-#### Display data using NavigationListener
+### Adding some `<PerformanceMark />`s
+
+We will need to add some `<PerformanceMark />` components to our React application in order to help `@shopify/react-performance` determine when your application is fully loaded and React has mounted your components.
```
michenly:
```suggestion
We will need to add some `<PerformanceMark stage="complete"/>` components to our React application in order to help `@shopify/react-performance` determine when your application is fully loaded and React has mounted your components.
```

michenly: I suggest this since `when your application is fully loaded and React has mounted your components` is only true with `stage="complete"`

packages/react-performance/README.md
```diff
@@ -17,7 +17,36 @@ $ yarn add @shopify/react-performance
...
+        <li>Smoky</li>
+      </ul>
+      <PerformanceMark stage="complete" id="cats" />
+    </section>
+  );
+}
+```
+
+In production apps, it is important to always have one `<PerformanceMark stage="complete" />` on each page level component. Otherwise, client-side navigations may not be recorded properly.
+
```
michenly: Maybe we need to add an example for `<PerformanceMark stage="usable"/>`

packages/react-performance/README.md
```diff
@@ -17,7 +17,36 @@ $ yarn add @shopify/react-performance
...
+        <li>Moe</li>
+        <li>Smoky</li>
+      </ul>
+      <PerformanceMark stage="complete" id="cats" />
+    </section>
+  );
+}
+```
+
+In production apps, it is important to always have one `<PerformanceMark stage="complete" />` on each page level component. Otherwise, client-side navigations may not be recorded properly.
```
michenly:
```suggestion
In production apps, it is important to always have one `<PerformanceMark stage="complete" />` on each page level component. Otherwise, client-side navigations may not be recorded properly. You should not use `usePerformanceMark("complete", "id")` in place of the component because...
```

packages/react-performance/README.md
```diff
@@ -17,7 +17,36 @@ $ yarn add @shopify/react-performance
...
+We will need to add some `<PerformanceMark />` components to our React application in order to help `@shopify/react-performance` determine when your application is fully loaded and React has mounted your components.
+
+```tsx
+//Cats.tsx
+import React from 'react';
+import {PerformanceMark} from '@shopify/react-performance';
+
+/**
+ * Represents an entire page of our application
+ */
```
michenly: I wonder if this should be

```
export function Cats() {
  return (
    <>
      <section>
        <h1>Cats</h1>
        <ul>
          <li>Kokusho</li>
          <li>Moe</li>
          <li>Smoky</li>
        </ul>
      </section>
      <PerformanceMark stage="complete" id="cats" />
    </>
  );
```

so the complete will include the render of `<section />` as well. (have I been doing this all wrong actually?)

packages/react-performance/README.md
```diff
@@ -17,7 +17,36 @@ $ yarn add @shopify/react-performance
...
+      </ul>
+      <PerformanceMark stage="complete" id="cats" />
+    </section>
+  );
+}
+```
+
+In production apps, it is important to always have one `<PerformanceMark stage="complete" />` on each page level component. Otherwise, client-side navigations may not be recorded properly.
+
+#### Displaying data using `useNavigationListener`
```
michenly:
```suggestion
### Displaying real time navigation data using `useNavigationListener`
```

packages/react-performance/README.md
```diff
@@ -26,9 +55,14 @@ To demonstrate, we'll create a component called `LastNavigationDetails` and use
 import React, {useState} from 'react';
 import {useNavigationListener, Navigation} from '@shopify/react-performance';
 
+/**
+ * A component which displays simple performance stats about the last navigation performed by the user
+ */
```
michenly: ❤️

packages/koa-performance/README.md
```diff
@@ -15,7 +15,7 @@ $ yarn add @shopify/koa-performance
 
 ## Usage
 
-### Basic server setup
+### Basic
 
 First add the middleware to your Koa app. We recommend using `koa-mount` or `koa-router` to restrict it to a specific endpoint.
```
michenly:
```suggestion
First add the middleware to your Koa app. This middleware will collect all metrics sent to '/client-metrics' and send a stats-d server. We recommend using `koa-mount` or `koa-router` to restrict it to a specific endpoint.
```

packages/react-performance/README.md
```diff
@@ -17,7 +17,36 @@ $ yarn add @shopify/react-performance
 
 The most basic way to use the tools in this package is to record information and display it locally to the user. In practice you usually only want to do this in development so that developers can easily see performance information.
 
-#### Display data using NavigationListener
+### Adding some `<PerformanceMark />`s
```
michenly: ### Adding metric tracking using `<PerformanceMark />`

michenly:
For `react-performance/README.md`

I would like to suggest the following structure.

```
## Quick-start

### Add metrics tracking to your application
#### Adding some `<PerformanceMark />`s
#### Tracking _usable_ vs _complete_

### Reporting performance data to a data aggregation server
#### Reporting data to a server
#### Parsing report JSON server-side

### Sending performance data to a monitoring & analytics server
#### Using Node (`@shopify/koa-performance`)
#### Using Rails (`quilt_rails`)

## API
```

packages/react-performance/README.md
```diff
@@ -50,36 +86,131 @@ export function LastNavigationDetails() {
...
+
+  return (
+    <section>
+      <h1>Cats</h1>
+      {/* mark the page as usable immediately upon rendering */}
+      <PerformanceMark stage="usable" id="cats" />
+      {data && (
+        <>
+          {/* mark the page as *complete* once we have our data */}
+          <PerformanceMark stage="complete" id="cats" />
```
lemonmade: Should we be defaulting to just using the hook?

packages/react-performance/README.md
lemonmade: Don't link directly to the README, but agreed to link to the local package directory

packages/react-performance/README.md
TheMallen: I don't think it matters too much, but possibly slightly more accurate

packages/react-performance/README.md
```diff
@@ -146,7 +277,80 @@ function ProductPage() {
...
+import {usePerformanceReport} from '@shopify/react-performance';
+import {ProductPage} from './ProductPage';
+
+function App() {
+  usePerformanceReport('/performance-report');
+  return <ProductPage />;
+}
+```
+
+To get sensical data, applications using `<PerformanceReport />` should be sure to have at least on performance mark on each top level page.
```
lemonmade: `at least one`. This is also duplicate content?

packages/react-performance/README.md
```diff
@@ -146,7 +277,80 @@ function ProductPage() {
...
+import {usePerformanceMark} from '@shopify/react-performance';
+
+function ProductPage() {
+  usePerformanceMark('complete');
+
+  return <div>cool product page</div>;
+}
+```
+
+The data sent to the `/performance-report` url should look something like:
```
lemonmade: This is an implementation detail, since we offer abstractions on both end we should not actively show the implementation details between them

packages/react-performance/README.md
michenly: The `...` is mostly in reference to a question I asked in a previous PR on why we dont use `usePerformanceMark` , I think the reason why is a hook version will get involved before the entire component finished rendering?

packages/react-performance/README.md
TheMallen: We have that below, titled `### Tracking _usable_ vs _complete_`

michenly: Yes, please ignore this comment, I realized that later on.

packages/react-performance/README.md
TheMallen: I'll add a tiny bit more context here, but it feels more like going super deep on when and where you might miss performance marks and the ramifications belongs more in an "FAQ" or something.

TheMallen: Also
> Where does Web include performance marks that aren't just in the pages?

From what I can see it literally *only* uses `usable` and `complete` (mostly `complete`, and our types currently disallow custom stages) and  I *think* only on page level components.

lemonmade: https://github.com/Shopify/web/blob/master/app/foundation/ErrorBoundary/ErrorBoundary.tsx#L55 and probably should typically put one in the `NotFound` component to be safe

michenly: so I didnt add this in online-store-web yet but I do kind of wonder. Would this also make sense to add for an embedded modal? or just a large modal?

packages/react-performance/README.md
TheMallen: I'm actually thinking of getting rid of the hook for this from the public API, I think you generally care where you put these and want to be able to conditionally render them, so they feel more like they want to be a component to me.

Vice versa, I don't see a reason to ever not use the hook version of `usePerformanceReport` so I'm using it in all the examples.

lemonmade: If we make the hook possible to use for "unfinished" cases, wouldn't that resolve the conditional rendering?

TheMallen: I'm not sure I prefer adding more API to the hook rather than just making the component the only API for it.

TheMallen: In a moment I'll push a big overhaul, part of which suggests only using the hook for `usePerformanceReport` and the component for `<PerformanceMark />`, let me know how you feel about it

packages/koa-performance/src/middleware.ts
```diff
@@ -47,7 +47,8 @@ export function clientPerformanceMetrics({
   statsdHost,
   statsdPort,
   prefix,
-  development,
+  // eslint-disable-next-line no-process-env
+  development = process.env.NODE_ENV === 'development',
```
michenly: if we are changing API, I kinda want to change this option from `development` to `disabledLogging`

packages/react-performance/src/performance-mark.ts
```diff
@@ -1,8 +1,6 @@
 import {usePerformanceEffect} from './performance-effect';
 
-export type Stage = 'complete' | 'usable';
-
-export function usePerformanceMark(stage: Stage, id: string) {
+export function usePerformanceMark(stage: string, id: string) {
```
michenly: why remove the type? I thought it was pretty useful.

TheMallen: The API actually supports any arbitrary string. I guess we could export the magic ones as an enum though

TheMallen: The API actually supports any arbitrary string. I guess we could export the magic ones as an enum though

packages/koa-performance/src/middleware.ts
TheMallen: I only made it optional because it is always passed the same value and was annoying in the examples. If rather just remove the API change than make it actually breaking

TheMallen: I only made it optional because it is always passed the same value and was annoying in the examples. If rather just remove the API change than make it actually breaking

packages/react-performance/src/performance-mark.ts
TheMallen: Added back as an enum, still using `string` as the type here so it properly supports any custom stage value

packages/koa-performance/src/middleware.ts
TheMallen: I'll keep this for now since it's just making an existing API optional, we can follow up in another PR to make breaking changes :)

PR #1140, opened 2019-10-23
✨ add documentation about performance and navigation event

michenly:


packages/performance/docs/events.md
```diff
@@ -0,0 +1,126 @@
...
+### GraphQL Event
+
+Use enum `EventType.GraphQL`
+
+The time spent resolving GraphQL queries during the navigation.
+
+This metric needs to be manually set up in Apollo Client.
+The setup can be done as a [ApolloLink](https://www.apollographql.com/docs/link/).
+
+```tsx
```
michenly: huh...so as I typed this out. I realized we should probably just make this link into a re-usable utility.

what does people think about between moving [`operationDetailsLink`](https://github.com/Shopify/web/blob/upgrade-packages/app/utilities/graphql-client/link/operation-details.ts#L21-L80) into `@shopify/react-graphql` and expect the consumer to add the performance part themselves. 

Or make the simplify version below into `@shopify/react-graphql`.

TheMallen: didn't @ismail-syed  work on that?

michenly: I found remanent of it in 3 different PR, but I guess non of it ever made it to production.
https://github.com/Shopify/quilt/pull/1018
https://github.com/Shopify/quilt/pull/1015
https://github.com/Shopify/quilt/pull/1005

packages/performance/docs/events.md
```diff
@@ -0,0 +1,126 @@
+# Events
+
+## Lifecycle Events
+
+### Time to First Byte
+
+Use enum `EventType.TimeToFirstByte`
```
TheMallen: > Use enum `EventType.TimeToFirstByte`

I find these statements confusing.  I'd prefer if we started this entire section of the docs explaining that there is an enum called `EventType` that all events have a status from, and then title the individual sections like:

```
### EventType.TimeToFirstBye

The time... (rest of the explanation)
```

packages/performance/README.md
```diff
@@ -94,3 +97,7 @@ The `Navigation` object represents a full navigation, either from a full-page re
   - `metadata`: an object with arbitrary key-value pairs providing additional context
 
 `Navigation` also provides a number of utility methods for gathering more actionable information, such as `eventsByType` for filtering events to a particular type, or `totalDownloadSize` for the cumulative size of all requested resources. Please consult the TypeScript definitions for a full listing of these methods.
+
+### Events
+
+Read about the list of events `Performance` and `Navigation` class supports and what they do [here](docs/events).
```
TheMallen: I don't think this section has enough explanation for why you would want to read about events. Maybe something like:

> The `Performance and `Navigation` classes both deal with `Event` objects with which contain information about the timing of specific milestones during the course of a user's browser session. 

I'd also sort of rather just nest the list of events in the same file since it feels weird to give them their own file but nothing else.

packages/performance/docs/events.md
lemonmade: Yeah, it's good to say that

> The `type` field on the event will be `EventType.TimeToFirstByte`

But the more important bit is a brief explanation and the link to learn about the event (also, weird that the link is just `Learn more`, no sentence, at the end of each section)

packages/performance/docs/events.md
```diff
@@ -0,0 +1,126 @@
...
+Use enum `EventType.LongTask`
+
+Any task that take 50 milliseconds or more.
+[Learn more](https://developer.mozilla.org/en-US/docs/Web/API/PerformanceLongTaskTiming)
+
+### Script Download Event
+
+Use enum `EventType.ScriptDownload`
+
+The time spent downloading a script resource.
```
lemonmade: I think it's probably worth discussing what other fields are here? In the case of script downloads, I think there's also some metadata available.

packages/performance/docs/events.md
lemonmade: Yeah can we maybe take this one out until we have that link available? I don't want people copy pasting this around.

packages/performance/docs/events.md
michenly: I would like the title to be a part of explanation rather just the type so the doc can still read like an article.
However, I still want to keep the type for the ease of searching later.

What if I do this structure? 

```
##### Time to First Byte (`EventType.TimeToFirstByte`)
The time until the server sends the first part of the response.
Learn more about this metric [here](https://developers.google.com/web/tools/chrome-devtools/network-performance/understanding-resource-timing).
```

lemonmade: In general, avoid using `here` as the linked word (bad for screen reader users). Otherwise, structure LGTM 👍 (of note, it might cause the anchor links for those titles to be kind of janky, but oh well)

PR #1145, opened 2019-10-24
Add usePrevious react hook

sylvhama:
## Description

Fixes https://github.com/Shopify/web/issues/20093

`web` is already using this hook at several places. Instead of copy pasting it over and over I guess it would be nice to put it here.

More details about this hook: https://usehooks.com/usePrevious/

## Checklist

### DIFF
--- a/packages/react-hooks/README.md
+++ b/packages/react-hooks/README.md
@@ -84,3 +84,21 @@ function MockComponent() {
   );
 }
 ```
+
+### `usePrevious()`
+
+This hook will store the previous value of a given variable.
+
+```tsx
+function Score({value}) {
+  const previousValue = usePrevious(value);
+  const newRecord = value > previousValue ? <p>We have a new record!</p> : null;
+
+  return (
+    <>
+      <p>Current score: {value}</p>
+      {newRecord}
+    </>
+  );
+}
+```
--- a/packages/react-hooks/src/hooks/previous.ts
+++ b/packages/react-hooks/src/hooks/previous.ts
@@ -1,0 +1,10 @@
+import {useRef, useEffect} from 'react';
+
+export default function usePrevious<T>(value: T) {
+  const ref = useRef<T>();
+  useEffect(() => {
+    ref.current = value;
+  }, [value]);
+
+  return ref.current;
+}
--- a/packages/react-hooks/src/hooks/tests/previous.test.tsx
+++ b/packages/react-hooks/src/hooks/tests/previous.test.tsx
@@ -1,0 +1,40 @@
+import * as React from 'react';
+import {mount} from '@shopify/react-testing';
+
+import usePrevious from '../previous';
+
+describe('usePrevious', () => {
+  function Score({value}) {
+    const previousValue = usePrevious(value);
+    const newRecord =
+      value > previousValue ? <p>We have a new record!</p> : null;
+
+    return (
+      <>
+        <p>Current score: {value}</p>
+        {newRecord}
+      </>
+    );
+  }
+
+  it('returns undefined initially', () => {
+    const value = 100;
+
+    const wrapper = mount(<Score value={value} />);
+
+    expect(wrapper).toContainReactText(`Current score: ${value}`);
+    expect(wrapper).not.toContainReactText('We have a new record!');
+  });
+
+  it('displays previous value after updating value', () => {
+    const value = 100;
+    const nextValue = 200;
+
+    const wrapper = mount(<Score value={value} />);
+
+    wrapper.setProps({value: nextValue});
+
+    expect(wrapper).toContainReactText(`Current score: ${nextValue}`);
+    expect(wrapper).toContainReactText('We have a new record!');
+  });
+});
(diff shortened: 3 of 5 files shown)


packages/react-hooks/README.md
```diff
@@ -84,3 +84,14 @@ function MockComponent() {
   );
 }
 ```
+
+### `usePrevious()`
+
+This hook will storage the previous value of a given value.
```
loic-d: will store

packages/react-hooks/src/hooks/previous.ts
```diff
@@ -0,0 +1,10 @@
+import {useRef, useEffect} from 'react';
+
+export default function usePrevious<T>(value: T) {
+  const ref = useRef<T>();
+  useEffect(() => {
+    ref.current = value;
+  }, [value]);
+
+  return ref.current;
+}
```
loic-d: Missing line break 💅

packages/react-hooks/README.md
```diff
@@ -84,3 +84,14 @@ function MockComponent() {
   );
 }
 ```
+
+### `usePrevious()`
+
+This hook will store the previous value of a given value.
```
loic-d: Should we say 'of a given variable'?

packages/react-hooks/src/hooks/tests/previous.test.tsx
```diff
@@ -0,0 +1,27 @@
...
+    return <div>{previousValue}</div>;
+  }
+
+  it('inits the previous value to undefined and displays it', () => {
+    const wrapper = mount(<MockComponent value={666} />);
+
+    expect(wrapper).toContainReactComponent('div', {children: undefined});
+  });
+
+  it('displays previous value after updating it', () => {
```
loic-d: ... after updating value?

packages/react-hooks/README.md
```diff
@@ -84,3 +84,14 @@ function MockComponent() {
...
 ```
+
+### `usePrevious()`
+
+This hook will store the previous value of a given variable.
+
+```tsx
+function MockComponent({value}) {
+  const previousValue = usePrevious(value);
+  return null;
```
cartogram: I'd love to see a more realistic example here. Something that shows a very minimal real use-case.

packages/react-hooks/src/hooks/tests/previous.test.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+import usePrevious from '../previous';
+
+describe('usePrevious', () => {
+  function MockComponent({value}) {
+    const previousValue = usePrevious(value);
+    return <div>{previousValue}</div>;
+  }
+
+  it('inits the previous value to undefined and displays it', () => {
```
cartogram: maybe just `returns undefined initially`?

sylvhama:
@cartogram @loic-d I have made some text updates based on your comments and I also created a better example. Let me know if you have further concerns :)

packages/react-hooks/README.md
```diff
@@ -84,3 +84,19 @@ function MockComponent() {
...
+### `usePrevious()`
+
+This hook will store the previous value of a given variable.
+
+```tsx
+function Score({value}) {
+  const previousValue = usePrevious(value);
+  return (
+    <>
+      <p>{value}</p>
```
loic-d: Nit but maybe we could have something like `<p>Current score: {value}</p>`

packages/react-hooks/README.md
```diff
@@ -84,3 +84,19 @@ function MockComponent() {
...
+      {previousValue < value && <p>New record!</p>}
```
loic-d: Nit but I would write the condition the other way around: "If the next scrore is greater than the previous one, we have a new record"

`value > previousValue && <p>We have a new record!</p>`

packages/react-hooks/src/hooks/tests/previous.test.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react';
+import {mount} from '@shopify/react-testing';
+
+import usePrevious from '../previous';
+
+describe('usePrevious', () => {
+  function MockComponent({value}) {
+    const previousValue = usePrevious(value);
```
loic-d: Should we re-use your 'real-world' example in the test?

packages/react-hooks/README.md
```diff
@@ -84,3 +84,19 @@ function MockComponent() {
...
+
+This hook will store the previous value of a given variable.
+
+```tsx
+function Score({value}) {
+  const previousValue = usePrevious(value);
+  return (
+    <>
+      <p>Current score: {value}</p>
+      {value > previousValue && <p>We have a new record!</p>}
```
ismail-syed: Nit. This coding style feels a bit off from how we generally handle conditionally rendering content.  I would suggest pulling it out 

```tsx
const newRecord = value > previousValue ? <p>We have a new record!</p> : null;

return ( 
...
  {newRecord}
...
) 
```

sylvhama: Done!

PR #1162, opened 2019-10-30
Fix translationKeyExists to work with onError handler

mgingras:
## Description
With custom onError handler in i18n `translate` doesn't raise an exception causing `translationKeyExists` to always return true.

Also considered testing if the resulting string from translate was empty AND we weren't using the default on error. i.e.

```ts
translationKeyExists(id: string) {
  try {
    const translation = this.translate(id);
    return translation !== '' || this.onError !== defaultOnError;
  } catch (error) {
    return false;
  }
}
```

## Type of change

- [x] `react-i18n` Patch: Bug/ Documentation fix (non-breaking change which fixes an issue or adds documentation)

## Checklist

### DIFF
--- a/packages/react-i18n/src/i18n.ts
+++ b/packages/react-i18n/src/i18n.ts
@@ -206,7 +206,7 @@ export class I18n {
 
   translationKeyExists(id: string) {
     try {
-      this.translate(id);
+      getTranslationTree(id, this.translations, this.locale);
       return true;
     } catch (error) {
       return false;
--- a/packages/react-i18n/CHANGELOG.md
+++ b/packages/react-i18n/CHANGELOG.md
@@ -6,7 +6,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ---
 
-<!-- ## [Unreleased] -->
+## [Unreleased]
+
+### Fixed
+
+- `translationKeyExists` on i18n works as expected with onError handler ([#1162](https://github.com/Shopify/quilt/pull/1162))
+
 
 ## [2.1.7] - 2019-11-12
 
--- a/packages/react-i18n/src/test/i18n.test.ts
+++ b/packages/react-i18n/src/test/i18n.test.ts
@@ -1451,8 +1451,8 @@ describe('I18n', () => {
 
   describe('#translationKeyExists', () => {
     it('returns true if the translation key exists', () => {
-      const mockResult = 'translated string';
-      translate.mockReturnValue(mockResult);
+      const mockResult = {hello: 'translated string'};
+      getTranslationTree.mockReturnValue(mockResult);
 
       const i18n = new I18n(defaultTranslations, defaultDetails);
       const result = i18n.translationKeyExists('hello');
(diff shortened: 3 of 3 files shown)


CameronGorrie:
There was [some discussion](https://shopify.slack.com/archives/CCNRSKLTH/p1570716399032000) that pointed to removing this utility altogether

mgingras:
Interesting, @CameronGorrie . The case we're trying to handle at the moment is where we have translations for a subset of things, and in the case we don't have a translation want to fall back to a default. Not sure how we'd handle it right now without the ability to check if a translation exists, and trying to translate blindly isn't great because it always throws an error.

packages/react-i18n/src/i18n.ts
```diff
@@ -204,11 +204,15 @@ export class I18n {
   }
 
   translationKeyExists(id: string) {
+    const errorHandler = this.onError;
     try {
+      (this.onError as any) = defaultOnError;
```
lemonmade: I know this works but it feels a little icky to me. I'd rather do a proper check for whether the key exists rather than relying on the exception for control flow, especially now that it requires weird mutations of self

CameronGorrie: This property should use the `private` modifier instead of being casted

mgingras: How do you do that? :)

mgingras: Tried again, lmk what you think. 😄

CameronGorrie: I think Chris and I submitted our comments around the same time so my point isn't valid now, but the reason you had to cast the property is because `onError` was declared using the `readonly` modifier so you couldn't reassign this member without tricking the compiler. Changing the modifier to `private` would have allowed reassignment without the cast. [More info](https://www.typescriptlang.org/docs/handbook/classes.html#public-private-and-protected-modifiers)

CameronGorrie:
I made a mistake porting this utility directly from the legacy translation library in web by not thinking too much about why it was being used. Perhaps there are cases when this utility is necessary but I'm thinking there's a solution to what you're trying to do.

> The case we're trying to handle at the moment is where we have translations for a subset of things, and in the case we don't have a translation want to fall back to a default

Would a function containing a dictionary of translations, a lookup, and optional fallback work? Maybe something like this:

```ts
function getTranslation(subset: string) {
  const dictionary = {
    subsetOne: i18n.translate('MyComponent.stringOne'),
    subsetTwo: i18n.translate('MyComponent.stringTwo'),
    default: i18n.translate('MyComponent.default'),
  }

  return dictionary[subset] || dictionary.default;
}
```

packages/react-i18n/src/i18n.ts
mgingras: Cool, thanks for the details

mgingras:
That would work, however then we're tracking what's defined both in the translation as well as in the business logic.

mgingras:
@CameronGorrie @lemonmade any further thoughts, as it stands this fn doesn't work, if this fix isn't the right approach, I'm happy to try something else, or we could remove it

CameronGorrie:
My vote would be to remove it. There's only a single usage in Web so I would assume other apps are not using it heavily

alxclark:
👋 We are currently using this fn for the Checkout Settings page in the following context 👇 

https://github.com/Shopify/web/blob/master/app/sections/Checkout/CheckoutSettings/components/CheckoutLanguage/CheckoutLanguage.tsx#L30-L31

We had to introduce this since calling `i18n.translate` with a key that doesn't exist results in the following error.

image

I am not sure that defining a `getTranslation` fn for all the locale in our context would be best since adding new translations would imply updating everytime a new locale is added.

PR #1163, opened 2019-10-30
Fix folder entries

TheMallen:
### What this?
This PR fixes a bug in `react-server-webpack-plugin` where any entrypoint nested in a server would not be detected. Previously, something like `server/index.ts` would not be detected and would be clobbered with a virtual module. Now it will work as expected.

## 🎩 help
- checkout this branch
- clone https://github.com/Shopify/shrink-ray/pull/512
- `yarn tophat` and answer the prompts `react-server-webpack-plugin` and `../shrink-ray`
- run the server, you should see the logs and the app should work.


packages/react-server-webpack-plugin/src/test/fixtures/client-entrypoint/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "react-server-webpack-plugin/client-entrypoint",
+  "name": "react-client-webpack-plugin/server-entrypoint",
```
cartogram: ?

TheMallen: oops lol

PR #1164, opened 2019-11-04
Remove all react-server-webpack-plugin fixtures

cartogram:
## Description
This PR is the first step to improving `react-server-webpack-plugin`'s tests. In this PR I am removing all of the fixtures, but in doing so we now have a space for building up an environment and then tearing it down during each test. The follow up PRs will use this structure to spawn a server, browser, etc... in order to more e2e style tests.

I've borrowed a lot from the `web-worker` package in this repo and will eventually propose some sort of shared utilities, but I don't want to start this process too soon.


packages/react-server-webpack-plugin/src/test/react-server-webpack-plugin.test.ts
```diff
@@ -201,3 +289,81 @@ function getModule(results: any, basePath: string) {
...
+  "version": "0.0.0",
+  "license": "MIT",
+  "description": "A fixture",
+  "dependencies": {
+    "webpack-virtual-modules": "^0.1.12"
+  },
+  "devDependencies": {
+    "@shopify/react-server": "^0.1.0",
+    "typescript": "~3.2.1",
+    "webpack": "^4.39.1"
```
lemonmade: Does it need to have dependencies?

packages/react-server-webpack-plugin/src/test/react-server-webpack-plugin.test.ts
```diff
@@ -28,117 +39,194 @@ describe('react-server-webpack-plugin', () => {
...
+        const customConfig = {
+          port: 3000,
+          host: '127.0.0.1',
+          assetPrefix: 'https://localhost/webpack/assets',
+          basePath: '.',
+        };
+
+        await withWorkspace(name, async ({workspace}) => {
+          await workspace.write('index.js', BASIC_JS_MODULE);
+          await workspace.write('package.json', BASIC_PACKAGE_JSON);
```
lemonmade: Not sure why you'd need a package.json for tests...

PR #1165, opened 2019-11-05
update `eslint-plugin-shopify` v32x

cartogram:
## Description

This PR updates `eslint-plugin-shopify` to version 32x. The main changes here are the grouping of imports and [requiring top-level describe](https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/require-top-level-describe.md) when using jest hooks in tests. Otherwise this is a really straightforward bump.

### DIFF
--- a/packages/dates/src/tests/is-less-than-one-minute-ago.test.ts
+++ b/packages/dates/src/tests/is-less-than-one-minute-ago.test.ts
@@ -1,4 +1,5 @@
 import {clock} from '@shopify/jest-dom-mocks';
+
 import {isLessThanOneMinuteAgo} from '../is-less-than-one-minute-ago';
 
 describe('isLessThanOneMinuteAgo()', () => {
--- a/test/each-test.ts
+++ b/test/each-test.ts
@@ -11,6 +11,7 @@ if (typeof window !== 'undefined') {
   addClosest(window);
 }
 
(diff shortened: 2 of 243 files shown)


packages/dates/src/tests/is-less-than-one-minute-ago.test.ts
```diff
@@ -1,4 +1,5 @@
 import {clock} from '@shopify/jest-dom-mocks';
+
```
lemonmade: Hmm I am pretty surprised we are enforcing this.

test/each-test.ts
```diff
@@ -11,6 +11,8 @@ if (typeof window !== 'undefined') {
   addClosest(window);
 }
 
-afterEach(() => {
-  destroyAll();
+describe('destroyAll()', () => {
```
lemonmade: No, this is wrong. This was intended to be a global `afterEach` hook.

cartogram: Im going to revert it and ignore the rule.

packages/dates/src/tests/is-less-than-one-minute-ago.test.ts
cartogram: Its not something I would have added, but if people want it I think its fine.

PR #1173, opened 2019-11-08
React i18n webpack plugin

michenly:
## Description
Part of https://github.com/Shopify/sewing-kit/issues/1572

## Type of change

This is a new package

## To 🎩 

In consuming project
remove the babel plugin from sewing-kit in `node_modules/@shopify/sewing-kit/dist/src/tools/webpack/config/rules.js`

```
// remove babel plugin from sewing-kit
yarn add @shopify/sewing-kit@0.111.1-remove-babel-beta.4 

// this branch up to `ReactI18nCacheBusterPlugin`
yarn add @shopify/react-i18n-webpack-plugin@0.0.0-cache-buster-beta.2
```

In quilt
- `yarn tophat react-i18n-webpack-plugin ../path-to-project`

Do performance testing using instructions here https://github.com/Shopify/sewing-kit/issues/1572#issuecomment-555037501 

-----------
#### Using CI
Baseline vs ReactI18nPlugin On CI
Build time: 10m 33s vs 9m 28s
Size: +31B vs +878,046 B (+1.1%)

#### Using local laptop
MacBook Pro 13" 2.7Ghz Core i7, 16 GM Memory

##### Baseline:
https://github.com/Shopify/web/pull/20935

✅ All 635 assets have valid integrity hashes

1st Run: no cache
[client (baseline)] Finished compiling (10m16s, 3613 MB heap used)

2nd Run: with cache
[client (baseline)] Finished compiling (5m37s, 3478 MB heap used)

3rd Run: with cache
[client (baseline)] Finished compiling (5m36s, 3731 MB heap used)

##### Webpack Plugin: `ReactI18nPlugin`
https://github.com/Shopify/web/pull/20927

✅ All 1389 assets have valid integrity hashes

1st Run: no cached
[client (baseline)] Finished compiling (15m06s, 4362 MB heap used)

2nd Run: with cache
[client (baseline)] Finished compiling (10m17s, 4903 MB heap used)

3rd Run: with cache
[client (baseline)] Finished compiling (8m31s, 3265 MB heap used)

4th Run: with cache
[client (baseline)] Finished compiling (8m15s, 4143 MB heap used)

5th Run: with cache
[client (baseline)] Finished compiling (11m22s, 3962 MB heap used)

##### Webpack Cache Buster Plugin + babel Plugin (web)
https://github.com/Shopify/web/pull/20927 (commit 3)
`ReactI18nCacheBusterPlugin` + `'@shopify/react-i18n-webpack-plugin/babel`

✅ All 1389 assets have valid integrity hashes

1st Run: no cached
[client (baseline)] Finished compiling (17m41s, 5723 MB heap used)

2nd Run: with cache
[client (baseline)] Finished compiling (6m23s, 4847 MB heap used)

3rd Run: with cache
[client (baseline)] Finished compiling (6m38s, 4981 MB heap used)

4th Run: with cache
[client (baseline)] Finished compiling (6m19s, 5216 MB heap used)

5th Run: with cache
[client (baseline)] Finished compiling (6m13s, 4199 MB heap used)

##### Index File Generation + babel (web)
https://github.com/Shopify/web/pull/21090
`node translations.js` + `plugins.push(['@shopify/react-i18n-webpack-plugin/babel', {mode: 'from-generated-index'}]);`

✅ All 1392 assets have valid integrity hashes

1st Run: no cached
[client (baseline)] Finished compiling (10m57s, 4599 MB heap used)

2nd Run: with cache
[client (baseline)] Finished compiling (7m20s, 6158 MB heap used)

3rd Run: with cache
[client (baseline)] Finished compiling (9m01s, 4850 MB heap used)

4th Run: with cache
[client (baseline)] Finished compiling (8m26s, 5350 MB heap used)

5th Run: with cache
[client (baseline)] Finished compiling (8m06s, 5160 MB heap used)

##### Webpack Plugin: `ReactI18nPlugin` with `compiler.hooks.beforeCompile.tapAsync`
✅ All 1389 assets have valid integrity hashes

1st Run: no cached
[client (baseline)] Finished compiling (23m29s, 4866 MB heap used)

2nd Run: with cache
[client (baseline)] Finished compiling (6m46s, 4046 MB heap used)

3rd Run: with cache
[client (baseline)] Finished compiling (7m11s, 5467 MB heap used)

### DIFF
--- a/packages/react-i18n-webpack-plugin/src/babel-plugin.ts
+++ b/packages/react-i18n-webpack-plugin/src/babel-plugin.ts
@@ -1,0 +1,169 @@
+import path from 'path';
+
+import glob from 'glob';
+import {TemplateBuilder} from '@babel/template';
+import Types from '@babel/types';
+import {Node, NodePath} from '@babel/traverse';
+import stringHash from 'string-hash';
+
+interface State {
+  program: NodePath<Types.Program>;
+}
+
+export default function injectWithI18nArguments({
+  types: t,
+  template,
+}: {
+  types: typeof Types;
+  template: TemplateBuilder<Types.ImportDeclaration | Types.ObjectExpression>;
+}) {
+  function fallbackTranslationsImport({id}) {
+    return template(`import ${id} from './translations/en.json';`, {
+      sourceType: 'module',
+    })() as Types.ImportDeclaration;
+  }
+
+  function i18nCallExpression({id, fallbackID, bindingName, translations}) {
+    return template(
+      `${bindingName}({
+        id: '${id}',
+        fallback: ${fallbackID},
+        translations(locale) {
+          const __shopify__i18n_translations = [${translations
+            .filter(locale => !locale.endsWith('en.json'))
+            .map(locale =>
+              JSON.stringify(path.basename(locale, path.extname(locale))),
+            )
+            .sort()
+            .join(', ')}];
+          if (__shopify__i18n_translations.indexOf(locale) < 0) {
+            return;
+          }
+          return (async () => {
+            const dictionary = await import(/* webpackChunkName: "${id}-i18n", webpackMode: "lazy-once" */ \`./translations/$\{locale}.json\`);
+            return dictionary && dictionary.default;
+          })();
+        },
+      })`,
+      {
+        sourceType: 'module',
+        plugins: ['dynamicImport'],
+        preserveComments: true,
+      },
+    )();
+  }
+
+  function addI18nArguments({
+    binding,
+    bindingName,
+    filename,
+    fallbackID,
+    insertImport,
+  }) {
+    const {referencePaths} = binding;
+
+    const referencePathsToRewrite = referencePaths.filter(referencePath => {
+      const parent: Node = referencePath.parent;
+      return (
+        parent.type === 'CallExpression' &&
+        (!parent.arguments || parent.arguments.length === 0)
+      );
+    });
+
+    if (referencePathsToRewrite.length === 0) {
+      return;
+    }
+
+    if (referencePathsToRewrite.length > 1) {
+      throw new Error(
+        `You attempted to use ${bindingName} ${
+          referencePathsToRewrite.length
+        } times in a single file. This is not supported by the Babel plugin that automatically inserts translations.`,
+      );
+    }
+
+    const translations = getTranslations(filename);
+
+    if (translations.length === 0) {
+      return;
+    }
+
+    insertImport();
+
+    referencePathsToRewrite[0].parentPath.replaceWith(
+      i18nCallExpression({
+        id: generateID(filename),
+        fallbackID,
+        bindingName,
+        translations,
+      }),
+    );
+  }
+
+  return {
+    visitor: {
+      Program(nodePath: NodePath<Types.Program>, state: State) {
+        state.program = nodePath;
+      },
+      ImportDeclaration(
+        nodePath: NodePath<Types.ImportDeclaration>,
+        state: State,
+      ) {
+        if (nodePath.node.source.value !== '@shopify/react-i18n') {
+          return;
+        }
+        const {specifiers} = nodePath.node;
+        for (const specifier of specifiers) {
+          if (
+            !t.isImportSpecifier(specifier) ||
+            (specifier.imported.name !== 'withI18n' &&
+              specifier.imported.name !== 'useI18n')
+          ) {
+            continue;
+          }
+
+          const bindingName = specifier.local.name;
+          const binding = nodePath.scope.getBinding(bindingName);
+
+          if (binding != null) {
+            const fallbackID = nodePath.scope.generateUidIdentifier('en').name;
+            const {filename} = this.file.opts;
+
+            addI18nArguments({
+              binding,
+              bindingName,
+              filename,
+              fallbackID,
+              insertImport() {
+                const {program} = state;
+                program.node.body.unshift(
+                  fallbackTranslationsImport({id: fallbackID}),
+                );
+              },
+            });
+          }
+        }
+      },
+    },
+  };
+}
+
+function getTranslations(filename: string): string[] {
+  return glob.sync(
+    path.resolve(path.dirname(filename), 'translations', '*.json'),
+    {
+      nodir: true,
+    },
+  );
+}
+
+// based on postcss-modules implementation
+// see https://github.com/css-modules/postcss-modules/blob/60920a97b165885683c41655e4ca594d15ec2aa0/src/generateScopedName.js
+function generateID(filename: string) {
+  const hash = stringHash(filename)
+    .toString(36)
+    .substr(0, 5);
+  const extension = path.extname(filename);
+  const legible = path.basename(filename, extension);
+  return `${legible}_${hash}`;
+}
--- a/packages/react-i18n-webpack-plugin/src/babel-plugin.ts
+++ b/packages/react-i18n-webpack-plugin/src/babel-plugin.ts
@@ -23,20 +23,35 @@ export default function injectWithI18nArguments({
     })() as Types.ImportDeclaration;
   }
 
-  function i18nCallExpression({id, fallbackID, bindingName, translations}) {
+  function translationsImport({id}) {
+    return template(`import ${id} from './translations';`, {
+      sourceType: 'module',
+    })() as Types.ImportDeclaration;
+  }
+
+  function i18nCallExpression({
+    id,
+    fallbackID,
+    translationID,
+    bindingName,
+    translations,
+  }) {
+    const translationArrayString = translationID
+      ? translationID
+      : `[${translations
+          .filter(locale => !locale.endsWith('en.json'))
+          .map(locale =>
+            JSON.stringify(path.basename(locale, path.extname(locale))),
+          )
+          .sort()
+          .join(', ')}]`;
+
     return template(
       `${bindingName}({
         id: '${id}',
         fallback: ${fallbackID},
         translations(locale) {
-          const __shopify__i18n_translations = [${translations
-            .filter(locale => !locale.endsWith('en.json'))
-            .map(locale =>
-              JSON.stringify(path.basename(locale, path.extname(locale))),
-            )
-            .sort()
-            .join(', ')}];
-          if (__shopify__i18n_translations.indexOf(locale) < 0) {
+          if (${translationArrayString}.indexOf(locale) < 0) {
             return;
           }
           return (async () => {
@@ -58,6 +73,7 @@ export default function injectWithI18nArguments({
     bindingName,
     filename,
     fallbackID,
+    translationID,
     insertImport,
   }) {
     const {referencePaths} = binding;
@@ -94,6 +110,7 @@ export default function injectWithI18nArguments({
       i18nCallExpression({
         id: generateID(filename),
         fallbackID,
+        translationID,
         bindingName,
         translations,
       }),
@@ -109,6 +126,8 @@ export default function injectWithI18nArguments({
         nodePath: NodePath<Types.ImportDeclaration>,
         state: State,
       ) {
+        const fromGeneratedIndex =
+          (state as any).opts.mode === 'from-generated-index' || false;
         if (nodePath.node.source.value !== '@shopify/react-i18n') {
           return;
         }
@@ -127,6 +146,9 @@ export default function injectWithI18nArguments({
 
           if (binding != null) {
             const fallbackID = nodePath.scope.generateUidIdentifier('en').name;
+            const translationID = fromGeneratedIndex
+              ? nodePath.scope.generateUidIdentifier('translation').name
+              : undefined;
             const {filename} = this.file.opts;
 
             addI18nArguments({
@@ -134,11 +156,18 @@ export default function injectWithI18nArguments({
               bindingName,
               filename,
               fallbackID,
+              translationID,
               insertImport() {
                 const {program} = state;
                 program.node.body.unshift(
                   fallbackTranslationsImport({id: fallbackID}),
                 );
+
+                if (fromGeneratedIndex) {
+                  program.node.body.unshift(
+                    translationsImport({id: translationID}),
+                  );
+                }
               },
             });
           }
--- a/packages/react-i18n-webpack-plugin/src/react-i18n-parser-plugin.ts
+++ b/packages/react-i18n-webpack-plugin/src/react-i18n-parser-plugin.ts
@@ -1,0 +1,146 @@
+import VirtualModulesPlugin from 'webpack-virtual-modules';
+import ParserHelpers from 'webpack/lib/ParserHelpers';
+import {CallExpression} from 'estree';
+
+import {buildI18nOptions} from './i18n-options';
+import {generateID} from './utilities';
+
+const PLUGIN_NAME = 'ReactI18nParserPlugin';
+const TRANSLATION_DIRECTORY_NAME = 'translations';
+const I18N_CALL_NAMES = ['useI18n', 'withI18n'];
+
+export interface Options {
+  fallbackLocale: string;
+  currentLocale?: string;
+}
+
+export class ReactI18nParserPlugin {
+  private options: Options;
+  private virtualModules: VirtualModulesPlugin;
+
+  constructor(options: Options, virtualModules: VirtualModulesPlugin) {
+    this.options = options;
+    this.virtualModules = virtualModules;
+  }
+
+  apply(parser) {
+    // get all the import and the identifier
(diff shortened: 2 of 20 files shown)


templates/ROOT_README.hbs.md
```diff
@@ -43,4 +43,4 @@ For Shopifolk, you can reach out to us in Slack in the `#web-foundation-tech` ch
 
 MIT &copy; [Shopify](https://shopify.com/), see [LICENSE.md](LICENSE.md) for details.
 
-<a href="http://www.shopify.com/"><img src="https://cdn.shopify.com/assets2/press/brand/shopify-logo-main-small-f029fcaf14649a054509f6790ce2ce94d1f1c037b4015b4f106c5a67ab033f5b.png" alt="Shopify" width="200" /></a>
+<a href="http://www.shopify.com/"><img src="https://cdn.shopify.com/assets2/brand-assets/shopify-logo-main-8ee1e0052baf87fd9698ceff7cbc01cc36a89170212ad227db3ff2706e89fd04.svg" alt="Shopify" width="200" /></a>
```
michenly: update because of https://github.com/Shopify/quilt/pull/1139

packages/react-i18n-webpack-plugin/src/react-i18n-webpack-plugin.ts
```diff
@@ -0,0 +1,264 @@
...
+    compiler.hooks.compilation.tap(
+      PLUGIN_NAME,
+      (
+        compilation: webpack.compilation.Compilation,
+        {
+          normalModuleFactory,
+        }: {normalModuleFactory: webpack.compilation.NormalModuleFactory},
+      ) => {
+        const handler = (parser: any) => {
+          new FindI18nImportPlugin().apply(parser);
```
lemonmade: I find it way more confusing to have this as a separate plugin, personally

packages/react-i18n-webpack-plugin/src/react-i18n-webpack-plugin.ts
```diff
@@ -0,0 +1,264 @@
...
+
+          // replace useI18n & withI18n call arguments
+          parser.hooks.evaluate
```
lemonmade: So to be clear, https://webpack.js.org/api/parser/#call doesn't work here?

packages/react-i18n-webpack-plugin/src/react-i18n-webpack-plugin.ts
```diff
@@ -0,0 +1,264 @@
...
+            .for('CallExpression')
+            .tap(PLUGIN_NAME, (originalExpression: CallExpression) => {
+              if (
+                parser.state.module.resource.indexOf('node_modules') !== -1 &&
+                !parser.state.module.resource.endsWith('tsx') &&
```
lemonmade: These two conditions don't make sense to me, can you explain why they are needed?

packages/react-i18n-webpack-plugin/src/react-i18n-webpack-plugin.ts
```diff
@@ -0,0 +1,264 @@
...
+                !parser.state.i18nImports
+              ) {
+                return;
+              }
+
+              const componentPath = parser.state.module.resource;
+              const componentDir = parser.state.module.context;
+              const identifierName = parser.state.i18nImports.get(
```
lemonmade: What if a single file uses `useI18n` and `withI18n`?

packages/react-i18n-webpack-plugin/src/react-i18n-webpack-plugin.ts
```diff
@@ -0,0 +1,264 @@
...
+
+              if (
+                !identifierName ||
+                originalExpression.callee.type !== 'Identifier'
+              ) {
+                return;
+              }
+
+              let expression: CallExpression | null = null;
+              if (originalExpression.callee.name === 'compose') {
```
lemonmade: I feel like we should never have to care if `compose` is called or not. Is this hook not called for every call expression?

packages/react-i18n-webpack-plugin/src/react-i18n-webpack-plugin.ts
```diff
@@ -0,0 +1,264 @@
...
+        normalModuleFactory.hooks.parser
+          .for('javascript/dynamic')
+          .tap('HarmonyModulesPlugin', handler);
+      },
+    );
+  }
+}
+
+// Return a list of translationFiles name	function i18nCallExpression(
+function getTranslationFiles(parser: any): string[] {
```
lemonmade: It seems odd to pass the parser here; don't you have the parent directory of the file being processed?

packages/react-i18n-webpack-plugin/src/react-i18n-webpack-plugin.ts
```diff
@@ -0,0 +1,264 @@
...
+
+        normalModuleFactory.hooks.parser
+          .for('javascript/dynamic')
+          .tap('HarmonyModulesPlugin', handler);
+      },
+    );
+  }
+}
+
+// Return a list of translationFiles name	function i18nCallExpression(
```
lemonmade: Looks like a weird artifact of copy-pasting from Github?

packages/react-i18n-webpack-plugin/src/react-i18n-webpack-plugin.ts
```diff
@@ -0,0 +1,264 @@
...
+    );
+  } catch (error) {
+    // do nothing if the directory does not exist
+  }
+
+  return translationFiles;
+}
+
+function getChunkName(id: string) {
+  return `${id}-i18n`;
```
lemonmade: Seems too small to be its own little function?

packages/react-i18n-webpack-plugin/src/react-i18n-webpack-plugin.ts
```diff
@@ -0,0 +1,264 @@
...
+                return;
+              }
+
+              const translationFiles = getTranslationFiles(parser);
+              if (translationFiles.length === 0) {
+                return;
+              }
+
+              // check if fall back exist
+              const fallBackExist = translationFiles.find(
```
lemonmade: `fallback` is one word, also do we just want `includes` here?

packages/react-i18n-webpack-plugin/src/react-i18n-webpack-plugin.ts
```diff
@@ -0,0 +1,264 @@
...
+                translationFile =>
+                  translationFile === `${this.options.fallbackLocale}.json`,
+              );
+
+              const fallBackFileRelativePath = path.join(
+                './',
```
lemonmade: IMO it is more confusing to just `path.join` here instead of just doing a template literal string

packages/react-i18n-webpack-plugin/src/react-i18n-webpack-plugin.ts
```diff
@@ -0,0 +1,264 @@
...
+                TRANSLATION_DIRECTORY_NAME,
+                `${this.options.fallbackLocale}.json`,
+              );
+
+              if (!fallBackExist) {
+                compilation.errors.push(
```
lemonmade: Do we need to error?

packages/react-i18n-webpack-plugin/src/react-i18n-webpack-plugin.ts
```diff
@@ -0,0 +1,264 @@
...
+                    `${identifierName} 's arguments was not automatically filled in because` +
+                    `fallback translation file was not found at ${fallBackFileRelativePath} \n`,
+                );
+                return;
+              }
+
+              // Add a top-level fallbackLocale import
+              const fallbackLocaleID = camelCase(this.options.fallbackLocale);
+
+              const fallbackFileExpression = ParserHelpers.requireFileAsExpression(
```
lemonmade: This seems to create a `require` statement, which I *think* doptimizes some things during bundling. Can it use a default import instead?

packages/react-i18n-webpack-plugin/src/react-i18n-webpack-plugin.ts
```diff
@@ -0,0 +1,264 @@
...
+                compilation.errors.push(
+                  `${componentPath}\n` +
+                    `${identifierName} 's arguments was not automatically filled in because` +
+                    `fallback translation file was not found at ${fallBackFileRelativePath} \n`,
+                );
+                return;
+              }
+
+              // Add a top-level fallbackLocale import
+              const fallbackLocaleID = camelCase(this.options.fallbackLocale);
```
lemonmade: Hmm I feel like we need a "safer" name; this looks like it will just be `en`, which has a non-zero chance of conflicting with an actual variable in scope.

packages/react-i18n-webpack-plugin/src/react-i18n-webpack-plugin.ts
michenly: It does not. This tap wont trigger. I really wish it did!

packages/react-i18n-webpack-plugin/src/react-i18n-webpack-plugin.ts
michenly: 😂

packages/react-i18n-webpack-plugin/src/react-i18n-webpack-plugin.ts
michenly: mm...we will be in trouble here...

packages/react-i18n-webpack-plugin/src/react-i18n-webpack-plugin.ts
michenly: It does not. In the case of `compose`, this hook is called once with compose with `withI18n` as argument but no again with `withI18n`.

packages/react-i18n-webpack-plugin/src/react-i18n-webpack-plugin.ts
michenly: It's up to us. IMO I would like to warm the consumer that the fallback does not exist and it should be fix.
I debate between error or warning but landed on error since this should NEVER happen (ie. a translation folder exist with .json files but no fallback)

packages/react-i18n-webpack-plugin/src/react-i18n-webpack-plugin.ts
michenly: What about `en-US`? If I am making `fallbackLocale` a option it may not be `en` in the future

packages/react-i18n-webpack-plugin/src/react-i18n-webpack-plugin.ts
michenly: maybe... 😹 just need to figure out how
but also, do we not want the optimization here?

packages/react-i18n-webpack-plugin/src/react-i18n-webpack-plugin.ts
michenly: I mainly need `parser.state.compilation.compiler.inputFileSystem` from parser.

packages/react-i18n-webpack-plugin/src/react-i18n-webpack-plugin.ts
michenly: first one being if the code is in `node_modules` we do not want to transform it. This one I am sure of.
the second one I can remove. Maybe someone would be using `jsx` or `js`

packages/react-i18n-webpack-plugin/src/react-i18n-webpack-plugin.ts
lemonmade: Would it be definitely wrong if they had no fallback? I don't think we know it will be, and it's possible it could be an explicit choice. In the end, we should just stick to whatever the current plugin does to cut some scope.

packages/react-i18n-webpack-plugin/src/react-i18n-webpack-plugin.ts
lemonmade: sure, `enUs` is not safe either. Any name that is just camelcasing could conflict with the app code. The usual approach is to use a variable name that is very unlikely to conflict (one with random characters at the end, double underscore prefix, etc)

packages/react-i18n-webpack-plugin/src/react-i18n-webpack-plugin.ts
lemonmade: We never want to deoptimize the way the module is compiled.

packages/react-i18n-webpack-plugin/src/react-i18n-webpack-plugin.ts
michenly: aw, I thought the problem was the `camelCase`, I see what you mean now.

packages/react-i18n-webpack-plugin/src/react-i18n-webpack-plugin.ts
michenly: Didnt really resolved this yet. But it does seem it will take some time to research. Going to concentrate on shipping this first and do this as a improvement toward the end.

packages/react-i18n-webpack-plugin/src/react-i18n-webpack-plugin.ts
GoodForOneFare: Confusing for me, too.  If we're trying to slim down the body of this function, there are better parts to refactor into helper functions.

michenly: ok, moving it back in!

packages/react-i18n-webpack-plugin/src/react-i18n-parser-plugin.ts
```diff
@@ -0,0 +1,148 @@
...
+
+        const componentPath = parser.state.module.resource;
+        const componentDir = parser.state.module.context;
+        const importIdentifiers:
+          | string[]
+          | undefined = parser.state.i18nImports.get(componentPath);
+
+        if (
+          !importIdentifiers ||
+          originalExpression.callee.type !== 'Identifier'
```
cartogram: Is there a reason not to use babel/types in these conditionals? For example, `t.isIdentifier(node)` from https://babeljs.io/docs/en/babel-types#identifier.

michenly: cool I will give that a try! thou more dependency could mean even slower...><...

cartogram: True! Not saying you should use that module, I was just curious 🤔

packages/react-i18n-webpack-plugin/src/babel-plugin.ts
```diff
@@ -0,0 +1,169 @@
...
+  }
+
+  return {
+    visitor: {
+      Program(nodePath: NodePath<Types.Program>, state: State) {
+        state.program = nodePath;
+      },
+      ImportDeclaration(
+        nodePath: NodePath<Types.ImportDeclaration>,
+        state: State,
```
cartogram: @michenly for options, you should be able to grab them off this object that is passed to each of the visitors.

michenly: nice! would you also know how to pass the option in from webpack config?

michenly: I got it! Thanks for the hint Matt!

GoodForOneFare:
I've tested out the `from-generated-index` version of this locally and:
- It does not regress warm cache build times in our largest app
- It embeds the correct locale strings in components
- It successfully triggers a webpack recompile when a translations file is added/removed

🎉 

So I would ❤️:
- A PR with the `from-generated-index` option added to the Babel plugin
- A helper function that other tools can call to generate translation index files
- (Follow-on PR) an option to make the helper function generate an index containing specific locales
- Documentation on when/how to use all of this 😄 

e.g.,
```
import {generateTranslationIndexes} from `@shopify/react-i18n`;

// First arg is "fallback"; always first in the translations array.
// Second arg is optional; if specified, only matching locales will be included in `index.js` files.
await generateTranslationIndexes('en', ['fr', 'jp']);
webpack(require('./webpack.config.js'));
```

michenly:
OK closing this experiment with notes and starting a proper PR for the final solution we agree on.

michenly:
Possible future improvement for `ReactI18nPlugin` suggested by @GoodForOneFare 

- Use readdir instead of readdirSync
- Sort the result of readdir (random order will bust caches)
- Do the !parser.state.i18nImports before checking for node_modules

Issue #1099, opened 2019-10-07
Enable cookies redirection look [Type: Bug, Package: @shopify/koa-shopify-auth]


Popesites:
## Overview

I've been experiencing a number of issues with the authentication process for my app since updating the shopify koa auth package to 3.1.36. 

One of these issues is the following --

On Safari, when I open a fresh browser window with no cookies / cache at all, sign into my Shopify store, and open up my app, I am sent to a redirection look. 

Here's what it looks like:

Enable cookies page (https://<app_domain>/auth/enable_cookies) -> click "Enable Cookies" -> user gets redirected to https://<app_domain>/admin/apps/<some_hash> -> user gets redirected to https://<app_domain>/auth/enable_cookies again

## Consuming repo

> What repo were you working in when this issue occurred?

https://github.com/Shopify/quilt/tree/master/packages/koa-shopify-auth

Popesites:
Here is a video displaying this behavior:

noctivityinc:
@Popesites were you able to figure this out?  Im seeing the same thing.

Popesites:
@noctivityinc nope, still nothing.

gil--:
Started noticing this on Safari iOS 13 where it'll redirect back and forth between the cookie screen and the app in the admin. Going to try to look into this soon as it seems to be affecting merchants.

justjoeyuk:
@dimapaloskin how did you force these attributes?

dimapaloskin:
@justjoeyuk clone [https://github.com/pillarjs/cookies](https://github.com/pillarjs/cookies) repo, change [cookies `sameSite` setup manually](https://github.com/pillarjs/cookies/blob/master/index.js#L174) (`header += "; samesite=none"`), and link the package to project. 

Note, you need to use `https` (or at least have `X-Forwarded-Proto: https` header), because `koa` [detect `secure` flag automatically](https://github.com/koajs/koa/blob/master/lib/context.js#L164).

BTW, I'm working on the PR to `koa-shopify-auth` right now to fix the problem.

justjoeyuk:
@dimapaloskin This redirection loop also appears to happen on Safari on Mac OS 😕

dimapaloskin:
@justjoeyuk Unfortunately I can't reproduce this bug in safari. What versions of safari and mac os you use?

Does an infinity loop occur with the `SameSite=None; Secure`?

dimapaloskin:
Probably it will help someone.

[...]

PR #1177, opened 2019-11-12
[koa-shopify-auth] Set cookies with SameSite=none option

dimapaloskin:
## Description

Fixes (issue #1099)

In the latest versions of the chrome `koa-shopify-auth` users faced the problem of infinity redirection. Looks like it's because of [new changes](https://blog.chromium.org/2019/10/developers-get-ready-for-new.html) in `SameSite` behavior. Looks like an app can not access cookies from iframe if cookies were set up outside iframe. To fix that now we need to use `SameSite=None; Secure` settings for each cookie we want to receive inside an iframe. 

I propose to use `sameSite: 'none'` option in all `cookies.set` calls.

## Type of change

- [ ] `koa-shopify-auth` Patch: Bug/ Documentation fix (non-breaking change which fixes an issue or adds documentation)

## Checklist

### DIFF
--- a/packages/koa-shopify-auth/src/auth/create-enable-cookies-redirect.ts
+++ b/packages/koa-shopify-auth/src/auth/create-enable-cookies-redirect.ts
@@ -9,7 +9,9 @@ export default function createEnableCookiesRedirect(path: string) {
 
   return function topLevelOAuthRedirect(ctx: Context) {
     // This is to avoid a redirect loop if the app doesn't use verifyRequest or set the test cookie elsewhere.
-    ctx.cookies.set(TEST_COOKIE_NAME, '1');
+    ctx.cookies.set(TEST_COOKIE_NAME, '1', {
+      sameSite: 'none',
+    });
     redirect(ctx);
   };
 }
--- a/packages/koa-shopify-auth/src/auth/create-oauth-start.ts
+++ b/packages/koa-shopify-auth/src/auth/create-oauth-start.ts
@@ -26,7 +26,9 @@ export default function createOAuthStart(
       return;
     }
 
-    ctx.cookies.set(TOP_LEVEL_OAUTH_COOKIE_NAME);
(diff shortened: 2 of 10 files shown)


packages/koa-shopify-auth/src/auth/create-enable-cookies-redirect.ts
```diff
@@ -9,7 +9,9 @@ export default function createEnableCookiesRedirect(path: string) {
 
   return function topLevelOAuthRedirect(ctx: Context) {
     // This is to avoid a redirect loop if the app doesn't use verifyRequest or set the test cookie elsewhere.
-    ctx.cookies.set(TEST_COOKIE_NAME, '1');
+    ctx.cookies.set(TEST_COOKIE_NAME, '1', {
+      sameSite: 'none',
```
rmtngh: secure: true
https://www.chromestatus.com/feature/5088147346030592

> (Cookies marked SameSite=None should also be marked Secure.)

gil--: Also is this a safe change? Per https://github.com/Shopify/shopify_app/pull/851#issue-361762137 we may need to check the browser agent against a whitelist/blacklist.

> Additionally, SameSite=None (which we do not use today) is unfortunately interpreted incorrectly by specific browsers, so this PR handles such situations as well.

rmtngh: The ideal solution in my opinion would be adding a config option that receives cookie options and letting the devs add the logic around this themselves and remove it whenever other browsers (Safari) decided to fix it!

```
createShopifyAuth({
    apiKey: <SHOPIFY_API_KEY>,
    secret: <SECRET>,
    // If browser != "Safari"
    cookieOptions: {
        sameSite: 'none',
        secure: true
    }
  }),
```

mkevinosullivan:

Thank you!

PR #1178, opened 2019-11-12
✨ added locale to missing translation error

michenly:
## Description

I was debugging something and found it to be really useful if the translation error also included which locale the missing error was from.

sample error message:
```
Error: Missing translation for key: title in locale: fr
```

## Type of change

## Checklist

### DIFF
--- a/packages/react-i18n/src/errors.ts
+++ b/packages/react-i18n/src/errors.ts
@@ -1,8 +1,8 @@
 import {Replacements} from './types';
 
 export class MissingTranslationError extends Error {
-  constructor(key: string) {
-    super(`Missing translation for key: ${key}`);
+  constructor(key: string, locale: string) {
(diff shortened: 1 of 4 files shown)


packages/react-i18n/src/errors.ts
```diff
@@ -1,8 +1,8 @@
 import {Replacements} from './types';
 
 export class MissingTranslationError extends Error {
-  constructor(key: string) {
-    super(`Missing translation for key: ${key}`);
+  constructor(key: string, locale?: string) {
+    super(`Missing translation for key: ${key} in locale: ${locale}`);
```
lemonmade: Seems weird to include the locale part if locale is undefined?

michenly: oops, that should not be there. Removing the ?!!!

PR #1188, opened 2019-11-21
Update babel plugin

michenly:
## Description

Fixes (issue #) https://github.com/Shopify/sewing-kit/issues/1572

## Type of change

## Checklist

## 🎩

Using this branch as consumer:
https://github.com/Shopify/web-proving-ground/pull/40

- in quilt `yarn tophat react-i18n ../../web-proving-ground`
- run `dev up && dev s`
- see dev mode using the original babel plugin
- check `build/server/main.js` and see `if (["fr"].indexOf(locale) < 0) {` in the translation function of `CustomerList`

- run `yarn build`
- see prod mode using the index generation 
- check `/build/server/main.js` and see `_translations__WEBPACK_IMPORTED_MODULE_0__[/* default */ "a"].indexOf(locale) < 0` in the translation function of `CustomerList`


michenly:
I am still adding doc.
But the code & test are good for review!

packages/react-i18n/src/babel-plugin/shared.ts
```diff
@@ -0,0 +1,23 @@
...
+export function getLocaleIds({
+  translationFilePaths,
+  fallbackLocale,
+}: {
+  translationFilePaths: string[];
+  fallbackLocale?: string;
+}): string[] {
+  return translationFilePaths
+    .map(filePath => path.basename(filePath, path.extname(filePath)))
+    .filter(locale => (fallbackLocale ? locale !== fallbackLocale : true))
```
GoodForOneFare: Does it make sense for this to be called without a `fallbackLocale`?  I think we should just throw an exception if this happens.

packages/react-i18n/src/babel-plugin/babel-templates.ts
```diff
@@ -0,0 +1,71 @@
...
+          fallbackLocale,
+        }),
+      );
+
+  return template(
+    `${bindingName}({
+        id: '${id}',
+        fallback: ${fallbackID},
+        translations(locale) {
+          if (${translationArrayString}.indexOf(locale) < 0) {
```
GoodForOneFare: I think this would return false-positives for hyphenated locales.  e.g.:
* `hi` and `hi-IN` are used in our apps
* If we have only `translations/hi-IN.json`
* `"['hi-IN']".indexOf('hi')` would be a bug

packages/react-i18n/src/babel-plugin/index.ts
```diff
@@ -114,45 +98,72 @@ export default function injectWithI18nArguments({
...
-              filename,
-              fallbackID,
-              insertImport() {
-                const {program} = state;
+          if (!binding) {
+            return;
+          }
+
+          const {mode} = state.opts;
+          const fromGeneratedIndex = mode === 'from-generated-index' || false;
```
GoodForOneFare: I think the result of `===` will always be true/false?

```suggestion
          const fromGeneratedIndex = mode === 'from-generated-index';
```

packages/react-i18n/src/babel-plugin/index.ts
```diff
@@ -114,45 +98,72 @@ export default function injectWithI18nArguments({
...
+          const fallbackLocale = DEFAULT_FALLBACK_LOCALE;
+
+          const fallbackID = nodePath.scope.generateUidIdentifier(
+            camelCase(fallbackLocale),
+          ).name;
+          const translationArrayID = fromGeneratedIndex
+            ? `__shopify__i18n_translations_${index}`
```
GoodForOneFare: I think only one `index.js` file can exist, so I'm not sure why `index` is needed?

There's [a check to bail when someone uses `useI18n` / `withI18n` multiple times](https://github.com/Shopify/quilt/blob/01cae2989c7bc430953e8a26bbee1ef20d64e2b6/packages/react-i18n/src/babel-plugin.ts#L79-L84), so I _think_ we can only hit this path once.

packages/react-i18n/README.md
```diff
@@ -394,12 +394,36 @@ useI18n({
...
+      /* webpackChunkName: "MyComponent_<hash>-i18n", webpackMode: "lazy-once" */ `./translations/${locale}.json`
+    );
     return dictionary;
   },
 });
 ```
 
+#### Generate Indexes
+
+Because `babel-loader`'s cache is based on an input source content hash, it is possible for the consumer to encounter a bug where adding new translation file after a cached exist does not change the output. To combat this, we are introducing a new mode for the babel plugin and a helper function that should be run before build.
```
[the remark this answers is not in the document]
GoodForOneFare: Maybe add some notes about what the generator is doing, and "We recommend that you .gitignore `translations/index.js`".

packages/react-i18n/src/babel-plugin/index.ts
```diff
@@ -114,45 +98,72 @@ export default function injectWithI18nArguments({
...
+          if (!binding) {
+            return;
+          }
+
+          const {mode} = state.opts;
+          const fromGeneratedIndex = mode === 'from-generated-index' || false;
+          const fallbackLocale = DEFAULT_FALLBACK_LOCALE;
+
+          const fallbackID = nodePath.scope.generateUidIdentifier(
+            camelCase(fallbackLocale),
```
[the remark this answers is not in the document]
michenly: change the fallbackLocale like `zh-TW` to `zhTW` to be use as a variable name

packages/react-i18n/CHANGELOG.md
```diff
@@ -8,6 +8,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 <!-- ## [Unreleased] -->
 
+## [2.2.0] - 2019-??-??
+
+### Added
+
+- added `generateTranslationIndexes` helper and `mode=from-generated-index` option for babel plugin. This can be use to resolved caching issue when adding new translation file ([#1188](https://github.com/Shopify/quilt/pull/1188))
```
GoodForOneFare:
```suggestion
- Added a `generateTranslationIndexes` helper, and `mode=from-generated-index` option for the Babel plugin. This can be use to resolve unwanted cache hits when adding new translation files ([#1188](https://github.com/Shopify/quilt/pull/1188))
```

packages/react-i18n/src/babel-plugin/index.ts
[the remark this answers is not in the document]
GoodForOneFare: Ah, makes sense 👍 Thank you!

packages/react-i18n/src/babel-plugin/babel-templates.ts
michenly: ['hi-IN'].indexOf('hi') = -1
['hi-IN', 'hi'].indexOf('hi') = 1

I think that make sense??? huh??

michenly: I can change it to use inclIded thou if that make more sense

packages/react-i18n/src/babel-plugin/index.ts
michenly: :+1: will remove

packages/react-i18n/src/babel-plugin/babel-templates.ts
GoodForOneFare: No, you're right.  I was forgetting that this is a template, and that `translationsArrayString` would be an actual array at the time of comparison.

michenly:
all of the requests had been addressed 😸

packages/react-i18n/CHANGELOG.md
```diff
@@ -8,6 +8,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 <!-- ## [Unreleased] -->
 
+## [2.2.0] - 2019-??-??
+
+### Added
+
+- Added a `generateTranslationIndexes` helper, and `mode=from-generated-index` option for the Babel plugin. This can be use to resolve unwanted cache hits when adding new translation files ([#1188](https://github.com/Shopify/quilt/pull/1188))
```
GoodForOneFare:
```suggestion
- Added a `generateTranslationIndexes` helper, and `mode=from-generated-index` option for the Babel plugin. This can be used to resolve unwanted cache hits when adding new translation files ([#1188](https://github.com/Shopify/quilt/pull/1188))
```

packages/react-i18n/src/babel-plugin/generate-index.ts
```diff
@@ -0,0 +1,51 @@
...
+  TRANSLATION_DIRECTORY_NAME,
+  DEFAULT_FALLBACK_LOCALE,
+  getLocaleIds,
+  toArrayString,
+} from './shared';
+
+export function generateTranslationIndexes() {
+  const fallbackLocale = DEFAULT_FALLBACK_LOCALE;
+
+  // execSync is use here instead of Node's glob because it perform much faster
```
GoodForOneFare:
```suggestion
  // `find` is used here instead of Node's glob because it performs much faster
```

packages/react-i18n/src/babel-plugin/generate-index.ts
```diff
@@ -0,0 +1,51 @@
...
+  // (20s vs 1s in web with ~750 translation folders and 21 langs)
```
GoodForOneFare: 👍 very nice stats ❤️

PR #1192, opened 2019-11-25
[@shopify/polyfills] Use src as rootDirectory 

ismail-syed:
## Description

Fixes https://discourse.shopify.io/t/specific-polyfill-imports-from-shopify-polyfills-are-broken/5249

## Checklist

### DIFF
--- a/packages/polyfills/package.json
+++ b/packages/polyfills/package.json
@@ -10,6 +10,9 @@
     "access": "public",
     "@shopify:registry": "https://registry.npmjs.org"
   },
+  "scripts": {
+    "build": "tsc --p tsconfig.json"
+  },
   "author": "Shopify Inc.",
   "repository": {
     "type": "git",
--- a/packages/polyfills/CHANGELOG.md
+++ b/packages/polyfills/CHANGELOG.md
@@ -7,6 +7,10 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+## [1.1.5] - 2019-11-25
+
+- Fix `Can't resolve '@shopify/polyfills/<polyfill>'` import error
+
 ## [1.1.0] - 2019-08-08
 
 - Use `url-polyfill` instead of `url-search-params-polyfill` as it polyfills `URL` as well, including `URL.searchParams`.
--- a/packages/polyfills/src/index.ts
+++ b/packages/polyfills/src/index.ts
@@ -32,7 +32,7 @@ export const polyfills: {[polyfill: string]: PolyfillDescriptor} = {
 export function mappedPolyfillsForEnv(
   env: 'node' | 'jest' | string[],
 ): {[key: string]: string} {
-  const prefix = `@shopify/polyfills/dist`;
+  const prefix = `@shopify/polyfills/dist/src`;
   const noop = `${prefix}/noop`;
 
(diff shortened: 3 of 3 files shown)


package.json
```diff
@@ -11,7 +11,7 @@
     "test": "NODE_ICU_DATA=node_modules/full-icu jest --maxWorkers=3 --coverage=false",
     "check": "lerna run check",
     "release": "lerna publish",
-    "clean": "rimraf './packages/*/dist'",
+    "clean": "rimraf './packages/*/dist' './packages/polyfills/tsconfig.tsbuildinfo'",
```
ismail-syed: Kinda icky but couldn't think of quicker options. The _more correct_ way to do this would be to add a `tsconfig.json#tsBuildInfoFile` path to every package. I don't have the bandwidth to do that right now, this is a _hot fix_-ey kinda PR to unblock consumers.

packages/polyfills/package.json
```diff
@@ -6,6 +6,9 @@
   "sideEffects": true,
   "main": "./dist/src/index.js",
```
GoodForOneFare: If everything's in `dist`, why is `src` still part of this path?

package.json
TheMallen: Why does this fix the problem?

packages/polyfills/package.json
```diff
@@ -6,6 +6,9 @@
...
   "types": "./dist/src/index.d.ts",
+  "scripts": {
+    "build": "tsc --p tsconfig.json"
```
GoodForOneFare: As far as I can tell, this script will never be called.

ismail-syed: It's to maintain consistency with other packages in the repo. It's for cases when you'd want to build just one single package.

packages/polyfills/package.json
ismail-syed: forgot to check-in this change 🤦‍♂

packages/polyfills/tsconfig.json
```diff
@@ -2,8 +2,8 @@
   "extends": "../tsconfig_base.json",
   "compilerOptions": {
     "outDir": "./dist",
-    "baseUrl": ".",
-    "rootDir": "."
+    "baseUrl": "./src",
```
GoodForOneFare: Why doesn't adding `src` to the polyfills prefix fix this without changing the tsconfig?

https://github.com/Shopify/quilt/blob/6a937f9576c6b4398ab1bee5ac6f41bbb4bfdd89/packages/polyfills/src/index.ts#L35

package.json
ismail-syed: Explicitly specifying the `tsconfig.json#tsBuildInfoFile` doesn't fix anything per-se. It just prevents the need for cleaning this lingering  `tsconfig.tsbuildinfo` since we explicitly specify where we want this file to go.

packages/polyfills/tsconfig.json
ismail-syed: I attempted that and re-ran tests in `email`. Tests still fail.

GoodForOneFare: Works for me.  Retest, please.

ismail-syed: Got it working, had to update it in sewing-kit's node_modules as well

PR #1149, opened 2019-10-25
[koa-shopify-auth] Apply prefix to cookie redirect url

timolins:
## Description

Fixes #1148

Apply given prefix to the cookie redirect url in `koa-shopify-auth`.

## Type of change

- [x]  Patch: Bug/ Documentation fix (non-breaking change which fixes an issue or adds documentation)
- [ ]  Minor: New feature (non-breaking change which adds functionality)
- [ ]  Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

- [ ] I have added a changelog entry, prefixed by the type of change noted above

PR #1196, opened 2019-11-26
Consistently use prefix for authentication flow in `koa-shopify-auth` to enable more generic usage

passionkind:
## Description

This fix targets the `create-enable-cookies` part of the authentication flow in `koa-shopify-auth`.

I have added the `prefix` parameter to the redirect url that is being used for `create-enable-cookies`. Here we return html code and request the client to redirect via `window.location.href = "/auth?shop=${shop}"`. 

Not having this prefix there breaks the authentication flow if the middleware is not directly mounted via koa's `.use` handler. 

Let's consider the regular authentication flow:

1. verify the request and start authentication flow on `prefix/auth` if verification fails, otherwise skip re-authentication
2. handle `prefix/auth` and continue with `prefix/enable_cookies`
3. handle `prefix/enable_cookies`and continue with `prefix/auth`
4. handle `prefix/auth` and continue with `prefix/auth/inline`
5. handle `prefix/auth/inline` and continue with `prefix/auth/callback`
6. handle callback and finish authentication flow

Not using the prefix for step 2. would lead to breaking out of the flow unless `koa-shopify-app` is explicitly mounted to the root of the application or router as described in the documentation. 

This is the case for an app I am building where I dynamically create the authentication middleware only and exactly when I need it and mount it on a path instead of adding it to koa via `.use`. This allows me to inject all parameters on demand, which adds a lot of flexibility.

Using the current handling differs from how the authentication flow is being handled in general throughout all other steps in the flow. And adding the prefix should not break anything as the middleware will pick it up anyways. So I see no reason to not use the prefix for this step.

This is an excerpt of how I am using this package:

```typescript
import Router from 'koa-router';

const router = new Router();

router.get('/shopify/:app/:actionOrPage?/:rest*', async (ctx, next) => {

  const authPrefix = `/shopify/${ctx.params.app}/setup`;

  return compose([
    // session middleware ...,
    actionOrPage !== 'setup'
      ? verifyRequest({ authRoute: `${authPrefix}/auth` })
      : createShopifyAuth({ prefix: authPrefix /*, ... further options */ }),
    // further middleware after we have ensured that the user is authenticated,
  ])(ctx, next);
});
```

## Type of change

- Patch: Bug / Enhancement in `@shopify/koa-shopify-auth`

## Checklist

- [x] I have added a changelog entry, prefixed by the type of change noted above (in `packages/koa-shopify-auth/CHANGELOG.md` but did not specify a version and date as I assume this will be done by repo maintainers)
- [x] updated and executed tests


passionkind:
duplicate of #1149

PR #1198, opened 2019-11-27
[graphql-testing] fixing fetch policy issues

patsissons:
## Description

the way `apollo-client`'s `QueryManager` handles the fetch policy can cause issues with react testing and `act` scoping. if you use `no-cache` or `network-only` there is some `apollo-client` code that forces the result to be marked as non-partial response. non-partial responses are automatically scheduled on the query observable, whereas partial responses are returned synchronously (find this code in `ObservableQuery`'s `getCurrentResult`).

This PR creates a new `TestingApolloClient` that patches all query functions to _normalize_ the `fetchPolicy` option for testing.

## Type of change

This should be a non-breaking change as no tests should be relying on network only fetching.

## Checklist

### DIFF
--- a/packages/graphql-testing/src/client.ts
+++ b/packages/graphql-testing/src/client.ts
@@ -1,16 +1,16 @@
 import {
   ApolloClient,
-  MutationOptions,
-  OperationVariables,
-  QueryOptions,
   SubscriptionOptions,
   WatchQueryOptions,
+  ApolloClientOptions,
+  ObservableQuery,
 } from 'apollo-client';
-import {FetchResult, Observable} from 'apollo-link';
+
+type GetCurrentQueryResult = ApolloClient<
+  any
+>['queryManager']['getCurrentQueryResult'];
 
 type AnyFetchPolicy =
-  | MutationOptions['fetchPolicy']
-  | QueryOptions['fetchPolicy']
   | SubscriptionOptions['fetchPolicy']
   | WatchQueryOptions['fetchPolicy'];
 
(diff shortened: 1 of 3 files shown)


packages/graphql-testing/src/client.ts
```diff
@@ -0,0 +1,75 @@
...
+// no-cache and network-only fetch policies cause an initial async event to be
+// produced which is not properly caught inside of a react-testing act scope.
+const invalidFetchPolicies: AnyFetchPolicy[] = ['no-cache', 'network-only'];
+
+// unset fetch policies that produce observable events after mounting so we
+// don't run into issues with events firing outside of an act scope
+function normalizeFetchPolicy<T extends AnyFetchPolicy>(
+  fetchPolicy: T | undefined,
+) {
+  return !fetchPolicy || invalidFetchPolicies.includes(fetchPolicy)
```
GoodForOneFare: As someone with very little GraphQL experience, this seems pretty intrusive :/  I'm guessing that it's not a big deal because we always provide mocked data in tests, so this will never cause a mismatch between live / test behaviour?

patsissons: if it's possible to set and a reset then that's probably preferred. i'll do a quick test now to see if that's possible.

patsissons: I have an idea to use the observable produced that would lead to a much less invasive fix.

patsissons: so that observable idea didn't work because the event processing happens between the query and `react-graphql` so i can't really pre-process the results to filter out the bad event. I have a new strategy up now that patches the `QueryManager`'s `getCurrentQueryResult` and detects the correct conditions where we need to override the `partial` value that causes the grief. The conditions are pretty limited so this should be incredibly non-invasive. LMK what you think.

patsissons: the reason why i'm making these changes is because i thought of a case where you would want to tests `network-only` omitting cache and my first version would likely break that test.

PR #1197, opened 2019-11-27
react-i18n - add `from-dictionary-index` plugin mode

GoodForOneFare:
## Description
Adds a `from-dictionary-index` plugin.  With this, folks with lots of CPU (or time) can generate multiple versions of their app.  Each with a specific locale embedded directly into its JS.  The advantages of this are:
* Bundlers like webpack don't get bloated with metadata about where to find async translations
* Users don't have to download multiple sets of translations

## 🎩 
* Check out the `solo-i18n-test` branch of web-rails-proving-ground
* Generate French index files using:
    * `node -e "require('@shopify/react-i18n/generate-dictionaries').generateTranslationDictionaries(['fr'])"`
* Clear out old transpilation artifacts
    * `rm -fr tmp/sewing-kit/cache/webpack`
* `dev run`
* Ensure French translations appear in the `customers` route
* Ensure _only_ French translations are embedded in JS files

Customers 2019-11-27 15-09-36

## Type of change

- [x] `@shopify/react-i18n Minor: New feature

## Checklist

- [x] I have added a changelog entry, prefixed by the type of change noted above

PR #1199, opened 2019-11-28
sewing-kit-koa now checks for matching locales

GoodForOneFare:
## Description
When provided with manifests containing a `locale` attribute, sewing-kit-koa will now attempt to match a user request to a set of assets with that locale's translations embedded into its JavaScript.

This is a companion to the react-i18n PR that supports per-locale builds: https://github.com/Shopify/quilt/pull/1197.  The advantages of this are:
* Bundlers like webpack don't get bloated with metadata about where to find async translations
* Users don't have to download multiple sets of translations

## Type of change

- [x] `@shopify/sewing-kit-koa` Minor: New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/packages/sewing-kit-koa/src/assets.ts
+++ b/packages/sewing-kit-koa/src/assets.ts
@@ -1,34 +1,12 @@
 import {join} from 'path';
 
 import {readJson} from 'fs-extra';
-import {matchesUA} from 'browserslist-useragent';
 import appRoot from 'app-root-path';
 
-export interface Asset {
-  path: string;
-  integrity?: string;
-}
+import {Manifest} from './types';
+import {Manifests} from './manifests';
 
-export interface Entrypoint {
-  js: Asset[];
-  css: Asset[];
-}
-
-export interface AsyncAsset {
-  id?: string;
-  file: string;
-  publicPath: string;
-  integrity?: string;
-}
-
-export interface Manifest {
-  name?: string;
-  browsers?: string[];
-  entrypoints: {[key: string]: Entrypoint};
-  asyncAssets: {[key: string]: AsyncAsset[]};
-}
-
-export type ConsolidatedManifest = Manifest[];
+export {Asset} from './types';
 
 interface Options {
   assetPrefix: string;
@@ -43,6 +21,7 @@ interface AssetSelector {
 }
 
 interface AssetOptions {
+  locale?: string;
   name?: string;
   asyncAssets?: Iterable<string | RegExp | AssetSelector>;
 }
@@ -52,24 +31,22 @@ enum AssetKind {
   Scripts = 'js',
 }
 
-const DEFAULT_MANIFEST_PATH = 'build/client/assets.json';
-
 export default class Assets {
   assetPrefix: string;
   userAgent?: string;
-  manifestPath: string;
+  private manifests: Manifests;
   private resolvedManifestEntry?: Manifest;
 
   constructor({assetPrefix, userAgent, manifestPath}: Options) {
     this.assetPrefix = assetPrefix;
     this.userAgent = userAgent;
-    this.manifestPath = manifestPath || DEFAULT_MANIFEST_PATH;
+    this.manifests = new Manifests(manifestPath);
   }
 
   async scripts(options: AssetOptions = {}) {
     const js = getAssetsFromManifest(
       {...options, kind: AssetKind.Scripts},
-      await this.getResolvedManifest(),
+      await this.getResolvedManifest(options.locale),
     );
 
     const scripts =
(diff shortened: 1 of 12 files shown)


packages/sewing-kit-koa/src/manifests.ts
```diff
@@ -0,0 +1,150 @@
...
+
+export class Manifests {
+  manifestPath: string;
+  private resolvedManifestEntry?: Manifest;
+
+  constructor(manifestPath: string | null = null) {
+    this.manifestPath = manifestPath || DEFAULT_MANIFEST_PATH;
+  }
+
+  async getResolvedManifest(
```
TheMallen: nit: `Manifests#getResolvedManifest` seems a bit redundant, we could shorten to something like `Manifests#get` or `Manifests#resolve`

packages/sewing-kit-koa/src/manifests.ts
```diff
@@ -0,0 +1,150 @@
...
+
+const DEFAULT_MANIFEST_PATH = 'build/client/assets.json';
+const MULTI_ASYNC_LANGUAGE_MANIFESTS = 'noLocales';
+
+export interface ResolveOptions {
+  locale?: string | undefined;
+}
+
+export class Manifests {
+  manifestPath: string;
```
TheMallen: same with some of this stuff, since this new class is well named and specific I think we can make a lot of stuff less wordy in here by deleting the word `manifest`

packages/sewing-kit-koa/src/manifests.ts
```diff
@@ -0,0 +1,150 @@
...
+      if (!manifests) {
+        manifests = [];
+        acc.set(locale, manifests);
+      }
+      manifests.push(manifest);
+    });
+  } else {
+    acc.get(MULTI_ASYNC_LANGUAGE_MANIFESTS)!.push(manifest);
+  }
+  // console.log(JSON.stringify([...acc], null, 2));
```
TheMallen: 🔥

packages/sewing-kit-koa/src/manifests.ts
```diff
@@ -0,0 +1,150 @@
...
+    manifest.identifier = {target: manifest.name};
+  }
+  return manifest;
+}
+
+function groupManifestsByLocale(
+  acc: Map<string, ConsolidatedManifest>,
+  manifest: Manifest,
+) {
+  const locales = manifest.identifier!.locales;
```
TheMallen: is this really optional? If not why don't we change the type (it's from this library isn't it?) if so, we should maybe do an assertion above with a nice error message if it's not there?

packages/sewing-kit-koa/src/assets.ts
```diff
@@ -52,24 +31,22 @@ enum AssetKind {
...
-const DEFAULT_MANIFEST_PATH = 'build/client/assets.json';
-
 export default class Assets {
   assetPrefix: string;
   userAgent?: string;
-  manifestPath: string;
+  private manifests: Manifests;
   private resolvedManifestEntry?: Manifest;
 
   constructor({assetPrefix, userAgent, manifestPath}: Options) {
```
TheMallen: Given that we create[ one instance of this per request](https://github.com/Shopify/quilt/blob/master/packages/sewing-kit-koa/src/middleware.ts#L36) is there a reason we don't add `locale` to the class itself? Then we could remove the params all over the place for `locale` and just use the class attribute. Feels weird that we have a mix of params and class attributes for things that are per request.

GoodForOneFare:
>  but on an app like web with many languages we might want to be aware of how it changes, especially once there are 100+. Do we have numbers or anything for that?

The cost is all up front: read a larger manifest on first request, and sort it into buckets.  For every other request:
- There's one extra map operation to see if the user's locale exists
- For locale / no locale flows, exactly the same number of browser checks are performed as before

So it feels to me like the runtime cost is one extra map access and maybe an extra function call ‾\\_(ツ)_/‾

packages/sewing-kit-koa/src/manifests.ts
GoodForOneFare: 👍 Good call.  I went with resolve.

packages/sewing-kit-koa/src/manifests.ts
GoodForOneFare: Chopped some stuff out 👍 Probably still a bit wordy in places, but I tried to let go 😄

packages/sewing-kit-koa/src/assets.ts
GoodForOneFare: Yah, this is nasty ☹️  It's an artefact of [how userAgent is available when the middleware creates Assets](https://github.com/Shopify/quilt/blob/5ba6cd69793a071950467c5f09c4dee9e93d15d0/packages/sewing-kit-koa/src/middleware.ts#L35-L45), but our apps tend to fetch locale info via GraphQL.

Locales are a niche feature right now, and I figure it's better to just add this verbose version of things before we put any more work into the API.

packages/sewing-kit-koa/src/manifests.ts
GoodForOneFare: It's optional for now because `identifier`'s presence depends on an unreleased version of sewing-kit 😅 

The mandatory part can come 🔜 🤞

PR #1201, opened 2019-12-01
Faster unit test CI job

GoodForOneFare:
## Description

Speed up CI's unit test job a bit.  Right now, it's the bottleneck on shipping, and very slow to give feedback about failing tests:

Travis CI snapshot showing the unit test job taking 2 minutes longer than the build job

This is because a `pretest` build step was added.  This makes sense for local test runs where e2e tests are part of the flow.  However, CI's unit test job excludes tests that need build artefacts, so there's no need to build.

## Is it fast, tho?
Faster.  

quilt - Travis CI 2019-12-01 17-51-57

Getting the build step faster is probably going to require Node 12 + [TypeScript support for threaded builds](https://github.com/Microsoft/TypeScript/issues/30235).

## Type of change
CI / infrastructure.

## Checklist


package.json
```diff
@@ -7,8 +7,10 @@
     "build": "yarn tsc --build packages",
     "lint": "eslint '**/*.{ts,tsx}'",
     "ci:lint-docs": "yarn generate docs && test -z \"$(git status --porcelain)\" || echo 'The root README has not been updated. Run `yarn generate docs` in the root of your quilt directory and try again.'",
-    "pretest": "yarn run build",
-    "test": "NODE_ICU_DATA=node_modules/full-icu jest --maxWorkers=3 --coverage=false",
+    "_test": "NODE_ICU_DATA=node_modules/full-icu jest",
```
GoodForOneFare: Yuck :/

newtack:
For what it's worth, we switched from Jest to Zora and our tests (just 35K lines of Typescript) went from 30 seconds to 2 seconds. It is very fast and we love it.

GoodForOneFare:
Thanks for the tip, @newtack!  Zora looks really interesting, and it's always great to learn about a new contender 😄 

In case anyone else stumbles across this PR, I'd like to make it clear that Jest is _not_ a bottleneck in Shopify's CI.  The low-hanging fruit here is TypeScript compilation.  To that end, we've been experimenting with [improving TypeScript's caching performance in CI](https://github.com/Shopify/quilt/pull/1207), and could probably just do a build with type checking disabled to really speed things up.

PR #1207, opened 2019-12-04
[Proof of concept] Speed up build/type-check CI job with a cache

GoodForOneFare:
## Description
Speed up CI using one weird file timestamp workaround.  vs the usual 10m+, this is pretty good:

quilt - Travis CI 2019-12-03 18-36-22

## Can you prove that this works, yo?
In the [Github diff for a test commit](https://github.com/Shopify/quilt/compare/ci-test-2), you can see changes to:
* `packages/csrf-token-fetcher/src/index.ts`
* `packages/koa-performance/src/index.ts`
* `packages/koa-performance/tsconfig.json`
* `packages/react-google-analytics/package.json`
* `packages/semaphore/src/Semaphore.ts`

In [the Travis build output for that commit](https://travis-ci.org/Shopify/quilt/jobs/620359665), you can see most projects building in 0 seconds, but touched projects taking a few seconds.  Selected highlights:

```
22:36:39 - Building project '/home/travis/build/Shopify/quilt/packages/csrf-token-fetcher/tsconfig.json'...

22:36:51 - Building project '/home/travis/build/Shopify/quilt/packages/koa-performance/tsconfig.json'...

22:36:59 - Building project '/home/travis/build/Shopify/quilt/packages/react-google-analytics/tsconfig.json'...

22:37:06 - Building project '/home/travis/build/Shopify/quilt/packages/semaphore/tsconfig.json'...
```

## Why not do this on master?
Ehhh, I'd like to try it out on branches for a while.  If the cache invalidation logic isn't right, it'll be caught by the master build.

Releases will still be painful, but branch work / rebasing will be much faster.

## Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,13 +8,19 @@ shared:
       - export PATH="$HOME/.yarn/bin:$PATH"
     install:
       - yarn install --production=false --frozen-lockfile
-    cache: yarn
+    cache:
+      directories:
+        - node_modules
+        - tsbuildcache
+    before_cache:
+      - node ./scripts/save-typescript-cache.js
 
 matrix:
   include:
     - <<: *node_container
       name: 'Node - build, type-check, and end-to-end tests'
       script:
+        - node ./scripts/restore-typescript-cache.js
         - yarn build --verbose
         - yarn test:ci --testPathPattern react-server-webpack-plugin address
     - <<: *node_container
--- a/scripts/restore-typescript-cache.js
+++ b/scripts/restore-typescript-cache.js
@@ -1,0 +1,87 @@
+const {execSync} = require('child_process');
+const fs = require('fs-extra');
+
+const {exec, gracefulExit} = require('./utilities');
+
+process.on('uncaughtException', gracefulExit);
+process.on('unhandledRejection', gracefulExit);
+
+const resetModifiedTimesCommand = 'touch -m -t 201001010000';
+
+/*
+ * This script is an ugly hack to speed up CI time :/  It is necessary because:
+ * - TypeScript's incremental build caching is based on file modification dates
+ * - Git always clones repos using the current date/time for all files
+ * - So in CI, TypeScript will always view a cache as 100% "stale"
+ *
+ * To work around this, the script:
+ * - Resets the modification dates of all config / src files to a time in the
+ *   far past
+ * - Grabs all changed paths for the branch
+ * - Tweaks the modification date of all changed paths
+ *
+ * This gets the cache / git in sync, albeit in a very gross manner 🤭
+ */
+
+async function run() {
+  const branchName = process.env.TRAVIS_PULL_REQUEST_BRANCH;
+  if (!branchName || branchName === 'master') {
+    console.log('Skipping build cache for master');
+    return;
+  }
+
+  await resetConfigModifiedDates();
+  await resetSourceModifiedDates();
+  await restoreDistFiles();
+  await touchModifiedFiles(await findBranchChanges());
+}
+
+run();
+
+async function restoreDistFiles() {
+  await fs.mkdirp('./tsbuildcache');
+
+  if (await fs.pathExists('./tsbuildcache/cache.tar')) {
+    await exec('tar -xf ./tsbuildcache/cache.tar');
+  }
+}
+
+async function resetConfigModifiedDates() {
+  await exec(`${resetModifiedTimesCommand} ./packages/tsconfig_base.json`);
+  await exec(
+    `find ./packages/*/tsconfig.json -exec ${resetModifiedTimesCommand} {} +`,
+  );
+  await exec(`find ./config -exec ${resetModifiedTimesCommand} {} +`);
+}
+
+async function resetSourceModifiedDates() {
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -8,13 +8,19 @@ shared:
       - export PATH="$HOME/.yarn/bin:$PATH"
     install:
       - yarn install --production=false --frozen-lockfile
-    cache: yarn
+    cache:
+      directories:
+        - node_modules
+        - tsbuildcache
+    before_cache:
+      - node ./scripts/save-typescript-cache.js
```
GoodForOneFare: There's no way to cache by glob, so we need to zip things up ourselves :/

scripts/restore-typescript-cache.js
```diff
@@ -0,0 +1,87 @@
...
+ * - Grabs all changed paths for the branch
+ * - Tweaks the modification date of all changed paths
+ *
+ * This gets the cache / git in sync, albeit in a very gross manner 🤭
+ */
+
+async function run() {
+  const branchName = process.env.TRAVIS_PULL_REQUEST_BRANCH;
+  if (!branchName || branchName === 'master') {
+    console.log('Skipping build cache for master');
```
ismail-syed: Outside the scope of this PR, but got me thinking a bit. 

Locally, running the cold cache builds are painfully long. If CI uploads master build caches to a _thing_ and locally `yarn run build` pulls the cache from _thing_, would that work to bring down cold-cache local builds.

scripts/save-typescript-cache.js
```diff
@@ -0,0 +1,10 @@
+const {exec, gracefulExit} = require('./utilities');
+
+process.on('uncaughtException', gracefulExit);
+process.on('unhandledRejection', gracefulExit);
+
+async function run() {
+  await exec(`tar -cf tsbuildcache/cache.tar packages/*/dist`);
```
ismail-syed: Do we need to cache the whole `packages/*/dist` directory? Can we get away with just caching the `.tsbuildinfo` files?

scripts/restore-typescript-cache.js
```diff
@@ -0,0 +1,87 @@
...
+  await exec(`find ./packages/*/src -exec ${resetModifiedTimesCommand} {} +`);
+}
+
+async function fetchMasterCommits() {
+  await exec(`git remote set-branches origin 'master'`);
+  await exec(`git fetch -v`);
+}
+
+async function findBranchChanges() {
+  await fetchMasterCommits();
```
TheMallen: if we were ever trying top merge into a non-master branch wouldn't this be fucky?

scripts/save-typescript-cache.js
TheMallen: ☝️ also curious about this

scripts/restore-typescript-cache.js
GoodForOneFare: Yeah, that's possible.

scripts/save-typescript-cache.js
GoodForOneFare: tsc depends on the presence of dist files.  Example:

```
$ yarn build
$ rm packages/semaphore/dist/src/Semaphore.js
$ yarn build --verbose

#### SNIP other output ####

[11:28:50] Project 'packages/semaphore/tsconfig.json' is out of date because output file 'packages/semaphore/dist/src/Semaphore.js' does not exist

[11:28:50] Building project '/Users/gord/src/github.com/Shopify/quilt/packages/semaphore/tsconfig.json'...
```

scripts/restore-typescript-cache.js
GoodForOneFare: Could you clarify what "top merge into a non-master branch" means, please?

scripts/save-typescript-cache.js
GoodForOneFare: tsc depends on the presence of dist files.  Example:

```
$ yarn build
$ rm packages/semaphore/dist/src/Semaphore.js
$ yarn build --verbose

#### SNIP other output ####

[11:28:50] Project 'packages/semaphore/tsconfig.json' is out of date because output file 'packages/semaphore/dist/src/Semaphore.js' does not exist

[11:28:50] Building project '/Users/gord/src/github.com/Shopify/quilt/packages/semaphore/tsconfig.json'...
```

GoodForOneFare:
> If this doesn't work as expected, is the worst-case scenario longer non-master CI runs or false negatives in CI tests because of muckey cache states?

My hope for a worst case is:
- Branch CI reports valid types
- master CI fails on merges

That's annoying, but can probably be fixed with tweaks to the script.

The worsterer case is that everything blows up like sewing-kit did when I added a ts cache :/  I *think* that was related to the weird out of sync non-beta-beatas publication thing that happened, though.  It's hard to know without merging this and seeing what happens :/

GoodForOneFare:
Closing because I think would provide inconsistent results :/  While poking around at sewing-kit's caches, I got some false positives after upgrading libraries that _should_ have caused type errors.  I suspect the CI cache would have to invalidate on any `yarn.lock`s changes to make this work.

PR #1209, opened 2019-12-04
Export Options from jest-koa-mocks

vsumner:
## Description
This PR simply exports the Options type from jest-koa-mocks

## Type of change
- [x] jest-koa-mocks Minor: New feature (non-breaking change which adds functionality)

## Checklist


vsumner:
I'm not sure who to add for reviews here. Since @GoodForOneFare is ATC today. I thought ...maybe?

Issue #1205, opened 2019-12-03
[quilt_rails] Expose header passing API from `reverse_proxy`


TheMallen:
## Problem Space

It's relatively common for apps to need some kind of initial (meta)data before they can do too much. With the current setup suggested by `quilt_rails` + `react-server` we push folks to use GraphQL for everything, which has a few unfortunate side-effects:
- to get simple trivial initial data (eg. custom locale preferences or other user specific metadata) we are forced to introduce waterfalling GraphQL requests which is a big performance no-no
- until we have our SSR GraphQL architecture sorted out it practically means that nothing can be server rendered (since the initial data request would not be able to be run on the server and nothing could be rendered before it is done)

In complex applications this gets worse, as we have seen in our largest React+GraphQL app where we actually have two different top level queries we need to run in series to figure out what sections we can render (and one we have to run before we even start rendering the app!)

Beyond the above, it's also a bit of DX bummer for extremely simple applications to need to setup graphQL before even the most trivial of data can be sent to their React app.

## Proposal

We already have a hook that we recommend devs use to access request headers inside the app

```ts
import {useRequestHeader} from '@shopify/react-network';

// ...

const someData = useRequestHeader('some-header');
```

On top of this, the gem we use internally to proxy requests to `react-server` exposes a `headers` option.

```ruby
reverse_proxy(headers: { foo: 'bar' })
```

If we were to just expose a similar option on `render_react` we could pretty easily support simple cases (specifically strings) being passed to the react app directly from Rails

```ruby
# ui_controller.rb

def index
  render_react(headers: {'x-custom-locale': current_user.custom_locale})
end
```

```tsx
// App.tsx
import {useRequestHeader} from '@shopify/react-network';

// ...
export function App() {
  const locale = useRequestHeader('x-custom-locale');
```

This setup would be a bit less nice for complex data, which is both a good thing (pushes users towards GraphQL for complex data needs), and a bad thing (might just mean they use it anyway and their code is worse). I think it's a sane API that will make 80% easy and 20% possible for our devs, and if we see a consistent need for data that it is a poor fit for we can do more then.

## Type
- [x] New feature

TheMallen:
@jordanAtShopify  heard you might be interested in helping out with this

jordanAtShopify:
@TheMallen Thanks for the awesome write-up!

PR #1210, opened 2019-12-05
Added header argument to `render_react` method in quilt_rails

jordanAtShopify:
## Description

Fixes https://github.com/Shopify/quilt/issues/1205 for quilt_rails gem

This feature allows developers to pass custom headers to `react-server` through a Rails controller using Quilt::ReactRenderable. 

Ex:
```
class ReactController < ApplicationController

  def index
    render_react(headers: { 'x-custom-header': 'header-value-a' })
  end
end
```

## Type of change

- [x] `quilt_rails` Minor: New feature (non-breaking change which adds functionality)

## Checklist


gems/quilt_rails/README.md
```diff
@@ -300,6 +300,67 @@ function App() {
...
+    render_react(headers: { 'x-custom-header': 'header-value-a' })
+  end
+end
+```
+
+You will need to serialize the result of the useRequestHeader hook in order for it to persist to the client
+
+```tsx
+// app/ui/foundation/CustomUniversalProvider.tsx
+import {createContext} from 'react';
```
TheMallen: @cartogram you seemed to be under the impression we were handling this automagically now

gems/quilt_rails/README.md
```diff
@@ -300,6 +300,67 @@ function App() {
...
+import {createUniversalProvider} from '@shopify/react-universal-provider';
+
+export const CustomContext = createContext<string | null>(null);
```
TheMallen: Assuming we don't already have a broken version of this as per my above question, I wonder if it's worth providing a `UniversalHeaderProvider` as part of `@shopify/react-network` that handles serializing headers that are grabbed by `useRequestHeader` during SSR.

TheMallen: (out of scope for this PR, but could be a follow up)

gems/quilt_rails/lib/quilt_rails/react_renderable.rb
```diff
@@ -6,9 +6,10 @@ module Quilt
   module ReactRenderable
     include ReverseProxy::Controller
 
-    def render_react
+    def render_react(headers: {})
       raise DoNotIntegrationTestError if Rails.env.test?
 
+      @headers = headers
```
TheMallen: I guess it's sort of the railsy thing to do to  stick this on the class instance rather than pipe it down as a parameter, but it feels a bit weird to me to do it here.

jordanAtShopify: I can definitely change this, first week using ruby and rails so still learning the conventions around things like this.

gems/quilt_rails/README.md
jordanAtShopify: @TheMallen I think a UniversalHeaderProvider would be a good idea, I implemented a basic version for the warehouse project but it probably makes sense to add to react-network if `useRequestHeader` will need to be serialized.

gems/quilt_rails/README.md
jordanAtShopify: @cartogram just wanted to check again to see if you had any concerns about this part? I couldn't get the `useRequestHeader` to persist without serializing it separately but if there is a way to do it correctly I can remove this section.

cartogram: I still haven't run into issues with the `useRequestHeader` when testing in a few projects we use it, but it does seem like this might be necessary for this use.

PR #1215, opened 2019-12-09
🐛Fix unformat number in react-i18n

adjnor:
## Description

Issue: https://github.com/Shopify/intl-new-commerce-patterns/issues/916

Fixes an issue with `unformatNumber` when the thousand separator symbol is a period.
Ideally, we would use the [formatToParts](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NumberFormat/formatToParts) method to fix the issue but that is still experimental.

Before:

out

After: 

out

## Type of change

- [x] react-i18n Patch: Bug/ Documentation fix (non-breaking change which fixes an issue or adds documentation)

## Checklist

### DIFF
--- a/packages/react-i18n/src/i18n.ts
+++ b/packages/react-i18n/src/i18n.ts
@@ -238,9 +238,13 @@ export class I18n {
   }
 
   unformatNumber(input: string): string {
-    const decimalSymbol = this.numberDecimalSymbol();
+    const {thousandSeparator, decimalSeparator} = this.numberSymbols();
 
-    const normalizedValue = normalizedNumber(input, decimalSymbol);
+    const normalizedValue = normalizedNumber(
+      input,
+      decimalSeparator,
+      thousandSeparator === PERIOD,
+    );
 
     return normalizedValue === '' ? '' : parseFloat(normalizedValue).toString();
   }
@@ -495,22 +499,30 @@ export class I18n {
     return decimal.length === 0 ? DECIMAL_NOT_SUPPORTED : decimal;
   }
 
-  private numberDecimalSymbol() {
-    return this.formatNumber(1.1, {
+  private numberSymbols() {
+    const formattedNumber = this.formatNumber(123456.789, {
       maximumFractionDigits: 1,
       minimumFractionDigits: 1,
-    })[1];
+    });
+    return getSeparators(formattedNumber);
   }
 }
 
-function normalizedNumber(input: string, expectedDecimal: string) {
+function normalizedNumber(
+  input: string,
+  expectedDecimal: string,
+  usesPeriodThousandSeparator?: boolean,
+) {
   const nonDigits = /\D/g;
 
   // For locales that use non-period symbols as the decimal symbol, users may still input a period
   // and expect it to be treated as the decimal symbol for their locale.
   const hasExpectedDecimalSymbol = input.lastIndexOf(expectedDecimal) !== -1;
   const hasPeriodAsDecimal = input.lastIndexOf(PERIOD) !== -1;
-  const usesPeriodDecimal = !hasExpectedDecimalSymbol && hasPeriodAsDecimal;
+  const usesPeriodDecimal =
+    !usesPeriodThousandSeparator &&
+    !hasExpectedDecimalSymbol &&
+    hasPeriodAsDecimal;
   const decimalSymbolToUse = usesPeriodDecimal ? PERIOD : expectedDecimal;
   const lastDecimalIndex = input.lastIndexOf(decimalSymbolToUse);
 
@@ -547,4 +559,16 @@ function dateTimeFormatter(
   options: Intl.DateTimeFormatOptions = {},
 ) {
   return new Intl.DateTimeFormat(locale, options);
+}
+
+function getSeparators(str: string) {
+  let thousandSeparator;
+  let decimalSeparator;
+  for (const char of str) {
+    if (isNaN(parseInt(char, 10))) {
+      if (thousandSeparator) decimalSeparator = char;
+      else thousandSeparator = char;
+    }
+  }
+  return {thousandSeparator, decimalSeparator};
 }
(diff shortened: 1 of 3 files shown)


packages/react-i18n/src/i18n.ts
```diff
@@ -548,3 +560,15 @@ function dateTimeFormatter(
 ) {
   return new Intl.DateTimeFormat(locale, options);
 }
+
+function getSeparators(str: string) {
+  let thousandSeparator;
+  let decimalSeparator;
+  for (const char of str) {
+    if (isNaN(parseInt(char, 10))) {
```
benoitzohar: What is the reason for using `parseInt()` here?

```suggestion
    if (isNaN(char)) {
```

`isNaN(char)` will return the same value as `isNaN(parseInt(char, 10))` and will be faster. 
Maybe I'm missing something?

packages/react-i18n/src/i18n.ts
```diff
@@ -548,3 +560,15 @@ function dateTimeFormatter(
 ) {
   return new Intl.DateTimeFormat(locale, options);
 }
+
+function getSeparators(str: string) {
+  let thousandSeparator;
+  let decimalSeparator;
+  for (const char of str) {
```
benoitzohar: How will that scale to ltr languages?

packages/react-i18n/src/i18n.ts
```diff
@@ -495,22 +499,30 @@ export class I18n {
     return decimal.length === 0 ? DECIMAL_NOT_SUPPORTED : decimal;
   }
 
-  private numberDecimalSymbol() {
-    return this.formatNumber(1.1, {
+  private numberSymbols() {
+    const formattedNumber = this.formatNumber(123456.789, {
```
benoitzohar:
```suggestion
    const formattedNumber = this.formatNumber(123456.7, {
```

Do you need 3 numbers after the decimal separator? since `maximumFractionDigits` and `maximumFractionDigits` are both `1`.

packages/react-i18n/src/i18n.ts
adjnor: Some languages use a space for the thousand operator (e.g. `fr-FR`). `isNaN(' ')` is `false` although it should be true so I'm using `parseInt(' ', 10)` to first convert the string into a number or `NaN`

packages/react-i18n/src/i18n.ts
adjnor: I'm not sure that there are any languages where numbers are written from right to left (I assume that's what you meant). Here is something I found about Arabic:
> Though Arabic words are written and read in RTL (right-to-left) directionality, numbers are read left-to-right (LTR), just as they are in English

https://www.industryarabic.com/numbers-in-arabic/

Would be good to confirm though

packages/react-i18n/src/i18n.ts
benoitzohar: TIL that for JavaScript, `' '` is a number. Nice catch!

packages/react-i18n/src/i18n.ts
adjnor: Good point, I'll change it to your suggestion

packages/react-i18n/src/i18n.ts
benoitzohar: Yes that's what I meant.
Ok, after digging a bit I can confirm that number are written left-to-right:
> Historically, the Western numbering system is derived from the Hindu-Arabic numbering system; popularized by Muslim mathematicians like Muḥammad ibn Mūsā al-Khwārizmī. When these books were translated from Arabic, however, Europeans who had a left-to-right language, still maintained the Arabic numbering order. Why Europeans did this nobody knowes.

[source](https://www.quora.com/Why-are-numerals-written-left-to-right-in-Arabic)

Also, interestingly: 
> In Standard Arabic, the smaller number should be read first. For example, 1436 would be six-and-thirty-and-four-hundred-and-a-thousand. The ordering now makes sense.

packages/react-i18n/src/test/i18n.test.ts
```diff
@@ -605,31 +605,39 @@ describe('I18n', () => {
...
-      formatNumber.mockImplementationOnce(() => '1,1');
+      formatNumber.mockImplementationOnce(() => '123.456,7');
 
       const formatted = '123.456,7891';
 
       expect(i18n.unformatNumber(formatted)).toBe(input.toString());
     });
 
+    it('handles number with space as the thousand symbol', () => {
+      formatNumber.mockImplementationOnce(() => '123 456,7');
```
adjnor: Could not find any examples of languages using a space for the thousand separator and a period for the decimal.
See: https://docs.oracle.com/cd/E19455-01/806-0169/overview-9/index.html

adjnor:
> LGTM, but I would add a test (or tests) for `getSeparators()`

@benoitzohar Actually, I'm not seeing any tests for helper functions such as `normalizedNumber` in `i18n.ts`. That makes sense to me since testing `unformatNumber` already tests for `normalizedNumber` and `getSeparators`.

benoitzohar:
@adjnor  It's true. 
However, it feels weird to mock `formatNumber` but not `getSeparators` and `normalizedNumber`.
I would be consistent in the tests, either mocking all the helpers to only test the function `unformatNumber` (and have tests for each helper), or not mock `formatNumber`.
But since `formatNumber()` is tested already, my favorite solution would be the first one (mock all helpers and add tests).

It might be nitpicking, let me know what you think about this.

adjnor:
As we discussed IRL, I'll update the PR to move the logic from `getSeparators` into `numberSymbols` since both functions are used for the same thing.

adjnor:
@lukashlavacka Good point, I added the `@memoize()` decorator to the `numberSymbols` method.
The failing CI doesn't seem to be related to my changes, should be fixed once [this PR](https://github.com/Shopify/quilt/pull/1229) ships

PR #1219, opened 2019-12-11
Include webpackChunkName in react-async ReadMe

jordanAtShopify:
## Description
Ran into the issue tracked in https://github.com/Shopify/sewing-kit/issues/1431 recently, while a clearer error message would be great I thought it was also a good idea to update the ReadMe so developers are less likely to run into the issue when first implementing the package.

## Type of change

- [x] react-async Patch: Bug/ Documentation fix (non-breaking change which fixes an issue or adds documentation)

## Checklist


GoodForOneFare:
And thank you! 😸

PR #1220, opened 2019-12-11
Add useToggle hook to react-hooks

BPScott:
## Description

Standardize some of the layout in the react-hooks package as we were sometimes using named and sometimes using default exports - now we always used named.

Add two new hooks that I've found very useful in polaris-react and have came in handy when writing stuff in web.

`useToggle()` is a little hook that abstracts away creating a boolean state which crops up a lot when you have to deal with toggling the visibility of modals and popovers and other conditional rendering. It's a pretty light touch but I've found myself having to reach for it quite often as many of these boolean states are needed in meaty components.

See the readme file for a usage demo.

## Type of change

## Checklist

---

EDIT: Disregard a


packages/react-hooks/README.md
```diff
@@ -102,3 +102,50 @@ function Score({value}) {
...
+      <p>Value: {activeText}</p>
+      <button onClick={toggleIsActive}>Toggle</button>
+    </>
+  );
+}
+```
+
+### `useForcibleToggle()`
+
+This hook will provide a boolean state value and a set of memoised callbacks to toggle it, force it to true and force it to false. It accepts one argument that is the initial value of the state.
```
TheMallen: I know it's not how we usually structure hook APIs, but the other functions feel so much like nice-to-haves that I wonder if it would be nicer DX to have a single hook which exposes the "bonus" callbacks as a third return value?

```tsx
const [active, toggle] = useToggle();
const [forceyActive, forceyToggle, {setTrue, setFalse}] = useToggle();
```

packages/react-hooks/src/hooks/index.ts
```diff
@@ -1,5 +1,7 @@
+export {useForcibleToggle} from './forcible-toggle';
 export {useLazyRef} from './lazy-ref';
-export {default as useTimeout} from './timeout';
```
TheMallen: 👏🏻 yessss death to `default`

packages/react-hooks/src/hooks/toggle.ts
```diff
@@ -0,0 +1,10 @@
+import {useState, useCallback} from 'react';
+
+export function useToggle(initialState: boolean) {
+  const [state, setState] = useState(initialState);
+  const toggle = useCallback(() => setState(state => !state), []);
+
+  // cast needed to say this returns a two item array with the items in
```
TheMallen: so annoying that this is the case  :<

tuple literal when?

BPScott: I found out that [`as const`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions) as added in TS 3.4 does the same thing but I'm kinda paranoid about support in older versions of TS in consumers

packages/react-hooks/README.md
BPScott: Sometimes you want just the forcing functions and not the toggle though, and then things start to look a little unintuitive:

```ts
// weird
const [forceyActive, , {setTrue, setFalse}] = useToggle();
// slightly less weird, but still weird
const [forceyActive, _, {setTrue, setFalse}] = useToggle();
```

perrupa:
Have you considered returning an object instead of an array?

```ts
const {value, toggle, setTrue, setFalse} = useToggle();
```

renaming can be done through destructuring, like so

```ts
const {
  value: modalIsVisible, 
  toggle: toggleModal, 
  setTrue: openModal,
  setFalse: hideModal,
} = useToggle();
```

BPScott:
> Have you considered returning an object instead of an array?

I want to evoke a familiarity with `useState`'s API which returns an array whos first item is the value. I don't want people to have to use destructuring renaming for the value as that's going to look clunky and is unfamiliar.

Note that the existing API for forcibleToggle is an object with the exception of the value

```
const [modalIsVisible, {
  toggle: toggleModal, 
  setTrue: openModal,
  setFalse: hideModal,
}] = useForcibleToggle();

Should we push that as the API as the primary way of doing this? i.e. would you prefer it if useToggle() had the above API  (renaming useForcibleToggle to useToggle, and finding a new name for the current useToggle)

**Comment by perrupa** [comment]
> Should we push that as the API as the primary way of doing this? i.e. would you prefer it if useToggle() had the above API (renaming useForcibleToggle to useToggle, and finding a new name for the current useToggle)

I think we should pick 1 API to expose, and run with it. Creating two hooks that do the same thing but expose slightly different APIs seems to discourage consistency. I like the idea of using `forcibleToggle`'s API but renaming it to `useToggle`.

File: packages/react-hooks/README.md (hunk 1 above)
Comment (TheMallen) [reply packages/react-hooks/README.md]: It seems to me that the only reason to have the `forcible` option in the first place is because the `useToggle` API takes away the power that `useState` gives you. 

One solution could just be having `useToggle` be a simple hook that wraps a `useState` call and gives you the `toggle` function, and then have `useToggleState` be the hook that both creates a state *and* gives you the toggle. Then you aren't being opionated about having "forcible" callbacks, you're just giving them a maximally flexible API, and a slightly more opinionated one built on that.

### example

When you don't need regular state setting at all use `useToggleState` for convenience

```tsx
  const [shown, toggleShown] = useToggleState();

  return (
    <Button onClick={toggleShown}>toggle thing be shown</Button>
```

When you need both toggling *and* regular setting just wrap your state setter

```tsx
const [thingShown, setThingShown] = useState(false);
const toggle = useToggle(setThingShown);

return (
<>
 <Button onClick={() => setThingShown(true)}>set it open</Button>
 <Button onClick={toggle}>toggle it</Button>
```

and if you don't need toggling just like, use state

```tsx
const [modalOpen, setModalOpen] = useState(false);

return (
 <Button onClick={() => setModalOpen(true)}>set it open</Button>
```

Flufd:
In the past I used `[state, toggleState, setState] = useToggleState(true)`
If you only need the `state` + `setState` parts then you can just use the regular `useState` so you don't get the weird missing elements

packages/react-hooks/README.md
BPScott: It's veeery subtle but  

```jsx
const [thingShown, setThingShown] = useState(false);
 <Button onClick={() => setThingShown(true)}>set it open</Button>
```

Results in a new handler being created every render, which if you pass it into a pure component will break it's pure optimisations, e.g. if Button was pure above, it would always rerender because the onClick prop differs every time. You'd need to define `const setThingShownTrue = useCallback(() => setThingShown(state => true), []);` and pass that to onClick to not break the pure check.

The utility of these two hooks is that it encapsulates those useCallback wrappings so people don't have to think about forgetting that useCallback, I think that's a key part of this.

TheMallen: `useCallback` is usually premature optimization IMO so I'd rather we didn't do it for simple stuff like this by default

TheMallen: been trying to discourage it in general ;)

BPScott:
The masses have spoken. Two Hooks instead of one is a little confusing and following the "return an array with two things" API stet state is considered clunky if the second item is an object, so now we have a pure object based API.

If you need everything:
```ts
const {
  value: isActive,
  toggle: toggleIsActive,
  setTrue: setIsActiveTrue,
  setFalse: setIsActiveFalse
} = useToggle(false)
```

If you need just the toggle:

```ts
const {value: isActive, toggle: toggleIsActive} = useToggle(false)
```

AndrewMusgrave:
cc/ @BPScott Slight suggestion on naming before this is released. This hook isn't just for toggling anymore. It provides useful methods for changing a boolean value, `useBoolean` better represents what the hook does imo

PR #1222, opened 2019-12-13
🐛 [react-form] Update reduceField() to check array equality when determining if a field is dirty

nickhoffman:
## Description

In react-form, dirty tracking doesn't work when a field's initial value is an array. This is because we compare arrays with `===`, which always returns `false` regardless of the arrays' contents. So after the field's value has changed once, comparing the changes always fails.

You can see this bug in action in [this codepen](https://codesandbox.io/s/optimistic-archimedes-78tjo). Follow the instructions in the output pane.

## Questions

1. I've fixed this with a naive solution: if a field's default value is an array, use `JSON.stringify` to compare it to the current value. This only works if all of the elements in the default value are scalars. Is this okay, or is there a better approach that we should use?

2. How can I test this? I didn't see any tests related to dirty tracking, so I'm not sure where to start.

## Type of change

- [x] react-form Patch: Bug fix (non-breaking change which fixes an issue or adds documentation)

## Checklist

### DIFF
--- a/packages/react-form/src/hooks/field/reducer.ts
+++ b/packages/react-form/src/hooks/field/reducer.ts
@@ -1,6 +1,7 @@
 import {useReducer, Reducer} from 'react';
 
 import {FieldState, ErrorValue} from '../../types';
+import {shallowArrayComparison} from '../../utilities';
 
 interface UpdateErrorAction {
   type: 'updateError';
@@ -62,10 +63,13 @@ export function reduceField<Value>(
     case 'update': {
       const newValue = action.payload;
       const {defaultValue} = state;
+      const isDirty = Array.isArray(defaultValue)
+        ? !shallowArrayComparison(defaultValue, newValue)
+        : defaultValue !== newValue;
 
       return {
         ...state,
-        dirty: defaultValue !== newValue,
+        dirty: isDirty,
         value: newValue,
         touched: true,
       };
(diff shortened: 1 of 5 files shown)


nickhoffman:
@TheMallen and @cartogram, what do you think of this?

packages/react-form/src/hooks/field/test/field.test.tsx
```diff
@@ -292,6 +294,86 @@ describe('useField', () => {
...
+      return {
+        value,
+        defaultValue,
+        error: undefined,
+        touched: true,
+        dirty,
+      };
+    }
+
+    describe('when the new value is the same as the default value', () => {
```
TheMallen: These tests are OK, and they do seem to be testing the right behaviour, but I'd prefer if they were testing the public API (`useField`) instead of `reduceField` specifically, were there any technical reasons that we couldn't do that?

nickhoffman: I started down that road at first and switched to these tests because putting the tests in `useField` required a lot of setup; more than I thought was reasonable.

nickhoffman:
@TheMallen, I've updated the logic to do a shallow comparison of arrays instead of using `JSON.stringify`. Could you 👀 it again, please?

packages/react-form/src/hooks/field/reducer.ts
```diff
@@ -48,6 +48,30 @@ export function updateErrorAction(error: ErrorValue): UpdateErrorAction {
   };
 }
 
+function shallowArrayComparison(arrA, arrB) {
```
TheMallen: nit: could we put this into [utilities](https://github.com/Shopify/quilt/blob/master/packages/react-form/src/utilities.ts)

TheMallen: Also, we should type them as `unknown[]`

nickhoffman: Sure thing. I've moved it there.

PR #1224, opened 2019-12-20
Use deep repo links

sambostock:
## Description

This changes how we link each package to a repository. Rather than link to the top-level mono-repo, we deep link.

Based on seeing how @babel/plugin-transform-flow-comments does it:

```json
  {
    "name": "@babel/plugin-transform-flow-comments",
    [...]
    "repository": "https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-flow-comments",
```

This originated from a discussion in Slack, about how some tooling doesn't handle packages in Quilt very well (e.g. Dependabot). @GoodForOneFare suggested we may be able to use this approach.

This affects **every** package.

## Type of change

## Checklist

### DIFF
--- a/packages/address-consts/package.json
+++ b/packages/address-consts/package.json
@@ -14,10 +14,7 @@
     "@shopify:registry": "https://registry.npmjs.org"
   },
   "author": "Shopify Inc.",
-  "repository": {
-    "type": "git",
-    "url": "git+https://github.com/Shopify/quilt.git"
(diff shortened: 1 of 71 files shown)


sambostock:
Not sure what the best way to handle the changelog section is. Does this warrant updating the changelog at all? If so, do we add an entry in every single package? Like, I guess this would be a patch level change?

PR #1227, opened 2020-01-03
Cookies must be `SameSite=None`

ragalie:
## Description

Chrome requires `SameSite=None` for third-party cookies starting in February. Updates documentation accordingly.

## Type of change

## Checklist

- [ ] ~I have added a changelog entry, prefixed by the type of change noted above~ I did not add a changelog entry since this is a change to the README, has no functionality implications and is already very visible.

### DIFF
--- a/packages/koa-shopify-auth/README.md
+++ b/packages/koa-shopify-auth/README.md
@@ -95,7 +95,7 @@ app.keys = [SHOPIFY_SECRET];
 
 app
   // sets up secure session data on each request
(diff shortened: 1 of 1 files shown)


gil--:
@ragalie Do we need to send `Secure: true` as well? Per https://web.dev/samesite-cookie-recipes/ 

> Cookies for cross-site usage must specify SameSite=None; Secure to enable inclusion in third party context.

PR #1233, opened 2020-01-12
Add `directory` key to package.json `repository`

sambostock:
### TL;DR

**#1224 should have made this change instead:**

```diff
 "name": "@shopify/{{name}}",
 "repository": {
   "type": "git",
   "url": "git+https://github.com/Shopify/quilt.git",
+  "directory": "packages/{{name}}",
 },
```

### Details

This follows up on #1224, which updated the repository fields, in an effort to improve support for tools such as Dependabot to deep-link into the repo:

```diff
 "name": "@shopify/{{name}}",
-"repository": {
-  "type": "git",
-  "url": "git+https://github.com/Shopify/quilt.git"
-},
+"repository": "https://github.com/shopify/quilt/tree/master/packages/{{name}}",
```

While these links do work, they still did not improve integration with Dependabot.

After speaking to Dependabot support, and being referred to [React's `package.json` `repository` entry](https://github.com/facebook/react/blob/64aae7b06fa47e126b0ba9c0ba9896caa803528e/packages/react/package.json#L23), it looks like specifying the additional `directory` key is the answer:

```diff
 "name": "@shopify/{{name}}",
-"repository": "https://github.com/shopify/quilt/tree/master/packages/{{name}}",
+"repository": {
+  "type": "git",
+  "url": "git+https://github.com/Shopify/quilt.git",
+  "directory": "packages/{{name}}",
+},
```

**This makes that change, fixes some other small inconsistencies, and adds tests along the way.**

## Type of change

## Checklist


test/consistent-package-json.test.ts
```diff
@@ -0,0 +1,135 @@
...
+    it('specifies a dependencies', () => {
+      expect(packageJSON.dependencies).not.toEqual({});
+    });
+
+    it('specifies a description', () => {
+      expect(packageJSON.description).not.toBeUndefined();
+    });
+
+    it('specifies files', () => {
+      if(packageName == 'polyfills') return; // FIXME
```
sambostock: The `polyfills` package does not contain the expected files:

```diff
 "files": [
-  "dist/*",
-  "~tsconfig.tsbuildinfo",
+  "**/*.js",
+  "**/*.d.ts",
 ]
```

test/consistent-package-json.test.ts
```diff
@@ -0,0 +1,135 @@
...
+
+      expect(packageJSON.homepage).toBe(expectedHomepage);
+    });
+
+    it('specifies the MIT license', () => {
+      expect(packageJSON.license).toBe(packageJSONTemplate.license);
+    });
+
+    it('specifies the expected main', () => {
+      if(packageName === 'graphql-persisted') return; // FIXME: Address this in graphql-persisted
```
sambostock: The `graphql-persisted` package does not specify an entry for `main`.

test/consistent-package-json.test.ts
```diff
@@ -0,0 +1,135 @@
...
+      expect(packageJSON.repository).toEqual(expectedRepository);
+    });
+
+    it('specifies scripts, including build', () => {
+      expect(packageJSON.scripts.build).toBe(packageJSONTemplate.scripts.build);
+    });
+
+    it('specifies if it has sideEffects', () => {
+      const exceptions = ['csrf-token-fetcher', 'dates', 'react-effect', 'react-google-analytics', 'useful-types']
+      if (exceptions.includes(packageName)) return; // FIXME
```
sambostock: The following packages have no `sideEffects` entry:
- `csrf-token-fetcher`
- `dates`
- `react-effect`
- `react-google-analytics`
- `useful-types`

test/consistent-package-json.test.ts
```diff
@@ -0,0 +1,135 @@
...
+
+      expect(packageJSON.sideEffects).toBe(Boolean(packageJSON.sideEffects));
+    });
+
+    it('specifies the expected types', () => {
+      if(packageName === 'graphql-persisted') return; // FIXME
```
sambostock: The `graphql-persisted` package fails to specify `types`.

test/consistent-package-json.test.ts
```diff
@@ -0,0 +1,142 @@
...
+  'version',
+];
+
+const ROOT_PATH = resolve(__dirname, '..');
+const packageJSONTemplatePath = join('templates', 'package.hbs.json');
+const packageJSONTemplate = readJSONSync(
+  join(ROOT_PATH, packageJSONTemplatePath),
+);
+
+// eslint-disable-next-line jest/valid-describe
```
sambostock: I had to add this cause the linter is trying to check that the first argument to `describe` is a `string`, which in this case it is, it's just dynamic and coming from a variable. The linter can't figure that out though, so it fails.

Looks like [others](https://github.com/jest-community/eslint-plugin-jest/issues/203) have run into this issue before.

packages/web-worker/package.json
```diff
@@ -14,11 +14,15 @@
...
+    "type": "git",
+    "url": "git+https://github.com/Shopify/quilt.git",
+    "directory": "packages/web-worker"
+  },
   "bugs": {
-    "url": "https://github.com/shopify/quilt/issues"
+    "url": "https://github.com/Shopify/quilt/issues"
   },
-  "homepage": "https://github.com/Shopify/quilt/blob/master/packages/web-workers/README.md",
+  "homepage": "https://github.com/Shopify/quilt/blob/master/packages/web-worker/README.md",
```
GoodForOneFare: Nice one.

test/consistent-package-json.test.ts
GoodForOneFare: This is just going to be `templates/package.hbs.json`, no?  I'd prefer just using that.

test/consistent-package-json.test.ts
```diff
@@ -0,0 +1,142 @@
...
+describe(packageJSONTemplatePath, () => {
+  it('has only known keys', () => {
```
GoodForOneFare: This feels a bit over-protective.  Can you describe a scenario in which this flags a mistake that wouldn't just be copy+pasted into the known keys?

test/consistent-package-json.test.ts
```diff
@@ -0,0 +1,142 @@
...
+      expect(packageJSON.main).toBe(packageJSONTemplate.main);
+    });
+
+    it('specifies name matching scope and path', () => {
+      const expectedName = compile(packageJSONTemplate.name);
+
+      expect(packageJSON.name).toBe(expectedName);
+    });
+
+    it('specifies the expected publishConfig', () => {
```
GoodForOneFare: Avoid vague terms like `expected` in test titles.

```suggestion
    it('specifies Shopify‘s public publishConfig', () => {
```

test/consistent-package-json.test.ts
```diff
@@ -0,0 +1,142 @@
...
+      };
+
+      expect(packageJSON.repository).toStrictEqual(expectedRepository);
+    });
+
+    it('specifies scripts, including build', () => {
+      expect(packageJSON.scripts.build).toBe(packageJSONTemplate.scripts.build);
+    });
+
+    it('specifies if it has sideEffects', () => {
```
GoodForOneFare:
```suggestion
    it('allows webpack tree-shaking', () => {
```

test/consistent-package-json.test.ts
```diff
@@ -0,0 +1,142 @@
...
+      const exceptions = ['csrf-token-fetcher', 'dates', 'react-effect', 'react-google-analytics', 'useful-types']
+      if (exceptions.includes(packageName)) return; // FIXME
+
+      expect(packageJSON.sideEffects).toBe(Boolean(packageJSON.sideEffects));
+    });
+
+    it('specifies the expected types', () => {
```
GoodForOneFare:
```suggestion
    it('declares type definitions', () => {
```

test/consistent-package-json.test.ts
```diff
@@ -0,0 +1,142 @@
...
+      if(packageName === 'graphql-persisted') return; // FIXME
+
+      expect(packageJSON.types).toBe(packageJSONTemplate.types);
+    });
+
+    it('specifies a version', () => {
```
GoodForOneFare: npm will refuse to publish without a version, so this feels over-protective.

test/consistent-package-json.test.ts
```diff
@@ -0,0 +1,142 @@
...
+    });
+
+    it('specifies the expected main', () => {
+      if(packageName === 'graphql-persisted') return; // FIXME: Address this in graphql-persisted
+
+      expect(packageJSON.main).toBe(packageJSONTemplate.main);
+    });
+
+    it('specifies name matching scope and path', () => {
+      const expectedName = compile(packageJSONTemplate.name);
```
GoodForOneFare: I would prefer just checking that:
```
expect(packageJson.name).toStrictEqual(`@shopify/${basename(packageJsonPath)}`)
```

Using `compile` makes more sense for other tests.  Here, it just adds a layer of indirection, and I spent too much time figuring out the actual test expectations.

test/consistent-package-json.test.ts
```diff
@@ -0,0 +1,142 @@
...
+  describe(packageJSONPath, () => {
+    it('specifies Shopify as author', () => {
+      expect(packageJSON.author).toBe(packageJSONTemplate.author);
+    });
+
+    it('specifies Quilt Issues as bugs URL', () => {
+      expect(packageJSON.bugs).toStrictEqual(packageJSONTemplate.bugs);
+    });
+
+    it('specifies a dependencies', () => {
```
GoodForOneFare:
```suggestion
    it('specifies dependencies', () => {
```

test/consistent-package-json.test.ts
```diff
@@ -0,0 +1,142 @@
...
+      expect(packageJSON.dependencies).not.toStrictEqual({});
+    });
+
+    it('specifies a description', () => {
+      expect(packageJSON.description).not.toBeUndefined();
+    });
+
+    it('specifies files', () => {
```
GoodForOneFare: If possible, give folks an idea of how `files` is used by npm.  e.g.,
```suggestion
    it('specifies publishable files', () => {
```

test/consistent-package-json.test.ts
```diff
@@ -0,0 +1,142 @@
...
+    });
+
+    it('specifies Quilt deep-link homepage', () => {
+      const expectedHomepage = compile(packageJSONTemplate.homepage);
+
+      expect(packageJSON.homepage).toBe(expectedHomepage);
+    });
+
+    it('specifies the MIT license', () => {
+      expect(packageJSON.license).toBe(packageJSONTemplate.license);
```
GoodForOneFare:
```suggestion
      expect(packageJSON.license).toBe('MIT');
```

test/consistent-package-json.test.ts
```diff
@@ -0,0 +1,142 @@
...
+
+    it('specifies files', () => {
+      if(packageName == 'polyfills') return; // FIXME
+
+      expect(packageJSON.files).toContain('dist/*');
+      expect(packageJSON.files).toContain('!tsconfig.tsbuildinfo');
+    });
+
+    it('specifies Quilt deep-link homepage', () => {
+      const expectedHomepage = compile(packageJSONTemplate.homepage);
```
GoodForOneFare: I know it's guaranteeing less, but I'd be happy enough with:
```
expect(packageJSON.homepage).toEndWith(`/${basename(packageJSONTemplatePath)}/README.md`)
```

test/consistent-package-json.test.ts
```diff
@@ -0,0 +1,142 @@
...
+    it('specifies the expected publishConfig', () => {
+      expect(packageJSON.publishConfig).toStrictEqual(
+        packageJSONTemplate.publishConfig,
+      );
+    });
+
+    it('specifies a repository deep-linking into the Quilt monorepo', () => {
+      const expectedRepository = {
+        ...packageJSONTemplate.repository,
+        directory: compile(packageJSONTemplate.repository.directory),
```
GoodForOneFare: Again, simplifying this to check for `packages/${basename(packageJsonPath)}` seems fine.

test/consistent-package-json.test.ts
sambostock: Cool. Should I go ahead and update it then?

test/consistent-package-json.test.ts
sambostock: I'm not sure what loosening this would gain us. I think the `compile` version is clearer, since it matches what is used in the other tests, and couples to the template as authoritative.

test/consistent-package-json.test.ts
sambostock: I had originally hardcoded things like this in this file, but then I started adding tests for the template as well. I realized that it made more sense to treat the template as the expected values, and ensure each package conforms to it.

If we hard code `MIT` here, it means that if we change the template down the road, the change will only be detected by the test suite after generating a new package and seeing it fail the suite. By coupling to the template, changing the license there would cause every package's license test to fail, which is likely what we want, since we'd probably want to update their licenses as well.

test/consistent-package-json.test.ts
sambostock: Would it not still make sense to have a `main` `index.js` that consists of something like

```javascript
export {apollo} from './apollo';
export {apollo} from './koa';
```

This would make it conform to the expected main entry point, while still offering the option to import the desired consumer directly. As a side benefit, it would also allow the `types` entry to be consistent.

test/consistent-package-json.test.ts
sambostock: I like this change, but I feel like including the key makes sense. Perhaps something like

```suggestion
    it('specifies if webpack can tree-shake it, via sideEffects', () => {
```

test/consistent-package-json.test.ts
sambostock: Fair. Should it be this file's responsibility to catch it before it gets to the publish stage though?

sambostock:
@GoodForOneFare, I've added a commit in which I switch to pre-compiling the expected template for each package, instead of using `compile` on a per-value basis. It does look much cleaner, while keeping things based on the template, though at the cost of generating the expected contents each time (probably fine).

sambostock:
I've addressed all exceptional packages and lint errors:
- `polyfills` now specifies `files: ["dist/*", "!tsconfig.tsbuildinfo"]`
- `csrf-token-fetcher`, `dates`, `react-effect`, `react-google-analytics`, and `useful-types` have had `sideEffects: false` added
- `graphql-persisted` now exposes a `main` entrypoint which re-exports the existing entrypoints
  ```typescript
  import * as apollo from './apollo';
  import * as koaMiddleware from './koa-middleware';

  export {apollo, koaMiddleware};
  ```

@GoodForOneFare, should these changes have accompanying changelog entries? Should they be in a separate PR, or are they fine here?

test/consistent-package-json.test.ts
```diff
@@ -0,0 +1,133 @@
...
+    // Template keys should match exactly. If updating it, remember to update tests!
+    expect(keys).toStrictEqual(KNOWN_TEMPLATE_KEYS);
+  });
+});
+
+readPackages().forEach(({packageName, packageJSON, packageJSONPath}) => {
+  const expectedPackageJSON = compileTemplate();
+
+  // eslint-disable-next-line jest/valid-describe
+  describe(packageJSONPath, () => {
```
GoodForOneFare: I get the intention of making a comprehensive suite that makes pinpointing errors very precise...  but the end result is a ton of output to sift through when I'm running the full suite (or trying to sift through CI logs).

I would prefer to see the test structure inverted, with each test proving a specific fact about all packages.  Yes, folks will have to grep a bit for error strings, but:
* The end result is less noisy during successful runs
* It's rare to touch >3 package.jsons in a PR, so the set of packages to sift through is small

-----

For context, here is a screencap of the test output with boilerplatey tests that focus on a single aspect of all packages:

A Quilt test run showing a failing 'all packages link to Quilt issue tracker'-style test.  The test has somewhat oblique output, but it should be easy enough to grep the codebase for it.

Contrast with the current output at my terminal's lowest zoom setting:

quilt 2020-01-14 00-49-59

That's ~70% of the output :/

test/consistent-package-json.test.ts
GoodForOneFare: Hookay!  I'm still on the fence, but let's see how it goes 😃

test/consistent-package-json.test.ts
GoodForOneFare: `compileTemplate` is fast.  Compiling lots of templates over and over again is fine if it makes test output more readable.

GoodForOneFare:
RE changelog entries, I guess they have to be part of this PR.  Otherwise, any merge gap between this and a changelog PR will result in someone else having to release everything.

packages/graphql-persisted/src/index.ts
```diff
@@ -0,0 +1,8 @@
+// MDN claims we should be able to export directly, but it doesn't seem to work
```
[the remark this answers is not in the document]
sambostock: Ok. I'll exclude `graphql-persisted` for now, remove the file I added, and instead open an issue to discuss this package's approach there.

sambostock:
@GoodForOneFare, I've done the following:
- Restored `graphql-persisted` to what it was, so it is now an exception to the `main` and `types` tests.
- Added selected changelog entries:
  - For the additions of `"sideEffects": false` across a couple packages
  - For the change to published `"files"` in `polyfills`
- Omitted changelog entries for "Shopify" capitalization, and the `directory` key, since I don't think consumers need to care.
- Rebased into meaningful commits, so the diffs are trivial to inspect.

While I had originally agreed with your point on consolidating the tests into _one test per key for all packages_, **I have switched it back to my original _one test per key per package_**. I feel strongly that this is a better developer experience. You brought up two main points:

- > The end result is less noisy during successful runs
  
  I did some testing, and as far as I can tell, the **giant output only shows up if running this single test file**, all on its own. When running a typical run, this file shows up just like any other suite (file name only):
  ```
  ...
  PASS packages/react-i18n/src/test/e2e/server.test.tsx
  PASS test/consistent-package-json.test.ts                <<<<
  PASS packages/i18n/src/test/pseudotranslate.test.ts
  ...
  ```
  I noticed, when looking at the CI output, the `console.log`, `console.warn`, and `console.error` output that makes it into the test output provides far more noise than this does. 😅 

- > It's rare to touch >3 package.jsons in a PR, so the set of packages to sift through is small
  
  This applies to PRs that change a package's `package.json`, but changes to the template, or which add a new field benefit greatly from knowing which package(s) failed. The consolidated output is effectively useless, since it just says there has been a failure and provides no info about where or how many occurrences there are. During my rebase, I had to redo some of my changes, and knowing where I needed to add `sideEffects` meant I didn't have to sift through 70+ `package.json` files.

Based on that, I think the benefits of granular failures far outweigh the cost.

PR #1235, opened 2020-01-14
Adds GIP currency code to CurrencyCode enum

mateus:
## Description

Adds the missing currency code value for `GIP` that is also present in [billing_graph_api/schema.graphql](https://github.com/Shopify/shopify/blob/32052207a433f679ad453b0de34925f1d815f61a/components/billing/app/graph_clients/billing_graph_api/schema.graphql#L389). This change is necessary in order to fix the following bug reported. 

Bugsnag: https://app.bugsnag.com/shopify/shopify-production/timeline?filters[error][0]=5e14e89ba14aea0018358f98

## Type of change

Updating the `CurrencyCode` enum to include the value `GIP`

## Checklist


mehdi-loup:
We should centralize the source of truth for currencies. Polaris has pipelines to create ruby gems and node modules, which could be leverage to have one source of truth servicing ruby/node apps

PR #1239, opened 2020-01-19
[several packages] Update jest matcher types

jgodson:
## Description

When working on https://github.com/Shopify/web/pull/22447, I ran into issues when using `toHaveReactProps`. For some reason it was no longer inferring the typescript types of the React Components. I believe it was due to [this change](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/39243/files) and due to the namespace no longer implementing the same interface it cause these issues.

This PR updates the `@types/jest` package and the `Matcher` interface in all packages to match.

## Type of change
I'm not sure what type of change this should be. It's a breaking change for Typescript projects, but the update to the types package is the real reason, not these packages themselves 

- [ ]  `ast-utilities` Minor: `jest.Matchers` type updated to match `@types/jest` version `25`
- [ ]  `graphql-testing` Minor: `jest.Matchers` type updated to match `@types/jest` version `25`
- [ ]  `react-i18n` Minor: `jest.Matchers` type updated to match `@types/jest` version `25`
- [ ]  `react-testing` Minor: `jest.Matchers` type updated to match `@types/jest` version `25`

## Checklist

### DIFF
--- a/packages/ast-utilities/src/matchers.ts
+++ b/packages/ast-utilities/src/matchers.ts
@@ -7,7 +7,7 @@ import {
 declare global {
   // eslint-disable-next-line @typescript-eslint/no-namespace
   namespace jest {
-    interface Matchers<R, T> {
+    interface Matchers<R, T = {}> {
       toBeFormated(expected: string): void;
     }
   }
--- a/packages/ast-utilities/src/matchers.ts
+++ b/packages/ast-utilities/src/matchers.ts
@@ -7,7 +7,7 @@ import {
 declare global {
   // eslint-disable-next-line @typescript-eslint/no-namespace
   namespace jest {
-    interface Matchers<R> {
+    interface Matchers<R, T> {
       toBeFormated(expected: string): void;
     }
   }
--- a/packages/ast-utilities/CHANGELOG.md
+++ b/packages/ast-utilities/CHANGELOG.md
@@ -7,6 +7,9 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Changed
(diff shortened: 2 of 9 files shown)


GoodForOneFare:
@lemonmade this change makes sense to me.  Could you help with Jason's question about what type of version bump this should be in the packages that it touches, please?

packages/ast-utilities/src/matchers.ts
```diff
@@ -7,7 +7,7 @@ import {
 declare global {
   // eslint-disable-next-line @typescript-eslint/no-namespace
   namespace jest {
-    interface Matchers<R> {
+    interface Matchers<R, T = {}> {
```
vsumner, line 10: Updated to match https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/jest/index.d.ts#L697

alexandcote:
@jgodson, can you check if you have the latest version of `@types/jest`

I think I already solve this in this PR.

https://github.com/Shopify/quilt/pull/1295

jgodson:
That fixed react testing, which this originally included as well. This updates the other packages

alexandcote:
My point is `Matchers<R>` should be valid because [T is optional](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/jest/index.d.ts#L697). I think we should only add `T` if we need it [like here](https://github.com/Shopify/quilt/pull/1239/files#diff-40fad76b75dae7a234fe469e933d30caR19).

jgodson:
Good point, it wasn't optional when I first put the PR up, but it appears that's the changes in jest 25. It shouldn't cause any issues though either in that case?

PR #1241, opened 2020-01-21
Introduce `useInterval` React Hook

sambostock:
## Description

**This adds a `useInterval` hook to `@shopify/react-hooks`.**

It wraps `setInterval` in a `useEffect`, similarly to the existing `useTimeout` hook, which wraps `setTimeout`.

## Type of change

- [x] `@shopify/react-hooks` Minor: New feature (non-breaking change which adds functionality)

## Checklist

- [x] I have squashed this PR into meaningful commits
- [ ] I have answered the questions about licensing

### DIFF
--- a/packages/react-hooks/src/hooks/interval.ts
+++ b/packages/react-hooks/src/hooks/interval.ts
@@ -1,0 +1,27 @@
+import {useEffect, useRef} from 'react';
+
+type IntervalCallback = () => void;
+type IntervalDelay = number | null;
+
+// Adapted from https://overreacted.io/making-setinterval-declarative-with-react-hooks
+// Also available at https://github.com/gaearon/overreacted.io/blob/master/src/pages/making-setinterval-declarative-with-react-hooks/index.md
+//
+// Copyright (c) 2020 Dan Abramov and the contributors.
+//
+export function useInterval(callback: IntervalCallback, delay: IntervalDelay) {
+  const savedCallback = useRef(callback);
+
+  useEffect(() => {
+    savedCallback.current = callback;
+  }, [callback]);
+
+  useEffect(() => {
+    function tick() {
+      savedCallback.current();
+    }
+    if (delay !== null) {
+      const id = setInterval(tick, delay);
+      return () => clearInterval(id);
+    }
+  }, [delay]);
+}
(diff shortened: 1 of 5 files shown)


packages/react-hooks/src/hooks/interval.ts
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+
+export function useInterval(callback: () => void, delay: number) {
+  React.useEffect(() => {
+    const id = setInterval(callback, delay);
+    return () => clearInterval(id);
```
GoodForOneFare: React clearing out the interval during rerenders will cause surprising behaviour.  e.g., a user editing form values will trigger a rerender, and the interval will be reset.

Here's a code sandbox to demonstrate the oddness: https://codesandbox.io/s/priceless-cdn-hlzc2  For as long as you type in the sandbox's input field, the interval won't fire.

carolopolo:
hey @sambostock ! i just worked on a custom `useInterval` hook for a project i'm working on. do you want to collaborate?

tleunen:
This article explains it well ;) https://overreacted.io/making-setinterval-declarative-with-react-hooks/

carolopolo:
> This article explains it well ;) https://overreacted.io/making-setinterval-declarative-with-react-hooks/

haha yeeeeeeep, i saw that for sure

sambostock:
I have adapted [Dan Abramov's `useInterval` implementation from his blog](https://overreacted.io/making-setinterval-declarative-with-react-hooks/), as shared above, **adding TypeScript types** and tests.

**The code snippets from his blog are [under the MIT license](https://github.com/gaearon/overreacted.io/blob/b71d15384aaf9b089dfbf9192a072eda77eb36a4/LICENSE-code-snippets)**. I believe that means we're fine to use it, however I'm unsure about this portion of the license:

> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

I'm not sure if that means we're fine, because this package is also [under the MIT license](https://github.com/Shopify/quilt/blob/d63d282685831b797e495cd4775adc857f640bc9/packages/react-hooks/package.json#L4), or if we specifically have to include the full license including

> Copyright (c) 2020 Dan Abramov and the contributors.

or if including just the copyright notice is good enough (in combination with already being license under MIT).

sambostock:
I've added a README entry with a code example, highlighting normal usage, as well as the `null` delay option.

I also mentioned the blog post, so it's clear it's the same hook readers may have come across before.

sambostock:
Ran this past @Robguay1. Looks like we're good, but he suggested adding the copyright notice in the file, which I have done.

PR #1242, opened 2020-01-22
koa-shopify-auth: update redirection page to use App Bridge action

tylerball:
## Description

Fixes https://github.com/Shopify/app-bridge/issues/1410

## Type of change

This PR updates the redirection script in `koa-shopify-auth` to use the App Bridge Redirect action instead of a bare postMessage call. I've :tophat:ed this in Safari.

cc @Shopify/app-bridge 

## Checklist

### DIFF
--- a/packages/koa-shopify-auth/src/auth/redirection-page.ts
+++ b/packages/koa-shopify-auth/src/auth/redirection-page.ts
@@ -1,18 +1,21 @@
-export default function redirectionScript({origin, redirectTo}) {
+export default function redirectionScript({origin, redirectTo, apiKey}) {
   return `
-    <script type="text/javascript">
+    <script src="https://unpkg.com/@shopify/app-bridge@^1"></script> <script type="text/javascript">
       document.addEventListener('DOMContentLoaded', function() {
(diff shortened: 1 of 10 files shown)


packages/koa-shopify-auth/src/auth/redirection-page.ts
```diff
@@ -1,18 +1,21 @@
...
       document.addEventListener('DOMContentLoaded', function() {
         if (window.top === window.self) {
           // If the current window is the 'parent', change the URL by setting location.href
           window.location.href = '${redirectTo}';
         } else {
           // If the current window is the 'child', change the parent's URL with postMessage
-          data = JSON.stringify({
-            message: 'Shopify.API.remoteRedirect',
-            data: { location: '${redirectTo}' }
+          var AppBridge = window['app-bridge'];
```
vividviolet: Looks great, is there any tests for this util?

tylerball: Nothing for this particular script, all this file does is render a route with this script to redirect the user out of the iframe in Safari so we can set a cookie.

PR #1255, opened 2020-02-03
set cookies to be secure

katiedavis:
## Description

Fixes issue where request is not considered secure. See: https://github.com/Shopify/shopify-demo-app-node-react/issues/129#issuecomment-580820979

## Type of change

- [ ] koa-shopify-auth Patch: Bug/ Documentation fix (non-breaking change which fixes an issue or adds documentation)

## Checklist

### DIFF
--- a/packages/koa-shopify-auth/CHANGELOG.md
+++ b/packages/koa-shopify-auth/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## 3.1.55 -2020-02-02
+
+- Package now forces cookies.secure to be true [1255](https://github.com/Shopify/quilt/pull/1255)
+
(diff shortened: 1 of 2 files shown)


TheMallen:
Should we also change the spots we set cookies to manually use `secure: true`? That way we can still advise partners to use security everywhere but be sure our cookies always are.

katiedavis:
@TheMallen There's actually a great PR open here: https://github.com/Shopify/quilt/pull/1251
I like the utility they built to check the browser etc. Should we review that and get it in?

packages/koa-shopify-auth/CHANGELOG.md
```diff
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## 3.1.56 -2020-02-02
+
+- Package now forces cookies.secure to be true [1255](https://github.com/Shopify/quilt/pull/1255)
+
+## 3.1.55 -2020-02-02
```
tylerball: I would just release em under the same version

Issue #1246, opened 2020-01-24
Add locale to performance statistics


GoodForOneFare:
## Overview
Since #1180, sewing-kit-koa has the ability to serve assets with a single locale's translations embedded directly into its JavaScript.  It would be great if we could segment our stats by this.

I think the path to follow is:
* Update react-performance 's report component and hook to accept / send a locale attribute
* Update koa-performance's middleware to process locale
* Update consumers to pass a locale into `<BasePerformanceReport>`

Links to those areas:
https://github.com/Shopify/quilt/blob/master/packages/react-performance/src/performance-report.ts#L37-L45
https://github.com/Shopify/quilt/blob/master/packages/koa-performance/src/middleware.ts#L78

PR #1260, opened 2020-02-05
Locale perf tracking

GoodForOneFare:
## Description

Fixes (issue #1246)

Client-side code can now pass a `locale` to our performance libs.  Server-middleware now sends on locale to stats trackers.

## Type of change

- [x] `@shopify/koa-performance` Minor: New feature (non-breaking change which adds functionality)
- [x] `@shopify/react-performance` - Minor: New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/packages/koa-performance/src/test/middleware.test.ts
+++ b/packages/koa-performance/src/test/middleware.test.ts
@@ -176,6 +176,54 @@ describe('client metrics middleware', () => {
         expect.objectContaining(additionalTags),
       );
     });
+
+    it('includes locale in distributions', async () => {
+      const context = createMockContext({
+        method: Method.Post,
+        requestBody: createBody({
+          connection: {
+            effectiveType: '3G',
+          },
+          locale: 'es',
+          events: [createLifecycleEvent()],
+        }),
+      });
+
+      await withEnv('production', async () => {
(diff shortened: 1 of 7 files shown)


packages/koa-performance/src/test/middleware.test.ts
```diff
@@ -493,6 +541,7 @@ type DeepPartial<T> = {
 function createBody({
   connection,
   events = [],
+  locale = undefined,
```
michenly: isn't the default always undefined ? :p

GoodForOneFare: D'oh!  Yes.  Fixed now 🙏

packages/react-performance/src/performance-report.ts
```diff
@@ -11,7 +11,10 @@ export interface ErrorHandler {
 
 export function usePerformanceReport(
   url: string,
-  {onError = noop}: {onError?: ErrorHandler} = {},
+  {
+    locale = undefined,
```
GoodForOneFare: I'm on the fence about whether this should be a more free-form `metadata`-like property that accepts/transmits app-specific custom tags.

If more values get crammed in here, I'd be happy to rework to make this more generic in a later release 😅

Issue #1234, opened 2020-01-13
`react-i18n` negative currency issue


Jurgyte:
## Overview
I use `I18n.formatCurrency` to display a discount value, which is a negative price. This function returns `$-3.45`, but it should be `-$3.45`.

---

## Checklist

- [ ] Please delete the labels section before submitting your issue
- [ ] I have described this issue in a way that is actionable (if possible)

movermeyer:
More context:

Some people expected this format, so we have previously used `$-3.45` for some regions.
However after some research (by `@winterweix`) we have recently decided that putting the `-` sign before the rest of the price is actually how we want it for all regions.

See the new Polaris guidelines: https://polaris.shopify.com/foundations/formatting-localized-currency#navigation

Aside: This has been fixed recently in [`shopify-i18n v1.1.4`](https://github.com/Shopify/shopify-i18n/pull/519)

PR #1264, opened 2020-02-06
Put minus sign in front of currency symbol formatting negative amounts

cejaekl:
## Description

Fixes https://github.com/Shopify/quilt/issues/1234

Changes the output of `i18n.formatCurrency(-3.25)` from `$-3.25` to `-$3.25`, to conform with our new currency formatting guidelines:

https://polaris.shopify.com/foundations/formatting-localized-currency

## Type of change

Changes `react-i18n`'s `i18n.formatCurrency()`.

- [X] `react-i18n` Patch: Minor change in formatting output.  Really too small to be considered "breaking".

## Checklist

### DIFF
--- a/packages/react-i18n/src/i18n.ts
+++ b/packages/react-i18n/src/i18n.ts
@@ -403,9 +403,13 @@ export class I18n {
       ...options,
     }).format(amount);
 
-    return shortSymbol.prefixed
+    const formattedWithSymbol = shortSymbol.prefixed
       ? `${shortSymbol.symbol}${formattedAmount}`
       : `${formattedAmount}${shortSymbol.symbol}`;
+
+    return amount < 0
+      ? `-${formattedWithSymbol.replace('-', '')}`
+      : formattedWithSymbol;
   }
 
   // Intl.NumberFormat sometimes annotates the "currency symbol" with a country code.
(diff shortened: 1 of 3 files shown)


packages/react-i18n/src/i18n.ts
```diff
@@ -403,9 +403,13 @@ export class I18n {
       ...options,
     }).format(amount);
 
-    return shortSymbol.prefixed
+    const formattedWithSymbol = shortSymbol.prefixed
       ? `${shortSymbol.symbol}${formattedAmount}`
       : `${formattedAmount}${shortSymbol.symbol}`;
+
+    return amount < 0
+      ? `-${formattedWithSymbol.replace('-', '')}`
```
ismail-syed: Are there any possible formats of `formattedWithSymbol` where a `-` would denote something that is not a minus symbol? I see `1 234,56 Kč CZK`, could there ever be a format of money that would result in `1 234,56 foo-bar`

cejaekl: I think we're "safe" to do this.
 - We should always have `-` as the `minus sign`.  Looking at the Unicode CLDR data, every single locale for which it defines a `minus_sign:` uses `-`
  - We should never have a `-` in this string that isn't a `minus sign`.  We're dealing with the "symbol" for the currency, or potentially the ISO-4217 currency code (which is guaranteed to be three alphabetic letters).

It's possible that a `-` might show up in the full-text name of the currency (e.g. `Guinea-Bisseau franc`), but we won't have the full name here, only the symbol or the three-letter code.

PR #1275, opened 2020-02-11
@shopify/dates v0.1.23-fixchrome80.2

ismail-syed:
Patch release for https://github.com/Shopify/quilt/pull/1272

Context: https://github.com/Shopify/quilt/pull/1272#issuecomment-584799390

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -278,6 +278,16 @@
     lodash "^4.17.4"
     lodash-decorators "^4.3.5"
 
+"@shopify/react-html@^8.0.10":
+  version "8.1.3"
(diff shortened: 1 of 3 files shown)


yarn.lock
```diff
@@ -278,6 +278,16 @@
     lodash "^4.17.4"
     lodash-decorators "^4.3.5"
 
+"@shopify/react-html@^8.0.10":
```
carysmills: Why is this updated?

ismail-syed: Since quilt is a mono-repo where some packages pull in `react-html`, the build I ran pulled in the latest `8.0.x` version that wasn't available when `@shopify/dates@0.1.23` was released. This has no impact on `@shopify/dates` since it doesn't touch `react-html`

ismail-syed: _It really doesn't matter_ that I checked these in but I like to be transparent for questions like this. I'm glad you asked!

carysmills: thanks for explaining :)

PR #1276, opened 2020-02-11
Fix react-google-analytics when script is blocked

cartogram:
## Description

This PR fixes a bug where the react-google-analytics `<Universal />` component breaks when the analytics script is blocked by the browser.

Note: The default privacy settings in Edge are strict enough to block this script. 

## Type of change

- [x] `react-google-analytics` Patch: Bug/ Documentation fix (non-breaking change which fixes an issue or adds documentation)

## Checklist

### DIFF
--- a/packages/react-google-analytics/src/Universal.tsx
+++ b/packages/react-google-analytics/src/Universal.tsx
@@ -38,13 +38,17 @@ export default function UniversalGoogleAnalytics({
   disableTracking,
 }: Props) {
   const setAnalytics = React.useCallback(
-    (googleAnalytics: UniversalAnalytics) => {
+    (googleAnalytics: UniversalAnalytics | Error) => {
       const normalizedDomain = getRootDomain(domain);
       const options = {
         cookieDomain: normalizedDomain,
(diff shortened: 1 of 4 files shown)


packages/react-google-analytics/src/Universal.tsx
```diff
@@ -38,14 +38,18 @@ export default function UniversalGoogleAnalytics({
...
+    (googleAnalytics: UniversalAnalytics | Error) => {
       const normalizedDomain = getRootDomain(domain);
       const options = {
         cookieDomain: normalizedDomain,
         legacyCookieDomain: normalizedDomain,
         allowLinker: true,
       };
 
+      if (googleAnalytics instanceof Error) {
+        return null;
```
TheMallen: should there be an `onError` callback like there is an `onLoad`?

lemonmade: Is there any API we need to give to let apps handle this gracefully?

cartogram:
Thanks both for the (very similar) reviews 👯 Updated with an `onError` prop with docs and a test.

PR #1277, opened 2020-02-12
Fix memory leak in @shopify/dates when SSR

helloneele:
## Description

This fixes the memory leak in the dates package that we were seeing when upgrading to any version after v 0.1.26 in Web.

More info is here: https://github.com/Shopify/web/issues/23454

## Type of change

## Checklist

### DIFF
--- a/packages/dates/src/utilities/formatDate.ts
+++ b/packages/dates/src/utilities/formatDate.ts
@@ -1,28 +1,45 @@
-import {memoize} from '@shopify/function-enhancers';
+const intl = new Map();
+const memoizedGetDateTimeFormat = function(locale, options) {
+  const key = dateTimeFormatCacheKey(locale, options);
+  if (intl.has(key)) {
+    return intl.get(key);
+  }
+  const i = new Intl.DateTimeFormat(locale, options);
+  intl.set(key, i);
+  return i;
+};
 
-const memoizedGetDateTimeFormat = memoize(
-  getDateTimeFormat,
-  dateTimeFormatCacheKey,
-);
+const browserFeatureDetectionDate = Intl.DateTimeFormat('en', {
+  hour: 'numeric',
+});
 
 interface FormatDateOptions extends Intl.DateTimeFormatOptions {
   hourCycle?: string;
 }
+
+const resolvedOptions: FormatDateOptions | undefined =
+  typeof browserFeatureDetectionDate.resolvedOptions === 'undefined'
+    ? undefined
+    : browserFeatureDetectionDate.resolvedOptions();
 
 export function formatDate(
   date: Date,
   locales: string | string[],
   options: FormatDateOptions = {},
 ) {
+  const hourCycleRequired =
+    resolvedOptions != null &&
+    options.hour12 != null &&
+    resolvedOptions.hourCycle != null;
+
+  if (hourCycleRequired) {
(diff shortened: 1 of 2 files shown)


packages/dates/src/utilities/formatDate.ts
```diff
@@ -1,9 +1,13 @@
-import {memoize} from '@shopify/function-enhancers';
-
-const memoizedGetDateTimeFormat = memoize(
-  getDateTimeFormat,
-  dateTimeFormatCacheKey,
-);
+const intl = new Map();
+const memoizedGetDateTimeFormat = function(locale, options) {
```
TheMallen: Why does this not cause leaks but the implementation from `function-enhancers ` does? Could we fix it there so that we don't risk this happening again in the future?

helloneele: The function enhancers implementation doesn't memoize when server side rendering. https://github.com/Shopify/quilt/blob/master/packages/function-enhancers/src/memoize.ts#L18

helloneele: I don't know what the reasoning was behind not memoizing when SSRing, I mostly just wanted to try this out if that fixes things. It does locally, but wanted to see it in production. Happy to change this later on.

michenly: SSR means the react app only render the one time. 
This mean memoize on server actually does not enhance performance, but leave behind memozing that will never be clear off.

this is the reasoning in theory, but maybe in practice there are some thing we are missing?

TheMallen: In the past we've had memory leaks from memoizing on the server, so that was the idea between not doing so. I fear this might make things worse instead of better.

TheMallen: That said, if it does work locally I suppose it's worth a shot.

helloneele: Explanation for not memoizing on the server makes sense. 👍 

> It worries me that we have no real rationale for why it would work though :(

I think the rationale for this is that there is a v8/Node.js issue which leads to `Intl.DateTimeFormat` not being properly cleaned up by the GC, and since we don't memoize we create more and more of those objects. So memoizing those objects minimizes the overall amount of objects we'll create which ends up in less unreachable objects for the GC.

michenly: if we have legit reasons to not skip over server memorization (which I assume could happen else where), then lets add an option to `memoize` to specifically memorize no matter what.

helloneele: I'll try what I have now in production, if that works I'm happy to change the memoizer. 👍

packages/dates/src/utilities/formatDate.ts
```diff
@@ -14,15 +18,14 @@ export function formatDate(
   locales: string | string[],
   options: FormatDateOptions = {},
 ) {
+  if (options.hour12 != null && options.hourCycle != null) {
```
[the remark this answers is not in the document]
helloneele: I'll just ship this to canary for a couple of minutes to see if my fix works, then revert it. I just want to see if my fix works first, then we can re-add what you added. Your fix introduced another `Intl.DateTimeFormat` that might be responsible for leaking memory.

PR #1282, opened 2020-02-14
Update act to use 16.9 version of act there were many improvements in…

vsumner:
…cluding async, nested act resolution, and more

## Description

Fixes (issue #)

## Type of change

## Checklist

### DIFF
--- a/packages/react-testing/src/root.tsx
+++ b/packages/react-testing/src/root.tsx
@@ -76,7 +76,6 @@ export class Root<Props> implements Node<Props> {
   private wrapper: TestWrapper<Props> | null = null;
   private element = document.createElement('div');
   private root: Element<Props> | null = null;
-  private acting = false;
 
   private render: Render;
   private resolveRoot: ResolveRoot;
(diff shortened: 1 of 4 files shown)


packages/react-testing/src/root.tsx
```diff
@@ -95,19 +94,23 @@ export class Root<Props> implements Node<Props> {
...
-    }
+    reactAct(() => {
+      result = action();
+    });
+    updateWrapper();
+    return result;
+  }
 
-    this.acting = true;
+  act<T>(action: () => T, {update = true} = {}): T {
```
TheMallen: ? why both?

vsumner: I found that the issue I ran into was on all internal `act` would behave inconsistently and fail because of the way we were resolving promises.

vsumner: That said. The "external" `act` needs to resolve those promises for external async actions.

IF we didn't need to run `updateWrapper` I'd love to just use `react-dom/test-utils` directly.

PR #1284, opened 2020-02-14
@shopify/react-form-state README adds mention of @shopify/react-form

alanthai:
## Description

Make the newer react form hook library more visible for those looking for a more functional approach for form state handling.

## Type of change
Documentation update

### DIFF
--- a/packages/react-form-state/README.md
+++ b/packages/react-form-state/README.md
@@ -5,7 +5,7 @@
 
 Manage react forms tersely and safely-typed with no magic.
 
-Check out [@shopify/react-form](https://github.com/Shopify/quilt/tree/master/packages/react-form) for our newer form hooks.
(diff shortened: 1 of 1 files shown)


packages/react-form-state/README.md
```diff
@@ -5,6 +5,8 @@
 
 Manage react forms tersely and safely-typed with no magic.
 
+Check out [@shopify/react-form](https://github.com/Shopify/quilt/tree/master/packages/react-form) for our newer form hooks.
```
michenly: ❤️ that you are adding this call out. If you can add some explanation to why someone would want to use `react-form` instead that would be even beter!

tenor

alanthai: Attempt 2:

`This library is now superseded by [@shopify/react-form](https://github.com/Shopify/quilt/tree/master/packages/react-form) as it allows you to write the preferred, functional, and hooks-driven React components over class-based ones.`

michenly: well, I mean to 👍 once the new text is in the code

PR #1285, opened 2020-02-15
add alwaysSendQuery option for PersistedLink

dsanders11:
## Description

Adds an option for `createPersistedLink` to always send the query, rather than waiting for the server to ask.

Two motivating use-cases:

### Development

This is useful for development, as it makes debugging easier when the query is visible in requests.

### SSR

It's also useful when doing SSR, as upload link speed isn't a factor there, so always sending the query avoids an extra round-trip with the API server.

### Usage

```javascript
createPersistedLink({
  alwaysSendQuery: process.env.NODE_ENV === 'development' || isServer
})
```

## Type of change

- [x] @shopify/graphql-persisted - Minor: New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/packages/graphql-persisted/src/apollo.ts
+++ b/packages/graphql-persisted/src/apollo.ts
@@ -9,6 +9,7 @@ import {
 import {CacheMissBehavior} from './shared';
 
 interface Options {
+  alwaysSendQuery?: boolean;
   idFromOperation?(operation: Operation): string | undefined | null;
 }
 
@@ -32,7 +33,10 @@ export class PersistedLink extends ApolloLink {
     }
 
     return new Observable(observer => {
-      const {idFromOperation = defaultIdFromOperation} = this.options;
+      const {
+        alwaysSendQuery = false,
+        idFromOperation = defaultIdFromOperation,
+      } = this.options;
       const id = idFromOperation(operation);
 
       if (typeof id !== 'string' || this.sendAlwaysIds.has(id)) {
@@ -46,7 +50,7 @@ export class PersistedLink extends ApolloLink {
 
       operation.setContext({
         http: {
-          includeQuery: false,
+          includeQuery: alwaysSendQuery,
           includeExtensions: true,
         },
       });
--- a/packages/graphql-persisted/src/apollo.ts
+++ b/packages/graphql-persisted/src/apollo.ts
@@ -9,7 +9,7 @@ import {
 import {CacheMissBehavior} from './shared';
 
 interface Options {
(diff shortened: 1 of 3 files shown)


dsanders11:
@alexandcote this PR could use a look if you have some time after the holidays, thanks!

alexandcote:
First of all, Thanks @dsanders11 for your contribution! Sorry for the delay... I will take care of reviewing this PR and assigning other folks. In the meantime, do you think you could rebase your PR on master?

Thanks a lot 💯

packages/graphql-persisted/src/apollo.ts
```diff
@@ -46,7 +50,7 @@ export class PersistedLink extends ApolloLink {
 
       operation.setContext({
         http: {
-          includeQuery: false,
+          includeQuery: alwaysSendQuery,
```
lemonmade, line 53: I'd probably change this option to match the context field name (`includeQuery`). The option name still bugs me a bit, because it technically applies to more than just queries, but Apollo uses "query" and "GraphQL document" interchangeably in a lot of places so it's probably fine to match.

gmcgibbon: Whatever we name it, it should also be documented in the readme like `idFromOperation` is.

dsanders11: Renamed to `alwaysIncludeQuery` (I think that's what you were suggesting, @lemonmade?). Originally it was "send" to align verbiage wise with the Koa middleware which has a `SendAlways` option.

@gmcgibbon, documented it in the readme.

dsanders11:
@alexandcote, rebased and addressed code review comments.

packages/graphql-persisted/src/apollo.ts
```diff
@@ -32,7 +33,10 @@ export class PersistedLink extends ApolloLink {
     }
 
     return new Observable(observer => {
-      const {idFromOperation = defaultIdFromOperation} = this.options;
+      const {
+        alwaysIncludeQuery = false,
```
lemonmade, line 37: I was proposing just `includeQuery`, since this directly maps to an option with that name in Apollo.

dsanders11: That name seems confusing to me? It loses the explicitness that the query will always be included. I'm not sure what's really gained by excluding the word always. Especially since, as I pointed out, the Koa middleware uses always to describe similar behavior.

gmcgibbon (approved):

lemonmade (approved):

PR #1286, opened 2020-02-18
Export formatDate function

helloneele:
## Description

This exports the `formatDate` function so we can reuse it in [`@shopify/react-i18n`](https://github.com/Shopify/quilt/blob/master/packages/react-i18n/src/i18n.ts#L314-L317) and in shopify web directly. This will reduce code duplication and hopefully also eliminate more memory leaks once we use it.

I'll put up a follow-up PR for the `@shopify/react-i18n` package that makes use of this function.
Follow up PR: https://github.com/Shopify/quilt/pull/1287

## Type of change

## Checklist

### DIFF
--- a/packages/dates/README.md
+++ b/packages/dates/README.md
@@ -30,6 +30,23 @@ const timeZone2 = 'America/Toronto';
 const newDate = applyTimeZoneOffset(date, timeZone1, timeZone2);
 ```
 
+### `formatDate`
+
+Takes in a date object and two additional parameters, the locale and an optional options object. Returns a new date string with the applied locale and options.
+
+```ts
+import {formatDate} from '@shopify/dates';
+
+const date = new Date('2020-02-18Z14:00');
+const locales = 'en';
+const options = {
+  timeZone: 'America/New_York',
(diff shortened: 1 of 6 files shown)


helloneele:
@michenly Oh, didn't realize there was documentation in the readme, sorry. Should be good now 🙂

packages/dates/README.md
```diff
@@ -30,6 +30,23 @@ const timeZone2 = 'America/Toronto';
...
+import {formatDate} from '@shopify/dates';
+
+const date = new Date('2020-02-18Z14:00');
+const locales = 'en';
+const options = {
+  timeZone: 'America/New_York',
+  hour: 'numeric',
+};
+
+const newDate = formatDate(date, locales, options); // 9 AM
```
michenly: The new date is just `9 AM`?

helloneele: Yeah, it's based on the options you pass it apparently. By doing `hour: 'numeric'` it'll just return the hour. If you don't specify hour it'll give you the date like this: `2/18/2020`

PR #1287, opened 2020-02-18
Use @shopify/dates formatDate function and memoize Intl.NumberFormat

helloneele:
## Description

Previous PRs: https://github.com/Shopify/quilt/pull/1277, https://github.com/Shopify/quilt/pull/1286

This makes use of the now exported `formatDate` function from the `@shopify/dates` package.

It also changes `Intl.NumberFormat.format()` to be memoized on SSR. This is because it is, same as `Intl.DateTimeFormat.format()`, leaking memory.

You can test this by doing `node --inspect leak.js` using node v10 or v12

```js
// leak.js

setInterval(() => {
  var b;
  for (let i = 0; i < 100; i++) {
    b = new Intl.NumberFormat("ja-JP", {
      style: "currency",
      currency: "JPY"
    }).format(Math.floor(Math.random() * Math.floor(1000000)));
  }
}, 1);

```

Right after start
image

After a minute
image

I opened an [issue in the node bug tracker](https://github.com/nodejs/node/issues/31914) to hopefully get the V8 fix backported to node v10/v12.

## Type of change

## Checklist

### DIFF
--- a/packages/react-i18n/src/utilities/translate.tsx
+++ b/packages/react-i18n/src/utilities/translate.tsx
@@ -17,6 +17,17 @@ const MISSING_TRANSLATION = Symbol('Missing translation');
 const PLURALIZATION_KEY_NAME = 'count';
 const SEPARATOR = '.';
 
+const numberFormats = new Map();
+export const memoizedNumberFormatter = function(locale, options = {}) {
+  const key = numberFormatCacheKey(locale, options);
+  if (numberFormats.has(key)) {
+    return numberFormats.get(key);
+  }
+  const i = new Intl.NumberFormat(locale, options);
+  numberFormats.set(key, i);
+  return i;
+};
+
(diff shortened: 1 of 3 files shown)


packages/react-i18n/src/utilities/translate.tsx
```diff
@@ -17,6 +17,17 @@ const MISSING_TRANSLATION = Symbol('Missing translation');
 const PLURALIZATION_KEY_NAME = 'count';
 const SEPARATOR = '.';
 
+const numberFormats = new Map();
+export const memoizedNumberFormatter = function(locale, options = {}) {
+  const key = numberFormatCacheKey(locale, options);
```
michenly: because we are using this at at least two places now, I think we should just add an `enabledOnServer` option to `memoize`. But that can also be done in a follow up PR.

PR #1292, opened 2020-02-21
Flush act after async resolve

vsumner:
## Description
When async actions resolve we need to flush any remaining acts.

## Type of change

## Checklist

### DIFF
--- a/packages/react-testing/src/root.tsx
+++ b/packages/react-testing/src/root.tsx
@@ -106,6 +106,8 @@ export class Root<Props> implements Node<Props> {
     this.acting = true;
 
     const afterResolve = () => {
+      // Flush any remaining act before updating the wrapper
+      act(() => {});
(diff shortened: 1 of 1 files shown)


packages/react-testing/src/root.tsx
```diff
@@ -106,6 +106,8 @@ export class Root<Props> implements Node<Props> {
     this.acting = true;
 
     const afterResolve = () => {
+      // Flush any remaining act before updating the wrapper
+      act(() => {});
```
TheMallen: You love to see it. 

Two things jump to mind:
- Can we add a test for the type of thing this is fixing so nobody deletes this despite the comment? 
- Let's use an explicit `noop` function rather than an inline one for added clarity
-

vsumner:
Didn't actually do anything.

PR #1296, opened 2020-02-24
Add option to use a custom comparator for dirty state checking

michaelmelanson:
## Description

This change adds an option to the `useField` hook in `react-form` to use a custom comparator for checking whether a field is dirty. This allows you to, for example, apply a deep equals function for deep equality checking if you field requires it:

```
  const myField = useField<Type>({
    value: initialValue,
    validates: () => false,
    dirtyStateComparator: _.isEqual
  });
```

This is required to resolve a bug we have an in upcoming Web feature (Slack me for more details) where we are storing a list of objects in a field, and currently we remain in the dirty state if the user manually resets the form to its default state. With this feature, we correctly detect that the form is again in its default state, and report that it's now clean.

**Note:** I recommend turning on "Hide whitespace changes" when reviewing this feature to make the diff readable.

## Type of change

- [x] `react-form` Minor: New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/packages/react-form/src/hooks/field/field.ts
+++ b/packages/react-form/src/hooks/field/field.ts
@@ -1,8 +1,12 @@
 import {useCallback, useEffect, useMemo, ChangeEvent} from 'react';
 import isEqual from 'fast-deep-equal';
 
-import {Validates, Field} from '../../types';
-import {normalizeValidation, isChangeEvent} from '../../utilities';
+import {Validates, Field, DirtyStateComparator} from '../../types';
+import {
+  normalizeValidation,
+  isChangeEvent,
+  defaultDirtyComparator,
+} from '../../utilities';
 
 import {
   updateAction,
@@ -15,6 +19,7 @@ import {
 export interface FieldConfig<Value> {
   value: Value;
   validates: Validates<Value>;
+  dirtyStateComparator?: DirtyStateComparator<Value>;
 }
 
 /**
@@ -105,10 +110,14 @@ export function useField<Value = string>(
   input: FieldConfig<Value> | Value,
   dependencies: unknown[] = [],
 ): Field<Value> {
-  const {value, validates} = normalizeFieldConfig(input);
+  const {
+    value,
+    validates,
+    dirtyStateComparator = defaultDirtyComparator,
+  } = normalizeFieldConfig(input);
   const validators = normalizeValidation(validates);
 
-  const [state, dispatch] = useFieldReducer(value);
+  const [state, dispatch] = useFieldReducer(value, dirtyStateComparator);
 
   const resetActionObject = useMemo(() => resetAction(), []);
   const reset = useCallback(() => dispatch(resetActionObject), [
@@ -238,12 +247,13 @@ export function useChoiceField(
 
 function normalizeFieldConfig<Value>(
   input: FieldConfig<Value> | Value,
+  dirtyStateComparator?: DirtyStateComparator<Value>,
 ): FieldConfig<Value> {
   if (isFieldConfig(input)) {
     return input;
   }
 
-  return {value: input, validates: () => undefined};
+  return {value: input, validates: () => undefined, dirtyStateComparator};
 }
 
 function isFieldConfig<Value>(input: unknown): input is FieldConfig<Value> {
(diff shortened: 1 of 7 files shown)


packages/react-form/src/hooks/field/reducer.ts
```diff
@@ -55,62 +55,77 @@ export type FieldAction<Value> =
...
+  dirtyStateComparator,
+}: {
+  dirtyStateComparator: DirtyStateComparator<Value>;
+}): Reducer<FieldState<Value>, FieldAction<Value>> {
+  return (state: FieldState<Value>, action: FieldAction<Value>) => {
+    switch (action.type) {
+      case 'update': {
+        const newValue = action.payload;
+        const {defaultValue} = state;
+        const dirty = dirtyStateComparator(defaultValue, newValue);
```
TheMallen: Any reason not to send the `dirtyStateComparator` as part of the action instead of as an option to creating the reducer?

michaelmelanson: Interesting, I didn't consider that. I've thought of it as being something you'd choose for the whole life of the field, which is why I baked it into the reducer. Is that something you think you'd want to choose on a per-action basis?

TheMallen: I don't think it is likely you would want to send it on a per-action basis, I guess it just feels slightly more like a piece of information that the `update` action cares about specifically more to me than a fact about the reducer overall, but I could really go either way.

michaelmelanson: I agree that the rest of the reducer doesn't care about it so it is a bit weird to have it available through all the actions. But I'm not really sure how to go about addressing that. Do you have any suggestions?

packages/react-form/src/hooks/field/field.ts
```diff
@@ -105,10 +110,14 @@ export function useField<Value = string>(
   input: FieldConfig<Value> | Value,
   dependencies: unknown[] = [],
 ): Field<Value> {
-  const {value, validates} = normalizeFieldConfig(input);
+  const {
+    value,
+    validates,
+    dirtyStateComparator = defaultDirtyComparator,
```
patsissons: i think i would consider pushing this default comparator assignment all the way down to `makeFieldReducer`. let the optional value reach as far as it can and assign its default once the value becomes necessary.

packages/react-form/src/hooks/field/field.ts
```diff
@@ -238,12 +247,13 @@ export function useChoiceField(
 
 function normalizeFieldConfig<Value>(
   input: FieldConfig<Value> | Value,
+  dirtyStateComparator?: DirtyStateComparator<Value>,
```
patsissons: is this needed?

packages/react-form/src/hooks/field/index.ts
```diff
@@ -7,6 +7,7 @@ export {
 } from './field';
 export {
   reduceField,
+  makeFieldReducer,
```
patsissons: do we need to export this?

packages/react-form/src/hooks/field/reducer.ts
patsissons: i think what you have here providing it as a factory parameter is fine

packages/react-form/src/hooks/field/reducer.ts
```diff
@@ -55,62 +55,77 @@ export type FieldAction<Value> =
...
-        ...state,
-        error: undefined,
-        value: newDefaultValue,
-        defaultValue: newDefaultValue,
-        touched: false,
-        dirty: false,
-      };
-    }
-  }
+export function reduceField<Value>(
```
patsissons: this and the `shallowFieldReducer` feel like they may belong somewhere else?

packages/react-form/src/hooks/field/reducer.ts
```diff
@@ -55,62 +55,77 @@ export type FieldAction<Value> =
...
-      return {
-        ...state,
-        dirty: isDirty,
-        value: newValue,
-        touched: true,
-      };
+export function makeFieldReducer<Value>({
+  dirtyStateComparator,
+}: {
+  dirtyStateComparator: DirtyStateComparator<Value>;
```
patsissons: IMO this should be an interface (maybe `ReducerOptions`?)

packages/react-form/src/hooks/field/reducer.ts
michaelmelanson: It feels like that to me too, but I just tried moving it out of that file and couldn't find a place that felt right because the call to `makeFieldReducer` would have to call back here anyway. 😕

PR #1298, opened 2020-02-26
Allow custom headers in the graphql proxy

alexandcote:
## Description

Fixes (issue #)

## Type of change

## Checklist


packages/koa-shopify-graphql-proxy/src/shopify-graphql-proxy.ts
```diff
@@ -16,6 +16,7 @@ export enum ApiVersion {
 
 interface DefaultProxyOptions {
   version: ApiVersion;
+  headers?: Record<string, string>;
```
tsov: Should this be `Record<string, string | number>` ?

packages/koa-shopify-graphql-proxy/README.md
```diff
@@ -85,3 +85,27 @@ app.use(
...
+
+If you need to pass headers to the proxied request, you can than add those headers during the configuration of the proxy.
+
+```javascript
+// server/index.js
+import koa from 'koa';
+import session from 'koa-session';
+import proxy, {ApiVersion} from '@shopify/koa-shopify-graphql-proxy';
+
+const app = koa();
```
tsov: I know the other examples have this lowercase, but this should technically be `app = new Koa()`, no?

packages/koa-shopify-graphql-proxy/README.md
```diff
@@ -85,3 +85,27 @@ app.use(
...
+
+app.use(session());
+
+app.use(() =>
+  proxy({
```
tsov: Maybe give an example where we pull an existing request header from context and pass it in the proxy. For example `X-Request-ID`

alexandcote:
We don't need this PR since `koa-better-http-proxy` already fowards the headers.

alexandcote:
We don't need this PR since `koa-better-http-proxy` already fowards the headers.

PR #1300, opened 2020-02-27
🔥 remove the mentioned of locale from quilt_rails and link the exampl…

michenly:
…e to react-server directly

### DIFF
--- a/gems/quilt_rails/README.md
+++ b/gems/quilt_rails/README.md
@@ -390,39 +390,16 @@ With SSR enabled React apps, state must be serialized on the server and deserial
 
 #### Customizing the node server
 
-By default, sewing-kit bundles in `@shopify/react-server-webpack-plugin` for `quilt_rails` applications to get apps up and running fast without needing to manually write any node server code. If what it provides is not sufficient, a custom server can be defined by adding a `server.js` or `server.ts` file to the app folder.
+By default, sewing-kit bundles in [`@shopify/react-server-webpack-plugin`](../../packages/react-server-webpack-plugin/README.md) for `quilt_rails` applications to get apps up and running fast without needing to manually write any node server code.
+
+If what it provides is not sufficient, a completely custom server can be defined by adding a `server.js` or `server.ts` file to the `app/ui` folder. The simplest way to customize the server is to export the object created by [`@shopify/react-server`](../../packages/react-server/README.md#node-usage)'s `createServer` call in `server.ts` file.
 
 ```
-└── app
+└── appeon
    └── ui
       └─- app.{js|ts}x
       └─- index.{js|ts}
       └─- server.{js|ts}x
-```
-
-```tsx
-// app/ui/server.tsx
-import '@shopify/polyfills/fetch';
-import {createServer} from '@shopify/react-server';
-import {Context} from 'koa';
-import React from 'react';
-
-import App from './app';
-
-// The simplest way to build a custom server that will work with this library is to use the APIs provided by @shopify/react-server.
-// https://github.com/Shopify/quilt/blob/master/packages/react-server/README.md#L8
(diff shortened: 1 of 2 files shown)


gems/quilt_rails/README.md
```diff
@@ -400,30 +400,9 @@ By default, sewing-kit bundles in `@shopify/react-server-webpack-plugin` for `qu
...
-    const whatever = /* do something special with the koa context */;
-    // any special data we add to the incoming request in our rails controller we can access here to pass into our component
-    return <App server someCustomProp={whatever} location={ctx.request.url} locale={locale} />;
-  },
-});
+Use [`@shopify/react-server`](../../packages/react-server/README.md) to create a simple custom node server.
 
-export default app;
-```
+Alternatively, if no customization on the server is needed, [`@shopify/react-server-webpack-plugin`](../../packages/react-server-webpack-plugin/README.md) can be use to have a virtual server created for you.
```
michenly: Move to linking the react-server doc. So the example it reference will always be up to date.

kvendrik: Thanks for updating the docs on this so quickly! Two thoughts:
- I find the instructions here a bit confusing. Won't quilt_rails set up the node server for you? Therefor, if you don't need to customize it you won't need to do anything right?
- Secondly, I feel like a missing piece here is why you might want to customize it. In my case I'm trying to get the locale to pass it into the i18n manager which doesn't seem to be in there by default. I feel like for common use-cases such as that we should have some kind of docs that explain how to do that.

michenly: * You are right. `quilt_rails` itself does not give you a node server, but by using `sewing-kit` it will automatically install `@shopify/react-server-webpack-plugin` and use it for you. So it does seem automatic from the outside, I just completely forgot about it 😬 

* I like the idea of doing common use-cases. Locale is not one thou as I think you can do it without. I will do some research but the only real need for customization I know of is web doing crazy auth things. But they dont use `quilt_rails`

gems/quilt_rails/README.md
```diff
@@ -390,7 +390,7 @@ With SSR enabled React apps, state must be serialized on the server and deserial
 
 #### Customizing the node server
 
-By default, sewing-kit bundles in `@shopify/react-server-webpack-plugin` for `quilt_rails` applications to get apps up and running fast without needing to manually write any node server code. If what it provides is not sufficient, a custom server can be defined by adding a `server.js` or `server.ts` file to the app folder.
+By default, sewing-kit bundles in [`@shopify/react-server-webpack-plugin`](../../packages/react-server-webpack-plugin/README.md) for `quilt_rails` applications to get apps up and running fast without needing to manually write any node server code. If what it provides is not sufficient, a completely custom server can be defined by adding a `server.js` or `server.ts` file to the app folder.
```
kvendrik: Love it, very clear.

gems/quilt_rails/README.md
```diff
@@ -400,30 +400,7 @@ By default, sewing-kit bundles in `@shopify/react-server-webpack-plugin` for `qu
...
-  render: (ctx, {locale}) => {
-    const whatever = /* do something special with the koa context */;
-    // any special data we add to the incoming request in our rails controller we can access here to pass into our component
-    return <App server someCustomProp={whatever} location={ctx.request.url} locale={locale} />;
-  },
-});
-
-export default app;
-```
+The simplest way to build a custom server that will work with this library is to use the APIs provided by [`@shopify/react-server`](../../packages/react-server/README.md).
```
kvendrik:
```suggestion
The simplest way to customize the server is to create a `server.tsx` file in your `app/ui` folder. This server file should export a [`@shopify/react-server`](../../packages/react-server/README.md) `createServer` call.
```

PR #1301, opened 2020-03-03
remove need to whitelist locales in @shopify/address

TayKangSheng:
## Description

This PR is to update @shopify/address to expand the number of supported locales. CountryService has expanded its list of SupportedLocales with all locales that are supported by ShopifyI18n and has also build in a default to fallback to `en` mechanism (https://github.com/Shopify/country-service/pull/409). Therefore we can remove the need to check against and maintain a list of supported locales on `@shopify/address-consts`. 

The motivation behind this work is to allow for `zh-CN` and `zh-TW` in the address form in credit card billing form. Fixes https://github.com/Shopify/intl-china/issues/110

Below is the list of languages supported by shopify/web, and its state of translations for address forms before and after this PR. 

| country | before this PR | after this PR |
| --- | --- | --- | 
| Dansk | ✅| ✅|
| Deutsch | ✅| ✅|
| English | ✅| ✅ |
| Español | ✅| ✅|
| Français | ✅| ✅|
| Italiano | ✅| ✅|
| Malay | ❌ | ✅|
| Nederlands | ✅| ✅|
| Norsk | ❌| ✅|
| Polski | ❌| ✅|
| Português | ✅| ✅|
| Português (BR) | ✅| ✅|
| Suomi | ❌| ✅|
| Türkçe | ❌| ✅|
| Svenska | ❌| ✅|
| čeština  | ❌| ✅|
| Hindi (हिन्दी) | ❌| ✅, but 1 field does not have translations in CountryDb |
| Thai | ❌| ✅|
| Japanese | ✅| ✅|
| Chinese (Simplified) | ❌| ✅|
| Chinese (Traditional) | ❌| ✅|
| Korean | ❌| ✅|

## Side note

Although the removal of some fields to `ResponseError` is unnecessary for this to work, however the interface for `ResponseError` is not aligned to [GraphQl specification](http://spec.graphql.org/June2018/#sec-Errors). So I have taken upon myself to fix this while adding the new fields to the interface. Let me know if this should be in another PR.  

## Type of change

- [x] `@shopify/address-consts` Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] `@shopify/address-mocks` Minor: New feature (non-breaking change which adds functionality)

## Checklist

- [x] 🎩 with shopify/web


packages/address/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@shopify/address",
-  "version": "2.7.17",
+  "version": "2.8.17",
```
ismail-syed: Again, same as https://github.com/Shopify/quilt/pull/1302/files#r388275151

packages/address/src/loader.ts
```diff
@@ -30,7 +29,7 @@ export const loadCountries: (
     | LoadCountriesResponse
     | ResponseError = await response.json();
 
-  if ('errors' in countries) {
+  if (!('data' in countries) && 'errors' in countries) {
```
ismail-syed: Why remove the `!('data' in countries)` guard?

packages/address/package.json
ismail-syed: Also, this version would be `2.8.0`

ismail-syed:
Give that this work requires changes from a few different PRs going it may be worth 🎩 ing all of them in web. We have a documented [guide](https://github.com/Shopify/quilt/blob/master/documentation/guides/release-and-deploy.md#releases-for-ing-changes). 

Though you might have to fight with lerna to get all the version correct since these PRs are all independent. For 🎩ing, I'd recommend 
* `git cherry-pick`ing all these changes into one branch
* Revert all the manual version bumps
* Run through the lerna release steps in the  [guide](https://github.com/Shopify/quilt/blob/master/documentation/guides/release-and-deploy.md#releases-for-ing-changes) for 🎩 ing
* Consume in the project you're testing this on and test.

packages/address/package.json
TayKangSheng: Right good catch! Totally slipped my mind.

packages/address/src/loader.ts
TayKangSheng: @ismail-syed Because CountryService will return both `data` and `errors` if the requested locale is supported by ShopifyI18n, but is not supported by CountryDb. I will include an example in the description. Thanks for the headsup.  In this case I'll update the LoadCountriesResponse signature to include a optional `errors` field.

TayKangSheng:
Initially I was under the impresssion that I have to split this PR, https://github.com/Shopify/quilt/pull/1302 and https://github.com/Shopify/quilt/pull/1303 into separate PRs and release the packages in order. However it turns out the release process is that all codes will be merged into master and everything should be released in one go. Therefore I am going to collapse the 2 other PRs into this PR. 

By having all the changes required for this feature in 1 single PR, it will be easier to review and 🎩 the code changes.

packages/address-consts/CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+- [Major] Add `errors` to `LoadCountriesResponse` and `LoadCountryResponse`;
```
ismail-syed: I would change `[Major]` to `[Breaking change]`

TayKangSheng:
1. `@shopify/address (v2.8.0)` requires `@shopify/address-consts (v2.0.0)` and `@shopify/address-mocks (v1.3.0)`
2. `@shopify/address-mocks (v1.3.0)` requires `@shopify/address-consts (v2.0.0)`

But doing `yarn dev-release` only bumps the version of each packages, without bumping the required dependencies (Which is understandable because I wouldn't expect Lerna to know which dependencies need to be bumped). However adding a commit to bump the dependencies will result in build error before the newly versioned packages are not released yet. 

Screenshot 2020-03-09 at 1 38 17 PM

All in all, it seems like I shouldn't be putting bumping all 3 packages into the same PR. The sequence should be: 
- merge PR to improve and bump `@shopify/address-consts`
- release `@shopify/address-consts` ([requires admin](https://github.com/Shopify/quilt/blob/master/documentation/guides/release-and-deploy.md#creating-releases))
- merge PR to improve and bump `@shopify/address-mocks` so that it can use the new `@shopify/address-consts` version. 
- release `@shopify/address-mocks` ([requires admin](https://github.com/Shopify/quilt/blob/master/documentation/guides/release-and-deploy.md#creating-releases))
- merge PR to improve and bump `@shopify/address` so that it can use the new `@shopify/address-consts` and `@shopify/address-mocks` version. 
- release `@shopify/address` ([requires admin](https://github.com/Shopify/quilt/blob/master/documentation/guides/release-and-deploy.md#creating-releases))

cc @ismail-syed

TayKangSheng:
@ismail-syed I've tested with the dev release for 🎩 ing. 

the released packages were 
- `@shopify/address@2.0.0-expand-locale-support-for-shopify-address.1`
- `@shopify/address-mocks@2.0.0-expand-locale-support-for-shopify-address.1`
- `@shopify/address-consts@2.0.0-expand-locale-support-for-shopify-address.1`

Shipit built it directly into this PR, i'm not sure what is the next step now. Can you let me know whats after this? do i just merge?

TayKangSheng:
> The following steps require admin access to the Shopify/quilt GitHub repo.

@ismail-syed How do I get the above packages released? My guess is that I can follow the steps of the [release-and-deploy.md#creating-releases](https://github.com/Shopify/quilt/blob/master/documentation/guides/release-and-deploy.md#creating-releases) and wait for an admin to merge the PR?

michenly:
@TayKangSheng these packages are being release.

TayKangSheng:
Thank you @michenly!! :grinning:

PR #1304, opened 2020-03-04
Deprecate @shopify/react-shopify-app-route-propagator

alexandcote:
## Description

Since this package moved to `@shopify/app-bridge-react`, we need to remove the `package.json` file otherwise lerna will try to cut a new release for this package.


packages/react-shopify-app-route-propagator/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+This package is now part of `@shopify/app-bridge-react`, starting from version 1.17.0
```
michenly: huh, you probably wont need this since its in the README anyway. But since its debatable its up to you!
(if you want, it can be deleted in the lerna release process)

PR #1305, opened 2020-03-04
[react-form] Bug fix: Use `dirtyStateComparator` from config.

michaelmelanson:
## Description

My previous pull request (#1296) to add the `dirtyStateComparator` option had a problem that snuck in while addressing a review comment. I made the mistake of adding the `dirtyStateComparator` as a function argument, which meant it didn't actually use the value with the same name from the field configuration, and then somehow didn't notice this during tophatting. 😕 

It meant that you had to use the flag like this:

```
  const myField = useField(
    originalValue, 
    undefined, //dependencies
    (first, second) => compare(first, second) // comparator
  );
```

... instead of inside the config block as intended:

```
  const myField = useField({
    value: originalValue,
    validates: () => undefined,
    dirtyStateComparator: (first, second) => compare(first, second),
  });
```

## Type of change

- [x] `react-form` Patch: Bug/ Documentation fix (non-breaking change which fixes an issue or adds documentation)

## Checklist

### DIFF
--- a/packages/react-form/CHANGELOG.md
+++ b/packages/react-form/CHANGELOG.md
@@ -7,6 +7,10 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


michaelmelanson:
No worries, it's not your fault. Fixed now anyway.

PR #1306, opened 2020-03-04
Improve use timeout

sambostock:
## Description

This updates the `useTimeout` hook in `shopify/react-hooks` to support the same API and behaviour as the `useInterval` hook added in #1241.

Specifically, it changes the behaviour of

- `callback`: Previously, any time this prop changes, the timeout is reset. Now, the `callback` is free to change any time before the timeout fires, since the `callback` is stored in a ref.
- `delay`: `null` can now be passed to clear the timeout.

## Type of change

- [x] `react-hooks` Minor: The change to `delay` accepting `null`, in addition to numbers, as well as making changing the callback not reset the timeout.

## Checklist

- [x] I have updated the README
- [x] I have rebased after #1241 was merged


packages/react-hooks/src/hooks/tests/timeout.test.tsx
```diff
@@ -18,38 +16,84 @@ describe('useTimeout', () => {
...
+    timer.runTimersToTime(HALF_A_SECOND);
+    expect(callback).not.toHaveBeenCalled();
+
+    const newCallback = jest.fn();
+    wrapper.setProps({callback: newCallback});
+
+    timer.runTimersToTime(HALF_A_SECOND);
+    expect(newCallback).toHaveBeenCalled();
+    expect(callback).not.toHaveBeenCalled();
+  });
```
sambostock: I'm not sure if this should be a breaking change or not. Technically, it does break the previous behaviour, though it would be weird behaviour to depend on, so I'm leaning towards `breaking`, but I want to confirm with reviewers.

### Previously

Changing the `callback` prop causes the timeout to be reset, effectively restarting the timer. If the timeout has already fired, a new timeout will be set, and the new callback will be run (in addition to the one that already ran).

### Now

Changing the `callback` does nothing. When the timeout fires, it will run whatever the current timeout is. If the timeout has already fired, nothing will happen, unless the `delay` is also changed.

GoodForOneFare:
Tagging in Web Foundation for review as I'm tried up in checkout's code red.

packages/react-hooks/src/hooks/tests/timeout.test.tsx
TheMallen: I kinda see this as a bugfix tbh

sambostock:
Thanks for the review @TheMallen!

I'm ok with the change to `callback` being considered a bugfix. Does the new behaviour of passing a `null` timeout to cancel (new backwards compatible feature) make it a minor version change then?

TheMallen:
@sambostock  makes sense to me!

packages/react-hooks/src/hooks/tests/timeout.test.tsx
sambostock: 👌 I've updated the Changelog to include two entries.
- Added support for `null` `delay`
- Fixed changing callback resetting the timeout

PR #1307, opened 2020-03-05
[react-form] reduceFields utilities

alanthai:
## Description

a new `reduceFields` function is exposed that works similar to `Array.reduce` but for tree structures, in particular to reduce over Fields. Since we have several methods that visit each Field node, a general `reduceFields` function is beneficial to simplify the codebase and remove duplication. `reduce` is a foundational function for which other functional methods such as `map`, `filter`, `some` and more are derived.

There is memoization problems with `dirty.ts` because the fields could be deeply nested. This will be fixed

## Type of change

## Checklist


packages/react-form/src/utilities.ts
```diff
@@ -68,32 +93,67 @@ export function propagateErrors(
   });
 }
 
-export function validateAll(fieldBag: {[key: string]: FieldOutput<any>}) {
-  const fields = Object.values(fieldBag);
-  const errors: FormError[] = [];
+// Reduce function similar to Array.reduce() for a tree-like FieldBag
+export function reduceFields<V>(
```
ismail-syed: Can we add more test coverage for this? Also, some docs in the readme.

packages/react-form/src/utilities.ts
```diff
@@ -32,6 +31,32 @@ export function mapObject<Output>(
   }, {}) as Output;
 }
 
+// Eg: set({a: 1}, ['b', 'c'], 2) // => {a: 1, b: {c: 2}}
+function setObject<T extends Object>(
```
ismail-syed: Can we add some test coverage for this as well?

alanthai: It is currently just an internal function (non-exported) for `getValue`, but if there is value later, I was thinking we could expose it.

alanthai:
@ismail-syed Docs are updated! :)

PR #1310, opened 2020-03-06
Use memoized number format for currency symbol

helloneele:
## Description

Fixes (issue #)

This should be the last step of eliminating memory leaks through the creation of new `Intl.DateTimeFormat` and `Intl.NumberFormat` and calling `format()` on them. There are no more case of a new `NumberFormat` instance being created in `quilt`.

I added some really basic tests to confirm that `formatCurrency()` is relying on `memoizedNumberFormatter()` (and `memoizedNumberFormatter` is already asserted to memoize correctly).

related to #1277 and #1287
related to https://github.com/nodejs/node/issues/31914

## Type of change

## Checklist

### DIFF
--- a/packages/react-i18n/src/i18n.ts
+++ b/packages/react-i18n/src/i18n.ts
@@ -333,7 +333,6 @@ export class I18n {
     return this.getCurrencySymbolLocalized(this.locale, currency);
   };
 
-  @memoize((currency: string, locale: string) => `${locale}${currency}`)
   getCurrencySymbolLocalized(locale: string, currency: string) {
     return getCurrencySymbol(locale, {currency});
   }
--- a/packages/react-i18n/src/index.ts
+++ b/packages/react-i18n/src/index.ts
@@ -3,7 +3,7 @@ export {I18nContext} from './context';
 export {I18n} from './i18n';
 export {useI18n} from './hooks';
 export {withI18n, WithI18nProps} from './decorator';
-export {translate} from './utilities';
+export {memoizedNumberFormatter, translate} from './utilities';
 export {
   I18nDetails,
   LanguageDirection,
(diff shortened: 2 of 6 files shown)


packages/react-i18n/src/i18n.ts
```diff
@@ -333,7 +333,6 @@ export class I18n {
     return this.getCurrencySymbolLocalized(this.locale, currency);
   };
 
-  @memoize((currency: string, locale: string) => `${locale}${currency}`)
```
patsissons, line 336: this is no longer required as our `getCurrencySymbol()` below is performing the memoization

packages/react-i18n/src/index.ts
```diff
@@ -3,7 +3,7 @@ export {I18nContext} from './context';
 export {I18n} from './i18n';
 export {useI18n} from './hooks';
 export {withI18n, WithI18nProps} from './decorator';
-export {translate} from './utilities';
+export {memoizedNumberFormatter, translate} from './utilities';
```
patsissons, line 6: we want to expose this memoize function so that `react-i18n` consumers can also avoid memory leaks if they need access to a number formatter.

packages/react-i18n/src/utilities/money.ts
```diff
@@ -29,12 +31,12 @@ export function getCurrencySymbol(
   return elements;
 }
 
-function formatCurrency(
+export function formatCurrency(
```
patsissons, line 34: exported here so we can test the function, but not re-exported in `utilities`

packages/react-i18n/src/utilities/translate.tsx
```diff
@@ -18,15 +18,18 @@ const PLURALIZATION_KEY_NAME = 'count';
 const SEPARATOR = '.';
 
 const numberFormats = new Map();
-export const memoizedNumberFormatter = function(locale, options = {}) {
+export function memoizedNumberFormatter(
+  locale: string,
+  options: Intl.NumberFormatOptions = {},
+) {
```
patsissons, line 24: I was mostly just adding types here, but this seemed like a natural adjustment to make. couldn't think of why we would want a const declaration over a function declaration.

PR #1311, opened 2020-03-06
Update dev release dist-tag to use beta

ismail-syed:
## Description

We got reports of our [dev release](https://github.com/Shopify/quilt/blob/master/documentation/guides/release-and-deploy.md#releases-for-ing-changes) workflow publishing to the `latest` tag instead of `dev`. shipit [doesn't support](https://github.com/Shopify/shipit-engine/blob/master/lib/snippets/publish-lerna-independent-packages#L7-L12) the `dev` dist-tag. This PR updates the dist-tag to use `beta`.

### DIFF
--- a/documentation/guides/release-and-deploy.md
+++ b/documentation/guides/release-and-deploy.md
@@ -52,16 +52,16 @@ git push origin master --follow-tags
 
 # Releases for 🎩ing changes?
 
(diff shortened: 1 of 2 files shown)


ismail-syed:
Going to merge this to prevent any accidental publishes that use the next tag

PR #1312, opened 2020-03-07
Solve a potential security vulnerability

alexandcote:
## Description
This PR bump `serialize-javascript` to solve a potential security vulnerability.
The new version of the `serialize-javascript` contains two breaking changes.

1. [`undefined` is now serialized](https://github.com/yahoo/serialize-javascript/releases/tag/v2.0.0)
2. [`Infinity` is now serialized](https://github.com/yahoo/serialize-javascript/releases/tag/v3.0.0)

The first breaking change can potentially break some of our apps. If you check how we deserialize, we use `JSON.parse`. Unfortunately, this code will throw.
```ts
JSON.parse({foo: undefined})
```
After some investigation, I discovered that `serialize-javascript` offers an [option that expects only valid JSON.](https://github.com/yahoo/serialize-javascript#optionsisjson)

>This option is a signal to serialize() that the object being serialized does not contain any function or regexps values. This enables a hot-path that allows serialization to be over 3x faster. If you're serializing a lot of data, and know its pure JSON, then you can enable this option for a speed-up.
>
>**Note: That when using this option, the output will still be escaped to protect against XSS.**

This small option ensures that the value serialized will be deserializable by `JSON.parse`. This also means that `{foo: undefined}` will never be a valid serialized value. 

**Bonus, the serialization should be 3x faster 🎉**

## Type of change

- [x] `@shopify/react-html` Minor: Fix a potential vulnerability issue with an internal package and ensure the serialized data will be deserializable.
- [x] `@shopify/react-serializer` Minor: Fix a potential vulnerability issue with an internal package and ensure the serialized data will be deserializable.
- [x] `@shopify/react-google-analytics` Minor: Bump a staled dependency.
- [x] `@shopify/react-server` Minor: Bump a staled dependency.
- [x] `@shopify/react-form-state` Minor: Bump a staled dependency.

## Checklist


packages/react-serialize/src/tests/utilities.test.ts
```diff
@@ -29,9 +29,41 @@ describe('getSerialized()', () => {
   });
 
   it('returns the deserialized data', () => {
-    const data = {foo: 'window.location = "http://dangerous.com"'};
+    const date = new Date();
+    const data = {
```
alexandcote: I added a test to demonstrate almost every use case.

packages/tsconfig.json
```diff
@@ -59,7 +59,6 @@
     {"path": "./react-serialize"},
     {"path": "./react-server"},
     {"path": "./react-server-webpack-plugin"},
-    {"path": "./react-shopify-app-route-propagator"},
```
alexandcote: This was something we forgot to do when we deprecated `react-shopify-app-route-propagator`

yarn.lock
```diff
@@ -10310,15 +10315,15 @@ tempfile@^1.1.1:
     uuid "^2.0.1"
 
 terser-webpack-plugin@^1.4.1:
-  version "1.4.1"
-  resolved "https://registry.yarnpkg.com/terser-webpack-plugin/-/terser-webpack-plugin-1.4.1.tgz#61b18e40eaee5be97e771cdbb10ed1280888c2b4"
-  integrity sha512-ZXmmfiwtCLfz8WKZyYUuuHf3dMYEjg8NrjHMb0JqHVHVOSkzp3cW2/XG1fP3tRhqEqSzMwzzRQGtAPbs4Cncxg==
+  version "1.4.3"
```
alexandcote: I bumped this package to get the new version of `serialize-javascript`.

packages/react-html/package.json
```diff
@@ -25,11 +25,10 @@
   "dependencies": {
     "@shopify/react-effect": "^3.2.11",
     "@shopify/react-hydrate": "^1.1.14",
-    "@shopify/react-serialize": "^1.0.22",
```
alexandcote: This package is not used anymore.

packages/react-serialize/CHANGELOG.md
```diff
@@ -4,10 +4,10 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
-<!-- Unreleased changes should go to UNRELEASED.md -->
+## Unreleased
```
michenly: actually, can we just deprecate `react-serialize` all together?

You can try out the new deprecation guide
https://github.com/Shopify/quilt/blob/master/documentation/guides/package-deprecation.md

packages/react-i18n/README.md
```diff
@@ -327,30 +327,53 @@ i18n.getTranslationTree('MyComponent.countries');
...
+This library uses the [`@shopify/react-effect`](https://github.com/Shopify/quilt/tree/master/packages/react-effect) package to allow translations to be extracted alongside other asynchronous side effects on the server. To make use of this, you will need to keep a reference to the `I18nManager` for your app. Then, import the `extract` function from `@shopify/react-effect`, and call it with your top-level component. Call the manager’s `extract` method to get an opaque representation of the translations that were loaded in that tree.
+
+We recommend you to use [`@shopify/react-html`](https://github.com/Shopify/quilt/tree/master/packages/react-html) to serialize the extracted translations.
```
michenly: actually, `@shopify/react-i18n-universal-provider` does a lot of these now

packages/react-serialize/CHANGELOG.md
```diff
@@ -4,10 +4,10 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/).
 
-<!-- Unreleased changes should go to UNRELEASED.md -->
+## ❗️ Deprecated on version 1.0.12 ❗️
```
michenly:
```suggestion
## ❗️ Deprecated on version 1.0.22 ❗️
```

michenly:
I think you are missing the `README.md` update 
`yarn plop docs` will take care of it

packages/react-i18n/README.md
alexandcote: Yes, My first version was using `@shopify/react-i18n-universal-provider` but then I was wondering if we should plug all the other packages into this readme or we should let the readme of `@shopify/react-i18n-universal-provider` explain everything.

Thought?

michenly: I can't think of any reason why someone would want to set this up themselves instead of using `@shopify/react-i18n-universal-provider`, so lets keep just link out to it.

alexandcote: Cool I updated the section and the example to use `@shopify/react-i18n-universal-provider`

alexandcote:
I'm 🛳️, any other comments will be addressed in a follow-up ✌️

PR #1316, opened 2020-03-10
Upgrade koa and its types to latest

vsumner:
## Description
This PR upgrades `koa` and `@types/koa` to the latest version.

## Type of change

## Checklist

### DIFF
--- a/packages/jest-dom-mocks/src/promise.ts
+++ b/packages/jest-dom-mocks/src/promise.ts
@@ -12,7 +12,7 @@ export default class Promise {
     }
 
     jest.useFakeTimers();
-    global.Promise = FakePromise;
+    global.Promise = FakePromise as any;
     this.isUsingFakePromise = true;
   }
 
--- a/yarn.lock
+++ b/yarn.lock
@@ -7576,14 +7576,7 @@ mime-db@1.43.0:
   resolved "https://registry.yarnpkg.com/mime-db/-/mime-db-1.43.0.tgz#0a12e0502650e473d735535050e7c8f4eb4fae58"
   integrity sha512-+5dsGEEovYbT8UY9yD7eE4XTc4UwJ1jBYlgaQQF38ENsKR3wj/8q8RFZrF9WIZpB2V1ArTVFUva8sAul1NzRzQ==
 
-mime-types@^2.1.12, mime-types@~2.1.19:
-  version "2.1.24"
-  resolved "https://registry.yarnpkg.com/mime-types/-/mime-types-2.1.24.tgz#b6f8d0b3e951efb77dedeca194cff6d16f676f81"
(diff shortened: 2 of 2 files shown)


yarn.lock
```diff
@@ -7563,13 +7571,25 @@ mime-db@1.40.0:
...
+  integrity sha512-+5dsGEEovYbT8UY9yD7eE4XTc4UwJ1jBYlgaQQF38ENsKR3wj/8q8RFZrF9WIZpB2V1ArTVFUva8sAul1NzRzQ==
+
+mime-types@^2.1.12, mime-types@~2.1.19:
   version "2.1.24"
   resolved "https://registry.yarnpkg.com/mime-types/-/mime-types-2.1.24.tgz#b6f8d0b3e951efb77dedeca194cff6d16f676f81"
   integrity sha512-WaFHS3MCl5fapm3oLxU4eYDw77IQM2ACcxQ9RIxfaC3ooc6PFuBMGZZsYpvoXS5D5QTWPieo1jjLdAm3TBP3cQ==
   dependencies:
     mime-db "1.40.0"
 
+mime-types@^2.1.18, mime-types@~2.1.24:
```
alexandcote: I think you can dedupe this one.

yarn.lock
```diff
@@ -1486,7 +1495,12 @@
   dependencies:
     "@types/node" "*"
 
-"@types/node@*", "@types/node@>=6":
+"@types/node@*":
+  version "13.9.0"
+  resolved "https://registry.yarnpkg.com/@types/node/-/node-13.9.0.tgz#5b6ee7a77faacddd7de719017d0bc12f52f81589"
+  integrity sha512-0ARSQootUG1RljH2HncpsY2TJBfGQIKOOi7kxzUY6z54ePu/ZD+wJA8zI2Q6v8rol2qpG/rvqsReco8zNMPvhQ==
+
+"@types/node@>=6":
```
alexandcote: Same here, I think you can dedupe.

vsumner: Screen Shot 2020-03-10 at 2 59 10 PM

I'm concerned about bumping it from 12 to 13. Maybe an apollo memory update would be better in another pr?

vsumner:
The fix for the promise type I pulled from https://github.com/DefinitelyTyped/DefinitelyTyped/issues/42084

packages/jest-dom-mocks/src/promise.ts
```diff
@@ -12,7 +12,7 @@ export default class Promise {
     }
 
     jest.useFakeTimers();
-    global.Promise = FakePromise;
+    global.Promise = FakePromise as any;
```
TheMallen: Why do we need to `as any` this now but not before?

alexandcote: I had the same question and I tried to remove it.

image

vsumner: `@types/node` `13` bump. This was introduced in that. See https://github.com/Shopify/quilt/pull/1316#issuecomment-597284951

PR #1318, opened 2020-03-10
✨add package deprecation document, and update current quilt according…

michenly:
* Added doc on how to deprecate package
* Fix project reference test so it will ignore any package without package.json (deprecated)
* Remove any reference to `react-effect-apollo` or `react-shopify-app-route-propagator`

### DIFF
--- a/documentation/guides/package-deprecation.md
+++ b/documentation/guides/package-deprecation.md
@@ -1,0 +1,42 @@
+# Deprecating a package
+
+## 1. Update `packages/{deprecate-package-name}/README.md` and `packages/{deprecate-package-name}/CHANGELOG.md`
+
+A note should be added to both of those files.
+
+Showing prominently
+
+- The last version supported.
+- Alternate to the deprecate package.
+
+## 2. Delete most files in `packages/{deprecate-package-name}`
+
+Leaving only `README.md` and `CHANGELOG.md`
+(It is important to delete `package.json` so lerna will no longer pick it up for future release.)
+
+## 3. Delete TS package project reference
+
+Delete the line referencing the deprecate package from [`packages/tsconfig.json`](../../packages/tsconfig.json)
+
+## 4. Global search deprecate package in quilt
+
+Do a global search in quilt for
+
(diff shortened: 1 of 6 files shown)


documentation/guides/package-deprecation.md
```diff
@@ -0,0 +1,42 @@
...
+## 1. Update `packages/{deprecate-package-name}/README.md` and `packages/{deprecate-package-name}/CHANGELOG.md`
+
+A note should be added to both of those files.
+
+Showing prominently
+
+- The last version supported.
+- Alternate to the deprecate package.
+
+## 2. Delete all files in `packages/{deprecate-package-name}`
```
vsumner: Maybe drop the `all`? Since it's not all?

documentation/guides/package-deprecation.md
```diff
@@ -0,0 +1,42 @@
...
+Do a global search in quilt for
+
+- Any package with `{deprecate-package-name}` listed as dependency
+- Any documentation that reference `{deprecate-package-name}`
+
+and update those accordingly.
+
+## 5.Update quilt README
+
+run `yarn plop docs` to remove `{deprecate-package-name}` from the main README
```
vsumner: ❤️

documentation/guides/package-deprecation.md
michenly: delete most? lol

packages/react-effect/README.md
```diff
@@ -84,15 +84,19 @@ You may optionally pass an options object that contains the following keys (all
   ```tsx
   import {renderToString} from 'react-dom/server';
   import {extract} from '@shopify/react-effect/server';
-  import {createApolloBridge} from '@shopify/react-effect-apollo';
+  import {HtmlContext, HtmlManager} from '@shopify/react-html/server';
```
vsumner: Nice catch!

michenly: doesn't really matter since the example is to show case `extract`, but nice to not mentioned anything that is depreacted.

PR #1321, opened 2020-03-11
✖️remove react-apollo and replace the usage with individual package. …

michenly:
…We are already using this under the hook so functionality wise there is no change.

I would like to get this into master soon so it will be easier to do 🎩 when I actually bring in functionality from `@apollo/react-hooks` to be use in `react-graphql`

## Description

Fixes (issue #)

## Type of change

## Checklist

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,7 +2,7 @@
 # yarn lockfile v1
 
 
-"@apollo/react-common@^3.1.3":
+"@apollo/react-common@>=3.0.0 <4.0.0", "@apollo/react-common@^3.1.3":
   version "3.1.3"
   resolved "https://registry.yarnpkg.com/@apollo/react-common/-/react-common-3.1.3.tgz#ddc34f6403f55d47c0da147fd4756dfd7c73dac5"
   integrity sha512-Q7ZjDOeqjJf/AOGxUMdGxKF+JVClRXrYBGVq+SuVFqANRpd68MxtVV2OjCWavsFAN0eqYnRqRUrl7vtUCiJqeg==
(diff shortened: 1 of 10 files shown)


yarn.lock
```diff
@@ -42,15 +42,6 @@
     ts-invariant "^0.4.4"
     tslib "^1.10.0"
 
-"@apollo/react-ssr@^3.1.3":
```
michenly: 🔥 🎉

yarn.lock
```diff
@@ -8910,17 +8901,6 @@ rc@^1.0.1, rc@^1.1.6, rc@^1.2.7:
...
-"react-apollo@>=3.0.0 <4.0.0", react-apollo@^3.1.2:
-  version "3.1.3"
-  resolved "https://registry.yarnpkg.com/react-apollo/-/react-apollo-3.1.3.tgz#5d8540b401bba36173b63e6c5e75fa561960c63e"
-  integrity sha512-orCZNoAkgveaK5b75y7fw1MSqSHOU/Wuu9rRFOGmRQBSQVZjvV4DI+hj604rHmuN9+WDABxb5W48wTa0F/xNZQ==
-  dependencies:
-    "@apollo/react-common" "^3.1.3"
-    "@apollo/react-components" "^3.1.3"
-    "@apollo/react-hoc" "^3.1.3"
-    "@apollo/react-hooks" "^3.1.3"
-    "@apollo/react-ssr" "^3.1.3"
```
michenly: this is the reason why this is actually not a big change at all.

michenly:
@vsumner I do debate over minor or path as well.
officially, the change of dependency does mean a minor, but because the package & its version its using under the hook did not change, I opt for a minor for now.

michenly:
I ended up doing minor dump for `react-graphql` and patch for `jest-mock-apollo`, thanks for the suggestion @vsumner !

Issue #1181, opened 2019-11-15
Improving TypeScript project reference automation 


ismail-syed:
In https://github.com/Shopify/quilt/pull/962 we introduced TypeScript project references in quilt. Today, we have to manually maintain the path references to sibling packages in the `/packages`. It would be best to automate this. - [contex](https://github.com/Shopify/quilt/pull/962#discussion_r324703977)

PR #1326, opened 2020-03-12
Fix every tsconfig references

alexandcote:
Fix: #1181 

## Description

This PR adds 2 tests

1. Ensure every internal dependency (dependencies from this package) used in a package are added as references in the `tsconfig.json` file.
2. Ensure every dependency list in the `package.json` of a package is used inside this package. The fail will fail if you have dependency you don't need.

I also fix every failing tests. I did each step in 2 separate commits.

### DIFF
--- a/packages/address/src/tests/AddressFormatter.test.ts
+++ b/packages/address/src/tests/AddressFormatter.test.ts
@@ -1,7 +1,7 @@
 import {fetch} from '@shopify/jest-dom-mocks';
 import {Address} from '@shopify/address-consts';
+import {mockCountryRequests} from '@shopify/address-mocks';
 
-import {mockCountryRequests} from '../../../address-mocks/src';
 import AddressFormatter from '../AddressFormatter';
 
 const address: Address = {
--- a/packages/address/src/tests/utilities.test.ts
+++ b/packages/address/src/tests/utilities.test.ts
@@ -1,4 +1,5 @@
-import {fixtures} from '../../../address-mocks/src/fixtures';
+import {fixtures} from '@shopify/address-mocks';
+
 import {renderLineTemplate} from '../utilities';
 
 const Canada = fixtures.country.EN.data.country;
--- a/packages/koa-shopify-webhooks/tsconfig.json
+++ b/packages/koa-shopify-webhooks/tsconfig.json
@@ -11,5 +11,5 @@
     "./src/**/*.tsx"
   ],
   "exclude": ["**/*.test.ts", "**/*.test.tsx"],
-  "references": [{"path": "../jest-koa-mocks"}, {"path": "../jest-dom-mocks"}]
+  "references": [{"path": "../network"}]
 }
--- a/packages/react-server-webpack-plugin/tsconfig.json
+++ b/packages/react-server-webpack-plugin/tsconfig.json
@@ -12,5 +12,5 @@
     "./src/**/*.tsx"
   ],
   "exclude": ["**/*.test.ts", "**/*.test.tsx"],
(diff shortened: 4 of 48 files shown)


packages/address/src/tests/AddressFormatter.test.ts
```diff
@@ -1,7 +1,7 @@
 import {fetch} from '@shopify/jest-dom-mocks';
 import {Address} from '@shopify/address-consts';
+import {mockCountryRequests} from '@shopify/address-mocks';
 
-import {mockCountryRequests} from '../../../address-mocks/src';
```
alexandcote: Use the package instead.

packages/address/src/tests/utilities.test.ts
```diff
@@ -1,4 +1,5 @@
-import {fixtures} from '../../../address-mocks/src/fixtures';
```
alexandcote: Use the package instead.

alexandcote:
@michenly / @ismail-syed I would like your thought on this PR before I complete the description. Do you see any value having something like this in our codebase?

Thanks 😃

packages/react-server-webpack-plugin/tsconfig.json
```diff
@@ -12,5 +12,5 @@
     "./src/**/*.tsx"
   ],
   "exclude": ["**/*.test.ts", "**/*.test.tsx"],
-  "references": [{"path": "../react-html"}, {"path": "../react-server"}]
+  "references": [{"path": "../react-server"}]
```
ismail-syed: Why is the `react-html` reference being removed? Did the script think it's not referenced?
It's definitely referenced in `packages/react-server-webpack-plugin/src/templates/client.ts`

packages/koa-shopify-webhooks/tsconfig.json
```diff
@@ -11,5 +11,5 @@
     "./src/**/*.tsx"
   ],
   "exclude": ["**/*.test.ts", "**/*.test.tsx"],
-  "references": [{"path": "../jest-koa-mocks"}, {"path": "../jest-dom-mocks"}]
+  "references": [{"path": "../network"}]
```
ismail-syed: * `jest-koa-mocks` is used in `packages/koa-shopify-webhooks/src/test/receive.test.ts`
* `jest-dom-mocks` is used in` packages/koa-shopify-webhooks/src/test/register.test.ts`

alexandcote: Probably the second test is not correct. Will check :)

packages/react-server-webpack-plugin/tsconfig.json
alexandcote: Because it's not part of the dependencies of the package. Probably a missing dependency?

alexandcote: The `templates` folder is excluded from `eslint` so again this is the reason why the package is not mandatory in the `package.json`.

[Reference](https://github.com/Shopify/quilt/blob/master/.eslintignore#L6)

ismail-syed: I don't believe we use these templates anywhere anyways. This fine to leave as you have it now.

packages/koa-shopify-webhooks/tsconfig.json
ismail-syed: Ok, I have pulled this locally and the references all seem to work so I'm OK with this.  These only just test files which we ignore during builds so that's fine.

alexandcote:
This PR is not ready for 👀

alexandcote:
@ismail-syed Let me know if you are good with this PR 😄

PR #1329, opened 2020-03-17
Add `hasCustomNameFormatter` method

kartiki975:
## Description
While working on [localizing names](https://github.com/Shopify/pos-channel/pull/1955)  in `pos-channel`, one of the requirements included changing the placement of the name fields in a layout. In particular, the `lastName` field must be placed before `firstName` field when a custom name formatter (with the format `lastName`+`firstName`) exists.

## Type of change

## Checklist

### DIFF
--- a/packages/react-i18n/src/i18n.ts
+++ b/packages/react-i18n/src/i18n.ts
@@ -27,7 +27,7 @@ import {
   Weekday,
   currencyDecimalPlaces,
   DEFAULT_DECIMAL_PLACES,
-  CUSTOM_NAME_FORMATTERS,
+  EASTERN_NAME_ORDER_FORMATTERS,
 } from './constants';
 import {
   MissingCurrencyCodeError,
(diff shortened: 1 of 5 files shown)


packages/react-i18n/src/i18n.ts
```diff
@@ -357,6 +357,13 @@ export class I18n {
     return firstName;
   }
 
+  hasCustomNameFormatter() {
```
ismail-syed: Can you add some documentation in the [readme](https://github.com/Shopify/quilt/tree/master/packages/react-i18n#i18n) and a changelog update for this as well.

kartiki975: @ismail-syed Do you think the name of this method is appropriate for the desired context? All of the current custom name formatters use the Eastern naming order; Wouldn't it be more clear to replace the `CustomNameFormatter` keyword in variables/methods to communicate that instead?

ismail-syed: Sorry, I don't think I follow. 

Are you suggesting to use something like the following over what's there now?

```
hasCustomNameFormatter(locale, language) {}
```

kartiki975: @ismail-syed the latest commit has the changes I implied. Please have a look! If it doesn't seem appropriate, I will revert them.

PR #1331, opened 2020-03-20
✨ make csrf less of with quilt_rails and graphql when needed

### LINKED ISSUE #1297: [graphql-universal-provider] automatically include `x-react-server-rendered`
**Update:** comments below detail an improved proposal

## Overview

`quilt_rails` provides a custom CSRF policy that checks for some headers to indicate that a request came from a server rendered page. Currently these need to be manually added with an `if (server)` check in the `clientOptions` function passed to the graphQL provider. Let's make this nicer by just automatically adding that header when we are on the server.

## Motivation

It's easy to miss the details of how this should work when reading our docs and get blocked on CSRF errors.

TheMallen:
closes #1297 

## Description
This PR adds a new simplified CSRF workaround strategy to `quilt_rails` for use with `@shopify/react-universal-graphql-provider` and other API setups. This strategy simple checks for the presence of an `x-shopify-react-xhr` header and throws if it is not found. It also updates the universal provider to automagically append this header to all requests.

We also update the documentation to first point users to use `null_session` if applicable and otherwise use the custom strategy if required.

## Why
After speaking with some folks who had trouble getting around CSRF errors adding GraphQL to their applications I was dissatisfied with our current APIs and documentation. Further, upon discussion with some users and @JackMc it seems like the weird setup we had with CSRF tokens on the client and a header on the server is unnecessary, and we can simply use a custom header since browsers do not allow them cross origin.

It is also clear that `null_session` is the ideal protection strategy and is generally what APIs should use if they are going to be external facing in any fashion or if they use token based authentication of some kind. As such I opted to update the docs to spell this out as the first option.

### DIFF
--- a/gems/quilt_rails/CHANGELOG.md
+++ b/gems/quilt_rails/CHANGELOG.md
@@ -5,7 +5,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+- Custom CSRF strategy to allow requests containing an `x-shopify-react-xhr: 1` header, deprecating the old custom CSRF strategy which was intended for use only in server rendering in favour of one for use both on the server and client.
 
 ## [1.10.0] - 2019-01-30
 
--- a/gems/quilt_rails/README.md
+++ b/gems/quilt_rails/README.md
@@ -404,18 +404,15 @@ If what it provides is not sufficient, a completely custom server can be defined
 
 #### Fixing rejected CSRF tokens for new user sessions
 
-If a React component calls back to a Rails endpoint (e.g., `/graphql`), Rails may throw a `Can't verify CSRF token authenticity` exception. This stems from the Rails CSRF tokens not persisting until after the first `UiController` call ends.
+When a React component sends HTTP requests back to a Rails endpoint (e.g., `/graphql`), Rails may throw a `Can't verify CSRF token authenticity` exception. This stems from the Rails CSRF tokens not persisting until after the first `UiController` call ends.
 
-To fix this:
+If your API **does not** require session data, the easiest way to deal with this is to use `protect_from_forgery with: :null_session`. This will work for APIs that either have no authentication requirements, or use header based authentication.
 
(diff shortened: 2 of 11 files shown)


TheMallen:
PS. Tophatted using shrink-ray roughly the following steps:
- use `yarn tophat` to get this branch of the universal provider into shrink-ray
- confirm that the `x-shopify-react-xhr` header is on graphql requests
- use the `:path` keyword argument in the gemfile of shrink-ray to get the local version of `quilt_rails` running in the app
- change `:null_session` to `Quilt::HeaderCsrfStrategy` in the graphql controller

gems/quilt_rails/CHANGELOG.md
```diff
@@ -5,7 +5,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+- Custom CSRF strategy to allow requests containing an `x-shopify-react-xhr: 1` header, deprecating the old custom CSRF strategy which was intended for use only in server rendering in favour of one for use both on the server and client
```
ismail-syed: In this line, I would add a link to this PR for reference

packages/react-graphql-universal-provider/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+### Added
+
+- The generated `ApolloClient` now automatically includes a `x-shopify-react-xhr: 1` header.
```
ismail-syed: Same here, link PR.

gems/quilt_rails/README.md
```diff
@@ -404,18 +404,36 @@ If what it provides is not sufficient, a completely custom server can be defined
...
+    result = MySchema.execute(query, operation_name: operation_name, variables: variables, context: context)
+
+    render(json: result)
+  end
+end
+```
+
+If your API **does** require session data, you may can use follow the following steps:
+
+- Add an `x-shopify-react-xhr` header to all GraphQL requests with a value of 1 (this is done automatically if you are using `@shopify/react-universal-graphql-provider`)
```
michenly:
```suggestion
- Add an `x-shopify-react-xhr` header to all GraphQL requests with a value of 1 (this is done automatically if you are using `@shopify/react-graphql-universal-provider`)
```

PR #1332, opened 2020-03-23
Allow assetname from context

TheMallen:
## What does this do?
This PR addresses some frustrations users have had trying to setup multiple sub-applications using `quilt_rails` and / or `react-server`. It allows the middleware output by `createRender` to accept an `assetName` in the form of a function of the form `(ctx: Koa.context) => string`. This allows serving multiple apps  conditionally without having to build your own render middleware from scratch.


michenly:
Can you please add some doc on how to use `assetName` to setup multiple sub-applications?

Issue #1334, opened 2020-03-24
[react-form] `useForm` form submission will ignore errors within `useList` list


sylvhama:
## Overview

On [the doc](https://github.com/Shopify/quilt/tree/master/packages/react-form#parameters-2) it says:
> The onSubmit handler will not be called if client validations fails.

However that's not the case we have errors within a `useList` list. 

See https://codesandbox.io/s/naughty-satoshi-gnwey

1. Change nothing, click on submit
2. Check the console: no log
3. Set a title, click on submit
4. Check the console: it submitted despite the list errors :confused: 

## Consuming repo

https://github.com/Shopify/web

PR #1335, opened 2020-03-24
[react-form] fix list validation on submit

sylvhama:
## Description

Fixes https://github.com/Shopify/quilt/issues/1334

In react-form, `useForm` form submission will ignore errors within `useList` list.

## Type of change

Return the value from `runValidation` in `useList`.

- [x] react-form Patch: Bug/ Documentation fix (non-breaking change which fixes an issue or adds documentation)

## Checklist


packages/react-form/src/hooks/test/form.test.tsx
```diff
@@ -216,6 +216,27 @@ describe('useForm', () => {
       });
     });
 
+    it('validates list fields with their latest values before submitting and bails out if any fail', async () => {
```
sylvhama: I've made a brand new test to make sure that the `submitSpy` does not get called only because of a list error.

packages/react-form/src/hooks/list/list.ts
```diff
@@ -178,7 +178,7 @@ export function useList<Item extends object>(
 
             const siblings = state.list.filter(listItem => listItem !== item);
 
-            runValidation(
+            return runValidation(
```
sylvhama: Since there was no return then https://github.com/Shopify/quilt/blob/master/packages/react-form/src/utilities.ts#L76 would never detect errors.

TheMallen: Good catch!

PR #1338, opened 2020-03-26
React hooks - add useMedia

sylvhama:
## Description

When using a design system such as [Polaris](https://polaris.shopify.com/), you might want to change props values based on a media query instead of doing custom css that could break if the design system changes its implementation. E.g. Imagine I might want to hide a [TextField](https://polaris.shopify.com/components/forms/text-field) label via its `labelHidden` prop. This is not meant to replace CSS media queries, it's really meant for specific scenarios.

I've adapted the code from https://usehooks.com/useMedia/

## Checklist

### DIFF
--- a/packages/react-hooks/src/hooks/media.ts
+++ b/packages/react-hooks/src/hooks/media.ts
@@ -1,0 +1,36 @@
+import {useState, useEffect} from 'react';
+
+export function useMedia<ValueType>(
+  queries: string[],
+  values: ValueType[],
+  defaultValue: ValueType,
+) {
+  const mediaQueryLists = queries.map(query => {
+    return window.matchMedia(query);
+  });
+
+  const getValue = () => {
+    const index = mediaQueryLists.findIndex(mediaQueryList => {
+      return mediaQueryList.matches;
+    });
+    return typeof values[index] === 'undefined' ? defaultValue : values[index];
+  };
+
+  const [value, setValue] = useState(getValue);
+
+  useEffect(() => {
+    const handler = () => {
+      setValue(getValue);
+    };
+    mediaQueryLists.forEach(mediaQueryList => {
+      mediaQueryList.addListener(handler);
+    });
+    return () => {
+      mediaQueryLists.forEach(mediaQueryList => {
+        mediaQueryList.removeListener(handler);
+      });
+    };
+  }, []);
+
+  return value;
+}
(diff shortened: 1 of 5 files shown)


packages/react-hooks/src/hooks/media.ts
```diff
@@ -0,0 +1,35 @@
...
+      setValue(getValue);
+    };
+    mediaQueryLists.forEach(mediaQueryList => {
+      mediaQueryList.addListener(handler);
+    });
+    return () =>
+      mediaQueryLists.forEach(mediaQueryList => {
+        mediaQueryList.removeListener(handler);
+      });
+  }, []);
```
sylvhama: Why does the linter does not like having an empty dependency array? What I want to simulate `componentDidMount` and `componentWillUnmount` .

packages/react-hooks/src/hooks/media.ts
```diff
@@ -0,0 +1,36 @@
+import {useState, useEffect} from 'react';
+
+export function useMedia<ValueType>(
+  queries: string[],
+  values: ValueType[],
+  defaultValue: ValueType,
+) {
+  const mediaQueryLists = queries.map(query => {
+    return window.matchMedia(query);
```
loic-d: We should return a default value (`false`?) if you are rendering on the server.

```
 if (typeof window === 'undefined') {
    return false;
  }
```

packages/react-hooks/src/hooks/media.ts
loic-d: You have to disabled the rule with:
`// eslint-disable-next-line react-hooks/exhaustive-deps`

packages/react-hooks/src/hooks/media.ts
qq99: `false.addEventListener(...)` below will 💥

loic-d: Sorry I missed that. You can return a noop object matching the signature of what you would expect on the client:

```
const noopMediaQueryList = {
  media: '',
  onchange: noop,
  addListener: noop,
  addEventListener: noop,
  removeListener: noop,
  removeEventListener: noop,
  dispatchEvent: () => false,
  matches: false,
};
```

sylvhama: Thanks I totally forgot about SSR!

sylvhama:
@qq99 for sure yours is muuuch simpler to use / understand. You're right that passing arrays is very confusing. Plus we can achieve the same goals with yours. Should we re-open your PR in quilt?

qq99:
That sounds good to me @sylvhama

Issue #411, opened 2018-11-16
Add common MIME types to @shopify/network


lemonmade:
We use the JSON, JS, and HTML ones fairly frequently.

PR #1344, opened 2020-03-31
:sparkles: Adding common MimeType base to @shopify/network

posth:
## Description

Fixes https://github.com/Shopify/quilt/issues/411

This PR adds an enum of common mime types to `@shopify/network`.  These choices were based off of the most frequently used mime types in `Shopify/web` plus the suggested ones from the issue.

It is not an exhaustive list and more values can be added.

## Type of change

- [ ] `@shopify/network` Minor: New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/packages/network/CHANGELOG.md
+++ b/packages/network/CHANGELOG.md
@@ -7,6 +7,8 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+- Minor: Added common mime types base ([#1344](https://github.com/Shopify/quilt/pull/1344))
+
 ## [1.4.0] - 2019-06-27
 
 - Added the following headers: 'X-XSS-Protection', 'X-Frame-Options', 'X-Download-Options', 'X-Content-Type-Options', 'Strict-Transport-Security', 'Referrer-Policy' ([#752](https://github.com/Shopify/quilt/pull/752))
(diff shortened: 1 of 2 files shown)


posth:
> I feel we should probably create a new `@shopify/files` package.
> 
> We already have this package internally in web. We can expose it in `quilt`.
> 
> The network package is for `common values related to dealing with the network.`
> 
> Mime types + exposed helpers like `getExtensionFromMimeType` are use extensively at runtime in the app for features like the image editor or rich media (not dealing with network specifically).
> 
> WDYT?

I think that's a good suggestion.  

This PR was initially just solving the issue of adding mime-types, but looking at `@shopify/files` inside `web`, some mime types on these PR were taken from there, and adding the extra helpers to a unique package makes sense.

Can move forward with the package, and link https://github.com/Shopify/quilt/issues/411 to it as well.

loic-d:
@posth Feel free to bring it up to #web-foundation-tech. I didn't see the original issue mentioned `@shopify/network`. I think it would be better in a new `files` package but we can definitely move forward with this change if it makes sense for everyone. Nice addition no matter where it is 😄 💯

posth:
> @posth Feel free to bring it up to #web-foundation-tech. I didn't see the original issue mentioned `@shopify/network`. I think it would be better in a new `files` package but we can definitely move forward with this change if it makes sense for everyone. Nice addition no matter where it is 😄 💯

Going to close this PR in favour of introducing a new package.  We can re-open this if it makes sense to just add the mime types, but we can continue onto the package PR.

PR #1345, opened 2020-03-31
🧹 clean up CHANGELOG.md files and add checks to keep them that way

### LINKED ISSUE #1341: Changelogs out of sync
I noticed `packages/react-hooks/CHANGELOG.md` still has an entry in the `UNRELEASED` section, despite that change having been published. This means the changelog is out of sync.

Ideally, a new patch version should be published documenting the change.

It is also possible this is not the only package whose `CHANGELOG.md` is out of sync.

### LINKED ISSUE #1342: Changelogs should be linted on publish
Quilt has many packages, and it is easy for someone to miss moving the `UNRELEASED` section of a `CHANGELOG.md` into a versioned section on publish.

**This has unfortunately already occurred, as documented in #1341.**

Ideally, we would solve this on multiple fronts:

- Update `CHANGELOG.md` automatically on publish
- Lint `CHANGELOG.md` on publish, preventing invalid publication
  _This could be done naively by checking for the presence of `## UNRELEASED`, vs `<!-- ## UNRELEASED -->`.

TheMallen:
closes #1342 
closes #1341 

## What this?
This PR is a continuation of @sambostock 's branch where he added some tests to help keep our changelogs in a healthier state. It adds two main things:

- a small test suite which helps make sure our changelogs are reasonably consistent
- a small script that runs before `yarn release` and reminds you to deal with `## Unreleased` headers

## Pictures 📸 

image
image


scripts/lint-changelogs.js
```diff
@@ -0,0 +1,66 @@
+/* global __dirname process */
```
TheMallen: This little script shares some stuff with the tests, we may eventually want to make a little changelog validation util package but for now I think it's OK

scripts/lint-changelogs.js
```diff
@@ -0,0 +1,66 @@
...
+
+const {readFileSync} = require('fs-extra');
+const glob = require('glob');
+
+const ROOT_PATH = resolve(__dirname, '..');
+
+const changelogs = readChangelogs();
+console.log(`🕵️‍♂️  Checking ${changelogs.length} changelogs`);
+
+const haveUnreleasedHeader = changelogs.filter(({packageChangelog}) =>
```
TheMallen: this check is pretty naive but it works pretty well to insure folks at least go fix the ## Unreleased header on packages they are trying to release

test/consistent-changelogs.test.ts
```diff
@@ -0,0 +1,138 @@
...
+  '### Added',
+  '### Changed',
+  '### Deprecated',
+  '### Removed',
+  '### Fixed',
+  '### Security',
+  // This is technically not part of Keep a Changelog spec
+  '### Chore',
+  // This isn't part of it either
+  '### Breaking Change',
```
michenly: Then let's just not allow those?

michenly:
> a small script that runs before yarn release and reminds you to deal with ## Unreleased headers

❤️ it

test/consistent-changelogs.test.ts
sambostock: Back when I wrote this, I didn't want to make the call on rewriting old changelog entries 😅

vsumner:
Screen Shot 2020-03-31 at 3 43 11 PM

^ bug we discussed in the channel based on node version.

PR #1350, opened 2020-04-02
adding 'form: auto' to formatCurrency

patsissons:
## Description

Adding a new `form: 'auto'` option to the `formatCurrency` utility function. `auto` will heuristically select `short` or `explicit` depending on the provided `currency` option and the pre-assigned `defaultCurrency` for the `I18n` instance. The table below describes how the heuristic operates (using `USD` and `CAD` as example currencies).

|`defaultCurrency`|`currency`|result|notes|
|-|-|-|-|
|`USD`|`USD`|`short`|exact match currencies use the `short` form|
|`USD`|`CAD`|`explicit`|non-matching currencies use the `explicit` form|
|`undefined`|`USD`|`short`|ambiguous matches use the `short` form|
|`USD`|`undefined`|`short`|ambiguous matches use the `short` form|

## Type of change

## Checklist


packages/react-i18n/src/i18n.ts
```diff
@@ -49,7 +49,7 @@ export interface NumberFormatOptions extends Intl.NumberFormatOptions {
 }
 
 export interface CurrencyFormatOptions extends NumberFormatOptions {
-  form?: 'short' | 'explicit';
+  form?: 'auto' | 'short' | 'explicit';
```
patsissons, line 52: since this is effectively automatically selecting based on locale, should we use `locale` instead of `auto`?

packages/react-i18n/CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 ### Added
 
 - Added korean eastern name formatter
+- Added `form: 'auto'` option to `formatCurrency`, to automatically select `short` or `explicit` based on `currency` and `defaultCurrency` ([#1350](https://github.com/Shopify/quilt/pull/1350))
```
alexandcote, line 15: This line should be above the last version.
You need to uncomment `<!-- ## [Unreleased] -->`

packages/react-i18n/README.md
```diff
@@ -126,6 +126,7 @@ The provided `i18n` object exposes many useful methods for internationalizing yo
 - `formatCurrency()`: formats a number as a currency according to the locale. Its behaviour depends on the `form:` option.
   - if `form: 'short'` is given, then a possibly-ambiguous short form is used, consisting of the bare symbol if the currency has a symbol, or the ISO 4217 code if there is no symbol for that currency. Examples: `CHF 1,25`, `€ 1,25 EUR`, `OMR 1.250`, `$ 1.25 USD`
   - if `form: 'explicit'` is given, then the result will be the same as for `short`, but will append the ISO 4217 code if it is not already present
+  - if `form: 'auto'` is given, then `explicit` will be selected the `currency` option does not match the `defaultCurrency`, otherwise `short` is selected
```
alexandcote, line 129:
```suggestion
  - if `form: 'auto'` is given, then `explicit` will be selected if the `currency` option does not match the `defaultCurrency`, otherwise `short` is selected
```

alexandcote: Do you also want to specify when `currency` or `defaultCurrency` is `undefined`?

packages/react-i18n/src/i18n.ts
alexandcote: I prefer `auto`.

packages/react-i18n/src/i18n.ts
```diff
@@ -364,6 +368,23 @@ export class I18n {
...
 
+  private formatCurrencyAuto(
+    amount: number,
+    options: Intl.NumberFormatOptions = {},
+  ): string {
+    // use the short format if we can't determine a currency match, or if the
+    // currencies match, use explicit when the currencies definitively do not
+    // match.
+    const formatShort =
+      !options.currency ||
```
alexandcote, line 379: I usually prefer to be explicit here and check if `options.currency == null`. 

If `currency` is an empty string, the condition will be true and I'm not sure if it should 🤔

patsissons: yea good point

packages/react-i18n/README.md
patsissons: yea maybe i'll add a secondary statement about undefined values.

packages/react-i18n/CHANGELOG.md
patsissons: oops, nice catch

atesgoral:
I'm confused as to why this PR is still open if it got merged already. @patsissons Can this be closed now?

gmcgibbon:
I closed the zombie PR, and now it says merged. 🧟

PR #1330, opened 2020-03-17
[react-hooks] useOnValueChange odd interaction with setState

qw-in:
## Description

Currently, any call to `setState` within the `useOnValueChange` callback will only be available in one render

[Codesandbox demo](https://codesandbox.io/s/shopify-useonvaluechange-behaviour-bgyeg)

Fixes (issue #)

## Type of change

- [X]  Patch: Bug/ Documentation fix (non-breaking change which fixes an issue or adds documentation)

## Checklist

- [x] I have added a changelog entry, prefixed by the type of change noted above

PR #1351, opened 2020-04-03
[react-hook] Improve useOnValueChange hook to use useEffect

alexandcote:
## Description

The implementation of `useOnValueChange` is incorrect. The hook should use `useEffect` to be called after the first render and after every update and should not block the render method of the component. The check should also be only executed if either `value` or `onChange` has changed.

## Type of change

This will be a breaking change because the timing of when the function is executed will change. Instead of being executed during the rendering, it will be executed after.

- [ ] `@shopify/react-hook` Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist


qw-in:
I concur #1330

PR #1352, opened 2020-04-03
[Bugfix] Fix getQueuingTime header extraction

kelhusseiny:
## Description

Fixes the extraction of `X-Rrequest-Start` header as the header comes from Nginx in a format  `X-Rrequest-Start t=1585939545.844`.
This will fix the emission to Datadog for Request Queuing Time.

## Type of change

## Checklist

### DIFF
--- a/packages/koa-metrics/src/timing.ts
+++ b/packages/koa-metrics/src/timing.ts
@@ -1,10 +1,10 @@
 import {Context} from 'koa';
 
 export function getQueuingTime(ctx: Context): number | null {
-  const requestQueuingTime = ctx.request.get('X-Request-Start');
-  if (requestQueuingTime) {
+  const requestStartHeader = ctx.request.get('X-Request-Start');
+  if (requestStartHeader) {
     try {
-      return parseInt(requestQueuingTime, 10);
+      return parseInt(requestStartHeader.replace('t=', ''), 10);
     } catch (err) {
       // this is a non-critical error, so we can continue execution.
     }
--- a/packages/koa-metrics/package.json
+++ b/packages/koa-metrics/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@shopify/koa-metrics",
-  "version": "0.3.12",
+  "version": "0.3.13",
   "license": "MIT",
   "description": "Aims to provide standard middlewares and instrumentation tooling for metrics in Koa.",
   "main": "dist/src/index.js",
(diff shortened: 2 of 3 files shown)


vpetkar:
Is the package automatically released after merging?

kelhusseiny:
@vpetkar I had this question for @jamesism. waiting for his answer

michenly:
https://github.com/Shopify/quilt/blob/master/.github/CONTRIBUTING.md#releasing

michenly:
I am not familiar with Nginx, can you provided some information on where you are seeing this mis-match?

packages/koa-metrics/src/timing.ts
```diff
@@ -1,10 +1,10 @@
 import {Context} from 'koa';
 
 export function getQueuingTime(ctx: Context): number | null {
-  const requestQueuingTime = ctx.request.get('X-Request-Start');
-  if (requestQueuingTime) {
+  const requestStartHeader = ctx.request.get('X-Request-Start');
+  if (requestStartHeader) {
     try {
-      return parseInt(requestQueuingTime, 10);
+      return parseInt(requestStartHeader.replace('t=', ''), 10);
```
michenly, line 7: since replace will take care of the case when t= is not there, this should be safe. But I would make it a minor change instead of patch.

kelhusseiny:
@michenly Here is the [reference](https://github.com/Shopify/cluster-service-ingress-nginx/blob/1b9c1e3600cf0c1059415bbf377b459ebf0c6c94/config/deploy/production/partials/_ingress-nginxcontrib.yml.erb#L224) of how the header is being set

packages/koa-metrics/src/timing.ts
kelhusseiny: According to README you sent, I have to page ATC from web foundation to make the release, right>?

kelhusseiny: You mean no need to release a new version?

michenly: You do need a new version, in the description of the issue thou, you checked off a patch version which is understandable that this is a bug to you. To this library and its consumer thou, the change you made is actually a new feature which would be a minor version bump. Do you agree?

kelhusseiny: gotcha! Agree 👍

PR #1353, opened 2020-04-04
 [react-form] Fixed getValue on empty array

alanthai:
- reduceField takes 4th argument to deal with non-field primitive/empty values
 - useDirty no longer uses useMemo

## Description

Fixes https://github.com/Shopify/web/pull/25308

`reduceFields` was previously unable to parse empty arrays and objects, making `getValue` not return the entire fieldValue since they do not loop over a field (empty arrays are not fields). A new argument `reduceEmptyFn` allows `getValue` to handle empty arrays.

## Type of change

## Checklist

### DIFF
--- a/packages/react-form/src/hooks/dirty.ts
+++ b/packages/react-form/src/hooks/dirty.ts
@@ -1,14 +1,6 @@
-import {useMemo} from 'react';
-
 import {FieldBag} from '../types';
-import {fieldsToArray} from '../utilities';
+import {reduceFields} from '../utilities';
 
 export function useDirty(fieldBag: FieldBag) {
-  const fields = fieldsToArray(fieldBag);
-
-  return useMemo(
-    () => fields.some(field => field.dirty),
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    fields,
-  );
+  return reduceFields(fieldBag, (dirty, field) => dirty || field.dirty, false);
 }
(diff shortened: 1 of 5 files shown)


alanthai:
@sylvhama I've tested this on web and the unit tests are now passing. There were some date related tests failing that are unrelated to my changes (timezome-related)

packages/react-form/src/hooks/dirty.ts
```diff
@@ -1,14 +1,6 @@
...
 
 export function useDirty(fieldBag: FieldBag) {
-  const fields = fieldsToArray(fieldBag);
-
-  return useMemo(
-    () => fields.some(field => field.dirty),
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-    fields,
-  );
+  return reduceFields(fieldBag, (dirty, field) => dirty || field.dirty, false);
```
alanthai, line 5: The problem with having `fields` as I had it is that `useList` fields are dynamic which causes improper memoization of fields (in the second argument of) `useMemo`.

We could alternately go back to using the more static `Object.values(fields)` where `fields` are those declared in `useForm` but it doesn't properly memoize nested fields or lists, making `useMemo` useless.

I would argue that removing hooks altogether would have little to no performance difference as shown here. In fact, removing the need for `useMemo` plus not needing to create a new array to watch is arguably more performant. In the long term, I would recommend deprecating `useDirty` and instead replace it with a utility function.

sylvhama:
@ismail-syed I also used `yarn tophat` on Shopify web project.

PR #1354, opened 2020-04-06
Adds useDebouncedValue hook

arthurgouveia:
## Description

Adds a new hook to use debounced values. [Here](https://github.com/Shopify/email/pull/3875/files#diff-64aafeed2cc2dcea9fe117fe5c7f6ee7R25)'s a use case.

## Type of change

## Checklist

### DIFF
--- a/packages/react-hooks/CHANGELOG.md
+++ b/packages/react-hooks/CHANGELOG.md
@@ -9,6 +9,10 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [1.6.1] - 2020-04-07
 
+### Added
+
+- Added `useDebouncedValue` hook ([#1354](https://github.com/Shopify/quilt/pull/1354))
+
 ### Fixed
 
 - `useOnChangeValue` is now executed in an `useEffect` and doesn't block the render method anymore. **This fix may cause timing issue in your project if you depended on the change handler happening synchronously.**
--- a/packages/react-hooks/CHANGELOG.md
+++ b/packages/react-hooks/CHANGELOG.md
@@ -5,13 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## [Unreleased] -->
-
-## [1.6.1] - 2020-04-07
+## [Unreleased]
 
 ### Added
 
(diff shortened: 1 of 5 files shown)


packages/react-hooks/src/hooks/debounced.ts
```diff
@@ -0,0 +1,29 @@
+import {useEffect, useRef, useState} from 'react';
+
+export const DELAY_COMMIT_SETTING_CHANGE = 500;
+
+export function useDebouncedValue<T>(
+  value: T,
+  delay = DELAY_COMMIT_SETTING_CHANGE,
+) {
+  const [state, setState] = useState<T>(value);
+  const updateRef = useRef<number>();
```
tleunen, line 10: Do we really need to keep the timeout in the ref?
If the `value` changes, the effect returned function will be called, therefore clear the timeout?

packages/react-hooks/src/hooks/debounced.ts
```diff
@@ -0,0 +1,29 @@
+import {useEffect, useRef, useState} from 'react';
+
+export const DELAY_COMMIT_SETTING_CHANGE = 500;
```
tleunen, line 3: No need to export this default value imo, also I would change its name for `defaultDelay` or something similar

packages/react-hooks/src/hooks/debounced.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export function useDebouncedValue<T>(
+  value: T,
+  delay = DELAY_COMMIT_SETTING_CHANGE,
```
tleunen, line 7: Would really be great to use the same signature as `useDeferredValue` so it's easier to migrate in the future... https://reactjs.org/docs/concurrent-mode-reference.html#usedeferredvalue

packages/react-hooks/CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [1.6.1] - 2020-04-07
 
+### Added
```
tleunen, line 12: This should be moved under the "Unreleased", not inside 1.6.1

packages/react-hooks/src/hooks/debounced.ts
```diff
@@ -0,0 +1,26 @@
+import {useEffect, useState} from 'react';
+
+interface DebouncedDelay {
```
tleunen, line 3: nit: `DebouncedOptions`?

packages/react-hooks/CHANGELOG.md
arthurgouveia: Thanks! It was but due a bad rebase and the new release that happened today, it got mixed up

Issue #1356, opened 2020-04-07
@shopify/react-csrf-universal-provider is broken with @shopify/react-html v9.3.0


jordanAtShopify:
## Overview
Due to the latest change in [@shopify/react-html v9.3.0](https://github.com/Shopify/quilt/blob/master/packages/react-html/CHANGELOG.md#930---2020-03-13) which upgraded serialize-javascript@^3.0.0 and enforced the JSON serialization in our universal providers (see https://github.com/Shopify/quilt/pull/1312), the crsf-universal-provider is broken as it tries to store
the crsfToken as a plain string which is not serializable. @shopify/react-csrf will also need updating, if only in documentation. If I change the provider to use

`<CsrfUniversalProvider value={{csrfToken}}>`
and
```
import {useCsrfToken} from '@shopify/react-csrf';
const {csrfToken} = useCsrfToken();
```

it works correctly but fails type checking.

## Checklist

- [x] Please delete the labels section before submitting your issue
- [x] I have described this issue in a way that is actionable (if possible)

jordanAtShopify:
Turns out this isn't an actual issue. I had some conflicting dependencies and (I think) multiple versions of serialize-javascript. Strange that it only broke when passing strings, but bumping some other dependencies fixed this for me. Sorry!

PR #1357, opened 2020-04-07
Fix type for react-csrf CsrfTokenContext to work properly with JSON encoded serialization

jordanAtShopify:
## Description

Fixes https://github.com/Shopify/quilt/issues/1356

This changes the type for `CsrfTokenContext` in react-csrf so that it can be properly JSON serialized and used with newest serialize-javascript@^3.0.0 which enforces JSON.parse for our universal providers. 

## Type of change

- [x] react-csrf Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] react-csrf-universal-provider Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

### DIFF
--- a/packages/react-csrf/src/hooks.ts
+++ b/packages/react-csrf/src/hooks.ts
@@ -5,9 +5,9 @@ import {CsrfTokenContext} from './context';
 export function useCsrfToken() {
   const csrf = useContext(CsrfTokenContext);
 
-  if (csrf == null) {
+  if (!csrf || csrf.csrfToken === undefined) {
     throw new Error('No CSRF token found in context.');
   }
 
-  return csrf;
+  return csrf.csrfToken;
 }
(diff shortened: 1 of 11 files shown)


packages/react-csrf/src/hooks.ts
```diff
@@ -5,9 +5,9 @@ import {CsrfTokenContext} from './context';
 export function useCsrfToken() {
   const csrf = useContext(CsrfTokenContext);
 
-  if (csrf == null) {
```
jordanAtShopify, line 8: This looks like it should have been updated when the type changed from `string | null` to `string | undefined` but was missed because of the fuzzy equality.

alexandcote:
Ohhhh nice found! This is totally my fault. I propose to solve the issue at the root level so directly inside the `react-html` instead. You can checkout `react-html/fix_serializer` to check the approach.
`

jordanAtShopify:
Closing in favor of https://github.com/Shopify/quilt/pull/1358

PR #1358, opened 2020-04-07
Ensure the serialized data is an object

jordanAtShopify:
## Description
Fixes https://github.com/Shopify/quilt/issues/1356
This makes sure setting value of the Serialized component to a plain string or number won't break the JSON.parse enforcing.

## Type of change

- [x] react-html Patch: Bug/ Documentation fix (non-breaking change which fixes an issue or adds documentation)

## Checklist


jordanAtShopify:
🤦‍♂ oops, this was my fault. Some dependency issues were causing my bug and not actually the JSON serializing. Super weird that the issues only showed up when passing a plain string... oh well. Sorry!

PR #1360, opened 2020-04-08
Removes "window" reference from "useDebouncedValue" hook

arthurgouveia:
## Description

Removes `window` dependency from the hook introduced on https://github.com/Shopify/quilt/pull/1354. Uses a ref to store the value and update it only if different from previous one.

## Type of change

(setting this as a Minor because it will be bundled with the next release so even though this might compromise SSR, nobody's using it yet, hence not being a breaking change)

## Checklist

### DIFF
--- a/packages/react-hooks/src/hooks/debounced.ts
+++ b/packages/react-hooks/src/hooks/debounced.ts
@@ -1,4 +1,4 @@
-import {useEffect, useState} from 'react';
+import {useEffect, useState, useRef} from 'react';
 
 interface DebouncedOptions {
   timeoutMs: number;
@@ -11,14 +11,20 @@ export function useDebouncedValue<T>(
   {timeoutMs}: DebouncedOptions = {timeoutMs: DEFAULT_DELAY},
 ) {
   const [state, setState] = useState<T>(value);
+  const stateRef = useRef<T>(state);
 
   useEffect(() => {
-    const timeout = window.setTimeout(() => {
+    if (stateRef.current === value) {
+      return;
+    }
+
+    const timeout = setTimeout(() => {
+      stateRef.current = value;
       setState(value);
     }, timeoutMs);
 
     return () => {
-      window.clearTimeout(timeout);
+      clearTimeout(timeout);
     };
   }, [value, timeoutMs]);
 
--- a/packages/react-hooks/README.md
+++ b/packages/react-hooks/README.md
@@ -27,8 +27,8 @@ function MyComponent() {
     },
   });
 
-  function handleSearchTextChange(value: string) {
-    setSearchValue(value);
+  function handleSearchTextChange(evt: React.KeyboardEvent) {
+    setSearchValue(evt.currentTarget.value);
   }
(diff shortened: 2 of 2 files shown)


packages/react-hooks/CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 - Added `useDebouncedValue` hook ([#1354](https://github.com/Shopify/quilt/pull/1354))
 
+### Fixed
+
+- Removed `window` for `[set|clear]Timeout` from `useDebouncedValue` hook ([#1360](https://github.com/Shopify/quilt/pull/1360))
```
ismail-syed, line 16: Where'd the `Added useDebouncedValue hook` changelog item go? Does this branch need a rebase?

tleunen, line 16: What about?
```suggestion
- Fixed SSR support by removing the reference to `window` in `[set|clear]Timeout` from `useDebouncedValue` hook ([#1360](https://github.com/Shopify/quilt/pull/1360))
```

packages/react-hooks/src/hooks/debounced.ts
```diff
@@ -13,12 +13,12 @@ export function useDebouncedValue<T>(
   const [state, setState] = useState<T>(value);
 
   useEffect(() => {
-    const timeout = window.setTimeout(() => {
+    const timeout = setTimeout(() => {
```
alexandcote, line 16: I also think we should only call this if `state` is different then `value. Otherwise, the will run even on the first render.

packages/react-hooks/src/hooks/debounced.ts
```diff
@@ -13,14 +13,16 @@ export function useDebouncedValue<T>(
...
+        setState(value);
+      }
     }, timeoutMs);
 
     return () => {
-      window.clearTimeout(timeout);
+      clearTimeout(timeout);
     };
-  }, [value, timeoutMs]);
+  }, [value, state, timeoutMs]);
```
alexandcote, line 25: You don't want to run this effect if the state change. We can probably use a ref instead.

packages/react-hooks/CHANGELOG.md
arthurgouveia: It is right above https://github.com/Shopify/quilt/pull/1360/files#diff-6c41914fef434be5eed762a9c9d10c5fR12.

Should I merge these two into the single `Added`? @tleunen ?

packages/react-hooks/src/hooks/debounced.ts
arthurgouveia: @tleunen guess we'll revert to the ref

packages/react-hooks/CHANGELOG.md
tleunen: Since it's the same version, I'd remove the log about fixing it

PR #1364, opened 2020-04-09
React hooks - add useMedia

sylvhama:
## Description

This is a new attempt to add a `useMedia` hook after https://github.com/Shopify/quilt/pull/1338. This one is inspired from https://github.com/Shopify/web/pull/14306, the code was first made by @qq99 <3

When using a design system such as Polaris, you might want to change props values based on a media query instead of doing custom css that could break if the design system changes its implementation. E.g. Imagine I might want to hide a TextField label via its labelHidden prop. This is not meant to replace CSS media queries, it's really meant for specific scenarios.

## Type of change

## Checklist


sylvhama:
If you want to tophat it, you can use [`yarn tophat react-hooks ../web`](https://github.com/Shopify/quilt/blob/4a700447969b911fbda901b194987419ba99ccf2/.github/CONTRIBUTING.md#testing-your-changes-in-a-local-project) and test https://github.com/Shopify/web/pull/25763

PR #1366, opened 2020-04-09
[react-i18n] fixing memoizedNumberFormatter signature

patsissons:
## Description

It is reasonable to want to create an `Intl.NumberFormat` instance with an `undefined` locale. This PR is patching `memoizedNumberFormatter` to support this use case. We can also support the case of a `string[]` of locales.

related to #1310

## Type of change

## Checklist

### DIFF
--- a/packages/react-i18n/src/test/utilities.test.tsx
+++ b/packages/react-i18n/src/test/utilities.test.tsx
@@ -125,6 +125,16 @@ describe('memoizedNumberFormatter', () => {
     expect(numberFormatterB).toBe(numberFormatterA);
   });
 
+  it('returns the same object when passed multiple locales as an array', () => {
+    const locales = ['en-US', 'en-CA'];
+    const numberFormatterA = memoizedNumberFormatter(locales);
+    expect(numberFormatterA).toBeInstanceOf(Intl.NumberFormat);
+
+    const numberFormatterB = memoizedNumberFormatter(locales);
+    expect(numberFormatterB).toBeInstanceOf(Intl.NumberFormat);
+    expect(numberFormatterB).toBe(numberFormatterA);
+  });
+
   it('returns different object when passed different arguments', () => {
     const numberFormatterA = memoizedNumberFormatter(locale);
     expect(numberFormatterA).toBeInstanceOf(Intl.NumberFormat);
--- a/packages/react-i18n/src/utilities/tests/money.test.ts
+++ b/packages/react-i18n/src/utilities/tests/money.test.ts
@@ -38,10 +38,6 @@ describe('formatCurrency()', () => {
 
   it('calls format on the memoized formatter', () => {
     const amount = 123;
-    const locale = 'en';
-    const options = {
-      currency: 'USD',
-    };
     const result = 'result';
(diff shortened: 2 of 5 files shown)


packages/react-i18n/src/utilities/tests/money.test.ts
```diff
@@ -38,10 +38,6 @@ describe('formatCurrency()', () => {
 
   it('calls format on the memoized formatter', () => {
     const amount = 123;
-    const locale = 'en';
-    const options = {
-      currency: 'USD',
-    };
```
patsissons, line 44: removing some unreferenced declarations

packages/react-i18n/src/test/utilities.test.tsx
```diff
@@ -125,6 +125,16 @@ describe('memoizedNumberFormatter', () => {
     expect(numberFormatterB).toBe(numberFormatterA);
   });
 
+  it('returns teh same object when passed multiple locales as an array', () => {
```
alexandcote, line 128:
```suggestion
  it('returns the same object when passed multiple locales as an array', () => {
```

packages/react-i18n/src/utilities/translate.tsx
```diff
@@ -45,10 +45,12 @@ export interface TranslateOptions<Replacements = {}> {
 }
 
 function numberFormatCacheKey(
-  locale: string,
-  options: Intl.NumberFormatOptions = {},
+  locales?: string | string[],
+  options?: Intl.NumberFormatOptions,
 ) {
-  return `${locale}${JSON.stringify(options)}`;
+  return `${
```
alexandcote, line 51: I would prefer we move the ternary outside of the interpolation for readability. Any strong opinion?

packages/react-i18n/src/utilities/translate.tsx
```diff
@@ -45,10 +45,12 @@ export interface TranslateOptions<Replacements = {}> {
...
+    Array.isArray(locales) ? locales.join('-') : locales
+  }-${JSON.stringify(options)}`;
```
alexandcote, line 53: `JSON.stringify(undefined)` returns `undefined` then adding undefined to the interpolation will do something like.

`fr-undefined` is it the behaviour we want?

patsissons: it's probably fine. i'm fine changing to `options ?? ''` if we want to use empty string instead

packages/react-i18n/src/utilities/translate.tsx
```diff
@@ -19,14 +19,14 @@ const SEPARATOR = '.';
...
+  locales?: string | string[],
+  options?: Intl.NumberFormatOptions,
 ) {
-  const key = numberFormatCacheKey(locale, options);
+  const key = numberFormatCacheKey(locales, options);
   if (numberFormats.has(key)) {
     return numberFormats.get(key)!;
   }
-  const i = new Intl.NumberFormat(locale, options);
+  const i = new Intl.NumberFormat(locales, options);
```
alexandcote, line 29: I guess passing `undefined` to `NumberFormat` is the same as passing an empty object?

Issue #503, opened 2019-02-06
React I18n / Expose client timezone on `I18n`


loic-d:
## Overview

Expose the client's timezone on `I18n`.

## Motivation

> What inspired this enhancement? What makes you think this should be in `quilt`?

In some scenarios we need to know the client's timezone to calculate time offsets between a shop's timezone and the browser's timezone.

This can easily be done:

```
export function getBrowserIanaTimeZone() {
  return Intl.DateTimeFormat().resolvedOptions().timeZone;
}
```

This could easily be exposed on `I18n` as a getter `clientTimezone` or something like that.

This will also be easier to test. At the moment we are mocking `Intl.DateTimeFormat().resolvedOptions().timeZone` in our tests.

```
let resolvedDateFormatOptions: jest.SpyInstance;
const initialDateFormatFormat = Intl.DateTimeFormat();
  resolvedDateFormatOptions = jest.spyOn(Intl, 'DateTimeFormat');
  resolvedDateFormatOptions.mockImplementation(() => ({
    format: initialDateFormatFormat.format,
    resolvedOptions: () => ({
      ...initialDateFormatFormat.resolvedOptions(),
      timeZone: 'America/Toronto',
    }),
  }));
```

If the client's timezone was directly exposed on `I18n`, it would be easier to mock.

### Area

- [x] Add any relevant `Area: <area>` labels to this issue

### Scope

- Is this issue related to a specific package?

  - [x] Tag it with the `Package: <package_name>` label.

---

## Checklist

- [x] Please delete the labels section before submitting your issue
- [x] I have described this issue in a way that is actionable (if possible)

ismail-syed:
I don't believe this is still an issue. Please re-open if it is.

PR #1367, opened 2020-04-09
[dates] exposing memoizedGetDateTimeFormat and adding getIanaTimeZone

patsissons:
## Description

I am exposing `memoizedGetDateTimeFormat()` so that consumers can avoid a memory leak when accessing a `Intl.DateTimeFormat` instance, as well as adding a new `getIanaTimeZone` that will extract the `timeZone` from the formatters `resovedOptions()`. This can be used to get the browser IANA timezone by simply calling without any parameters.

resovles #503

## Type of change

to not work as expected)

## Checklist

### DIFF
--- a/packages/dates/src/tests/utilities.test.ts
+++ b/packages/dates/src/tests/utilities.test.ts
@@ -1,4 +1,4 @@
-import {formatDate} from '../utilities/formatDate';
+import {dateTimeFormatCacheKey, formatDate} from '../utilities/formatDate';
 
 describe('formatDate()', () => {
   it('formats dates as expected for various locales', () => {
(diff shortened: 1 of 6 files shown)


packages/dates/src/utilities/formatDate.ts
```diff
@@ -1,13 +1,16 @@
-const intl = new Map();
-const memoizedGetDateTimeFormat = function(locale, options) {
+const intl = new Map<string | undefined, Intl.DateTimeFormat>();
+export function memoizedGetDateTimeFormat(
+  locale?: string | string[],
```
alexandcote, line 3:
```suggestion
  locales?: string | string[],
```

packages/dates/src/utilities/formatDate.ts
```diff
@@ -50,8 +53,10 @@ export function formatDate(
 }
 
 function dateTimeFormatCacheKey(
-  locales: string | string[],
+  locales?: string | string[],
   options?: Intl.DateTimeFormatOptions,
 ) {
-  return `${locales}-${JSON.stringify(options)}`;
+  return `${
+    Array.isArray(locales) ? locales.join('-') : locales
```
alexandcote, line 60: The same comment in this PR, I would prefer we move the ternary outside of the interpolation.

packages/dates/src/utilities/formatDate.ts
```diff
@@ -50,8 +53,10 @@ export function formatDate(
...
+  }-${JSON.stringify(options)}`;
```
alexandcote, line 61: Same question with the `undefined` edgecase.

packages/dates/src/tests/utilities.test.ts
```diff
@@ -37,3 +37,47 @@ describe('formatDate()', () => {
...
+    );
+  });
+
+  it('creates a key for a multiple locales without options', () => {
+    expect(dateTimeFormatCacheKey([locale, otherLocale])).toBe(
+      `${locale}-${otherLocale}-{}`,
+    );
+  });
+
+  it('sorts locales when there are multiple', () => {
```
alexandcote, line 78: 👌

PR #1371, opened 2020-04-14
:sparkles: Adding base package for mime types

posth:
## Description

First part to help fix https://github.com/Shopify/quilt/issues/411

Follow-up PR to add common mime types will be: https://github.com/Shopify/quilt/pull/1372

Based off a comment on https://github.com/Shopify/quilt/pull/1344 [here](https://github.com/Shopify/quilt/pull/1344#pullrequestreview-388275429).

We take the `@shopify/files` [package](https://github.com/Shopify/web/tree/master/packages/%40shopify/files) that exists in `@shopify/web`, and expose it as a new package in quilt.

This enables consumers to grab the common mime types it exposes, and two helpers to dynamically grab file extensions/mime types.

## Type of change

## Checklist

### DIFF
--- a/packages/mime-types/README.md
+++ b/packages/mime-types/README.md
@@ -1,0 +1,51 @@
+# `@shopify/mime-types`
+
+[![Build Status](https://travis-ci.org/Shopify/quilt.svg?branch=master)](https://travis-ci.org/Shopify/quilt)
+[![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](LICENSE.md) [![npm version](https://badge.fury.io/js/%40shopify%2Fmime-types.svg)](https://badge.fury.io/js/%40shopify%2Fmime-types.svg) [![npm bundle size (minified + gzip)](https://img.shields.io/bundlephobia/minzip/@shopify/mime-types.svg)](https://img.shields.io/bundlephobia/minzip/@shopify/mime-types.svg)
+
+Mime type consistency
+
+## Installation
+
+```bash
+$ yarn add @shopify/mime-types
+```
+
+## Usage
+
+This package exposes utilities to dynamically get MIME types from file names and/or get file extensions based on common MIME types.
+
+The types of files currently supported are:
+
+```
+  'image/jpeg',
+  'image/png',
+  'image/gif',
+  'application/pdf',
+  'video/mp4',
+  'video/quicktime',
+  'model/gltf-binary',
+  'application/x-mpegURL',
+```
+
+This helps keeping MIME types and file extensions consistent.
+
+### `getMimeTypeFromFilename`
+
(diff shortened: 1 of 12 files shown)


packages/files/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Installation
+
+```bash
+$ yarn add @shopify/files
+```
+
+## Usage
+
+This package exposes utilities and enums related to handling files.
```
michenly, line 14: handling files in what way? Reading of the file? Writing of the file? What type of files? Is this a file a a consumer are using in the application, a file a end user uploaded?

packages/files/src/get-extension-from-mime-type.ts
```diff
@@ -0,0 +1,25 @@
+import {MimeType} from './types';
+
+export function getExtensionFromMimeType(mimeType: MimeType) {
```
michenly, line 3: so this library only handle media files?

packages/files/README.md
posth: It would be to get MIME types from file names, or get file extensions from MIME types - those are the utilities exposed.

Currently it's mostly supporting media files, but https://github.com/Shopify/quilt/pull/1372 would ideally follow for other more common file types - this would fix https://github.com/Shopify/quilt/issues/411

I've updated the `README.md`, let me know if that's a bit clearer for consumers

michenly:
Thanks for the updated `README.md`, this is defiantly moving in the direction that makes more sense to me. I would like to see some usage (how and when to use the utilities) documentation.

I also question the library name. `files` feels to be too generic for what this utility currently do.
Either rename the library to be more specific, or update the documentation with what other types of utilities in the future belong in this library.

posth:
> I would like to see some usage (how and when to use the utilities) documentation.

Updated the `README.md`

> I also question the library name. `files` feels to be too generic for what this utility currently do.
> Either rename the library to be more specific, or update the documentation with what other types of utilities in the future belong in this library.

With https://github.com/Shopify/quilt/pull/1372, can update the documentation with the other types on the `README.md`.

`files` is what was used in `web` as that package, do you have an alternative in mind?

packages/mime-types/README.md
```diff
@@ -0,0 +1,51 @@
...
+```
+
+### `getExtensionFromMimeType`
+
+Takes in a `MimeType` and returns a string filename extension that matches the inputted `MimeType`.
+
+```ts
+import {getExtensionFromMimeType, MimeType} from '@shopify/files';
+
+getExtensionFromMimeType(MimeType.Pdf); // application/pdf
```
loic-d, line 50: Return value should be `.pdf` according to https://github.com/Shopify/quilt/pull/1371/files#diff-725363ff63fd976a5c2497fea33f0ba2R18

PR #1372, opened 2020-04-14
:sparkles: Adding common mime types

posth:
## Description

>Note: https://github.com/Shopify/quilt/pull/1371 will have to be merged as this branches off of it.

Second part to fix https://github.com/Shopify/quilt/issues/411

We add more common mime types to the new `@shopify/files` package.

## Type of change

## Checklist


packages/mime-types/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@shopify/mime-types",
-  "version": "0.0.0",
+  "version": "0.0.1",
```
loic-d, line 3: I think the version bump happens when you do a release.

```
Follow the prompts to choose a version for each package. If you are releasing a new package, we encourage you to version it 1.0.0 to start with.
```

https://github.com/Shopify/quilt/blob/master/documentation/guides/release-and-deploy.md

posth: Ahh that's good to know thank you

PR #1378, opened 2020-04-16
Update yarn.lock

vsumner:
## Description
The yarn lock changed after I pulled latest master and `yarn installed`. Committing the changes.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -77,7 +77,7 @@
   dependencies:
     "@apollographql/graphql-language-service-types" "^2.0.0"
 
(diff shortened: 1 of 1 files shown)


yarn.lock
```diff
@@ -11095,14 +11083,6 @@ tough-cookie@^3.0.1:
     psl "^1.1.28"
     punycode "^2.1.1"
 
-tough-cookie@~2.4.3:
```
dahukish, line 11098: we no longer need tough cookie?

PR #1383, opened 2020-04-17
[react-form] add allErrors property to Field

sjreive:
## Description

This PR adds an optional 'allErrors' property returned by useField in react-form.

The purpose of this change is to allow access to all error messages resulting from runValidation() if multiple validations fail for a single field. The format of this property is `allErrors: ['error msg 1', 'error msg 2']`

I modified the updateErrorAction to accept the whole array of validation error messages, and set the allErrors property in the reducer function. The runValidation function still returns the error message from the first failing validator.

## Type of change

- [x] react-form Minor: New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/packages/react-form/CHANGELOG.md
+++ b/packages/react-form/CHANGELOG.md
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## [Unreleased] -->
+## [Unreleased]
+
+### Added
+
+- Add optional `allErrors` property to `Field` type definition that stores all error messages resulting from `runValidation` [#1383](https://github.com/Shopify/quilt/pull/1383)
 
 ## [0.5.4] - 2020-04-13
 
--- a/packages/react-form/src/hooks/field/field.ts
+++ b/packages/react-form/src/hooks/field/field.ts
@@ -123,13 +123,13 @@ export function useField<Value = string>(
 
   const runValidation = useCallback(
     (value: Value = state.value) => {
-      const error = validators
+      const errors = validators
         .map(check => check(value, {}))
         .filter(value => value != null);
 
-      if (error && error.length > 0) {
-        const [firstError] = error;
-        dispatch(updateErrorAction(firstError));
+      if (errors && errors.length > 0) {
+        const [firstError] = errors;
+        dispatch(updateErrorAction(errors));
         return firstError;
       }
 
--- a/packages/react-form/src/hooks/field/reducer.ts
+++ b/packages/react-form/src/hooks/field/reducer.ts
@@ -1,14 +1,14 @@
 import {useReducer, Reducer} from 'react';
 
 import {FieldState, ErrorValue, DirtyStateComparator} from '../../types';
-import {defaultDirtyComparator} from '../../utilities';
+import {defaultDirtyComparator, shallowArrayComparison} from '../../utilities';
 
 export interface ReducerOptions<Value> {
   dirtyStateComparator?: DirtyStateComparator<Value>;
 }
 interface UpdateErrorAction {
   type: 'updateError';
-  payload: ErrorValue;
+  payload: ErrorValue[] | ErrorValue;
 }
 
 interface ResetAction {
@@ -45,7 +45,9 @@ export function newDefaultAction<Value>(value: Value): NewDefaultAction<Value> {
   };
 }
 
-export function updateErrorAction(error: ErrorValue): UpdateErrorAction {
+export function updateErrorAction(
+  error: ErrorValue[] | ErrorValue,
+): UpdateErrorAction {
   return {
     type: 'updateError',
     payload: error,
@@ -88,10 +90,22 @@ export function makeFieldReducer<Value>({
       }
 
       case 'updateError': {
-        return {
-          ...state,
-          error: action.payload,
-        };
+        const payload = Array.isArray(action.payload)
+          ? action.payload
+          : [action.payload];
+        const [firstError] = payload;
+        if (shallowArrayComparison(payload, state.allErrors)) {
+          return {
+            ...state,
+            error: firstError,
+          };
+        } else {
+          return {
+            ...state,
+            error: firstError,
+            allErrors: payload,
+          };
+        }
       }
 
       case 'reset': {
@@ -103,6 +117,7 @@ export function makeFieldReducer<Value>({
           value: defaultValue,
           dirty: false,
           touched: false,
+          allErrors: [],
         };
       }
 
@@ -138,5 +153,6 @@ export function initialFieldState<Value>(value: Value): FieldState<Value> {
     error: undefined,
     touched: false,
     dirty: false,
+    allErrors: [],
   };
 }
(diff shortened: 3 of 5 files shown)


packages/react-form/src/hooks/field/reducer.ts
```diff
@@ -88,10 +90,19 @@ export function makeFieldReducer<Value>({
...
+          const [firstError] = action.payload;
+          return {
+            ...state,
+            error: firstError,
+            allErrors: action.payload,
+          };
+        } else {
+          return {
+            ...state,
+            error: action.payload,
```
michaelmelanson, line 103: I think you should set the `allErrors` field in both the cases.

```suggestion
            error: action.payload,
            allErrors: [action.payload],
```

packages/react-form/src/hooks/field/field.ts
```diff
@@ -129,7 +129,7 @@ export function useField<Value = string>(
 
       if (error && error.length > 0) {
```
michaelmelanson, line 130: You might think about renaming `error` to `errors` here.

packages/react-form/src/hooks/field/reducer.ts
sjreive: good call, I have made this change 👍

packages/react-form/src/hooks/field/test/field.test.tsx
```diff
@@ -191,6 +191,57 @@ describe('useField', () => {
...
+            secondFailingValidator,
+          ],
+        };
+        const wrapper = mount(<TestField config={fieldConfig} />);
+        const newValue = faker.commerce.product();
+
+        wrapper.find('input')!.trigger('onChange', changeEvent(newValue));
+        wrapper.find('input')!.trigger('onBlur', blurEvent());
+
+        wrapper.debug();
```
vsumner, line 236: Please remove :)

packages/react-form/CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## [Unreleased] -->
```
vsumner, line 8: Please uncomment.

packages/react-form/src/hooks/field/test/field.test.tsx
sjreive: 🤦‍♀️Ty for catching, have removed

packages/react-form/CHANGELOG.md
sjreive: Done :)

packages/react-form/src/hooks/field/reducer.ts
```diff
@@ -138,5 +150,6 @@ export function initialFieldState<Value>(value: Value): FieldState<Value> {
     error: undefined,
     touched: false,
     dirty: false,
+    allErrors: [],
```
jared-dykstra, line 153: Because, `allErrors` is a non-primitive type, it cannot be consumed in the same way as `dirty`, or `touched` or other primitive values.  What is the probability of at least one consumer using the `allErrors` array like this:

```tsx
React.useEffect(() => {
  /* AllErrors is assumed to have changed. Do something imperative, like a Toast message */
}, [field.allErrors]
```

My concern is that it could create a really poor DX.  Developers would expect to be able to detect meaningful changes to allErrors, but it will change all the time, as new array instances are created even if the contents of the array aren't meaningfully different.

I also understand there would be some real reluctance to make `useField` depend on an immutability library, as that would affect a large percentage of bundle sizes.

jared-dykstra: I intended to click "Comment", not "Request Changes".   Please feel free to dismiss my review if it's not relevant.

sjreive: @jared-dykstra I see your point. Do you have any suggestion for a change that wouldn't have such an impact of developer experience ?

jared-dykstra: I think it's a bit tricky to solve.  Some ideas:
1. Make the reducer immutable, using immer.js.
   * This would be a rather significant change due to deps and bundle size implications, so there is plenty of reason to *not* go this route
   * However, more and more libraries are embracing immutable data
1. Stick with primitive types
   * `allErrors` would have to be exposed a different way, perhaps as the return value of `runValidation()`  However, that would provide a Developer Experience which encourages imperative programming.  It wouldn't mesh very will with React's Functional Programming paradigm.
1. Carefully manage the array instance:
   1. Create a function to update the previous value of `allErrors` by comparing each array element: If everything is the same, re-use the previous instance.  If anything has changed, create a new array instance.

IMO, the best option is to make the state immutable.  The 2nd best option is to carefully manage the array instance.

I'd recommend adding at least one test:

packages/react-form/src/hooks/field/test/field.test.tsx
```diff
@@ -191,6 +191,100 @@ describe('useField', () => {
...
+        wrapper.find('input')!.trigger('onBlur', blurEvent());
+
+        const allErrorsFirstValidation = wrapper.find('p').props.children;
+
+        wrapper.find('input')!.trigger('onChange', changeEvent(newValue));
+        wrapper.find('input')!.trigger('onBlur', blurEvent());
+
+        const allErrorsSecondValidation = wrapper.find('p').props.children;
+
+        expect(allErrorsFirstValidation).toBe(allErrorsSecondValidation);
```
jared-dykstra, line 286: Nice!

* I'd like to change the expect from `tobe` to `toStrictEqual`, as that would be closer to the quality checks in `useEffect`'s deps list.
* Also, can we have a second test for the happy-path, essentially the same, but testing the equality check when `allErrorsFirstValidation === allErrorsSecondValidation` when both arrays have zero elements?

sjreive: Done :)

packages/react-form/src/hooks/field/reducer.ts
```diff
@@ -88,10 +90,33 @@ export function makeFieldReducer<Value>({
       }
 
       case 'updateError': {
-        return {
-          ...state,
-          error: action.payload,
-        };
+        if (Array.isArray(action.payload)) {
```
michaelmelanson, line 93: You could combine the array and non-array cases here, with something like this:

```
const errors = Array.isArray(action.payload) ? action.payload : [action.payload];
```

Then from the non you can use `errors` instead of `action.payload`.

alexandcote: Like @michaelmelanson said,

you can do : 
```ts
const payload = Array.isArray(action.payload) ? action.payload : [action.payload];
const [firstError] = payload;
if (shallowArrayComparison(action.payload, state.allErrors)) {
  return {
      ...state,
      error: firstError,
  };
} else {
  return {
    ...state,
    error: firstError,
    allErrors: action.payload,
  };
}
```

packages/react-form/src/hooks/field/test/field.test.tsx
```diff
@@ -191,6 +191,149 @@ describe('useField', () => {
...
+        wrapper.find('input')!.trigger('onBlur', blurEvent());
+
+        const allErrorsSecondValidation = wrapper.find('p').props.children;
+
+        expect(allErrorsFirstValidation).toStrictEqual(
+          allErrorsSecondValidation,
+        );
+      });
+
+      it('does not update allErrors if there are no failing validations', () => {
```
alexandcote, line 291: Do we want to clear the `allErrors` array when the use fix the error?

PR #1391, opened 2020-04-20
[Date.Humanize] Humanize future dates

jaxee:
## Description

Improving `Date.Humanize` to take into consideration future dates (specifically: `Tomorrow at {time}`).

Splits the `humanizeDate` method into `humanizeFutureDate` and `humanizePastDate`. The previous functionality lives in `humanizePastDate` and the new functionality lives in `humanizeFutureDate`.

This change sets up the humanize method(s) to easily add more future date humanization.

Questions:
- Is the `CHANGELOG.md` correct?
- Should we be differentiating between Today (past) & Today (future)?
- Are there other future date considerations that we can add now?

## Type of change

- [x] react-i18n - Minor: New feature (non-breaking change which adds functionality)

## Checklist

cc @patrickkidney @jppellerin

### DIFF
--- a/packages/react-i18n/src/i18n.ts
+++ b/packages/react-i18n/src/i18n.ts
@@ -1,10 +1,12 @@
 import {
   formatDate,
+  isFutureDate,
   isLessThanOneHourAgo,
   isLessThanOneMinuteAgo,
   isLessThanOneWeekAgo,
   isLessThanOneYearAgo,
   isToday,
+  isTomorrow,
   isYesterday,
   TimeUnit,
 } from '@shopify/dates';
@@ -443,6 +445,14 @@ export class I18n {
   }
 
   private humanizeDate(date: Date, options?: Intl.DateTimeFormatOptions) {
+    if (isFutureDate(date)) {
+      return this.humanizeFutureDate(date, options);
+    } else {
+      return this.humanizePastDate(date, options);
+    }
+  }
+
+  private humanizePastDate(date: Date, options?: Intl.DateTimeFormatOptions) {
     if (isLessThanOneMinuteAgo(date)) {
       return this.translate('date.humanize.lessThanOneMinuteAgo');
     }
@@ -492,6 +502,27 @@ export class I18n {
         date: monthDay,
         time,
       });
+    }
+
+    return this.formatDate(date, {
+      ...options,
+      style: DateStyle.Short,
+    });
+  }
+
+  private humanizeFutureDate(date: Date, options?: Intl.DateTimeFormatOptions) {
+    const time = this.formatDate(date, {
+      ...options,
+      hour: 'numeric',
+      minute: '2-digit',
+    }).toLocaleLowerCase();
+
+    if (isToday(date)) {
+      return time;
+    }
+
+    if (isTomorrow(date)) {
+      return this.translate('date.humanize.tomorrow', {time});
     }
 
     return this.formatDate(date, {
--- a/packages/react-i18n/src/test/i18n.test.ts
+++ b/packages/react-i18n/src/test/i18n.test.ts
@@ -1192,151 +1192,173 @@ describe('I18n', () => {
         ).toBe('Dec. 20, 2012');
       });
 
-      it('formats a date less than one minute ago', () => {
-        const today = new Date('2012-12-20T00:00:00-00:00');
-        const lessThanOneMinuteAgo = new Date(today.getTime());
-        lessThanOneMinuteAgo.setSeconds(-5);
-        clock.mock(today);
-        const i18n = new I18n(defaultTranslations, {
-          ...defaultDetails,
-          timezone,
+      describe('past dates', () => {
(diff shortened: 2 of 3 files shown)


packages/react-i18n/src/i18n.ts
```diff
@@ -500,6 +510,27 @@ export class I18n {
     });
   }
 
+  private humanizeFutureDate(date: Date, options?: Intl.DateTimeFormatOptions) {
+    const time = this.formatDate(date, {
+      ...options,
```
kyledurand, line 515: Should this be moved two lines down so consumers can override hour and minute settings?

packages/react-i18n/src/i18n.ts
```diff
@@ -500,6 +510,27 @@ export class I18n {
...
+    if (isToday(date)) {
+      return time;
+    }
+
+    if (isTomorrow(date)) {
+      return this.translate('date.humanize.tomorrow', {time});
+    }
+
+    return this.formatDate(date, {
+      ...options,
```
kyledurand, line 529: Same thing here

packages/react-i18n/src/i18n.ts
```diff
@@ -500,6 +510,27 @@ export class I18n {
     });
   }
 
+  private humanizeFutureDate(date: Date, options?: Intl.DateTimeFormatOptions) {
+    const time = this.formatDate(date, {
+      ...options,
+      hour: 'numeric',
+      minute: '2-digit',
```
kyledurand, line 517: Do we have a `DateStyle.Time` or something we can use here?

packages/react-i18n/src/i18n.ts
```diff
@@ -500,6 +510,27 @@ export class I18n {
...
+    }).toLocaleLowerCase();
+
+    if (isToday(date)) {
+      return time;
+    }
+
+    if (isTomorrow(date)) {
+      return this.translate('date.humanize.tomorrow', {time});
```
kyledurand, line 525: This translation key already existed?

alex-page:
Hey @jaxee sorry I have to remove myself from reviewing this PR I have a bunch of work I need to wrap up this week. Let me know if you need help finding another person to look over this. I think Kyle has done a good bit of work getting this started.

packages/react-i18n/src/i18n.ts
jaxee: The time implementation in `Humanize` is the same as before the PR.

I think it is a good idea. If allowing hour and minute overrides is something we want to do in `Humanize` - could it be done in a follow up PR?

packages/react-i18n/src/i18n.ts
jaxee: See [this comment](https://github.com/Shopify/quilt/pull/1391#discussion_r413034933)

packages/react-i18n/src/i18n.ts
jaxee: The implementation for `DateStyle.Time` is a little different:
```
{
    hour: '2-digit',
    minute: '2-digit',
}
```
I think the difference would be:
numeric - `3:23 pm`
2-digit - `03:23 pm`

packages/react-i18n/src/i18n.ts
jaxee: My understanding is that the translations are provided by the consuming library (because I couldn't find `Yesterday at {time}' anywhere). Is there a place I should be putting translations?

packages/react-i18n/src/i18n.ts
kyledurand: I'm down with that 👍

packages/react-i18n/src/test/i18n.test.ts
```diff
@@ -1192,151 +1192,173 @@ describe('I18n', () => {
...
-          defaultTranslations,
-          i18n.locale,
-        );
       });
 
-      it('formats a date less than one year ago', () => {
-        // Offset from 2012-12-20 to 2012-07-20
-        const fiveMonthsInMilliseconds =
-          (30 + 31 + 30 + 31 + 31) * 24 * 60 * 60 * 1000;
+      describe('future dates', () => {
```
alexandcote, line 1329: Do you want to add a test for today but in the future?

packages/react-i18n/src/i18n.ts
alexandcote: I think we probably want to enforce good practice here. I think I prefer the current implementation for this reason but maybe @patsissons have other thoughts.

packages/react-i18n/src/i18n.ts
alexandcote: @patsissons Do you remember why we choose `DateStyle.Time` to have 2 digits for the hour by any chance?

packages/react-i18n/src/test/i18n.test.ts
jaxee: Yep! I can definitely add that test 👍

PR #1394, opened 2020-04-21
Fix available permits in Semaphore

tleunen:
## Description

While debugging an issue with a mutex (Semaphore(1)), I noticed it was possible to increase the number of available permits when the consumer calls `acquire()` several times before they get released.

## Type of change
- [x] `@shopify/semaphore` Patch: Fixed an issue with incorrect number of permits when the semaphore has multiple tries to be acquired, then released

## Checklist

### DIFF
--- a/packages/semaphore/src/Semaphore.ts
+++ b/packages/semaphore/src/Semaphore.ts
@@ -31,7 +31,6 @@ export class Semaphore {
 
   acquire(): Promise<Permit> {
     if (this.availablePermits > 0) {
-      this.availablePermits--;
       return Promise.resolve(this.createPermit());
     } else {
       const deferred: Deferred = {};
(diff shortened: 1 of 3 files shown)


packages/semaphore/src/Semaphore.ts
```diff
@@ -44,15 +43,18 @@ export class Semaphore {
...
       async (): Promise<any> => {
         this.availablePermits++;
 
-        if (this.deferreds.length) {
+        if (this.deferreds.length > 0) {
           const deferred = this.deferreds.shift()!;
           deferred.resolve!(this.createPermit());
           await deferred.promise;
         }
+
```
atesgoral, line 57: Runaway empty line
```suggestion
```

atesgoral:
🚢

PR #1396, opened 2020-04-22
Add support for useMediaEffect

tleunen:
## Description

In our project, we prefer using `useLayoutEffect` to prevent the the user from seeing the UI flashing from one state to another. This is done because `useLayoutEffect` will rerender synchronously.

It's easier to review if you hide the whitespace changes.

## Type of change

- [x] `react-hooks` Minor

## Checklist

### DIFF
--- a/packages/react-hooks/src/hooks/tests/media.test.tsx
+++ b/packages/react-hooks/src/hooks/tests/media.test.tsx
@@ -2,15 +2,9 @@ import React from 'react';
 import {matchMedia, mediaQueryList} from '@shopify/jest-dom-mocks';
 import {mount} from '@shopify/react-testing';
 
-import {useMedia} from '../media';
+import {useMedia, useMediaLayout} from '../media';
 
-function MockComponent({mediaQuery}: {mediaQuery: string}) {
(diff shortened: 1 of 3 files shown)


packages/react-hooks/src/hooks/tests/media.test.tsx
```diff
@@ -2,13 +2,7 @@ import React from 'react';
...
-import {useMedia} from '../media';
-
-function MockComponent({mediaQuery}: {mediaQuery: string}) {
-  const matchedQuery = useMedia(mediaQuery);
-  const message = matchedQuery ? 'matched' : 'did not match';
-  return <div>{message}</div>;
-}
+import {useMedia, useMediaLayout} from '../media';
 
 describe('useMedia', () => {
```
sylvhama, line 7: 1. Could you rename to `'useMedia and useMediaLayout'`?
2. Also you could use `describe.each` instead of doing a `forEach` on line 16

tleunen: Done! Thank you! TIL about `describe.each`.

PR #1398, opened 2020-04-22
[react-graphql] Add fetchMore param to useQuery hook result documentation

attila-berki:
## Description

Add the missing `fetchMore` to the `useQuery` hook result params documentation.

image

## Type of change

- [x] react-graphql Patch: Bug/ Documentation fix (non-breaking change which fixes an issue or adds documentation)

## Checklist


packages/react-graphql/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## [Unreleased] -->
```
alexandcote, line 8: Can you uncomment this line, please 😃

attila-berki:
> The doc was never a complete set of types (ie. its missing `called`)
> I wonder for maintenance sake, maybe we should just link to https://github.com/Shopify/quilt/blob/master/packages/react-graphql/src/hooks/types.ts#L23-L27

I don't think we should link it to the type definition, it is not so straight forward to figure out what is available based on that. 
If we always expose everything that comes from `apollo` than maybe linking the `apollo` docs and only mentioning the additional stuff would be better. What do you think?
https://www.apollographql.com/docs/react/data/queries/

attila-berki:
Hi @michenly 👋 

Should I still merge this PR, or you would like to go to the other direction?

Thanks.

michenly:
This is important to know that the type here is based on apollo API but slightly changed, specially since this library accept a range of apollo dependencies and there can be feature added in a minor update which cause the API listed in the README to be different. 

Your repo's apollo version + type should really be the source of truth here since linking to apollo docs will have similar problem where it can only display the API from the latest version and not from the version of apollo in your application.

packages/react-graphql/CHANGELOG.md
```diff
@@ -5,7 +5,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## [Unreleased] -->
+## [Unreleased]
+
+- Add `fetchMore` param to `useQuery` hook result documention ([#1398](https://github.com/Shopify/quilt/pull/1398)).
```
alexandcote, line 10: I don't think you need this because we will not release a new version for doc changes.

PR #1399, opened 2020-04-22
Add `innerWidth` mock to jest-dom-mocks

theodoretan:
## Description

I'm currently using `Object.defineProperty` to modify the innerWidth of the window but it can lead to test flakiness due to state leakage.

Instead, add this to the mock so we can restore it after the test runs to prevent bugs from popping up.

## Type of change

- [x] jest-dom-mocks Minor: New feature (non-breaking change which adds functionality)

## Checklist


packages/jest-dom-mocks/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## [Unreleased] -->
 
+## [Unreleased]
+
+- Added mock for `innerWidth` to dimentions ([#1399](https://github.com/Shopify/quilt/pull/1399))
```
theodoretan, line 12: Typo here

PR #1403, opened 2020-04-23
[react-form] allErrors will return an empty array if there are no errors

sjreive:
Added a check so that if error is undefined, allErrrors  on useField will be an empty array.


packages/react-form/src/hooks/field/reducer.ts
```diff
@@ -103,7 +104,7 @@ export function makeFieldReducer<Value>({
           return {
             ...state,
             error: firstError,
-            allErrors: payload,
+            allErrors: allErrors,
```
alexandcote, line 107: Since the key and the variable have the same name you can simply do this
```suggestion
            allErrors,
```

PR #1404, opened 2020-04-26
Hooks argument type fixes

CameronGorrie:
## Description

Issue noticed in Shopify/web (TypeScript 3.8) when upgrading `react-graphql@6.x` the `usePreload`, `usePrefetch`, and `useKeepFresh` argument types from `react-async` require a second argument:

Screen Shot 2020-04-26 at 1 07 45 PM

In many cases the hook is used by passing through the async/query component w/o additional arguments. If I understand correctly the `NoInfer` type gives the hook options precedence, so the second argument cannot be omitted when calling the hook because it cannot be inferred as undefined.

## Type of change

## Checklist

### DIFF
--- a/packages/react-async/src/hooks.ts
+++ b/packages/react-async/src/hooks.ts
@@ -25,7 +25,7 @@ export type KeepFreshable<KeepFreshOptions extends object> = Pick<
 export function usePreload<PreloadOptions extends object>(
   ...args: IfAllOptionalKeys<
     PreloadOptions,
-    [Preloadable<PreloadOptions>, NoInfer<PreloadOptions>?],
+    [Preloadable<PreloadOptions>, PreloadOptions?],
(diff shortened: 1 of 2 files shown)


packages/react-async/src/hooks.ts
```diff
@@ -25,7 +25,7 @@ export type KeepFreshable<KeepFreshOptions extends object> = Pick<
 export function usePreload<PreloadOptions extends object>(
   ...args: IfAllOptionalKeys<
     PreloadOptions,
-    [Preloadable<PreloadOptions>, NoInfer<PreloadOptions>?],
+    [Preloadable<PreloadOptions>, PreloadOptions?],
```
ismail-syed, line 28: If you hack in these type changes locally into web's typing for `react-async` does the codebase still type-check successfully?

CameronGorrie: I tried it out with the tophat command and `dev type-check` passes

Issue #1400, opened 2020-04-22
Add a way to pass basic data to node via rails, document it's use and places not to use it


TheMallen:
## Problem
Many apps want to be able to get *basic* data from the server that is unlikely / will never change over the course of the session before they render any React components. In some cases they need to do this *to decide whether they even want to render the React app* (looking at you `Shopify/Web`). We basically tell folks to "just use graphQL / fetch" for these things when they are using `quilt_rails`, but often they end up going through a lot of contortions to not do that or use it for things that it really shouldn't be used for.

**At the very least, if all you're getting is something like a shop_id that will never change during the user's session, it is bad to force an extra HTTP request to do so.**

## Current art
Presently the only real way to get data from your Rails app to your React app via quilt_rails without making any extra HTTP requests is:
- adding extra HTTP headers to the proxied request 
- having the app manually grab them using `react-html`'s `getRequestHeader`
- serializing/deserializing the extracted data using `useSerialized` 
- providing that data down into the react app via props or context

This is a lot of boilerplate to add for something that is relatively a common want, and as a result what usually ends up happening is they do in fact "just use GraphQL" and add an extra blocking network hop to all their renders.

## Proposal
- add an optional argument to `render_react` which serializes  data into JSON and attaches it as a header to the proxied request (for example: `x-quilt-data: '{"some": "serialized", "json": "object"}'`
- add code to the render middleware generated by `react-server` which extracts this data and passes it as a prop to the rendered app
- 🎉 

## Caveats
We definitely do not want this to be seen as the default way to pass *all* data, so we must be sure to document that it is specifically suited to *data that does not change during a session and is needed for every route of the app*.

PR #1411, opened 2020-04-30
Better patterns for sharing headers/data

ismail-syed:
## Description

Closes https://github.com/Shopify/quilt/issues/1400

This PR improves the way we add basic data and headers to the React server via Rails. This approach essentially follows whats documented in https://github.com/Shopify/quilt/issues/1400

## Checklist

### DIFF
--- a/gems/quilt_rails/README.md
+++ b/gems/quilt_rails/README.md
@@ -326,65 +326,40 @@ function App() {
 export default App;
 ```
 
-##### Example: sending headers from Rails controller
+##### Example: Sending headers and custom data from Rails controller
+
+In some cases you may want to send custom headers or basic data from Rails to your React server. Quilt facilitates this case by providing consumers with a `data` argument on the `render_react` call.
+
+**Note:** The data passed should be data that is unlikely or will never change over the course of the session before they render any React components.
 
 ```ruby
 class ReactController < ApplicationController
   include Quilt::ReactRenderable
 
   def index
-    render_react(headers: { 'x-custom-header': 'header-value-a' })
+    render_react(data: { 'x-custom-header': 'header-value-a', 'some_id': 123 })
   end
 end
 ```
 
-You will need to serialize the result of the useRequestHeader hook in order for it to persist to the client
-
-```tsx
-// app/ui/foundation/CustomUniversalProvider.tsx
-import {createContext} from 'react';
-import {createUniversalProvider} from '@shopify/react-universal-provider';
-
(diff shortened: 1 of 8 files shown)


gems/quilt_rails/README.md
```diff
@@ -326,67 +326,42 @@ function App() {
 export default App;
 ```
 
-##### Example: sending headers from Rails controller
+##### Example: Sending headers and custom data from Rails controller
```
TheMallen, line 329: Should these maybe be two different sections?

gems/quilt_rails/README.md
```diff
@@ -326,67 +326,42 @@ function App() {
...
+
+**Note:** The data passed should be data that is unlikely or will never change over the course of the session before they render any React components.
 
 ```ruby
 class ReactController < ApplicationController
 
   def index
-    render_react(headers: { 'x-custom-header': 'header-value-a' })
+    render_react(data: { 'x-custom-header': 'header-value-a', 'some_id': 123 })
```
TheMallen, line 340: I don't think `x-custom-header` should be passed into `data`, should it?

gems/quilt_rails/lib/quilt_rails/react_renderable.rb
```diff
@@ -6,29 +6,29 @@ module Quilt
   module ReactRenderable
     include ReverseProxy::Controller
 
-    def render_react(headers: {})
+    def render_react(headers: {}, data: {})
```
TheMallen, line 9: I think the example above should be using `headers` as well as `data`

gems/quilt_rails/README.md
ismail-syed: Yah, they should. Good catch

gems/quilt_rails/README.md
ismail-syed: No, it shouldn't. This is wrong. In this case, the `x-custom-header` header would be hidden behind the `x-quilt-data` header which is not what we'd want for a header.

gems/quilt_rails/README.md
ismail-syed: Tried separating these. Lots of redundancy so I just kept them together.

gems/quilt_rails/README.md
ismail-syed: To support the separation of headers/data while ensuring it's easily accessible on the client, I've merged the headers and data in the `x-quilt-data` response header to the react-server.

PR #1412, opened 2020-04-30
✨ magic-entries-webpack-plugin

TheMallen:
## Description
closes [sewing-kit#2009](https://github.com/Shopify/sewing-kit/issues/2009)

This PR adds a webpack plugin to quilt that allows convention based "magic" webpack entrypoints. The README and the tests should be sufficient to understand how it works.

## Stuff that should happen after
- See if there's a way we can get virtual modules in react-server to work with this
- Add to sewing-kit
- Pull out common workspace utilities that are in 3 tests now into a shared utility

### DIFF
--- a/packages/magic-entries-webpack-plugin/CHANGELOG.md
+++ b/packages/magic-entries-webpack-plugin/CHANGELOG.md
@@ -5,8 +5,8 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [0.1.0]
 
 ### Added
 
-- `@shopify/magic-entries-webpack-plugin` package
+- `@shopify/magic-entries-webpack-plugin` package [[#1412](https://github.com/Shopify/quilt/pull/1412)]
--- a/packages/magic-entries-webpack-plugin/CHANGELOG.md
+++ b/packages/magic-entries-webpack-plugin/CHANGELOG.md
@@ -1,0 +1,12 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
+and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+
+### Added
+
+- `@shopify/magic-entries-webpack-plugin` package
--- a/packages/magic-entries-webpack-plugin/README.md
+++ b/packages/magic-entries-webpack-plugin/README.md
@@ -1,0 +1,115 @@
+# `@shopify/magic-entries-webpack-plugin`
+
+[![Build Status](https://travis-ci.org/Shopify/quilt.svg?branch=master)](https://travis-ci.org/Shopify/quilt)
+[![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](LICENSE.md) [![npm version](https://badge.fury.io/js/%40shopify%2Fmagic-entries-webpack-plugin.svg)](https://badge.fury.io/js/%40shopify%2Fmagic-entries-webpack-plugin.svg)
+
+A webpack plugin that automatically sets up webpack entrypoints using filename conventions
+
+## Installation
+
+```bash
+$ yarn add @shopify/magic-entries-webpack-plugin
+```
+
+## Motivation
+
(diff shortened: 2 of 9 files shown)


TheMallen:
Just as a note, as part of this I checked how easy it would be to get the virtual modules react-server-webpack-plugin sets up to work with this and the answer is not very. I think we would likely need to fork the virtual modules plugin since it seems to do it's magic mostly via monkey patching `InputFileSystem` and since it does not do that to `readdir` it cannot "see" virtual modules.

packages/magic-entries-webpack-plugin/README.md
```diff
@@ -0,0 +1,115 @@
+# `@shopify/magic-entries-webpack-plugin`
+
+[![Build Status](https://travis-ci.org/Shopify/quilt.svg?branch=master)](https://travis-ci.org/Shopify/quilt)
+[![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](LICENSE.md) [![npm version](https://badge.fury.io/js/%40shopify%2Fmagic-entries-webpack-plugin.svg)](https://badge.fury.io/js/%40shopify%2Fmagic-entries-webpack-plugin.svg)
+
+A webpack plugin that automatically sets up webpack entrypoints using filename conventions
```
ismail-syed, line 6: `.` at the end of this line?

packages/magic-entries-webpack-plugin/README.md
```diff
@@ -0,0 +1,115 @@
...
+```
+
+### Server and client presets
+
+Some universal applications (such as react applications using server-side-rendering) may need to have a different set of entrypoints for the client and server environments.
+
+To provide for this common usecase, the plugin comes with presets for client and server accessible via the `.client` and `.server` static methods.
+
+```js
+// the same as the default behaviour except it looks for files matching '*.entry.server.{jsx,js,ts,tsx}'
```
ismail-syed, line 80: Does `entry`  need to be in the file name for this to match?

ismail-syed: Spoke too soon, just read the `Customization` section below.

packages/magic-entries-webpack-plugin/src/test/utilities.ts
```diff
@@ -0,0 +1,48 @@
...
+    await fs.emptyDir(this.root);
+    await fs.remove(this.root);
+  }
+}
+
+const rootFixtureDirectory = path.resolve(__dirname, '../fixtures');
+
+export async function createWorkspace({name}: {name: string}) {
+  const fixtureDirectory = path.join(rootFixtureDirectory, name);
+  await fs.mkdirp(fixtureDirectory);
```
ismail-syed, line 32: Can `mkdirp` be done in the `Workspace` constructor instead?

packages/magic-entries-webpack-plugin/src/test/utilities.ts
```diff
@@ -0,0 +1,48 @@
...
+  await fs.writeFile(path.join(fixtureDirectory, '.gitignore'), '*');
```
ismail-syed, line 33: It seems a bit strange to `.gitignore` all these manually. Maybe use a naming convention for the fixture directories and `.gitignore` them in quilts root `.gitignore`. If nested `.gitignore`s is a thing maybe a `.gitignore` in this plugin directory.

packages/magic-entries-webpack-plugin/CHANGELOG.md
```diff
@@ -0,0 +1,12 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
+and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
```
ismail-syed, line 8: Might as well just add a `0.1.0` release header here?

TheMallen:
@ismail-syed  you can add it using the `plugins` key on any webpack config, so you can tophat it pretty easily in any sewing-kit project using the `webpack` plugin

packages/magic-entries-webpack-plugin/src/test/utilities.ts
TheMallen: Looking at it again, we could do that but if we did we'd need to use sync versions of everything which is bad especially since with tests things often want to run in parallel. I suppose we could move `createWorkspace` onto `Workspace.create` which might be a bit nicer.

packages/magic-entries-webpack-plugin/src/test/utilities.ts
TheMallen: We have the same util as this in a couple other libraries, so I think manually gitignoring all of them might be weird. I don't really see a problem with us writing the gitignore, as we're already messing around in the filesystem in these folders anyway

Issue #1387, opened 2020-04-19
Safari 13.1 now blocks all third-party cookies by default


arenatheme:
https://community.shopify.com/c/Shopify-APIs-SDKs/Safari-13-1-and-embedded-apps/td-p/688416

pratiknikam:
I think everyone is facing this issue now. I didn't hear any solution from Shopify team yet :( 

In my case Safari 13.1 is causing cookie redirect loop.

trytada:
This would be important to prioritize quickly. Solution is discussed here https://medium.com/@sleyhane/shopify-embedded-apps-and-safari-81903dec20f7, it's currently making users that use Safari unable to use most apps using this auth module. If someone has an idea how to apply that fix to the module, that'd be amazing.

trytada:
We're also seeing Cookie Redirect Loop on Safari 13.1

pratiknikam:
@lemonmade would love to hear from you on this one.

lemonmade:
@pratiknikam and others, sorry for the delay. We have a couple folks working on getting this resolved, and we'll be able to provide guidance very soon. Keep an eye on this issue or the [community post](https://community.shopify.com/c/Shopify-APIs-SDKs/Safari-13-1-and-embedded-apps/td-p/688416/page/2).

trytada:
@lemonmade , this is great to hear this is getting worked on. I believe there should also be some different handling of cookies based on the user agent, as discussed here: https://www.chromium.org/updates/same-site/incompatible-clients

We're excited to see support for this!

Milbergs:
@lemonmade - looking forward to the fix, we are currently submitting our app for the review and we need to fix this ASAP

ReinisL:
I don't know if it's related to this issue, but what was causing the redirect loop for me was that I was missing these options in the afterAuth method. 

```
ctx.cookies.set("shopOrigin", shop, {
  httpOnly: false,
  secure: true, <----
  sameSite: "none", <----
});
```

pratiknikam:
@ReinisL no this one is different I believe. Can you test app with above code in Safari 13.1?

ReinisL:
@pratiknikam yes. I don't have redirect issue anymore. Tested on multiple Safari 13.1 browsers

pratiknikam:
That's super weird. Are you doing anything special with the way you handled cookies?

[...]

PR #1413, opened 2020-05-01
[koa-shopify-auth] Fixes ITP 2.3 and Safari 13.1 enable cookies loop

ayronshopify:
## Description
Package: **koa-shopify-auth**
Fixes https://github.com/Shopify/quilt/issues/1387

Intelligent Tracking Prevention (ITP) prevents websites from tracking users and with Safari 13.1 blocking all third-party cookies by default, we're finding that users are experiencing an endless loop of seeing the `Enable Cookies` screen, clicking the button, and being redirected back to the `Enable Cookies` screen again. This is due to the fact that we are not currently using the Storage Access API by calling `document.requestStorageAccess()` which is the way to ensure these cookies can now be saved/accessed. 

## Type of change

We are using some of [the code](https://github.com/Shopify/shopify_app/tree/master/app/assets/javascripts/shopify_app) implemented by `shopify_app` as this is already working and thoroughly tested there. 

We're keeping the `Enable Cookies` page which is surfaced at the top-level:
image

Once the user clicks the button they will be redirected to a new page in an iframe which will make the `document.requestStorageAccess()` call needed to save the cookies. 
image

Older browsers/Chrome will probably not see either of these screens, depending on the version checking done in the `StorageAccessHelper`. Newer Firefox users will see the `Enable Cookies` screen. Newer Safari users will see both the `Enable Cookies` screen and the `Browser` screen. Once the flow has been completed Older/Chrome/Firefox users shouldn't see any other screens when accessing their apps. However, newer Safari users will always see the `Browser` screen before they can access the app. 

New Safari Flow:
safari-13-fix

## Checklist


packages/koa-shopify-auth/src/auth/client/itp-helper.ts
```diff
@@ -0,0 +1,42 @@
+const itpHelper = `(function() {
```
ragalie, line 1: Might be worth putting a comment in that these came from `shopify_app` gem and are tested there

ismail-syed:
Can you please add a changelog update for this package.

packages/koa-shopify-auth/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## [Unreleased] -->
 
+## [3.1.57] - 2020-05-01
```
ismail-syed, line 10: This needs to be a `3.1.61` since the last released version was [3.1.60](https://github.com/Shopify/quilt/blob/master/packages/koa-shopify-auth/package.json#L3). The changelogs to keep a good track of peer dep bumps via lerna 😢

packages/koa-shopify-auth/src/auth/client/itp-helper.ts
```diff
@@ -0,0 +1,43 @@
+// Copied from https://github.com/Shopify/shopify_app
```
TheMallen, line 1: It's probably fine to leave this as is (especially because it is copied from somewhere), but if we felt like making it a bit more readable we could actually have this as  TypeScript with it's own tsconfig that compiles it for direct script tag use if we wanted. Maybe a good future improvement.

packages/koa-shopify-auth/src/auth/create-request-storage-access.ts
```diff
@@ -0,0 +1,82 @@
...
+  return function requestStorage(ctx: Context) {
+    const {query} = ctx;
+    const {shop} = query;
+
+    if (shop == null) {
+      ctx.throw(400, Error.ShopParamMissing);
+      return;
+    }
+
+    ctx.body = `
```
TheMallen, line 28: Eventually we might want to introduce tagged template literals or something so we can make all of these a bit nicer, but this is fine for now.

packages/koa-shopify-auth/src/auth/client/storage-access-helper.ts
```diff
@@ -0,0 +1,151 @@
...
+      }
+
+      StorageAccessHelper.prototype.grantedStorageAccess = function() {
+        try {
+          sessionStorage.setItem('shopify.granted_storage_access', true);
+          this.setCookie('shopify.granted_storage_access=true');
+          if (!document.cookie) {
+            throw 'Cannot set third-party cookie.'
+          }
+          this.redirectToAppTargetUrl();
```
eugene-kim, line 72: Hi @ayronshopify @ismail-syed @TheMallen 

I'm a Shopify partner who's trying to understand this authorization process and I'm confused about this line and was hoping to get some clarification.

Why do we redirect the user to `this.redirectData.appTargetUrl` rather than `/auth`, which is the [`oAuthStartPath`](https://github.com/Shopify/quilt/blob/master/packages/koa-shopify-auth/src/auth/index.ts#L42) in the [`createShopifyAuth`](https://github.com/Shopify/quilt/blob/master/packages/koa-shopify-auth/src/auth/index.ts#L31) function?

The logic that checks the `shopify.granted_storage_access` cookie is associated with the `/auth` path and so it seems to me that we should be redirecting the user to `/auth`

Thank you!

ragalie: Hi! Good question. I think the reason is that we may have already completed OAuth. If we're already completed OAuth, and now we also have storage access, then the app is ready to load, so it doesn't make sense to do OAuth again.

If we haven't completed OAuth yet, then when we hit the app target URL we'll realize we need to do OAuth, and then redirect to OAuth anyway, so that case is also covered.

Does that make sense?

kavika-1:
The redirect always goes to "/auth" now on safari, for example.  In our case, we set the prefix property, and the redirect causes a 404.  By prefix, we mean we use the "prefix" in the example provided for koa-shopify-auth...
````
app.use(
  shopifyAuth({
    // if specified, mounts the routes off of the given path
    // eg. /shopify/auth, /shopify/auth/callback
    // defaults to ''
    prefix: '/shopify',
````
In the past we could use the referer (sic) to retrieve the path, but now that this redirect happens client side, we no longer receive/have access the referer in order to patch this up.  Any way to preserve or send along the prefix?

PR #1415, opened 2020-05-04
[@shopify/react-i18n]  📝 Add documentation for React Native support

macdonaldr93:
## Description

We've been trying to get this to work in our React Native apps. Figured out that support was added a while ago but supporting React Native was not the intended use-case. LUCKILY, it works for React Native too.

## Type of change

Add documentation for React Native support in @shopify/react-i18n.

- [x] @shopify/react-i18n Patch: Bug/ Documentation fix (non-breaking change which fixes an issue or adds documentation)

## Checklist


packages/react-i18n/CHANGELOG.md
```diff
@@ -5,6 +5,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+- Add documentation about React Native support ([#1415](https://github.com/Shopify/quilt/pull/1415))
```
michenly, line 10: You actually don't need this as we don't publish a new version for documentation or testing changes.

GoodForOneFare:
Is there a way to fail Quilt CI if i18n stops working with React Native?

macdonaldr93:
@GoodForOneFare that's a good point. One way I could see this test working is to mock the dynamic import function to throw an error and write a test for that. That's essentially why we have to use this other mode. What do you think?

GoodForOneFare:
Short of spinning up a React Native app, I doubt there's an effective way to test it 😞 The good news is that RN is using (IMO) the least breakable version of i18n compilation, so less of a concern than when I first saw the PR 😄

macdonaldr93:
@GoodForOneFare I actually just discovered another dependency... I don't know why I didn't have to do this originally.

I had to add a script to generate the index files.

```
const {
  generateTranslationDictionaries,
} = require('@shopify/react-i18n/generate-dictionaries');

const SUPPORTED_LOCALES = ['en'];

(async () => {
  await generateTranslationDictionaries(SUPPORTED_LOCALES, {
    fallbackLocale: 'en',
  });
})();
```

I don't mind adding this to the docs but is there a way to run this without having to generate the indexes in a separate step.

GoodForOneFare:
I think [the usage guide's scripts](https://github.com/Shopify/quilt/blob/master/packages/react-i18n/README.md#statically-embedding-locale-specific-translations) cover that.

perpetuatingcuriosity:
From the usage guide:
> […] generate translations/index.js files containing specific locale data using the @shopify/react-i18n/generate-dictionaries helper. e.g., the _following code **generates three versions of an application** with English, French, and German content using webpack_.

React Native doesn’t use webpack, and given app size is a limiting factor on mobile, it would not be beneficial to have 3 different versions of the application. 

@michenly Can you clarify what happens when webpack isn’t used?

michenly:
@perpetuatingcuriosity absolutely nothing when you don't want language specific build and didn't use the example code in the guide.

The script @macdonaldr93 had above will generate all the index files (it has to be done as a separate step), then update your babel-plugin integration to included `{mode: 'from-generated-index'}` will make sure the babel loader read from the index file instead of async load them.

I am a bit confuse by this statement thou `React Native doesn’t use webpack, and given app size is a limiting factor on mobile, it would not be beneficial to have 3 different versions of the application.`

Because this to me sound like you do want language specific build, you just need to make sure to load the French app for French user.

GoodForOneFare:
`await generateTranslationDictionaries(['en', 'fr', 'es'])` should embed multiple languages, but I don't think it has ever been used in production.

perpetuatingcuriosity:
> Because this to me sound like you do want language specific build, you just need to make sure to load the French app for French user.

On mobile we would embed all the translations into the app. We have a single build and then make sure to load the French translations for the french user – the rest of the app would stay the same. 

Will using `from-dictionary-index` produce multiple builds of our app – one for each locale? I might be misunderstanding sth. 

Re: webpack: React Native uses [metro](https://facebook.github.io/metro/) to bundle js.

michenly:
the `await generateTranslationDictionaries(['en', 'fr', 'es'])` @GoodForOneFare mentioned above should included all three languages in one app. But like Gord mentioned, it had never been use in production before in this matter and you should test it.

macdonaldr93:
Okay that's good to know! And yeah, the metro bundler is part of the problem... It doesn't support dynamic imports so we can't use the automatically generated version of this plugin.

I guess all the languages are loaded at once and then selectively shown, instead of only loading the relevant language. That might be problematic but I can't see a way of resolving this without dynamic imports.

macdonaldr93:
Updated and rebased. Let me know if this is good to go.

PR #1417, opened 2020-05-05
Update dom mocks setup docs

ddayguerrero:
## Description

Fixes documentation

Documentation change: Updating the setup instructions for our jest mocking utilities. Currently, we recommend invoking `ensureMocksReset` in the `beforeEach` setup. Instead, let's instruct developers to use it in the `afterEach` setup.

**Why?**
Based on the `mock` in before, `clean up` after testing principle.

Our implementation even [hints](https://github.com/Shopify/quilt/blob/master/packages/jest-dom-mocks/src/index.ts#L69) at resetting after tests.

**Notes**

Open for discussion! 😇

**Affected package**: `jest-dom-mocks`

## Type of change

## Checklist

### DIFF
--- a/packages/jest-dom-mocks/CHANGELOG.md
+++ b/packages/jest-dom-mocks/CHANGELOG.md
@@ -7,6 +7,12 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## [Unreleased] -->
 
+## [2.9.1] - 2020-05-05
+
+### Added
+
+- Fixed setup instructions instructions for `ensureMocksReset`
+
 ## [2.9.0] - 2020-04-23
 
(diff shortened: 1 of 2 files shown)


packages/jest-dom-mocks/CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## [Unreleased] -->
 
+## [2.9.1] - 2020-05-05
```
GoodForOneFare, line 10: I think Quilt packages avoid version bumps for doc only changes.  Because it causes a lot of unnecessary version bumps in consumers.

alexandcote: Yes, you can probably remove the entry in the changelog and simply update the doc.

Issue #1370, opened 2020-04-14
Generate package script doesn't update packages/tsconfig.json


posth:
## Overview

When running `yarn generate:package` to create a new package, it doesn't automatically update the `packages/tsconfig.json` file with a new entry to the path of the created package.

### Scope

Related to all new packages.

---

## Checklist

- [ ] Please delete the labels section before submitting your issue
- [ ] I have described this issue in a way that is actionable (if possible)

PR #1418, opened 2020-05-05
Generator / Update global tsconfig when new package is added

loic-d:
## Description
This PR fixes https://github.com/Shopify/quilt/issues/1370

When creating a new package with `yarn generate:package`, we were not properly adding the new package's path to the global `tsconfig.json`. This had to be done manually.

🎩 Instructions:
- Generate a new package with `yarn generate:package`.
- Give your package a name (`foo` for example) and answer the other questions.
- Open `packages/tsconfig.json` and make sure `{"path": "./foo"},` has been added to the list of references (alphabetically). 

## Type of change

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "release-beta": "lerna publish --dist-tag beta",
     "clean": "rimraf './packages/*/dist'",
     "generate": "plop",
-    "generate:package": "plop package && yarn plop docs",
+    "generate:package": "plop package && yarn plop docs && prettier --write \"./packages/tsconfig.json\"",
     "tophat": "tophat"
   },
   "repository": {
@@ -72,7 +72,7 @@
     "jest-enzyme": "^7.1.2",
     "jest-transform-graphql": "^2.1.0",
     "lerna": "^2.9.0",
-    "plop": "^2.0.0",
+    "plop": "^2.6.0",
     "prettier": "~2.0.4",
     "puppeteer": "^1.20.0",
     "react": "16.12.0",
--- a/plopfile.js
+++ b/plopfile.js
@@ -4,7 +4,7 @@ const path = require('path');
 const jsPackageNames = getPackageNames('js');
 const gemNames = getPackageNames('ruby');
 
-module.exports = function(plop) {
+module.exports = function (plop) {
   plop.setGenerator('package', {
     description: 'create a new package from scratch',
     prompts: [
@@ -60,6 +60,19 @@ module.exports = function(plop) {
         path:
           'packages/{{kebabCase name}}/src/test/{{properCase name}}.test.ts',
         templateFile: 'templates/test.hbs.ts',
+      },
+      {
+        type: 'modify',
+        path: 'packages/tsconfig.json',
+        transform: (file, {name}) => {
+          const tsConfig = JSON.parse(file);
+          const kebabCase = plop.getHelper('kebabCase');
+          tsConfig.references.push({path: `./${kebabCase(name)}`});
+          tsConfig.references.sort(({path: firstPath}, {path: secondPath}) =>
+            firstPath.localeCompare(secondPath),
+          );
+          return JSON.stringify(tsConfig);
+        },
       },
     ],
   });
--- a/yarn.lock
+++ b/yarn.lock
@@ -1216,18 +1216,26 @@
     "@types/yargs" "^15.0.0"
     chalk "^3.0.0"
 
-"@mrmlnc/readdir-enhanced@^2.2.1":
-  version "2.2.1"
-  resolved "https://registry.yarnpkg.com/@mrmlnc/readdir-enhanced/-/readdir-enhanced-2.2.1.tgz#524af240d1a360527b730475ecfa1344aa540dde"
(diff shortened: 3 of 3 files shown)


plopfile.js
```diff
@@ -4,7 +4,7 @@ const path = require('path');
 const jsPackageNames = getPackageNames('js');
 const gemNames = getPackageNames('ruby');
 
-module.exports = function(plop) {
+module.exports = function (plop) {
```
loic-d, line 7: Not sure why VS Code is auto-formatting on save like this 🤔

plopfile.js
```diff
@@ -61,6 +61,19 @@ module.exports = function(plop) {
...
+        type: 'modify',
+        path: 'packages/tsconfig.json',
+        transform: (file, {name}) => {
+          const tsConfig = JSON.parse(file);
+          tsConfig.references.push({path: `./${name}`});
+          tsConfig.references = tsConfig.references.sort(
+            ({path: firstPath}, {path: secondPath}) =>
+              firstPath.localeCompare(secondPath),
+          );
+          return JSON.stringify(tsConfig);
```
loic-d, line 74: `transform` must return a string (or a `Promise` resolving to a string)

package.json
```diff
@@ -72,7 +72,7 @@
     "jest-enzyme": "^7.1.2",
     "jest-transform-graphql": "^2.1.0",
     "lerna": "^2.9.0",
-    "plop": "^2.0.0",
+    "plop": "^2.6.0",
```
loic-d, line 75: `transform` function wasn't supported in `2.0.0`

package.json
```diff
@@ -17,7 +17,7 @@
     "release-beta": "lerna publish --dist-tag beta",
     "clean": "rimraf './packages/*/dist'",
     "generate": "plop",
-    "generate:package": "plop package && yarn plop docs",
+    "generate:package": "plop package && yarn plop docs && prettier --write \"./packages/tsconfig.json\"",
```
loic-d, line 20: Run prettier on the updated `/packages/tsconfig.json`

plopfile.js
```diff
@@ -61,6 +61,19 @@ module.exports = function(plop) {
           'packages/{{kebabCase name}}/src/test/{{properCase name}}.test.ts',
         templateFile: 'templates/test.hbs.ts',
       },
+      {
+        type: 'modify',
+        path: 'packages/tsconfig.json',
+        transform: (file, {name}) => {
```
loic-d, line 67: `file` is `string`.

plopfile.js
```diff
@@ -61,6 +61,19 @@ module.exports = function(plop) {
...
+          const tsConfig = JSON.parse(file);
+          tsConfig.references.push({path: `./${name}`});
```
alexandcote, line 69: I think the package name needs to be kebab case. If you write the name correctly like `react-foo` this will work. But if you do `react foo` the package directory will be named `react-foo` and the tsconfig will add `react foo` as reference.

Maybe we don't want to support that 🤷🏼‍♂️

yarn.lock
```diff
@@ -1216,18 +1216,26 @@
     "@types/yargs" "^15.0.0"
     chalk "^3.0.0"
 
-"@mrmlnc/readdir-enhanced@^2.2.1":
-  version "2.2.1"
-  resolved "https://registry.yarnpkg.com/@mrmlnc/readdir-enhanced/-/readdir-enhanced-2.2.1.tgz#524af240d1a360527b730475ecfa1344aa540dde"
-  integrity sha512-bPHp6Ji8b41szTOcaP63VlnbbO5Ny6dwAATtY6JTjh5N2OLrb5Qk/Th5cRkRQhkWCt+EJsYrNB0MiL+Gpn6e3g==
+"@nodelib/fs.scandir@2.1.3":
```
alexandcote, line 1219: When I pull your branch and ran `yarn` the `yarn.lock` file changed.

package.json
loic-d: I think it makes sense. This way we don't rely on our committed templates being properly prettified.

plopfile.js
loic-d: Good point. Now forcing kebab case with `plop.getHelper('kebabCase')`.

package.json
loic-d: 👀  a bit more and you don't have access to the package's name here.

yarn.lock
loic-d: Both @mathieulag and I don't have a local change when running `yarn` 🤔 
Can you try again after deleting the repo? Thanks!

plopfile.js
```diff
@@ -61,6 +61,20 @@ module.exports = function(plop) {
...
         templateFile: 'templates/test.hbs.ts',
       },
+      {
+        type: 'modify',
+        path: 'packages/tsconfig.json',
+        transform: (file, {name}) => {
+          const tsConfig = JSON.parse(file);
+          const kebabCase = plop.getHelper('kebabCase');
+          tsConfig.references.push({path: `./${kebabCase(name)}`});
+          tsConfig.references = tsConfig.references.sort(
```
mathieulag, line 71: `sort` is [in-place](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort), so no need to use the return value.

PR #1423, opened 2020-05-07
✨💥  pass data, url to app component by default in react-server-webpack-plugin

### LINKED ISSUE #1115: `react-server-webpack-plugin` should pass URL for location on client and server
Right now, we pass in `location={ctx.request.url}` to the application during SSR as part `@shopify/react-server-webpack-plugin`. There are four problems with this:

1. It probably won't line up with what our own react router will accept
2. It is a pain to parse information out of
3. It does not contain any information about the origin of the request
4. It is not automatically passed down during the client render, which may be unexpected

I think the best way to resolve this is to:

1. Pass `url={ctx.request.URL}` instead, which is a proper `URL` object
2. Pass in this value on the client also as a convenience

Any thoughts @TheMallen @michenly?

TheMallen:
Closes #1115

### What do

This PR revamps what we pass into the user's app in our react-server plugin.  It is a breaking change, and since we 

In detail it:
- Moves `x-quilt-data` header parsing into it's own middleware
- Changes the name of the `x-quilt-data` serialization key to just `quilt-data` (no need for x when it is no longer a header)
- Passes the full WHATWG `URL` object into the app component in `@shopify/react-server-webpack-plugin` as an automatic prop
- Deserializes and passes `quilt-data` into the app component as well
- Updates relevant documentation

### Tophatting 🎩

It's tough to write sane tests for this, but I tophatted this pretty thoroughly using `hydrox-again` following this method:
- checkout this branch
- `yarn build`
- `cp ./packages/react-server ../hydrox-again/node_modules/@shopify`
- `cp ./packages/react-server-webpack-plugin ../hydrox-again/node_modules/@shopify`
- `cp ./packages/react-server-webpack-plugin ../hydrox-again/node_modules/@shopify/sewing-kit/node_modules`
- change the `App.tsx` to use the new props
- run the app

I also prepared this [branch](https://github.com/Shopify/hydrox-again/pull/15) that makes all the changes necessary to get this working and also removes some cruft. Once this is actually released we can use this branch to actually install the new version.

### DIFF
--- a/packages/react-server-webpack-plugin/CHANGELOG.md
+++ b/packages/react-server-webpack-plugin/CHANGELOG.md
@@ -9,8 +9,40 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ### Changed
 
-- Generated entrypoints no longer render the App component with a `location` prop. Apps can instead get the same data from the new `url` prop's `href` attribute.
-- Generated entrypoints no longer render the `App` component with a `server` prop. Whether an app is rendered on the server can instead be trivially inferred from `typeof window === 'undefined'`
+- ⚠️ Generated entrypoints no longer render the App component with a `location` prop. Apps can instead get the same data from the new `url` prop's `href` attribute.
+- ⚠️ Generated entrypoints no longer render the `App` component with a `server` prop. Whether an app is rendered on the server can instead be trivially inferred from `typeof window === 'undefined'`
+
+Apps which were using the `location` prop will need to use the `url` prop instead. Where before an app component might look like this
+
+```tsx
+export function App({location}: {location: string}) {
+  return (
+    <div className={styles.App}>
+      <Router location={location}>
+        <Frame>
+          <Routes />
+        </Frame>
+      </Router>
+    </div>
+  );
+}
+```
+
+It would now look like this
+
+```tsx
+export function App({url}: {url: URL}) {
+  return (
+    <div className={styles.App}>
+      <Router location={url}>
+        <Frame>
+          <Routes />
+        </Frame>
+      </Router>
+    </div>
+  );
+}
+```
 
 ### Added
 
--- a/packages/react-server-webpack-plugin/CHANGELOG.md
+++ b/packages/react-server-webpack-plugin/CHANGELOG.md
@@ -5,7 +5,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## [Unreleased] -->
+## [Unreleased]
+
+### Changed
+
+- Generated entrypoints no longer render the App component with a `location` prop. Apps can instead get the same data from the new `url` prop's `href` attribute.
+- Generated entrypoints no longer render the `App` component with a `server` prop. Whether an app is rendered on the server can instead be trivially inferred from `typeof window === 'undefined'`
+
+### Added
+
+- Generated entrypoints now pass a `data` prop to the `App`. This prop contains the deserialized data from the `x-quilt-data` header which is used by `quilt_rails` to pass data directly to React
+- Generated entrypoints now pass a `url` prop to the `App`. This prop is a full WHATWG compliant `URL` object representing the url for the request that react-server responded to.
 
 ## [2.2.17] - 2019-11-29
 
--- a/packages/react-server-webpack-plugin/README.md
+++ b/packages/react-server-webpack-plugin/README.md
@@ -95,7 +95,7 @@ The plugin passes some props to your application automatically.
 ```typescript
 interface DefaultProps {
   /*
-   The full WHATWG URL object for the initial request
+   The full WHATWG URL object for the initial request. On the client this will *not* update reactively. It is always the URL for the initial request.
   */
   url: URL;
 
--- a/packages/react-server-webpack-plugin/README.md
+++ b/packages/react-server-webpack-plugin/README.md
@@ -15,7 +15,7 @@ $ yarn add @shopify/react-server-webpack-plugin
 
 ### With sewing-kit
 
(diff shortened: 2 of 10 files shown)


packages/react-server-webpack-plugin/src/react-server-webpack-plugin.ts
```diff
@@ -110,10 +116,15 @@ function clientSource() {
     ${HEADER}
     import React from 'react';
     import ReactDOM from 'react-dom';
-    import {showPage} from '@shopify/react-html';
-    import App from 'index';
+    import {showPage, getSerialized} from '@shopify/react-html';
+
+    import App from './index';
```
michenly, line 121: this line is failing for me as it cannot find `./index`

packages/react-server-webpack-plugin/src/react-server-webpack-plugin.ts
```diff
@@ -110,10 +116,15 @@ function clientSource() {
...
+
     const appContainer = document.getElementById('app');
-    ReactDOM.hydrate(React.createElement(App), appContainer);
+    const data = getSerialized('quilt-data');
+    const url = new URL(window.location.href);
+
+    ReactDOM.hydrate(React.createElement(App, {data, url: url}), appContainer);
```
michenly, line 127:
```suggestion
    ReactDOM.hydrate(React.createElement(App, {data, url}), appContainer);
```

packages/react-server-webpack-plugin/src/react-server-webpack-plugin.ts
TheMallen: 🤔  I thought I had left this as 'index' instead of '/index'm if you change it back does it work?

packages/react-server-webpack-plugin/README.md
```diff
@@ -88,6 +88,52 @@ started react-server on localhost:PORT
...
+// index.jsx
+import React from 'react';
+
+export default function App({url, data}: {url: Url, data: Record<string, any>}) {
+  const {href} = url;
+  if (href.endsWith('/about')) {
+    return <div>this is an about page</div>;
+  }
+
+  return (
```
michenly, line 124: running into a small integration issue with react-router
```
import {Router} from '@shopify/react-router';
<Router location={url}>
```
👆 this fail type check because `location` is typed `location?: string;`

we can use `url.href` or change the type to `location?: string | URL;` and everything would still work

packages/react-server-webpack-plugin/src/react-server-webpack-plugin.ts
michenly: yup changing it back to `index` works fine.

michenly:
A thought I am having as I integrate it with the magic provider.
I wonder if we should just has `@shopify/react-router` build into `@shopify/react-server` ?

Once we have `@shopify/react-router` provider setup, presumably whenever consumer want url information they can call the [various hooks react-router now has](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/hooks.md#usehistory) to use it. Unless there are a lot of app we want to support not using react-router?

packages/react-server-webpack-plugin/README.md
TheMallen: url.href is the right choice, that's what I did in the branch I linked

packages/react-server/src/render/render.tsx
```diff
@@ -77,14 +78,14 @@ export function createRender(render: RenderFunction, options: Options = {}) {
...
-      const [, Serialize] = useSerialized<Data>('x-quilt-data');
+      const [, Serialize] = useSerialized<Data>('quilt-data');
 
       return (
         <AsyncAssetContext.Provider value={asyncAssetManager}>
           <HydrationContext.Provider value={hydrationManager}>
             <NetworkContext.Provider value={networkManager}>
               {children}
-              <Serialize data={() => ctx.headers['x-quilt-data']} />
+              <Serialize data={() => ctx.state.quiltData} />
```
ismail-syed, line 88: I don't see a changelog update for this so I presume it doesn't need one. No user impact so makes sense.

packages/react-server-webpack-plugin/src/react-server-webpack-plugin.ts
ismail-syed: Yah, I recall that the webpack plugin needed `index` instead of `./index`. It was a weird thing I explored/debugged ~8-10 months ago when we first started the plugin.

packages/react-server-webpack-plugin/CHANGELOG.md
```diff
@@ -5,7 +5,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## [Unreleased] -->
+## [Unreleased]
+
+### Changed
+
+- Generated entrypoints no longer render the App component with a `location` prop. Apps can instead get the same data from the new `url` prop's `href` attribute.
```
ismail-syed, line 12: This would constitute a breaking change. Can we highlight this more with some ⚠️, maybe a code snippet comparison?

packages/react-server-webpack-plugin/README.md
michenly: Yes, I understand that I can use `url.href`. The point I was trying to make it we don't have to.

Our `Router` only use `location` to pass to `react-router`'s `StaticRouter`

Their doc show example of the following

```
<StaticRouter location={req.url} context={context}>
        <App />
</StaticRouter>
```

which mean with a type change `Router` can accept object as well as string.

TheMallen:
@ismail-syed probably the relevant spot for more migration explanation would be sewing-kit since it will be hooking up the plugin?

ismail-syed:
> the relevant spot for more migration explanation would be sewing-kit 

Good point. Publishing this as a breaking change would address all the concerns I had in mind.

packages/react-server-webpack-plugin/README.md
TheMallen: ahhh, sorry I misunderstood your comment. Seems like it makes sense to update the type indeed then!

Issue #1426, opened 2020-05-08
[koa-shopify-auth] 3.1.61 breaks support for prefix on safari


kavika-1:
## Overview

Thanks for v3.1.61 to address safari issues with cookies and storage.  However, when using shopifyAuth "prefix", the safari browser 404s on safari 13.1 (and likely other versions).

This is because the redirect always goes to "/auth/..." without the original prefix.  By prefix, we mean we use the "prefix" in the example provided for koa-shopify-auth...
````
app.use(
  shopifyAuth({
    // if specified, mounts the routes off of the given path
    // eg. /shopify/auth, /shopify/auth/callback
    // defaults to ''
    prefix: '/shopify',
````
So in the example above, on safari, 3.1.61 redirects to "/auth/enable_cookies..." instead of "/shopify/auth/enable_cookies...", and thus shopifyAuth doesn't serve that non-prefixed route.

The trick is that these redirects are now are hard-coded client-side.  For instance, in client/request-storage-access.ts:
````
    var targetInfo = {
          myshopifyUrl: "https://${shop}",
          hasStorageAccessUrl: "/auth/inline?shop=${shop}",
          doesNotHaveStorageAccessUrl: "/auth/enable_cookies?shop=${shop}",
````

It would be possible to derive the prefix client-side from the window.location.path.  However, hopefully there is a better or more robust way to get the prefix server-side, and have the server inject that into the client code snippet.

eugene-kim:
I think this is a duplicate of this issue: https://github.com/Shopify/quilt/issues/1148

kavika-1:
Thanks for pointing out [#1148].  Definitely good to be aware of that, as it is similar.  I believe in this case it is a new set of code that basically ended up with the same issue, e.g. "Respect `prefix`...".  Will post a PR for review.

Here's a community thread on this, https://community.shopify.com/c/Shopify-APIs-SDKs/Safari-13-1-and-embedded-apps/td-p/688416/highlight/false/page/3#, suggesting where to do the work.

Issue #1148, opened 2019-10-25
Respect `prefix` in enable cookie route


timolins:
## Overview

The redirect introduced with #1022 doesn't apply the prefix. This means all requests get forwarded to `/auth` instead of `/my-prefix/auth`. 

## Consuming repo

> What repo were you working in when this issue occurred?

koa-shopify-auth

scotttrinh:
@timolins 

I think you need to sign the CLA to get your PR merged. I'm running into this issue now, too, and would love to help get this over the finish line.

timolins:
I think you a right, but I wasn't able to find more information about this. They should probably add a few words about this in [`CONTRIBUTING.md`](https://github.com/Shopify/quilt/blob/master/.github/CONTRIBUTING.md).

scotttrinh:
@timolins 

If you click through the details on the failed check I think it'll walk you through it, but I haven't done it so it's a bit opaque to me, too. Definitely worth mentioning in the contribution guide!

timolins:
Thanks for the hint! Just signed the CLA and the bot is happy now.

eugene-kim:
I'm running into this issue as well and it's great see that there's a fix ready to go. Is there anything preventing the PR from being merged besides the review and the branches being out of sync?

scotttrinh:
I have a fork with @timolins' fix merged published to npm while we're waiting if anyone here cares to use it. As soon as this lands in quilt, I plan on deprecating it, so if you're like me and just need to get it working, [@cala/koa-shopify-auth](https://www.npmjs.com/package/@cala/koa-shopify-auth) can hold you over for now.

junaid33:
Facing the same issue. @shopify/koa-shopify-auth works sometimes and sometimes it says page not found. This issue is causing our application to be rejected from the Shopify App Store. The package is updated but we are using a prefix.

TheMallen:
I believe this was fixed in #1440. Feel free to reopen if the issue persists.

dooman87:
@TheMallen I believe this issue has been fixed only partially. I created a PR #1498 which should finally fix this. Appreciate if you could have a look.

PR #1428, opened 2020-05-10
[koa-shopify-auth] add prefix support post pull [#1413]

kavika-1:
## Description

Fixes [#1148][#1426]

Apply prefix in koa-shopify-auth to various Safari/ITP redirects.

[#1413] addresses several issues with Safari by bringing in some code from `shopify_app`.  However, that code was not `koa-shopify-auth` `prefix` aware.  This small set of changes helps address this.

NOTE: There are similar PRs [#1149] and [#1196], hopefully this simply addresses the same issue post [#1413].

Quick discusion: While most of the redirects are related to prefixing "/auth" routes, there is one additional target, `appTargetUrl`, to which the prefix could be added to as well, since otherwise, upon client side redirect, the server has no idea where this came from.  Note that the referer (sic) header is now blocked on safari, and even with the Referrer-Policy set, only the origin is provided not the path, so even that method cannot be used to infer the prefix upon return.

## Type of change

## Checklist


packages/koa-shopify-auth/src/auth/client/request-storage-access.ts
```diff
@@ -1,12 +1,12 @@
...
   return `(function() {
       function redirect() {
         var targetInfo = {
           myshopifyUrl: "https://${shop}",
-          hasStorageAccessUrl: "/auth/inline?shop=${shop}",
-          doesNotHaveStorageAccessUrl: "/auth/enable_cookies?shop=${shop}",
-          appTargetUrl: "/?shop=${shop}"
+          hasStorageAccessUrl: "${prefix}/auth/inline?shop=${shop}",
+          doesNotHaveStorageAccessUrl: "${prefix}/auth/enable_cookies?shop=${shop}",
+          appTargetUrl: "${prefix}/?shop=${shop}"
```
ragalie, line 9: I don't think we should set `prefix` here. The path that the application lives on is orthogonal to the paths that the OAuth logic lives on. (You can tell this because the `verifyRequest` middleware can be applied separately from the `shopifyAuth` middleware.)

In a separate PR, however, I'd be open to a couple alternatives:
1. A new `applicationUrl` configuration option that would specify the path to direct the user when cookie access is enabled.
2. Logic to pass a `returnTo` value through query params. (We do this in `shopify_app`, though it's pretty complex IIRC.)

kavika-1: Thanks for the feedback; I'll update the PR shortly

kavika-1: Ok, appTargetUrl prefix undone.

simonhaenisch:
I've been getting frustrated with koa-shopify-auth and the speed that fixes are made... but even more frustrating is that real issues are left uncommented for weeks, and community PRs take so long to be reviewed or merged.

So I created a fork and I did the same fix already 11 days ago:

https://github.com/simonhaenisch/koa-shopify-auth/commit/08f2c56241bc50d2b7e807359e29138d1488c3da

(except some whitespace and the missing test, it's actually the exact same set of changes 😄)

I can confirm that this fix works with the 3 different browsers (Chrome, Firefox, Safari) I tested with, using different cookie settings.

ragalie:
@simonhaenisch I'm sorry that working with us and the library has been frustrating :( We're definitely trying to do a better job maintaining this library. I hope that we can earn back your trust and support, starting by getting this fix into the mainline. Thanks for the feedback on this.

kavika-1:
@ragalie, i signed the CLA a few days back, and if i try to sign again, it says "To sign our CLA 1 field needs to be changed / GitHub username has already signed a CLA".  also, for some reason it shows a strange email address synthesized from my local network address, instead of my official public email.  any way to restart that process?

otherwise, will work on item 2.

TheMallen:
@kavika-1 I suspect the CLA check tries to use the email from the *commits* themselves which would explain why it is showing two versions of you... I'd double check that `git config --get user.email` returns what you expect 🤔 

@ragalie not sure how we should proceed if the CLA but is not working after he checks that :/

kavika-1:
@TheMallen Darn, indeed.  I set up a new macos environment, and indeed the user.email (and name) are not set.  But don't know how to fix the past?

TheMallen:
@kavika-1  I had a similar problem when I refromated my macbook recently, so I figured :P Sorry for all the trouble.

You should be able to do some history rewriting against your branch / clone (it will let you force push to your branch just not master). Github has a [little guide ](https://help.github.com/en/github/using-git/changing-author-info) for how to do it actually.

If it doesn't work for you I can do it for you, but I'd need to know what email to change it to and to make sure I have your permission as I wouldn't want to mess around with the history of your contributions lightly.

kavika-1:
@TheMallen grateful for your experience, i'll give that rewrite a try!

kavika-1:
@TheMallen sorry seems i didn't quite get those instructions to work properly, and it rewrote the whole history?  sorry 'bout that.  i am not sure how to revert that change if possible at all, so i'll go to plan "B" and create a new branch/pull request unless you have a magic command to undo that :/

TheMallen:
@kavika-1  sorry didn't get back to you in time, but I see you have a new branch that is passing CLA, so thanks again for the hard work :)

kavika-1:
@TheMallen thanks for your assistance and encouragement, The hard work was self-inflicted!

PR #1434, opened 2020-05-13
[enzyme-utilities] Trigger now always return the value

alexandcote:
## Description
`trigger` doesn't return the value returned by the promise. We now return the value even if it's a promise.

## Type of change

## Checklist


packages/enzyme-utilities/src/index.ts
```diff
@@ -50,9 +50,9 @@ export function trigger(wrapper: AnyWrapper, keypath: string, ...args: any[]) {
   });
 
   if (isPromise(returnValue)) {
-    return Promise.resolve(promise as Promise<any>).then(ret => {
+    return Promise.resolve(promise).then(() => {
```
TheMallen, line 53: I don't get why we wrap it if we aren't using the resolved value

TheMallen: After chatting in slack this makes sense

PR #1437, opened 2020-05-13
Updates usePrefetch options

hilaryml:
## Description

This pr updates the generic for `usePrefetch` to allow `fetchPolicy` to be passed in `options`.

This is done to support a web pr that introduced a `cache-first fetch policy` for prefetch rather than using the `network-only` default: https://github.com/Shopify/web/pull/26887. 

## Type of change

- [x] `react-graphql` [Patch: Bug/ Documentation fix (non-breaking change which fixes an issue or adds documentation)

## Checklist

### DIFF
--- a/packages/react-graphql/src/async/query.ts
+++ b/packages/react-graphql/src/async/query.ts
@@ -22,7 +22,9 @@ export function createAsyncQuery<Data, Variables, DeepPartial>({
     return useAsync(resolver, {assets: AssetTiming.NextPage}).load;
   }
 
-  function usePrefetch(options: VariableOptions<Variables>) {
+  function usePrefetch(
+    options: VariableOptions<Variables> & Pick<QueryProps, 'fetchPolicy'>,
+  ) {
     const load = usePreload();
     return useBackgroundQuery(load, options);
   }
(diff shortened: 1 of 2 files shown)


packages/react-graphql/src/async/query.ts
```diff
@@ -22,7 +22,9 @@ export function createAsyncQuery<Data, Variables, DeepPartial>({
     return useAsync(resolver, {assets: AssetTiming.NextPage}).load;
   }
 
-  function usePrefetch(options: VariableOptions<Variables>) {
+  function usePrefetch(
+    options: VariableOptions<Variables> & Pick<QueryProps, 'fetchPolicy'>,
```
hilaryml, line 26: wasn't sure whether it would be better to do `options: VariableOptions<Variables> & Omit<WatchQueryOptions, 'query'>` instead.

lemonmade: I think it's fine to start here, everyone was fine with no options for a long time, and I bet they will be happy with just this one extra option for a long time from now :)

hilaryml: Ok great. Thanks Chris!

PR #1439, opened 2020-05-13
[enzyme-utilities] trigger now update the wrapper even if you don't await the promise

alexandcote:
## Description
`trigger` doesn't update the wrapper if the callback returns a promise and you don't resolve this promise. Now `trigger` update the wrapper before the promise is resolved and after the promise is resolved.

## Type of change

## Checklist

### DIFF
--- a/packages/enzyme-utilities/src/index.ts
+++ b/packages/enzyme-utilities/src/index.ts
@@ -49,6 +49,8 @@ export function trigger(wrapper: AnyWrapper, keypath: string, ...args: any[]) {
     }
   });
 
+  updateRoot(wrapper);
+
   if (isPromise(returnValue)) {
     // `promise` here refer to the the `act` promise above.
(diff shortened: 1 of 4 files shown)


packages/enzyme-utilities/src/index.ts
```diff
@@ -49,6 +49,8 @@ export function trigger(wrapper: AnyWrapper, keypath: string, ...args: any[]) {
     }
   });
 
+  updateRoot(wrapper);
```
TheMallen, line 52: baha, this makes a tonne of sense. Would you mind contributing it to https://github.com/Shopify/preact-testing as well? :P

alexandcote: You already do it 🎉: https://github.com/Shopify/preact-testing/blob/f6025a84ba067b7e8def28383ef6d2739162d21f/src/root.tsx#L172

PR #1441, opened 2020-05-14
Add useForceUpdate hook

tleunen:
## Description

We have a valid usecase of a `useForceUpdate` in my project so I'm adding it here for others to use if they need to. This shouldn't be recommended tho, but sometimes there're no other choices.

## Type of change

- [x] `react-hooks` Minor: New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/packages/react-hooks/src/hooks/force-update.ts
+++ b/packages/react-hooks/src/hooks/force-update.ts
@@ -1,0 +1,6 @@
+import {useState, useCallback} from 'react';
+
+export function useForceUpdate() {
+  const [, setState] = useState({});
+  return useCallback(() => setState({}), []);
+}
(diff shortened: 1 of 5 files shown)


packages/react-hooks/src/hooks/force-update.ts
```diff
@@ -0,0 +1,6 @@
+import {useState, useCallback} from 'react';
+
+export function useForceUpdate() {
```
alexandcote, line 3: Can we simply do this: 
```
const [,setState] = useState();
return setState;
```

tleunen: Answered on Slack, but copying my reply here for transparency.

`setState(undefined)` wouldn't work because we absolutely need to provide a new value to trigger a new render. At best, as explained by this issue https://github.com/facebook/react/issues/17474, setting the state with the same value might trigger en 2nd render, but not more than that because the component is not even supposed to rerender when the value is the same.

Another way to write it is with `useReducer`, but I thought it would be best to keep it simple with `useState`.
```
const [, forceUpdate] = useReducer(() => ({}), {});
return forceUpdate;
```

WDYT? Prefer the reducer way?

PR #1442, opened 2020-05-14
Accept Firefox params

### LINKED ISSUE #1430: quilt_rails - accept Firefox metrics that do not contain connection info
## Overview
`Quilt::Performance::Report` currently rejects request data that contains an empty `connection` value:
https://github.com/Shopify/quilt/blob/e2f834b77009b91db37d526442ad0febd7855bcd/gems/quilt_rails/lib/quilt_rails/performance/report.rb#L10-L12

As [MDN's `Navigator.connection` docs](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/connection) suggest, this causes most Firefox requests to be rejected because:

> This feature is behind the dom.netinfo.enabled preference (needs to be set to true). To change preferences in Firefox, visit about:config

## Motivation
We'd like to accept stats for all browsers 😅 

## Workaround

Right now, I'm patching quilt_rails with:

```rb
require 'quilt_rails'

module Quilt
  module Performance
    module Reportable
      def process_report(&block)
        # quilt_rails throws an exception if connection is empty, but Firefox
        # defaults to not sending connection information.
        # See more at the Firefox's flag note in:
        # https://developer.mozilla.org/en-US/docs/Web/API/Navigator/connection
        report_params[:connection] = {effectiveType: 'unknown'} unless report_params[:connection].present?

        Client.send!(Report.from_params(report_params), &block)
      end
    end
  end
end
```

b-ryu:
## Description

Fixes #1430 

Firefox doesn't attach `connection` parameters to performance data by default, causing the endpoint to reject the request, since `connection` is a required param.

Changed it so that `connection` is no longer a required param, and the value defaults to `{effectiveType: 'unknown'}`.

Updated tests and changelog. Tophatted changes with the following steps:

1. Check out this branch in `quilt`
1. `dev cd shrink-ray`
2. In `Gemfile`, replace `quilt_rails` entry with `gem "quilt_rails", path: <relative_path_to_local_quilt_rails>`
3. `dev up && dev run`
4. Install Firefox if you don't already have it
5. Navigate to `https://shrink-ray.myshopify.io/repos/generated/` in Firefox
6. Open the network tab in the console and verify that the performance report request is a `success`, and that your request body had an empty `connection` parameter (see screenshots)
7. Verify that, in the terminal, you can see the logged distributions and that `browser_connection_type` is `unknown`

## Screenshots

| Before | After |
| --- | --- |
| Screen Shot 2020-05-14 at 12 31 53 PM |  Screen Shot 2020-05-14 at 1 04 49 PM |

In development Shrink-Ray, you can also verify the distributions logged to the terminal:

Screen Shot 2020-05-14 at 1 39 05 PM
 

## Type of change

## Checklist

### DIFF
--- a/gems/quilt_rails/CHANGELOG.md
+++ b/gems/quilt_rails/CHANGELOG.md
@@ -7,9 +7,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 <!-- ## [Unreleased] -->
 
+## [1.12.1] - 2020-05-13
+
+- Fixed: Performance endpoint now accepts data that lacks a `connection` parameter, replacing it with a stand-in value
+
 ## [1.12.0] - 2020-05-07
 
-- Add: Expose a `data` arugment on `render_react` to share data to the React server using the `X-Quilt-Data` header ([#1411](https://github.com/Shopify/quilt/pull/1411))
+- Add: Expose a `data` argument on `render_react` to share data to the React server using the `X-Quilt-Data` header ([#1411](https://github.com/Shopify/quilt/pull/1411))
 
 ## [1.11.1] - 2020-03-24
 
--- a/gems/quilt_rails/test/quilt_rails/performance_reportable_test.rb
+++ b/gems/quilt_rails/test/quilt_rails/performance_reportable_test.rb
@@ -405,15 +405,32 @@ module Quilt
       assert_equal(times_called, 4)
     end
 
-    def test_raises_parameter_missing_errors_when_mandatory_params_are_missing
+    def test_replaces_omitted_connection_param_with_default_value
       @params = ActionController::Parameters.new(
-        navigations: [],
-        events: []
+        'navigations' => [],
+        'events' => [{
+          'type' => 'ttfb',
+          'start' => 2,
+          'duration' => 1000,
+        }, {
+          'type' => 'kitty-cat cuddle',
+          'start' => 100,
+          'duration' => 999999,
+        }],
       )
 
-      assert_raises ActionController::ParameterMissing do
-        process_report
-      end
+      StatsD
+        .expects(:distribution)
+        .with('time_to_first_byte', 2, tags: {
+          browser_connection_type: 'unknown',
+        })
+      StatsD
+        .expects(:distribution)
+        .with('kitty-cat cuddle', 100, tags: {
+          browser_connection_type: 'unknown',
+        })
+
+      process_report
     end
 
     private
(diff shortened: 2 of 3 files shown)


gems/quilt_rails/CHANGELOG.md
```diff
@@ -7,9 +7,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
 <!-- ## [Unreleased] -->
 
+## [1.12.1] - 2020-05-13
+
+- Fixed: Performance endpoint now accepts data that lacks a `connection` parameter, replacing it with a stand-in value
+
 ## [1.12.0] - 2020-05-07
 
-- Add: Expose a `data` arugment on `render_react` to share data to the React server using the `X-Quilt-Data` header ([#1411](https://github.com/Shopify/quilt/pull/1411))
+- Add: Expose a `data` argument on `render_react` to share data to the React server using the `X-Quilt-Data` header ([#1411](https://github.com/Shopify/quilt/pull/1411))
```
ismail-syed, line 16: 😅

gems/quilt_rails/CHANGELOG.md
```diff
@@ -7,9 +7,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 <!-- ## [Unreleased] -->
 
+## [1.12.1] - 2020-05-13
```
ismail-syed, line 10: We usually do the version bumps in the changelogs directly on master unless you plan to deploy this right after. Given that this is just for the gem, this should be ok as is.

b-ryu: So for future PRs, would I leave the changelog untouched, or just omit the version bump? I guess the latter would break the lint rule that's set up

ismail-syed: You omit the version header changes. You add to changelog entries (the line items). The lint rule in quilt runs when we run the release on master.

gems/quilt_rails/test/quilt_rails/performance_reportable_test.rb
```diff
@@ -405,15 +405,32 @@ def test_process_report_calls_distribution_callback_for_each_distribution
...
       @params = ActionController::Parameters.new(
-        navigations: [],
-        events: []
+        'navigations' => [],
+        'events' => [{
+          'type' => 'ttfb',
+          'start' => 2,
+          'duration' => 1000,
+        }, {
+          'type' => 'kitty-cat cuddle',
```
TheMallen, line 416: I like that you kept this adorable test data 😆

PR #1445, opened 2020-05-15
[quilt_rails] Parse plaintext data as JSON

### LINKED ISSUE #1431: quilt_rails - accept JSON metrics in text/plain format
## Overview
When sending metrics using `navigator.sendBeacon(url, JSON.stringify(myStats))`, Firefox uses Content-Type `text/plain`:

Firefox network dev tools view of a metrics request that appears to be sent as JSON content.  However, the request header view shows text/plain content-type.

I would like to discuss if it makes sense to have quilt_rails parse plain as JSON content.

## Motivation

I'm trying to avoid working around this problem on the client side because:
* Workarounds typically involve [using a `Blob`](https://stackoverflow.com/a/47122503) to set the content type, but this [is flagged by Chrome's CORS checks](https://stackoverflow.com/a/52054424)
* [Sending as `FormData`](https://stackoverflow.com/a/59878273) works for the browser, but trips Rails' CORS checks
  * [A suggested workaround is disabling forgery protection](https://github.com/rails/rails/issues/22965#issuecomment-172983268), but I'd prefer not to leave all endpoints unprotected just to get metrics
* More selfishly, any client-side workaround is extra bytes, and I'm trying to keep the overhead of performance metrics low 😅 

## Workaround
For now, I'm patching quilt_rails like this:

```rb
require 'quilt_rails'

module Quilt
  module Performance
    module Reportable
      def process_report(&block)
        # Munge plain content into JSON because:
        # - Firefox's `sendBeacon` always uses Content-Type=text/plain
        # - Chrome throws security errors if explicit content type headers are applied (https://stackoverflow.com/a/45622855)
        # - Sending as form-data trips Rails' HTTP Origin header checks ("didn't match request.base_url")

        report_params = if request.content_type == 'text/plain'
          ActionController::Parameters.new(JSON.parse(request.body.read))
        else
          params.transform_keys! { |key| key.underscore.to_sym }
          report_params = params
        end

        Client.send!(Report.from_params(report_params), &block)
      end
    end
  end
end
```

b-ryu:
## Description

Fixes #1431 

When using certain APIs, Firefox will send data to the performance endpoint with `Content-Type: text/plain`. This PR makes it so we parse any incoming plaintext data as JSON.

**Minor fix:** fixed the changelog date from the previous `quilt_rails` release
**Minor fix:** fixed a link in one of the docs

Tophatted with the following steps:
1. Check out this branch in `quilt`
1. `dev cd shrink-ray`
2. In `Gemfile`, replace `quilt_rails` entry with `gem "quilt_rails", path: <relative_path_to_local_quilt_rails>`
3. `dev up && dev run`
4. Install Firefox if you don't already have it (you could use Chrome but it's a lot easier in Firefox)
5. Navigate to `https://shrink-ray.myshopify.io/repos/generated/` in Firefox
6. Open the network tab in the console and find the request to the performance endpoint
7. Right-click the request to "Edit and Resend", change `Content-Type` from `application/json` to `text/plain`, and resend the request
7. Verify that you get a successful response
8. In the terminal, verify that the distributions are properly logged

## Screenshots
| | Before | After |
| --- | --- | --- |
| cURL commands (`text/plain` and `application/json`) | Screen Shot 2020-05-19 at 10 47 31 AM | Screen Shot 2020-05-19 at 11 04 54 AM |
| Distribution logs (`text/plain`) | Screen Shot 2020-05-19 at 10 58 54 AM | Screen Shot 2020-05-19 at 11 03 14 AM |

## Type of change

- [x] `quilt_rails` Patch: Bug/ Documentation fix (non-breaking change which fixes an issue or adds documentation)

## Checklist

### DIFF
--- a/gems/quilt_rails/CONTRIBUTING.md
+++ b/gems/quilt_rails/CONTRIBUTING.md
@@ -19,7 +19,7 @@ All notable changes should be included in the [`CHANGELOG.md`](CHANGELOG.md).
 
 ## Releasing
 
-**Note:** Gem releases will _not_ publish Node packages. If the gem depends on changes in the Node library, you _must_ [publish a Node release first](.github/contributing.md#releasing).
+**Note:** Gem releases will _not_ publish Node packages. If the gem depends on changes in the Node library, you _must_ [publish a Node release first](documentation/guides/release-and-deploy.md).
 
 1. `cd gems/quilt_rails`
 1. Update `version.rb` and `CHANGELOG.md` to your new desired version
--- a/gems/quilt_rails/CONTRIBUTING.md
+++ b/gems/quilt_rails/CONTRIBUTING.md
@@ -19,7 +19,7 @@ All notable changes should be included in the [`CHANGELOG.md`](CHANGELOG.md).
 
 ## Releasing
 
-**Note:** Gem releases will _not_ publish Node packages. If the gem depends on changes in the Node library, you _must_ [publish a Node release first](documentation/guides/release-and-deploy.md).
+**Note:** Gem releases will _not_ publish Node packages. If the gem depends on changes in the Node library, you _must_ [publish a Node release first](../../documentation/guides/release-and-deploy.md).
 
 1. `cd gems/quilt_rails`
 1. Update `version.rb` and `CHANGELOG.md` to your new desired version
--- a/gems/quilt_rails/CONTRIBUTING.md
+++ b/gems/quilt_rails/CONTRIBUTING.md
@@ -19,7 +19,7 @@ All notable changes should be included in the [`CHANGELOG.md`](CHANGELOG.md).
 
 ## Releasing
 
-**Note:** Gem releases will _not_ publish Node packages. If the gem depends on changes in the Node library, you _must_ [publish a Node release first](../../documentation/guides/release-and-deploy.md).
+**Note:** Gem releases will _not_ publish Node packages. If the gem depends on changes in the Node library, you _must_ [publish a Node release first](../../.github/CONTRIBUTING.md#releasing).
 
 1. `cd gems/quilt_rails`
 1. Update `version.rb` and `CHANGELOG.md` to your new desired version
--- a/gems/quilt_rails/lib/quilt_rails/performance/reportable.rb
+++ b/gems/quilt_rails/lib/quilt_rails/performance/reportable.rb
@@ -4,7 +4,9 @@ module Quilt
   module Performance
     module Reportable
       def process_report(&block)
-        params.transform_keys! { |key| key.underscore.to_sym }
+        if request.content_type == 'text/plain'
+          self.params = ActionController::Parameters.new(JSON.parse(request.body.read))
+        end
         Client.send!(Report.from_params(params), &block)
       end
     end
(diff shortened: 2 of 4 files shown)


gems/quilt_rails/lib/quilt_rails/performance/reportable.rb
```diff
@@ -4,8 +4,12 @@ module Quilt
   module Performance
     module Reportable
       def process_report(&block)
-        params.transform_keys! { |key| key.underscore.to_sym }
```
b-ryu, line 7: We seem to already do this in [`Report.from_params`](https://github.com/Shopify/quilt/blob/e00ab07c32658211eac3c27b31cba87f24d5cfc7/gems/quilt_rails/lib/quilt_rails/performance/report.rb#L11)

gems/quilt_rails/CONTRIBUTING.md
```diff
@@ -19,7 +19,7 @@ All notable changes should be included in the [`CHANGELOG.md`](CHANGELOG.md).
 
 ## Releasing
 
-**Note:** Gem releases will _not_ publish Node packages. If the gem depends on changes in the Node library, you _must_ [publish a Node release first](.github/contributing.md#releasing).
+**Note:** Gem releases will _not_ publish Node packages. If the gem depends on changes in the Node library, you _must_ [publish a Node release first](../../.github/CONTRIBUTING.md#releasing).
```
TheMallen, line 22: Oh oops.

gems/quilt_rails/lib/quilt_rails/performance/reportable.rb
TheMallen: That makes sense.

gems/quilt_rails/test/quilt_rails/performance_reportable_test.rb
```diff
@@ -433,12 +434,47 @@ def test_replaces_omitted_connection_param_with_default_value
...
+      )
+
+      StatsD
+        .expects(:distribution)
+        .with('time_to_first_byte', 2, tags: {
+          browser_connection_type: 'unknown',
+        })
+      StatsD
+        .expects(:distribution)
+        .with('kitty-cat cuddle', 100, tags: {
```
TheMallen, line 461: 🐱 🐈

gems/quilt_rails/lib/quilt_rails/performance/reportable.rb
```diff
@@ -4,8 +4,12 @@ module Quilt
...
-        Client.send!(Report.from_params(params), &block)
+        report_params = if request.content_type == 'text/plain'
+          ActionController::Parameters.new(JSON.parse(request.body.read))
```
TheMallen, line 8: I think the more "rubyist" way to do this would be to pull the massaging here out into it's own private method that looks something like:

```ruby
def normalized_params
  return params unless request.content_type == 'text/plain'
  
  ActionController::Parameters.new(JSON.parse(request.body.read))
end
```

and then use that in the `Client.send` here. Personally, I'm about even on either implementation, but I think what I suggested would be more idiomatic to ruby/rails

PR #1448, opened 2020-05-20
🔧 normalize webpack version ranges to dedupe some stuff

TheMallen:
I ran into some dupes in a `sewing-kit` PR. Noticed that the source was some differing webpack requirements in quilt. I don't believe any of these are necessary to be pinned / tilda-d where they were so I am opening up the versioning a bit and standardizing all of them.

I picked the upper and lower bounds based on the current range of webpack versions we used between quilt and sewing-kit. The E2E tests for the plugins should help make sure the unification of the versions in the quilt lockfile doesn't break anything.

### DIFF
--- a/packages/magic-entries-webpack-plugin/package.json
+++ b/packages/magic-entries-webpack-plugin/package.json
@@ -26,7 +26,7 @@
   "dependencies": {
     "glob": "^7.1.6",
     "glob-to-regexp": "^0.4.1",
-    "webpack": "^4.43.0"
+    "webpack": ">4.25.1, <=4.5.0"
   },
   "files": [
     "dist/*",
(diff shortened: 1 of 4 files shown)


TheMallen:
@alexandcote  I think this is the sustainable fix for the issue with my PR on sk

packages/magic-entries-webpack-plugin/package.json
```diff
@@ -26,7 +26,7 @@
   "dependencies": {
     "glob": "^7.1.6",
     "glob-to-regexp": "^0.4.1",
-    "webpack": "^4.43.0"
+    "webpack": ">4.25.1, <=4.5.0"
```
alexandcote, line 29: Can it be a `peerDependencies`?

PR #1454, opened 2020-05-23
[@shopify/react-i18n] Add i18n.formatUnit()

BPScott:
## Description

This PR is mostly a curiosity and a warning to others that head down this path, and is probably something to come back to in the future rather than something we want to merge now (May 2020).

EDIT update Aug 2021: We've now dropped support for node 10 and are using an up-to-date typescript version. That means I've been able to rebase this PR and act upon the "todo in the future" comments. We're just waiting till our browser support matches the browsers this is available in.

[Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat) added the concept of formatting units (see https://v8.dev/features/intl-numberformat#units for an example).

```js
const formatter = new Intl.NumberFormat('en', {
  style: 'unit',
  unit: 'kilobyte',
});
formatter.format(1.234);
// → '1.234 kB'
```

That sounded handy so I added a `i18n.formatUnit()` helper in the same way that we already have a `i18n.formatCurrency()` helper.

This looks like this is supported in all browsers we support with the exception of older safari versions - [it was added in Safaris 14.1 and ios 14.7](https://caniuse.com/mdn-javascript_builtins_intl_numberformat_numberformat_options_unit_parameter) so we probably don't want to merge this till that browser support is fully present so we don't encourage it's usage

## Type of change

- [ ] @shopify/react-i18n Minor: New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/packages/react-i18n/src/i18n.ts
+++ b/packages/react-i18n/src/i18n.ts
@@ -50,23 +50,6 @@ import {
 export interface NumberFormatOptions extends Intl.NumberFormatOptions {
   as?: 'number' | 'currency' | 'percent' | 'unit';
   precision?: number;
-  // At time of writing TS does not have these definitions. They may be removed
-  // once TS adds them onto the type for Intl.NumberFormatOptions.
-  // TS PR: https://github.com/microsoft/TypeScript/pull/38013
-  // Intl.NumberFormatOptions info see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat
-  //
-  // Strictly speaking there is a defined set of valid values for unit, however
-  // combinatorial complexity makes it unfeasable to list them all out here.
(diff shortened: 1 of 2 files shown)


BPScott:
TS 4.0 is gonna ship with updated lib definitions so once we're using that we can remove all the unit/unitDisplay types and their comments from NumberFormatOptions and the `@ts-ignore`s from the tests

trishrempel (approved):
Just happened to come across this. It looks like there is [browser compatibility](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat#browser_compatibility) for the unit parameters now.

I'd suggest adding `formatUnit` to the README before shipping.

BPScott:
Going to close this. It's old and stale.

PR #1459, opened 2020-05-25
Change date.humanize.today to 'Today at {time}'

toyelowo:
## Description

Humanize future 'today' date as `Today at {time}`

## Type of change

- [x] **react-i18n** - Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

### DIFF
--- a/packages/react-i18n/CHANGELOG.md
+++ b/packages/react-i18n/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+<!-- ## [Unreleased] -->
+
+### Changed
+
+- Updated `I18n#humanizeDate` to humanize today's date as `Today at {time}` [[#1459](https://github.com/Shopify/quilt/pull/1459)]
+
+  Please see the [migration guide](./migration-guide.md) for more information.
+
 ## [3.0.0] - 2020-04-23
 
 ### Changed
--- a/packages/react-i18n/migration-guide.md
+++ b/packages/react-i18n/migration-guide.md
@@ -2,6 +2,19 @@
 
 This is a concise summary of changes and recommendations around updating `@shopify/react-i18n` in consuming projects. For a more detailed list of changes, see [the changelog](./CHANGELOG.md).
 
+## [Unreleased]
+
+🛑Breaking change - New translation key is needed for future 'today' date formatted with `DateStyle.Humanize`. Consumers will need to add the translation key as outlined below.
+
+```json
+"date": {
+  "humanize": {
+    ...
+    "today": "Today at {time}"
+  }
+},
+```
+
 ## [3.0.0] - 2020-04-23
 
 🛑Breaking change - New translation key is needed for future dates formatted with `DateStyle.Humanize`. Consumers will need to add the translation key as outlined below.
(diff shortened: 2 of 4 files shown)


packages/react-i18n/CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [3.1.0] - 2020-05-25
```
alexandcote, line 8:
```suggestion
<!-- ## [Unreleased] -->
```

packages/react-i18n/src/i18n.ts
```diff
@@ -474,7 +474,7 @@ export class I18n {
     }).toLocaleLowerCase();
 
     if (isToday(date)) {
-      return time;
+      return `Today at ${time}`;
```
alexandcote, line 477: The package should already do this for you. Maybe I can help you configure the package if you want.

toyelowo: Yeah that'll be awesome. Thanks, I'll close the PR then

alexandcote, line 477: We should not hard code the string here. We should use the translation file to translate this into multiple languages. For example, in French, the text should be `Aujourd'hui à ${time}`. 

Updating this is considered a breaking change because the consumer needs to add a new translation key. I'm wondering if there is a reason why we didn't add this version in the first place. Maybe @TheMallen has more context since he works more than me on this package.

yarn.lock
```diff
@@ -11410,6 +11410,11 @@ typescript@^3.7.5:
   resolved "https://registry.yarnpkg.com/typescript/-/typescript-3.7.5.tgz#0692e21f65fd4108b9330238aac11dd2e177a1ae"
   integrity sha512-/P5lkRXkWHNAbcJIiHPfRoKqyd7bsyCma1hZNUGfn20qm64T6ZBlrzprymeu918H+mB/0rIg2gGK/BXkhhYgBw==
 
+ua-parser-js@^0.7.17:
```
alexandcote, line 11413: I think you can revert this change.

packages/react-i18n/CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+<!-- ## [Unreleased] -->
+
+### Changed
+
+- Updated `I18n#humanizeDate` to humanize today's date as `Today at {time}` [[#1459](https://github.com/Shopify/quilt/pull/1459)]
```
alexandcote, line 12: Can you add the link to the migration guide like above 😃

packages/react-i18n/src/i18n.ts
```diff
@@ -518,7 +518,7 @@ export class I18n {
     }).toLocaleLowerCase();
 
     if (isToday(date)) {
-      return time;
+      return this.translate('date.humanize.today', {time});
```
alexandcote, line 521: 💯

packages/react-i18n/migration-guide.md
```diff
@@ -2,6 +2,19 @@
...
+
+```json
+"date": {
+  "humanize": {
+    ...
+    "today": "Today at {time}"
+  }
+},
+```
+
```
jaxee, line 17: ✨ so good

PR #1462, opened 2020-05-26
Remove records

ismail-syed:
They're being moved to web-foundations' decision records to consolidate our org-wide records.  See https://github.com/Shopify/web-foundation/pull/160


atesgoral:
@ismail-syed Is this still happening?

atesgoral (approved):
🔥

PR #1468, opened 2020-05-28
Short date style should use numeric instead of 2-digit for day

### LINKED ISSUE #1467: DateStyle.Short uses wrong format for day
## Overview

`DateStyle.Short` formats day values with single digits with zero padding. There shouldn't be padding.

Incorrect:
Jun 09, 2020

Correct:
Jun 9, 2020

## Consuming repo

web

atesgoral:
## Description

Fixes #1467

react-i18n's `DateStyle.Short` uses zero padding in day values. It should actually be just a single digit when the day number has a single digit.

## Type of change

- [x] react-i18n Patch: Bug/ Documentation fix (non-breaking change which fixes an issue or adds documentation)

## Checklist

### DIFF
--- a/packages/react-i18n/src/test/i18n.test.ts
+++ b/packages/react-i18n/src/test/i18n.test.ts
@@ -1156,14 +1156,14 @@ describe('I18n', () => {
     });
 
     it('formats a date using DateStyle.Short', () => {
-      const date = new Date('2012-12-20T00:00:00-00:00');
+      const date = new Date('2012-12-09T00:00:00-00:00');
       const i18n = new I18n(defaultTranslations, {
         ...defaultDetails,
         timezone,
       });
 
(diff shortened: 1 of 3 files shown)


packages/react-i18n/src/test/i18n.test.ts
```diff
@@ -1156,14 +1156,14 @@ describe('I18n', () => {
     });
 
     it('formats a date using DateStyle.Short', () => {
-      const date = new Date('2012-12-20T00:00:00-00:00');
+      const date = new Date('2012-12-09T00:00:00-00:00');
```
jim-chou-shopify, line 1159: nit: suffix ISO date string with `Z`

atesgoral: There are like 20 instances of `-00:00` in this test file and I'd rather not switch the convention in this PR. Can be a separate quick PR to do that though.

PR #1469, opened 2020-05-28
[@shopify/react-form] Undirty form state

### LINKED ISSUE #1463: Add function to manually "undirty" all fields in a form [react-form]
Currently dirty state of `react-form` works best in GraphQL applications and is tricky to manage in others. This is because dirty state is implicit based on the difference between `default` and current `value`s. 

This has caused confusion both internally and externally (see issue #799 for example)

We should provide a simple utility on `useDirty` and by extension `useForm` that manually resets all the default values to their current values.

b-ryu:
## Description

Fixes #1463 

Some consumers of `@shopify/react-form` would like an easier way of "undirtying" their form state, especially after a successful submission. This PR introduces:
- An optional boolean param/config **`makeCleanAfterSubmit`** (defaults to `false`) for `useSubmit` and `useForm`, which, if `true`, will clear the dirty state of the fields upon a successful submission (i.e. set default values to be the most recently submitted values)
- `useForm`'s returned `Form` object also now contains a bound **`makeClean`** callback in parity with `reset`
- For other use cases, we provide a simple utility **`makeCleanFields`**

**Minor fixes:** Polish docs/tests

## Tophatting

Tophatted with the following steps:

<details>
<summary>Playground code
</summary>

```tsx
import {hot} from 'react-hot-loader/root';
import React from 'react';
import {PlaygroundProvider} from 'foundation/PlaygroundProvider';
import {Page, Card, TextField, Button} from '@shopify/polaris';
import {
  useField,
  useForm,
  submitSuccess,
  makeCleanFields,
} from '@shopify/react-form';

export default hot(function Playground() {
  const fields = {
    field1: useField('field1_default'),
    field2: useField('field2_default'),
    field3: useField('field3_default'),
  };
  const {
    fields: {field1, field2, field3},
    dirty,
    submit,
    makeClean,
  } = useForm({
    fields,
    onSubmit: async (values) => {
      console.log('You submitted the form!:', values);
      console.log('It should also undirty all the fields in the form');
      return submitSuccess();
    },
    makeCleanAfterSubmit: true,
  });

  return (
    <PlaygroundProvider>
      <Page title="@shopify/react-form: undirtyForm demo">
        <Card title={dirty ? 'This form is dirty' : 'This form is clean'}>
          <Button onClick={makeClean}>Undirty Form (callback)</Button>
          <Button onClick={() => makeCleanFields(fields)}>
            Undirty Form (utility)
          </Button>
          <TextField
            autoComplete={false}
            label={`Field 1 (current default: ${field1.defaultValue})`}
            {...field1}
          />
          <TextField
            autoComplete={false}
            label={`Field 2 (current default: ${field2.defaultValue})`}
            {...field2}
          />
          <TextField
            autoComplete={false}
            label={`Field 3 (current default: ${field3.defaultValue})`}
            {...field3}
          />
          <Button onClick={submit}>Submit</Button>
        </Card>
      </Page>
    </PlaygroundProvider>
  );
});

```
</details>

1. Pull this branch down in `quilt`
2. While in `quilt`, `yarn tophat packages/react-form ../web`
3. `dev cd web`, and paste the code above into `playground/Playground.tsx`
4. `yarn playground`
5. Go to `localhost:8080` in your browser and make sure that the code works as expected. Optionally, mess around with the playground code and check other use cases

## Type of change

- [x] `@shopify/react-form` Minor: New feature (non-breaking change which adds functionality)

## Checklist


packages/react-form/src/utilities.ts
```diff
@@ -205,3 +205,7 @@ export function defaultDirtyComparator<Value>(
     ? !shallowArrayComparison(defaultValue, newValue)
     : defaultValue !== newValue;
 }
+
+export function undirtyFields(fieldBag: FieldBag) {
```
b-ryu, line 209: I thought about potentially making this a hook (using `useVisitFields`, à la `useReset`) but decided against it for a couple reasons:
- Unlike `useDirty`, `useSubmit`, or `useReset`, "undirtying" fields didn't strike me as "core" form functionality
- Unlike `useReset`, whose memoized callback I can see being passed into child components like buttons, undirtying fields is probably something that would be kept close to submit logic. If consumers really need to pass in a `undirtyFields` callback, they can probably roll their own hook using `useCallback`

I'm pretty new to hooks, so if anyone has thoughts about this please let me know.

packages/react-form/README.md
```diff
@@ -55,11 +56,18 @@ import React from 'react';
...
 
 function MyComponent() {
-  const {fields, submit, submitting, dirty, reset, submitErrors} = useForm({
+  const {
+    fields: {title},
+    submit,
+    submitting,
+    dirty,
+    reset,
+    submitErrors,
```
TheMallen, line 65: should a bound version of `undirty` be returned from here?

packages/react-form/src/utilities.ts
TheMallen: I think this is the right choice for the utility, though I'd like to see a bound version returned from `useForm` for parity with `reset`.

packages/react-form/src/hooks/test/form.test.tsx
```diff
@@ -333,6 +332,148 @@ describe('useForm', () => {
...
     });
+
+    describe('undirty', () => {
+      it('does not undirty fields after successful submit by default', async () => {
+        const promise = Promise.resolve(submitSuccess());
+        const wrapper = mount(<ProductForm data={fakeProduct()} />);
+
+        wrapper
+          .find(TextField, {label: 'title'})!
+          .trigger('onChange', 'tortoritos, the chip for turtles!');
```
TheMallen, line 343: 🐢

packages/react-form/README.md
b-ryu: I could add that in 👍

ismail-syed:
I'm getting the following error's in the playground when 🎩 ing 🤔 


Screen Shot 2020-05-29 at 1 03 00 PM

b-ryu:
> I'm getting the following error's in the playground when 🎩 ing 🤔
> 
> Screen Shot 2020-05-29 at 1 03 00 PM

I think I also got that earlier. For me, just `Ctrl-C` and running it again worked for me 🤔

packages/react-form/README.md
```diff
@@ -865,7 +903,7 @@ A value returned by `reduceFn` iterating through all the fields in the form, and
 
 ##### Examples
 
-Here is how to calculate if the entire form is dirty
+Here is how to calculate if the entire form is dirty (though in practice you could use `useDirty` to the same effect):
```
TheMallen, line 906: Good addition

PR #1477, opened 2020-05-29
react-testing / Replace .props() with .props in documentation

loic-d:
## Description

This PR updates `react-testing` documentation.

`.props()` on a `wrapper` is not valid since `props` [is a getter and not a function](https://github.com/Shopify/quilt/blob/153d270592373acc0d835625e2824599480cdbd0/packages/react-testing/src/root.tsx#L44).

## Type of change

- [x] `react-testing` Patch: Bug/ Documentation fix (non-breaking change which fixes an issue or adds documentation)

## Checklist


michenly:
No changlog needed for doc change. ( I really should update the template...)

PR #1483, opened 2020-05-29
[@shopify/react-network] Get rid of unused function

### LINKED ISSUE #931: Add an easy util for getting the local from accept-language header

b-ryu:
## Description

This PR gets rid of an unused/unexported function `hasDocumentCookie` (there seems to be an identical function in [`@shopify/react-cookie`](https://github.com/Shopify/quilt/blob/master/packages/react-cookie/src/utilities.ts), but the one in this package isn't being used anywhere).

## Type of change

- [x] `@shopify/react-network` Patch: Bug/ Documentation fix (non-breaking change which fixes an issue or adds documentation)

## Checklist

### DIFF
--- a/packages/react-network/CHANGELOG.md
+++ b/packages/react-network/CHANGELOG.md
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## [Unreleased] -->
+## [Unreleased]
+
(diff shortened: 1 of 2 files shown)


TheMallen:
Are you sure this isn't used anywhere ourside of this repo? Seems tough to verify. It would probably need to be a major bump because we are removing an API entirely.

b-ryu:
> Are you sure this isn't used anywhere ourside of this repo? Seems tough to verify. It would probably need to be a major bump because we are removing an API entirely.

Oh I kinda just assumed that since it wasn't exported through `index.ts` consumers wouldn't really be using it. I can just close this PR if that's not the case.

PR #1484, opened 2020-05-29
Upgrade webpack-virtual-modules

ismail-syed:
## Description

This PR upgrades `webpack-virtual-modules` and also adds a test the `magic-entries-webpack-plugin` to verify that it creates magic entries from virtual modules. This ability was provided free by upgrading the `webpack-virutual-modules` plugin

## Type of change

## Checklist

### DIFF
--- a/packages/magic-entries-webpack-plugin/package.json
+++ b/packages/magic-entries-webpack-plugin/package.json
@@ -33,5 +33,9 @@
   "files": [
     "dist/*",
     "!tsconfig.tsbuildinfo"
-  ]
+  ],
+  "devDependencies": {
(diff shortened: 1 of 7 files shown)


packages/magic-entries-webpack-plugin/package.json
```diff
@@ -33,5 +33,9 @@
   "files": [
     "dist/*",
     "!tsconfig.tsbuildinfo"
-  ]
+  ],
+  "devDependencies": {
```
ismail-syed, line 37: Adding this for the editor to pick up these modules in tests.

PR #1487, opened 2020-06-02
[@shopify/react-i18n-universal-provider] Better details merge

### LINKED ISSUE #1473: i18n details does not always use available details.
## Overview

Following up from [issue](https://github.com/Shopify/signage/pull/84)

We should do a smarter merge we do for i18n details in [universal provider](https://github.com/Shopify/quilt/blob/master/packages/react-i18n-universal-provider/src/I18nUniversalProvider.tsx#L28-L31). In a way that the locale (or any other i18n details) should never end up being undefined if those values are available in the serialized data.

## Consuming repo

...

## Labels

- [ ] Add the `Type: Bug` label to this issue.

### Area

- [ ] Add any relevant `Area: <area>` labels to this issue

### Scope

  - [ ] Tag it with the `Package: <package_name>` label.

---

## Checklist

b-ryu:
## Description

Fixes #1473 

Our previous method of consolidating serialized i18n details and the override values allows for `undefined` fields to overwrite otherwise defined fields. In all cases, if a field is defined in either/both the serialized and override objects, then it should be defined in the resulting merge object.

## Type of change

- [x] `@shopify/react-i18n-universal-provider` Patch: Bug (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/packages/react-i18n-universal-provider/README.md
+++ b/packages/react-i18n-universal-provider/README.md
@@ -39,13 +39,18 @@ I18nDetails {
 import {I18nUniversalProvider} from '@shopify/react-i18n-universal-provider';
 
 function App({locale}: {locale?: string}) {
-  return <I18nUniversalProvider locale={locale}>{/* rest of the app */}</I18nUniversalProvider>;
+  return (
+    <I18nUniversalProvider locale={locale}>
+      {/* rest of the app */}
+    </I18nUniversalProvider>
+  );
 }
 ```
 
 ### Possible Issues
 
-####  Missing i18n manager error
+#### Missing i18n manager error
+
 ```
 Error: Missing i18n manager. Make sure to use an <I18nContext.Provider /> somewhere in your React tree from the @shopify/react-i18n hook.
 ```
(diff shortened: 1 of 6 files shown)


packages/react-i18n-universal-provider/README.md
```diff
@@ -66,7 +71,7 @@ npx yarn-deduplicate --packages @shopify/react-effect yarn.lock
 ```
 
 ```bash
-$ yarn why @shopify/react-i18n # ensure no duplicate / unmet dependencies 
-yarn list  # ensure no duplicate / unmet dependencies 
+$ yarn why @shopify/react-i18n # ensure no duplicate / unmet dependencies
+yarn list  # ensure no duplicate / unmet dependencies
```
michenly, line 75: just to double check, is this file all stylistic update?

b-ryu: Yeah, I didn't actually change anything myself; my editor just auto-linted the code on save.

packages/react-i18n-universal-provider/src/utilities.ts
```diff
@@ -0,0 +1,13 @@
+function pruneUndefinedFields(obj = {}) {
+  Object.keys(obj).forEach(key => obj[key] === undefined && delete obj[key]);
```
TheMallen, line 2: nit: it feels a bit weird that this both mutates the object and also returns it, usually I think we'd want to either return a totally new object (especially in react contexts where object identity matters) or only mutate it and not return anything (so that it is clear that we are relying on side-effects)

It's not a big deal since this function isn't really exposed anywhere though.

b-ryu: Yeah I see what you mean, I guess I did it like this because it made the destructuring in the other function a bit more concise, and also deleting undefined fields seemed like an inconsequential mutation to me. I'll change it so that it just returns a new object just for cleanliness' sake.

PR #1489, opened 2020-06-03
✖️  deprecate react-server-webpack-plugin

michenly:
## Description

Fixes (issue #) this is the attempt to address the dependency issue between quilt_rails and react-server-webpack-plugin.

I am deprecating `react-server-webpack-plugin` and moving it into `react-server` 

## Type of change

- [x]  Minor: New feature (non-breaking change which adds functionality) for react-server

## Checklist


packages/react-server/src/webpack-plugin/index.ts
```diff
@@ -0,0 +1 @@
+export {ReactServerPlugin} from './webpack-plugin';
```
TheMallen, line 1: I think we need to add a `webpack-plugin.js` to the root of the package if we want to be able to get at this (see other packages with multiple sub-packages)

michenly: its already included as `packages/react-server/webpack.d.ts` and `packages/react-server/webpack.js` and reference in `packages/react-server/package.json`

michenly: thou I might rename them...

Issue #854, opened 2019-08-15
[@shopify/react-server] Expose an API for nicer error pages


ismail-syed:
## Overview

https://github.com/Shopify/quilt/pull/850 addressed https://github.com/Shopify/quilt/issues/838 but we should expose an API to handle error pages. 

Things we should be able to cover with this API: 

- Display a **production** error page when server rendering fails in prod.
- Display a **development** error page with stack trace information
- Make the default error pages nicer than they are now.

ismail-syed:
Might be worth exploring, but pulling this out into a react component would be nice 
https://github.com/Shopify/sewing-kit/blob/master/packages/sewing-kit/src/tools/webpack/dev/development-server/middleware/warmup-markup.ts

PR #1490, opened 2020-06-03
Support custom error pages for production SSR errors in react-server

ismail-syed:
## Description

Related to https://github.com/Shopify/quilt/issues/854

This PR enables `@shopify/react-server` applications to provide custom error pages which we can now use on production SSR errors. It does this by: 

* Exposing a `renderError` option in `react-server`'s `createServer`. `renderError` is similar to the standard `render` option we pass, except it's for rendering a consumer provided `Error` component. If `renderError` is not set, then react-server defaults to returning a static HTML error page as a sane default/fallback. 
* `react-server-webpack-plugin` has also been updated to utilizes the new `renderError` API in `react-server.` The plugin checks for the existence of an `error` component. If an `error` component exists, it gets imported in the server and passed to `react-server`. This also creates a virtual client entry point for the `error` component for client-side hydration. 

## 🎩 

1. checkout this branch and run `yarn run build`
2. `cd packages/react-server && yarn link`
3. `cd packages/react-server-webpack-plugin && yarn link` 
4. checkout the `quilt_rails-prod-ssr-error` branch in `shrink-ray` 
5. `yarn link @shopify/react-server && yarn link @shopify/react-server-webpack-plugin`
6.  Run the following to run the rails server in production mode 

```
$ NODE_ENV=production SK_SIMULATE_PRODUCTION=1 bundle exec rake assets:precompile
$ NODE_ENV=production SK_SIMULATE_PRODUCTION=1 dev run
```
7. in a second window run `node build/server/main.js`
8. Open the app and verify that you see the following error page. This is an error page setup in the `shrink-ray` branch. 

Screen Shot 2020-06-04 at 3 15 05 AM

10. Delete the error component file: `app/ui/error.tsx`
11. Re-due steps 6, 7, 8 
12. Verify that you see the following screenshot. This is the default fallback error page provided by `react-server`. 

Screen Shot 2020-06-04 at 1 28 25 PM

## Type of change

## Checklist


packages/react-server/src/render/error/fallback-error-markup.ts
```diff
@@ -0,0 +1,68 @@
...
+                          We received an error from the server. Please check
+                          again later.
+                        </p>
+                      </div>
+                    </div>
+                  </div>
+                </div>
+                <div class="Polaris-EmptyState__ImageContainer">
+                  <img
+                    src="https://cdn.shopify.com/s/files/1/0757/9955/files/empty-state.svg"
```
ismail-syed, line 54: Note, this is the not right empty state image I plan to use for the fallback error page. 

**wrong one:**
image
 

**right one:**
image


### Why aren't you using the right one?

I put the expected svg in a [google cloud storage](https://console.cloud.google.com/storage/browser/cdn.shopifycloud.com/web-foundation/images). Getting this image publicly accessible is WIP.

packages/react-server/src/render/error/fallback-error-markup.ts
```diff
@@ -0,0 +1,68 @@
+export const fallbackErrorMarkup = `
```
TheMallen, line 1: Might be nicer to have this as an actual template, but I know the pathing would be a bit of a pain.

michenly:
I try 🎩  and it running into a strange error where my browser would download the document instead of opening it in the browser. I tried both Chrome and Firefox and its the same behaviour. Also checked that this does not occur with shrink-ray master.

Screen Shot 2020-06-04 at 11 41 23 AM

packages/react-server/src/render/error/fallback-error-markup.ts
ismail-syed: PR has been updated

packages/react-server-webpack-plugin/src/react-server-webpack-plugin.ts
```diff
@@ -129,7 +153,7 @@ function clientSource() {
   `;
 }
 
-function noSourceExists(
+export function noSourceExists(
```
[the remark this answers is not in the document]
ismail-syed: Good catch!

packages/react-server/src/render/error/fallback-error-markup.ts
ismail-syed: yah, i felt that pain and went with this approach instead

ismail-syed:
Tested and verified in shrink-ray. Merging and cutting a release.

PR #1491, opened 2020-06-04
[@shopify/react-network, @shopify/react-server, quilt_rails] Allow client-side access of request headers

### LINKED ISSUE #1472: The header method does not work on client side react app
## Overview

Follow up from https://github.com/Shopify/signage/pull/84
The [header method](https://github.com/Shopify/quilt/tree/master/gems/quilt_rails#example-sending-custom-headers-from-rails-controller) described in the quilt rail doc has value on the server side render but does not have value on client side.

I am not sure what the solution is, but we should at least update the doc to included a serialization step when using header such as the [example here](https://github.com/Shopify/signage/pull/84#issuecomment-635504555).

## Consuming repo

...

## Labels

- [ ] Add the `Type: Bug` label to this issue.

### Area

- [ ] Add any relevant `Area: <area>` labels to this issue

### Scope

  - [ ] Tag it with the `Package: <package_name>` label.

---

## Checklist

b-ryu:
## Description

Fixes #1472 

The hook `useRequestHeader` in `@shopify/react-network` only has value on server-side renders, since it relies on a context object available only on the server. This PR allows for that hook to access the same headers on subsequent client-side renders. It does this by tracking headers accessed via `useRequestHeader` server-side and serializing those headers for client use. 

Note that since `useRequestHeader` depends on the initial server-side render to know which headers to serialize, it is still not useful for components that are not rendered initially on the server (i.e. conditionally rendered components).

- In **`@shopify/react-network`**, `useRequestHeader` now tracks accessed headers in server-side renders, attaching them to `NetworkManager`'s `accessedHeaders` member in server context. On client-side renders, it retrieves those headers from serialized `quilt-data`. Consequently, it now also depends on `@shopify/react-html`
- In **`@shopify/react-server`**, `NetworkManager`'s `accessedHeaders` are serialized and stored under the `accessedHeaders` field in serialized `quilt-data`
- In **`quilt_rails`**, our `X-Quilt-Data` header now only concerns "data", as opposed to data and headers together

**Pending documentation changes:**
> `useRequestHeader` should only be used in top-level components, or at the very least, components that can be expected to be rendered on the server

>Consumers of this gem and related Quilt packages are encouraged to pass data pertinent to rendering to the `data` param in `render_react` rather than `headers` (since such data is guaranteed to be serialized and available client-side, regardless of whether or not it was accessed on the first server-side render)

## Type of change

## Checklist


gems/quilt_rails/lib/quilt_rails/react_renderable.rb
```diff
@@ -37,10 +37,9 @@ def proxy(headers, data)
       end
 
       begin
-
         reverse_proxy(
           url,
-          headers: headers.merge('X-CSRF-Token': form_authenticity_token, 'X-Quilt-Data': headers.merge(data).to_json)
+          headers: headers.merge('X-CSRF-Token': form_authenticity_token, 'X-Quilt-Data': data.to_json)
```
TheMallen, line 42: 👏

packages/react-network/src/hooks.ts
```diff
@@ -28,7 +29,22 @@ export function useCspDirective(
...
+
+  if (network) {
+    // Server: get it from context
+    // It should also store/serialize it for later client-side renders
+    return network.getHeader(header);
+  } else {
+    // Client: get it from serialized data
+    // If not present (i.e. component was not initially rendered on server) return undefined
+    try {
+      return (getSerialized('quilt-data') as {
```
TheMallen, line 41: This will technically work, but using `getSerialized` here will do a bunch of DOM parsing shenanigans inline in the render loop every time the component rerenders. As a result I don't think we want to go with this approach to getting the data from serialized. Instead we likely want to use a context provider and the `useSerialized` hook.

packages/react-network/src/manager.ts
```diff
@@ -45,7 +46,14 @@ export class NetworkManager {
   }
 
   getHeader(header: string) {
-    return this.requestHeaders[header.toLowerCase()];
+    const headerKey = header.toLowerCase();
```
TheMallen, line 49: these changes def make sense 👍

packages/react-server/src/render/render.tsx
```diff
@@ -85,7 +85,12 @@ export function createRender(render: RenderFunction, options: Options = {}) {
           <HydrationContext.Provider value={hydrationManager}>
             <NetworkContext.Provider value={networkManager}>
               {children}
-              <Serialize data={() => ctx.state.quiltData} />
+              <Serialize
+                data={() => ({
+                  ...ctx.state.quiltData,
+                  accessedHeaders: networkManager.accessedHeaders,
```
TheMallen, line 91: I don't think we want to serialize this as part of the `data` object. I'd expect it to have it's own key.

b-ryu:
@TheMallen and I discussed this a bit more offline during our 1-on-1. Basically, the two possible approaches to this are:
1. Manually deserializing (and memoizing) the requested headers in the `useRequestHeader` hook (basically this PR at the time of writing with some polish)
2. Providing some sort of universal provider component (i.e. `HeaderUniversalProvider`) that manages that serialization for us and provides header-related context to the rest of the app, which `useRequestHeader` would read from

Option 2 is cleaner and probably more performant in the long run (having a single deserialization client-side as opposed to one for every single `useRequestHeader` instance). The downside is that the consumer would have to wrap their app in another provider which adds more complexity relatively speaking.

b-ryu:
> Looks great, but could we also add `NetworkUniversalProvider` to the code generated by `react-server-webpack-plugin`?

Just relating some things discussed offline: since this provider relies on its `headers` prop to determine which headers to serialize, it's probably best to leave it out of the plugin for now. It'd be weird to define which headers to serialize in a plugin config. ATM the expected use case is that the consumer is responsible for wrapping their own app in the provider.

PR #1494, opened 2020-06-08
♻️ move generator to `sewing_kit`

michenly:
## Description

Ad title

Related sewing-kit PR https://github.com/Shopify/sewing-kit/pull/2133

## Type of change

## Checklist


gems/quilt_rails/docs/api.md
```diff
@@ -241,19 +241,14 @@ The `Performance::Reportable` mixin uses [https://github.com/Shopify/statsd-inst
 
 #### `quilt_rails:install`
 
-This is equivalent to running `rails generate sewing_kit:install` and `rails generate quilt:install`
+This is equivalent to running `rails generate sewing_kit:install` (after installing sewing_kit gem) and `rails generate quilt:install`
```
gmcgibbon, line 244: If we are expecting to invoke `sewing_kit:install` here we could just add a dependency on `sewing_kit` and constrain it to the upcoming release.

gems/quilt_rails/lib/generators/quilt_rails/install_generator.rb
```diff
@@ -2,9 +2,16 @@
...
-    def run_all_generators
-      generate("sewing_kit:install")
-      generate("quilt:install")
+    def setup_sewing_kit_gem
+      say("Add sewing_kit gem")
+      gem("sewing_kit")
+    end
+
+    def run_react_app_generator
+      after_bundle do
```
gmcgibbon, line 11: You don't get this for free. It might be better to just declare a dependency on `sewing_kit`.

michenly: `quilt_rails` are meant to be decouple if someone want to use it that way (external developer who have no access to sewing-kit) so I dont want to have quilt_rails be dependent on sewing-kit if someone does not need it.

I can remove this task and having the consumer run the two generator manually from the documentation maybe?

gmcgibbon: If that's the case then the sewing_kit generator should be optional, yes. This generator doesn't serve much purpose at this point I think. Maybe we should remove it?

michenly: I mainly kept it around so it wont be a breaking change. It certainly looks like it is necessary now to do a breaking change and remove it.

gems/quilt_rails/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Changed
+
+- Move SewingKit generator task into sewing-kit. The generator command remains the same. ([#1494](https://github.com/Shopify/quilt/pull/1494))
```
gmcgibbon, line 12: Should also mention the removal of `gems/quilt_rails/lib/generators/quilt_rails/install_generator.rb `.

PR #1509, opened 2020-06-13
quilt_rails update

michenly:
- [x] Remove installation of `react-server` because it is included as of sk v0.131
- [x] set javascript path before copying over app file. (from https://github.com/Shopify/shopify-rails/pull/67/files)
- [x] add in a demo app generator that has a more complete example of a typical react app
- [x] remove mention of dev in the doc and replace it with `rails new` & `rails server`
- [x] remove automatic passing of `X-CSRF-Token` in quilt_rails
- [x] Fix `react-graphql-universal-provider` doc to remove mentioned of csrf and use URL for graphql uri


gems/quilt_rails/test/generators/quilt/react_app/react_app_generator_test.rb
```diff
@@ -12,6 +12,13 @@ class QuiltReactAppGeneratorTest < Rails::Generators::TestCase
     prepare_destination
   end
 
+  test "set javascript path" do
+    run_generator
+    assert_file "config/application.rb" do |app_config|
+      assert_match "config.javascript_path = \"ui\"", app_config
```
michenly, line 18: @gmcgibbon the reason why this is failing is because it cannot find `config/application.rb`
what should happen if there if no `config/application.rb`?
unless that is just a problem of my test case?

gems/quilt_rails/test/generators/quilt/demo_app/demo_app_generator_test.rb
```diff
@@ -0,0 +1,24 @@
...
+
+class QuiltDemoAppGeneratorTest < Rails::Generators::TestCase
+  tests Quilt::DemoAppGenerator
+  destination File.expand_path("./tmp", File.dirname(__FILE__))
+
+  setup do
+    prepare_destination
+  end
+
+  test "creates the main React app" do
```
michenly, line 15: really not sure why all the test are failing 😕

gems/quilt_rails/lib/generators/quilt/demo_app/demo_app_generator.rb
```diff
@@ -0,0 +1,71 @@
...
+    end
+
+    def add_polaris_sk_config
+      file_path = "config/sewing-kit.config.ts"
+
+      gsub_file(
+        file_path,
+        "plugins: [", "plugins: [#{File.read(File.expand_path(find_in_source_paths('polaris_sass_plugin.ts')))}"
+      )
+    end
```
michenly, line 37: a bit of a flaky solution

gems/quilt_rails/test/generators/quilt/demo_app/demo_app_generator_test.rb
gmcgibbon: The dummy app isn't present in the generator's test directory by default. It looks like this project sources files from two locations as well, which is redundant. Try changing `test/support/generator_test_helper.rb` to something like this:
```
module GeneratorTestHelpers
  DUMMY_PATH = File.expand_path("../dummy", __dir__)

  def copy_from_dummy(*path)
    dummy_path = File.join(DUMMY_PATH, *path)
    destination_path = File.join(destination_root, *path)

    FileUtils.mkdir_p(File.dirname(destination_path))
    FileUtils.cp(dummy_path, destination_path)
  end
end
```

And then calling it in the test cases like:
```
setup do
    prepare_destination
    copy_from_dummy("config", "application.rb")
  end
```
You'll need to update `QuiltReactAppGeneratorTest` to use `copy_from_dummy("config", "routes.rb")` with this approach too.

gems/quilt_rails/lib/generators/quilt/demo_app/demo_app_generator.rb
gmcgibbon: If the config file was JSON it would be more language agnostic. As it stands you'll likely have to resort to pattern matching in some way to do this.

gems/quilt_rails/lib/generators/quilt/demo_app/demo_app_generator.rb
```diff
@@ -0,0 +1,69 @@
...
+      unless File.exist?(file_path)
+        inject_into_file(
+          file_path,
+          File.read(File.expand_path(find_in_source_paths('path_import.ts'))),
+          after: "import {Plugins, Env} from '@shopify/sewing-kit';"
+        )
+
+        gsub_file(
+          file_path,
+          "plugins: [", "plugins: [#{File.read(File.expand_path(find_in_source_paths('polaris_sass_plugin.ts')))}"
```
michenly, line 32: > As it stands you'll likely have to resort to pattern matching in some way to do this.

 @gmcgibbon for the pattern matching in both `gsub_file` and `inject_into_file`, can I use Regex in place of plain text?

gems/quilt_rails/CHANGELOG.md
```diff
@@ -7,6 +7,17 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 <!-- ## [Unreleased] -->
 
+### Changed
+
+- Remove automatic passing of `X-CSRF-Token` in the header. With [csrf header strategy](./README.md#fixing-rejected-csrf-tokens-for new-user-sessions) you should not need this value for GraphQL request. If absolutely needed, use [custom headers method](./README.md#example:-sending-custom-headers-from-rails-controller) to pass the value manually. ([#1509](https://github.com/Shopify/quilt/pull/1509))
```
ismail-syed, line 12: [csrf header strategy](./README.md#fixing-rejected-csrf-tokens-for new-user-sessions) isn't resolving the hyperlink.

michenly: https://github.com/Shopify/quilt/tree/master/gems/quilt_rails/README.md#fixing-rejected-csrf-tokens-for-new-user-sessions is what I was aiming for...there must be a typo in there...

michenly: fixed!

gems/quilt_rails/lib/generators/quilt/demo_app/demo_app_generator.rb
gmcgibbon: Yes. See thor's documentation: https://rubydoc.info/github/wycats/thor/master/Thor%2FActions:gsub_file

PR #1512, opened 2020-06-16
[@shopify/jest-mock-apollo] Provide more useful errors

### LINKED ISSUE #617: Better error when a mock mutation is missing
## Overview
I was running a test where I forgot to mock the graphql mutation associated with it. I got the error `[error] Cannot read property 'data' of undefined` which wasn't helpful in finding what was causing the failure of test. A better error message will help in identifying the error faster.

## Type

- [ ] New feature

## Motivation
The fact that the error message I got wasn't good at letting me know what actually was failing motivated me to ask for this. 

## Checklist

b-ryu:
## Description

Fixes #617 

In our `@shopify/jest-mock-apollo` package, errors pertaining to missing mocks (queries, mutations) are not easily visible to the consumer and are obfuscated by more confusing messages (e.g. `Cannot read property 'data' of undefined`).  This PR fixes this by propagating errors using the intended error API instead of as query data.

Since this affects return behaviour it should probably require a major version bump.

## Screenshots

| Before (broken test for illustration) | After (broken test for illustration) |
| --- | --- | 
| Screen Shot 2020-06-22 at 6 28 10 PM | Screen Shot 2020-06-22 at 4 28 14 PM |

## Tophat

I've added tests to cover the missing cases, but you can play around with the tests to see the different behaviours:

1. `dev cd quilt && git fetch --all && git checkout 617-test-error`
2. Navigate to `packages/jest-mock-apollo/src/test/index.test.tsx`
3. Break one of the non-error tests by mocking an empty/insufficient client
4. `dev test packages/jest-mock-apollo/src/test/index.test.tsx`
5. The test should fail, and you should get useful error messages in the console like in the screenshot above

## Type of change

- [x] `@shopify/jest-mock-apollo` Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

### DIFF
--- a/packages/jest-mock-apollo/src/MockApolloLink.ts
+++ b/packages/jest-mock-apollo/src/MockApolloLink.ts
@@ -38,38 +38,41 @@ export default class MockApolloLink extends ApolloLink {
             : mockForOperation;
       }
 
-      let result: ExecutionResult | Error;
+      let result: ExecutionResult;
 
       if (response == null) {
         let message = `Can’t perform GraphQL operation '${operationName}' because no valid mocks were found`;
 
         if (typeof mock === 'object') {
           const operationNames = Object.keys(mock);
-          // We will provide a more helpful message when it looks like they just provided data,
-          // not an object mapping names to fixtures.
-          const looksLikeDataNotFixtures = operationNames.every(
-            name => name === name.toLowerCase(),
-          );
 
-          message += looksLikeDataNotFixtures
-            ? ` (it looks like you tried to provide data directly to the mock GraphQL client. You need to provide your fixture on the key that matches its operation name. To fix this, simply change your code to read 'mockGraphQLClient({${operationName}: yourFixture}).'`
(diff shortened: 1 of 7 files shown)


packages/jest-mock-apollo/src/MockApolloLink.ts
```diff
@@ -79,7 +77,7 @@ export default class MockApolloLink extends ApolloLink {
             ),
           };
         } catch (error) {
-          result = error;
+          result = normalizeError(error);
```
TheMallen, line 80: would it make more sense to just re`throw` the normalized error here? Then the test would actually fail regardless of assertions. It feels a bit weird to me to require users to look at a runtime value and in their component to see they messed up their mock format.

b-ryu: I sorta tried to leave general functionality as intact as possible. I didn't know if there's a use case where a consumer would want to test their component's error handling, in which case it might be better to let them explicitly define their pass-fail behaviour through assertions? Would maybe just logging the error suffice?

PR #1515, opened 2020-06-16
🐛 update react-network apply context to only set cookie if the value …

michenly:
…had been changed

## Description

Fixes (issue #) https://github.com/Shopify/signage/issues/145

## Type of change

## Checklist

### DIFF
--- a/packages/react-network/src/server.ts
+++ b/packages/react-network/src/server.ts
@@ -26,6 +26,11 @@ export function applyToContext<T extends Context>(
   Object.entries(cookies).forEach(([cookie, options]) => {
     const {value, ...cookieOptions} = options;
 
+    // only set cookie that had been changed
+    if (ctx.cookies.get(cookie) === value) {
+      return;
+    }
+
     // missing 'none` in `sameSite`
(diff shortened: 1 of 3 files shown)


packages/react-network/src/server.ts
```diff
@@ -26,6 +26,11 @@ export function applyToContext<T extends Context>(
   Object.entries(cookies).forEach(([cookie, options]) => {
     const {value, ...cookieOptions} = options;
 
+    // only set cookie that had been changed
+    if (ctx.cookies.get(cookie) === value) {
```
TheMallen, line 30: this might want to actually be `if (ctx.cookies.get(cookie) == value || ctx.cookies.get(cookie, options) == value)`, I'm not exactly sure how `options` interacts with `cookies.get`

TheMallen: checked this out, options for `cookies.get` not the same as `cookies.set`
https://github.com/pillarjs/cookies#cookiesget-name--options--

so it should be
```
if (ctx.cookies.get(cookie) === value || ctx.cookies.get(cookie, {signed: true}) === value)
```

Issue #1516, opened 2020-06-16
[react-network] Empty cookies are still being set


kvendrik:
Some cookies still get passed through

```
curl -IXGET "https://signage.myshopify.io/tools/signage-maker" --cookie "_cookieA=this%2Bhas%2Ba%2Bplus; _cookieB=" --silent | grep "_cookieA\|_cookieB" 

set-cookie: _cookieB=; path=/; expires=Thu, 01 Jan 1970 00:00:00 GMT; HttpOnly; secure
```

[This](https://github.com/Shopify/quilt/blob/30a0dabc9ba416ba41cc690fa761e5fa7af14b58/packages/react-network/src/server.ts#L31-L34) needs to be changed to `typeof rawCookieValue  !== 'undefined' && ...` because currently it skips the equals check if the cookie is empty.

TheMallen:
fixed by #1517

PR #1517, opened 2020-06-16
[react-network] fixes bug where cookie is passed through when empty

kvendrik:
## Description

Fixes (issue #) https://github.com/Shopify/quilt/issues/1516

Fixes bug where cookie is passed through when empty due to broken undefined check.

## Type of change

## Checklist


packages/react-network/src/server.ts
```diff
@@ -29,7 +29,10 @@ export function applyToContext<T extends Context>(
     // only set cookies that have been changed
     // decode URI as the manager's cookie value is also decoded
     const rawCookieValue = ctx.cookies.get(cookie);
-    if (rawCookieValue && decodeURIComponent(rawCookieValue) === value) {
+    if (
+      typeof rawCookieValue !== 'undefined' &&
```
michenly, line 33: this will let false & null pass through, is that the intent?

michenly: I guess the type for rawCookieValue can only be string or undefined...

kvendrik: Yeah exactly 😄

TheMallen, line 33: is there a reason we have to use typeof instead of what we usually do that checks for both null and undefined (`!= null`)?

michenly: 😂  I ask similar question cuz this syntax was confusing me. 
as it turn out thou, rawCookieValue is type as string | undefined.

kvendrik: `!= null` works too I have no string preference. I'll change it

PR #1518, opened 2020-06-17
Fix an issue with bundle size increased

alexandcote:
## Description

We had reports that the new version of `@shopify/polyfills` [increase the bundle size by about 20kb](https://shrink-ray.shopifycloud.com/repos/brochure2/commits/b6a838e9800d1f32c9ff71c7cd19b43e96d319a6/entrypoints). After some investigation, using directly the `core-js` import was the cause of the issue.

Instead of only include the polyfills needed, the new version includes every polyfill.

| Before | After |
| -- | -- |
| image | image |

## Type of change
Reverting to the original strategy until we move to `core-js@3`

- [x] `@shopify/polyfills Patch: Bug (non-breaking change which fixes an issue)

## Checklist


IlanaB:
can this be merged? i'm going to close https://github.com/Shopify/brochure2/pull/21128 in the meantime

Issue #1475, opened 2020-05-28
SSR false queries called on the server


CameronGorrie:
## Overview
The [query component re-exported w/types from `react-graphql`](https://github.com/Shopify/quilt/blob/master/packages/react-graphql/src/Query.tsx#L13) server side renders queries that should be opted out via the `ssr: false` option. This was observed when upgrading to `react-apollo v3.1.4` and `react-graphql v6.1.4` and can be reproduced in Shopify/web wherever we have legacy components using the HOC query implementation (ping me for additional reproduction steps).

The custom query hook implementation from `react-graphql` opts out of server side rendering as expected so we could fix forward by exporting an HOC from `react-graphql` that uses our query hook.

Related:
https://github.com/apollographql/react-apollo/pull/3515#issuecomment-572630300
https://github.com/apollographql/apollo-client/pull/5839
https://github.com/apollographql/apollo-client/issues/4814

## Consuming repo

> What repo were you working in when this issue occurred?

Shopify/Web

---

## Checklist

- [x] Please delete the labels section before submitting your issue
- [ ] I have described this issue in a way that is actionable (if possible)

PR #1519, opened 2020-06-17
Query component uses hook

CameronGorrie:
## Description

Fixes Shopify/quilt#1475

Client-side queries are being executed on the server when using the Apollo Query component re-exported from `react-graphql`. This PR replaces the declarative  component with our own implementation which uses the custom useQuery hook and handles client-only queries correctly.

## Type of change

- [x] `react-graphql` Patch: Bug (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/packages/react-graphql/package.json
+++ b/packages/react-graphql/package.json
@@ -25,7 +25,6 @@
   "homepage": "https://github.com/Shopify/quilt/blob/master/packages/react-graphql/README.md",
   "dependencies": {
     "@apollo/react-common": ">=3.0.0 <4.0.0",
-    "@apollo/react-components": ">=3.0.0 <4.0.0",
     "@apollo/react-hooks": ">=3.0.0 <4.0.0",
     "@shopify/async": "^2.1.5",
     "@shopify/react-async": "^3.1.19",
--- a/packages/react-graphql/src/Query.tsx
+++ b/packages/react-graphql/src/Query.tsx
@@ -1,13 +1,19 @@
-import React from 'react';
-import {Query as ApolloQuery} from '@apollo/react-components';
 import {OperationVariables} from 'apollo-client';
+import {DocumentNode} from 'graphql-typed';
 
-import {QueryProps} from './types';
+import {useQuery, QueryHookResult, QueryHookOptions} from './hooks';
 
-// eslint-disable-next-line react/prefer-stateless-function
-class QueryTypeClass<
-  Data = any,
-  Variables = OperationVariables
-> extends React.Component<QueryProps<Data, Variables>> {}
+interface QueryComponentOptions<Data, Variables> extends QueryHookOptions {
+  children: (result: QueryHookResult<Data, Variables>) => JSX.Element | null;
+  query: DocumentNode<Data, Variables>;
+}
 
-export const Query: typeof QueryTypeClass = ApolloQuery as any;
+export function Query<Data = any, Variables = OperationVariables>({
+  children,
+  query,
+  ...options
+}: QueryComponentOptions<Data, Variables>) {
+  const result = useQuery<Data, Variables>(query, options);
+
+  return children(result);
+}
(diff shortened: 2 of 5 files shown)


packages/react-graphql/src/Query.tsx
```diff
@@ -1,13 +1,19 @@
...
+  query: DocumentNode<Data, Variables>;
+}
 
-export const Query: typeof QueryTypeClass = ApolloQuery as any;
+export function Query<Data = any, Variables = OperationVariables>({
+  children,
+  query,
+  ...options
+}: QueryComponentOptions<Data, Variables>) {
+  const result = useQuery<Data, Variables>(query, options);
```
CameronGorrie, line 16: I'm having trouble with the query options type:

Screen Shot 2020-06-17 at 6 34 39 PM

@lemonmade wondering if you would mind explaining how [this type](https://github.com/Shopify/quilt/blame/master/packages/react-graphql/src/hooks/query.ts#L29) works?

lemonmade: I don't know if it actually does what it says it does, but this type is trying to say that "if all the variable values can be nullish, then this function takes one required argument (the query), and one optional argument (the options, including variables). If at least one variable is required, this function takes two required arguments (the query and the options).

CameronGorrie: Thanks Chris, the explanation makes sense. I'm still having issues passing the query hook options through to the hook. The variables generic type in this case should be the `QueryHookOptions` object since I'm extending it in my `QueryComponentOptions` interface. I'll ask around :)

packages/react-graphql/src/Query.tsx
```diff
@@ -1,13 +1,26 @@
...
 
-export const Query: typeof QueryTypeClass = ApolloQuery as any;
+export function Query<Data = any, Variables = OperationVariables>({
+  children,
+  query,
+  ...options
+}: QueryComponentOptions<Data, Variables>) {
+  const opts = [options] as IfAllNullableKeys<
+    Variables,
+    [QueryHookOptions<Data, NoInfer<Variables>>?],
```
vsumner, line 19: Should there be a `?` in there?

packages/react-graphql/src/tests/Query.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+    },
+  ],
+};
+
+describe('<Query />', () => {
+  it('returns the data', async () => {
+    const graphQL = createGraphQL({PetQuery: mockData});
+    const renderPropSpy = jest.fn(() => null);
+
+    await mountWithGraphQL(<Query query={petQuery}>{renderPropSpy}</Query>, {
```
vsumner, line 31: Nice! This is a great start. I'd also love to add a way to test the SSR issue that this component solves. But I don't think that is required for this PR.

packages/react-graphql/package.json
```diff
@@ -25,7 +25,6 @@
   "homepage": "https://github.com/Shopify/quilt/blob/master/packages/react-graphql/README.md",
   "dependencies": {
     "@apollo/react-common": ">=3.0.0 <4.0.0",
-    "@apollo/react-components": ">=3.0.0 <4.0.0",
     "@apollo/react-hooks": ">=3.0.0 <4.0.0",
```
vsumner, line 28: Should this have been removed?

packages/react-graphql/src/Query.tsx
CameronGorrie: Yes it should. Chris added some context regarding the query options type here: https://github.com/Shopify/quilt/pull/1519#discussion_r442494059

packages/react-graphql/src/tests/Query.test.tsx
CameronGorrie: I could add a set of tests to ensure SSR behaviour. Do you feel these are implementation details of `useQuery` that are already tested though? https://github.com/Shopify/quilt/blob/master/packages/react-graphql/src/hooks/tests/query-ssr.test.tsx

packages/react-graphql/package.json
CameronGorrie: Yes. Before this package was just re-exporting the `Query` component from Apollo which was imported from `@apollo/react-components`. Now that we're defining our own component the import can be safely removed.

packages/react-graphql/src/tests/Query.test.tsx
vsumner: I was referring to the double call on ssr., but I think we have good coverage overall.

PR #1521, opened 2020-06-17
✅ port expanded e2e tests from preact library

TheMallen:
When writing the preact variant of the testing library I had written an expanded e2e test suite for it to ensure that everything worked. This PR ports those tests back to the react library. 

These will be super helpful for us when we try to refactor it into a common core with individual libraries on top.


packages/react-testing/src/tests/e2e.test.tsx
```diff
@@ -1,218 +1,471 @@
...
-
-      return clicked ? (
-        <div>Clicked!</div>
-      ) : (
-        <Clickable onClick={setClicked.bind(null, true)} />
-      );
-    }
+  it('can output structured debug strings', () => {
+    const wrapper = mount(
+      // eslint-disable-next-line @shopify/jsx-prefer-fragment-wrappers
```
vsumner, line 33: In web I disabled this for all test files. Would it be valuable to do that here?

Issue #1504, opened 2020-06-10
`parseGid` does not take params into account


d9su:
## Overview

`parseGid` reports "invalid gid" error if the GID contains params, because [the regex it uses doesn't take those into account](https://github.com/Shopify/quilt/blob/master/packages/admin-graphql-api-utilities/src/index.ts#L1). For example `gid://shopify-deliveries/Delivery/1912?model=Delivery` would be reported as invalid.

To fix it, we can either adjust the regex, or manually trim the content after `?`. We should also update `composeGid` to add the ability for to compose with params.

## Consuming repo

https://github.com/Shopify/deliveries/issues/984

---

## Checklist

- [x] Please delete the labels section before submitting your issue
- [x] I have described this issue in a way that is actionable (if possible)

d9su:
Should be closed by #1524

PR #1524, opened 2020-06-18
Accept query params in gid

jsaubry:
## Description

[This PR in Core](https://github.com/Shopify/shopify/pull/244251) will be adding query params to some GIDs, `gid://shopify/OnlineStoreProductPage/1?sessionId=xyz`. We don't need to be able to extract those query params in the front-end, they are only there to distinguish objects with the same ID. This is required to make sure Apollo doesn't poison its cache with unrelated objects.

More context can be found on [this issue](https://github.com/Shopify/shopify/issues/236662).

Also related to https://github.com/Shopify/quilt/issues/1504

## Type of change

- [x] admin-graphql-api-utilities Minor: New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/packages/admin-graphql-api-utilities/CHANGELOG.md
+++ b/packages/admin-graphql-api-utilities/CHANGELOG.md
@@ -7,6 +7,10 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## [Unreleased] -->
 
+## [Unreleased]
+
+- Let `parseGid` parse GIDs containing query parameters.
+
 ## [0.0.4] - 2019-01-09
 
 - Start of Changelog
--- a/packages/admin-graphql-api-utilities/src/index.ts
+++ b/packages/admin-graphql-api-utilities/src/index.ts
@@ -1,7 +1,7 @@
-const GID_REGEXP = /\/(\w+(-\w+)*)$/;
+const GID_REGEXP = /\/(\w+(-\w+)*)(?:\?(?:[\w-_]+=[\w-_]+&)*(?:[\w-_]+=[\w-_]+))?$/;
 
 export function parseGid(gid: string): string {
-  // appends forward slash to help identify invalid id
+  // prepends forward slash to help identify invalid id
   const id = `/${gid}`;
   const matches = GID_REGEXP.exec(id);
   if (matches && matches[1] !== undefined) {
--- a/packages/admin-graphql-api-utilities/src/test/index.test.ts
+++ b/packages/admin-graphql-api-utilities/src/test/index.test.ts
@@ -28,6 +28,24 @@ describe('admin-graphql-api-utilities', () => {
       const gid = `gid://shopify/Section/${id}`;
       expect(parseGid(gid)).toStrictEqual(id);
     });
+
+    it('returns the id portion of a gid with a query parameter', () => {
+      const id = v4();
+      const gid = `gid://shopify/Section/${id}?foo=bar`;
+      expect(parseGid(gid)).toStrictEqual(id);
+    });
+
+    it('returns the id portion of a gid with multiple query parameters', () => {
+      const id = v4();
+      const gid = `gid://shopify/Section/${id}?foo=bar&baz=0`;
+      expect(parseGid(gid)).toStrictEqual(id);
+    });
+
+    it('returns the id portion of a gid with a query parameter with extended character set', () => {
+      const id = v4();
+      const gid = `gid://shopify/Section/${id}?foo-a=bar_baz`;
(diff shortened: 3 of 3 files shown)


BPScott:
Would we also ever want to expand `composeGid` so that you can create gids with query parameters?

jsaubry:
I agree with you Evan, I think it would be preferable to have a better API, but I'm not planning here to to anything breaking. However, I think that even if we were to use/return an URI object, we would still need to craft a regex to extract the numeric id from it.

> Would we also ever want to expand composeGid so that you can create gids with query parameters?

This is out of the scope of what I need, to I can't really answer that question. However, if you think that both features need to come together I can add it to this PR, it should be quite simple

eapache:
>  I think that even if we were to use/return an URI object, we would still need to craft a regex to extract the numeric id from it.

I don't think so? Instead of a regex it would just be `uri.path.last.to_i`.

alexandcote:
Thanks for your contribution. 💯

The regex probably needs to support some special character for example, `-`, `$`, `%`, etc...
For that reason, I think we should loose the regex. I propose something like
```
\/(\w+(-\w+)*)((?:\?.*)?)$
```

Then if you want to validate that the query params are valid and you would like to return the query params, you can use `new URLSearchParams(match[3])`.

I think you can probably change the API to return the query params if you add an argument to the function. Something like this: 

```
interface ParsedGID {
  id: string;
  params: URLSearchParams;
}

export function parseGid(gid: string, parseParams: false): string
export function parseGid(gid: string, parseParams: true): ParsedGID
export function parseGid(gid: string, parseParams = false): ParsedGID | string {
 ...
}
```

jsaubry:
> Then if you want to validate that the query params are valid and you would like to return the query params, you can use new URLSearchParams(match[3]).

I'll do the query param validation.

However, I don't think I can do the proposed change to the API without going through breaking the API. I don't really know the TS type system, but I suspect that having a "larger" return value is a breaking change and I would prefer to not have to do breaking changes..

BPScott:
> @BPScott , this is out of the scope of what I need, to I can't really answer that question. However, if you think that both features need to come together I can add it to this PR, it should be quite simple

No need for it in this PR, I'm not sure of a use-case atm :)

> I think you can probably change the API to return the query params if you add an argument to the function

I'm not a fan of overloading functions so they return different things based on args, I'd rather have a distinct `parseGidWithParams` that would return a thing of shape `ParsedGID`,  while keeping `parseGid` always returning `string`

jsaubry:
> I'll do the query param validation.

I changed my mind again. After playing with `URLSearchParams` and reading the doc, there's no clear ways of using it to validate urls/params. It does some weird things when you pass it garbage.

So, in conclusion, I won't update the PR and I would like to merge it like it is right now. It's not perfect, it doesn't play nice with url encoded payloads, but to get something robust and still simple, we would need to pull in extra dependencies. However, I don't know how this project like external dependencies.

Opinions?

jsaubry:
**Patch note**

I've been talking more with Brian from OnlineStore Editor and Alex, and everyone seems a little bit uncomfortable about using a regex rather than a real URI parser. I 100% agree with with this sentiment, but I think it would be really hard to replicate the current behavior we have with a strict URI parser. Some [tests](https://github.com/Shopify/quilt/blob/f81b1ff9e16a6b7139319ef4315cd9eae25de4d9/packages/admin-graphql-api-utilities/src/test/index.test.ts#L16) already shows that we accept/expect GID not using an URI format.

From my discussion with Brian
> For instance I've seen a lot of people use ids like `someTestId` without the true structure in tests.

I think this means it would be really hard to use a better URI parser.

However, I still changed the PR a little, I added the characters `-` and `_` as valid characters for query params. To be fully URI compliant we would need even more character, but we know that we will soon be needing those 2 characters in the OnlineStore Editor.

eapache:
> we accept/expect GID not using an URI format.

What? Why? Is core returning GIDs in these formats (it shouldn't) or has the method been hijacked for things that aren't actually GIDs?

jsaubry:
I have no ideas. But if you look at those 2 tests

```typescript
    it('throw Error for an invalid id', () => {
      const key = 'gid://shopify/Section/';
      ['@', '-1', '!1'].forEach(id =>
        expect(() => parseGid(`${key}/${id}`)).toThrow(
          `Invalid gid: ${key}/${id}`,
        ),
      );
    });

    it('returns the id portion of an unprefixed gid', () => {
      ['1', '1a', v4()].forEach(id => expect(parseGid(id)).toStrictEqual(id));
    });
```

it clearly shows that we expect this function to be able to work on GIDs **and** unprefixed GIDs.

PR #1528, opened 2020-06-19
Fix typo in ReactServerNoResponseError message

paarthmadan:
## Description
Fixed a typo in one of the error messages present in the `quilt_rails` package. The error is raised in the `react_renderable` file under the name of `ReactServerNoResponseError`. The initial message had a typo for "presists" which I would imagine is supposed to be "persists" (though it may be a term I'm not aware of :) )

## Type of change
Spelling fix in `quilt_rails`

- [x] `quilt_rails` Patch: Bug (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/gems/quilt_rails/CHANGELOG.md
+++ b/gems/quilt_rails/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 <!-- ## [Unreleased] -->
 
+### Fixed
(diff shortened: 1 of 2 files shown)


paarthmadan:
@ismail-syed Sorry for the basic question, but, does this change require a changelog entry, and would it be considered a breaking change. This is a fairly small update, but I wouldn't want an external service that relies on the message of the error, for instance, to break.

Just want to better understand if this is good to merge into master 😄

TheMallen:
@paarthmadan  as a typo it should be ok for it to be a patch

paarthmadan:
Thanks for the information, @TheMallen – that helps! I've added a `changelog` entry under the "Fixed" heading.

PR #1530, opened 2020-06-22
[react-graphql] Update to documentation and when to use Preload, Prefetch, KeepFresh

oluwatimio:
## Description

The documentation does not explicitly say when we need to use `MyComponent.Preload` or not.
## Type of change
- [x] react-graphql Patch: Documentation fix (non-breaking change which fixes an issue or adds documentation)

## Checklist


packages/react-async/README.md
```diff
@@ -173,6 +173,31 @@ const MyModalComponent = createAsyncComponent({
...
+```tsx
+const MyComponent = createAsyncComponent({
+  load: () => import(/* webpackChunkName: 'MyComponent' */ './MyComponent'),
+  usePreload: () => usePreload(MyQuery),
+});
+
+<MyComponent.Preload />;
+```
+
+When relying on the library by not including usePreload; in this case, we do not need to have.
```
TheMallen, line 195: This sentence is a bit unclear and fragemented,  maybe something like:

 When relying on the default behaviour we do not need to render `MyComponent.Preload`

oluwatimio: Thank you @TheMallen! .. made the change

packages/react-async/README.md
```diff
@@ -173,6 +173,37 @@ const MyModalComponent = createAsyncComponent({
...
+```tsx
+const MyComponent = createAsyncComponent({
+  load: () => import(/* webpackChunkName: 'MyComponent' */ './MyComponent'),
+  usePreload: () => usePreload(MyQuery),
+});
+
+<MyComponent.Preload />;
+```
+
+When relying on the default behaviour we do not need to render `MyComponent.Preload`
```
attila-berki, line 195: I would also mention the `defer` property and the fact that it is ignored / overwritten by the `.Preload` / `.Prefetch` components.

oluwatimio:
Just merged master into this branch. Are we good to merge this @TheMallen and @attila-berki ?

PR #1533, opened 2020-06-23
Update README.md

gustavodegeus:
## Description
Found a extra `:` while following the readme instructions.

Fixes (issue #)

## Type of change
- [X]  Patch: Bug (non-breaking change which fixes an issue)

## Checklist


michenly:
I just removed the demo app instruction in favor of shopify-app coming up. But thanks for helping out!

PR #1536, opened 2020-06-24
Use China CDN for sewing kit assets if countryCode is set to CN

yishus:
## Description

Fixes (issue #)

## Type of change

## Checklist

### DIFF
--- a/packages/sewing-kit-koa/src/middleware.ts
+++ b/packages/sewing-kit-koa/src/middleware.ts
@@ -37,6 +37,7 @@ export default function middleware({
       assetPrefix,
       userAgent: ctx.get(Header.UserAgent),
       manifestPath,
+      countryCode: ctx.get('X-Ip-Metadata'),
     });
 
     setAssets(ctx, assets);
--- a/packages/sewing-kit-koa/src/assets.ts
+++ b/packages/sewing-kit-koa/src/assets.ts
@@ -3,7 +3,7 @@ import {join} from 'path';
 import {readJson} from 'fs-extra';
 import appRoot from 'app-root-path';
 
-import {Manifest} from './types';
+import {Asset, Manifest} from './types';
 import {Manifests} from './manifests';
 
 export {Asset} from './types';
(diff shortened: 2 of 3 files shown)


shioyama:
Thanks @yishus ! This is great, helps clarify what this change would involve. We'll discuss and come back with more feedback asap.

packages/sewing-kit-koa/src/middleware.ts
```diff
@@ -37,6 +37,7 @@ export default function middleware({
       assetPrefix,
       userAgent: ctx.get(Header.UserAgent),
       manifestPath,
+      countryCode: ctx.get('X-Ip-Metadata'),
```
shioyama, line 40: Assuming it will be the same as the header on core, `X-Ip-Metadata` will not be simply the country code but a string containing many things, including the country code.

Something like this:

```
asn:16276;anon:0;anonVpn:0;hostProvider:0;pubProxy:0;torExit:0;countryCode:CA
```

Here's the [code in storefront-renderer](https://github.com/Shopify/storefront-renderer/blob/eb55c8d6a8553bd11f4f4ad5c9abef3220195784/app/middlewares/ip_metadata_middleware.rb#L6) which parses out the country code from it.

yishus:
@shioyama ah i was thinking of cleaning this up and trying it out with web before sharing with you both, lemme get back soon also after those are done

shioyama:
Ok no worries! It's fine for this to be rough right now, it's mostly serving as a proof that this is possible and how it would look. 🙂

jaryl:
This looks good, though we might want the switching logic to be configurable at the app level (at least an on/off switch). I don't know how many apps are built on quilt, but we might want to be more conservative in our approach because we don't know what sort of issues might arise if an app that isn't prepared for CDN switching starts using this.

At this point, we might want to discuss what mechanism we have to turn on/off CDN switching. We don't have access to beta flags directly, so maybe env vars?

yishus:
Closing this for now

Issue #1522, opened 2020-06-17
Explore `@sewing-kit` for quilt's builds


TheMallen:
## Overview
We should rebuild quilt's build setup from the ground up to using the `@sewing-kit/*` ecosystem of packages (referred to in general as sewing-kit-next) being championed and prototyped by @lemonmade

## Motivation
There are a few reasons to do this:
- Organizationally, we want to be able to recommend a build-tool for other npm package publishing repos at Shopify (especially multi-package repos) and sewing-kit `0.x` doesn't handle this usecase well (if at all)
- @lemonmade  has been working on this stuff in relative isolation for a while and we need some on-ramps to be able to gain understanding before we can even begin to help
- Quilt itself has a poor build situation. We can't build assets other than ts in this repo and we only build legacy commonjs files. Sewing-kit-next can help us on both counts. **We should see immediate impact in the form of greater tree-shakeability, and thus reduced bundle sizes for our consumers**

TheMallen:
## Resources
The hardest part of this is that, since the tool is currently a one-man show under active development, there is little documentation and relatively few examples to look at. We will be able to help make up this gap, once we have worked with and on it a bit more,  but for now it will mostly be a "read the code and figure it out" situation unfortunately. 

### Repos to look at

Luckily, @lemonmade  has made extensive use dogfooding the sewing-kit-next in many of the packages that are being used in some more experimental projects. This means there are some repos with similar structures to quilt (in one case an actual fork / re-imagining) that we can learn from.

-the [sewing-kit-next monorepo]( https://github.com/lemonmade/sewing-kit), useful to see the source but also as an example of a setup since it currently builds itself
- the [alternate quilt monorepo](https://github.com/lemonmade/quilt), useful as a starting point to see some of the existing packages already converted and how a shared plugin might work for us
- the [https://github.com/lemonmade/remote-ui](remote-ui) repo, just another example really

ismail-syed:
I've tagged this issue with the Explore label and renamed to highlight this work as our exploration. I've open'd https://github.com/Shopify/quilt/issues/1592 and added it to our board to cut an official release in our build phase.

PR #1538, opened 2020-06-24
[quilt] Build Quilt using sewing-kit-next

b-ryu:
## Description

Related to #1522

This PR integrates [sewing-kit-next](https://github.com/lemonmade/sewing-kit) into Quilt's builds.

- `yarn build`
- `yarn test`
- `yarn lint`

### Some notes
- I've noticed some intermittent issues with Watchman when trying to run tests, especially right after a fresh build. Usually running `watchman shutdown-server` and `dev up` after restarting the terminal after a while fixes things. I suspect it may have something to do with the volume of built files (**note:** look into configuring Watchman)
- This PR bumps the maximum memory allowed to VS Code's builtin TS plugin. Adding a `tsconfig.json` to the workspace root causes the plugin to traverse the _entire_ workspace when you open a `.ts` file not included in one of the referenced projects/packages (opening a `.ts` file in a project/package loads the TS for that package only, as before). This behaviour also occurs in other [similarly structured workspaces](https://github.com/lemonmade/quilt), the difference being that `Shopify/quilt` is relatively large so the plugin appears to run out of memory halfway, causing it to hang and crash repeatedly. Increasing the memory limit seems to fix the problem, though it might be something to look into further


.eslintrc.js
```diff
@@ -1,18 +1,21 @@
 module.exports = {
   extends: [
-    'plugin:@shopify/typescript',
-    'plugin:@shopify/typescript-type-checking',
-    'plugin:@shopify/react',
-    'plugin:@shopify/jest',
-    'plugin:@shopify/prettier',
+    'plugin:@sewing-kit/typescript',
```
ismail-syed, line 3: Where do these configs live?

b-ryu: [Here](https://github.com/lemonmade/sewing-kit/tree/main/packages/eslint-plugin/src/config), I believe.

ismail-syed: Found it, https://github.com/lemonmade/sewing-kit/tree/main/packages/eslint-plugin/src/config

Not related to quilt. It's interesting that there's a`@sewing-kit/eslint-plugin` vs `@sewing-kit/plugin-eslint`. One is a plugin for sewing-kit and one is an eslint plugin.

ismail-syed: Sorry, github is being wonky with this big PR. Comment thread didn't' refresh and the page is freezing.

.vscode/settings.json
```diff
@@ -23,6 +23,7 @@
     "build": true,
     "**/*.js": true
   },
+  "typescript.tsserver.maxTsServerMemory": 8000,
```
ismail-syed, line 26: Can you share some re-producible steps for the issues you had with VSCode, please? I can try to take a deeper look.

b-ryu: Sure thing:
1. Pull this branch down and open it in VS Code
2. Get rid of the `typescript.tsserver.maxTsServerMemory` setting fix
3. Close all tabs in and close VS Code
4. Open Quilt in VS Code
5. Open `sewing-kit.config.ts` in the workspace root
6. On my machine at least, the builtin TS/JS plugin hangs and never stops initializing, as noted in the bottom toolbar. The plugin eventually restarts itself a couple times and then crashes, leaving a warning popup

Screen Shot 2020-07-22 at 2 52 57 PM

In VS Code you can use the command `>Open TS Server log` to see the plugin logs.

config/sewing-kit/index.ts
```diff
@@ -0,0 +1,73 @@
...
+      }
+    }),
+  };
+}
+
+export function quiltPackage({binaryOnly = true, jestEnv = 'jsdom'} = {}) {
+  return createComposedProjectPlugin<Package>('Quilt.Package', [
+    javascript(),
+    typescript(),
+    react(),
```
ismail-syed, line 35: We have some quilt packages that don't use any react (eg. `address-consts`). Do they still need the `react` plugin?

b-ryu: Probably not, I can make that configurable via arguments instead 👍

config/sewing-kit/index.ts
```diff
@@ -0,0 +1,73 @@
+import {
+  Package,
+  createComposedProjectPlugin,
+  createProjectTestPlugin,
+  createProjectBuildPlugin,
+} from '@sewing-kit/plugins';
+import {react} from '@sewing-kit/plugin-react';
```
ismail-syed, line 7: Something to keep in mind for the future. It would be nice to have some of these common plugins re-exported  from `@sewing-kit/plugins` so devs can do something like: 

```tsx
import {
   // other stuff
  react, 
  javascript, 
  typescript,
 } from '@sewing-kit/plugins';
```

instead of 

```tsx
import {react} from '@sewing-kit/plugin-react';
import {javascript} from '@sewing-kit/plugin-javascript';
import {typescript} from '@sewing-kit/plugin-typescript';
```

@lemonmade do you foresee any issues with this?

config/sewing-kit/index.ts
```diff
@@ -0,0 +1,73 @@
...
+      ) {
+        return [pluginName, {loose: true}];
+      } else {
+        return plugin;
+      }
+    }),
+  };
+}
+
+export function quiltPackage({binaryOnly = true, jestEnv = 'jsdom'} = {}) {
```
ismail-syed, line 31: What's this `binaryOnly` flag for? It's defaulted to true and I don't see it being used anywhere else?

b-ryu: I think I just copied that from Chris' [Quilt config](https://github.com/lemonmade/quilt/blob/main/config/sewing-kit/index.ts). Wasn't sure if we wanted `.mjs` and `.esnext` files in our builds. I can get rid of it and hardcode that to be `true` instead for now.

ismail-syed: Let's leave them out for now. Adding support for `.mjs` and `.esnext` feels out of scope in validating sewing-kit-next in quilt.

packages/address-consts/package.json
```diff
@@ -24,7 +24,7 @@
   },
   "homepage": "https://github.com/Shopify/quilt/blob/master/packages/address-consts/README.md",
   "files": [
-    "dist/*",
+    "build/*",
```
ismail-syed, line 27: Is there a reason why these are all changing to `build`? Anyway we can keep it as `dist`?

packages/address-consts/package.json
```diff
@@ -3,8 +3,8 @@
   "version": "2.0.1",
   "license": "MIT",
   "description": "Constants and types relating to `@shopify/address`",
-  "main": "dist/src/index.js",
-  "types": "dist/src/index.d.ts",
+  "main": "index.js",
```
ismail-syed, line 6: Is outputting an `index.js` in the root a defaults from sewing-kit next? I personally like when all the build output is in one folder that way it's easier to undo/clear caches. Probably worth a discussion with the team though.

packages/address-consts/package.json
b-ryu: Looking through the plugin code I don't believe it's a hard requirement, but I think it's the preferred default. I'll see if keeping it as `dist` changes anything 👍

packages/address-consts/package.json
b-ryu: I can look into this; I think it's the default, not sure if it's configureable (based on `PackageEntryOptions` in SK-next). FWIW, I added a `yarn clean` script to `package.json` that clears away any built files/entrypoints.

tsconfig.json
```diff
@@ -0,0 +1,76 @@
+{
```
ismail-syed, line 1: Does sewing-kit-next expect the `tsconfig.json` to live in the root dir?

packages/address-consts/package.json
ismail-syed: After seeing the use case for multiple entries into a package (eg. `web-workers`), this sounds like a reasonable tradeoff. 

```tsx
   pkg.entry({root: './src/index'});
   pkg.entry({name: 'babel', root: './src/babel-plugin'});
   pkg.entry({name: 'webpack', root: './src/webpack-parts'});
   pkg.entry({name: 'worker', root: './src/worker'});
```

This whole automation of a package entry is really neat!

package.json
```diff
@@ -1,65 +1,77 @@
...
+  "author": "Shopify Inc.",
   "workspaces": [
     "packages/*"
   ],
+  "scripts": {
+    "_test": "NODE_ICU_DATA=node_modules/full-icu yarn sk test",
+    "build": "sk build",
+    "check": "lerna run check",
+    "ci:lint-docs": "yarn generate docs && test -z \"$(git status --porcelain)\" || echo 'The root README has not been updated. Run `yarn generate docs` in the root of your quilt directory and try again.'",
+    "clean": "rimraf './packages/*/build' './packages/*/*.d.ts' './packages/*/*.js' './packages/*/*.esnext' './packages/*/*.mjs' '.sewing-kit'",
```
ismail-syed, line 24: Clearing all these files manually feels like a guessing game to me. We should rely on sewing-kit to handle this. 

There are a number of files that match the following  `'./packages/*/*.d.ts' './packages/*/*.js'` that are removed in this PR. Did any of these get accidentally picked up by a `yarn run clean` or are all of them covered by sewing-kit's `pkg.entry` automation?

I don't expect an answer for this 😅 Just more things to consider during our validation of sewing-kit-next.

packages/address-consts/package.json
ismail-syed: Something to verify. Do to generated root level entry points respect the output folder in the tsconfig. 

* Update the output folder in `packages/web-worker/tsconfig.json` to `dist`
* Run a build 
* Verify that all the root level entry points generated for the following include a `./dist` 

```tsx
  pkg.entry({root: './src/index'});
  pkg.entry({name: 'babel', root: './src/babel-plugin'});
  pkg.entry({name: 'webpack', root: './src/webpack-parts'});
  pkg.entry({name: 'worker', root: './src/worker'});
```

package.json
b-ryu: I think I deliberately got rid of some `packages/*/*.d.ts` and `packages/*/*.js` files (also added them to `.gitignore`) since we didn't need them anymore with our new generated entrypoints, but it's probably worth it for me to go back and double-check that they were all actually just entrypoints 👍

tsconfig.json
b-ryu: Not certain, but I [believe so](https://github.com/lemonmade/sewing-kit/blob/main/packages/plugin-typescript/src/plugin-typescript.ts#L324)? I remember getting some errors about not finding a `quilt/tsconfig.json` when playing around with with different `tsconfig.json` setups.

config/sewing-kit/index.ts
TheMallen: disagree

config/sewing-kit/index.ts
TheMallen: I agree the core plugins are annoying to remember what all the package names are

lemonmade: This is what composed plugins are for (e.g., https://github.com/lemonmade/quilt/blob/main/packages/sewing-kit-plugins/src/index.ts#L29), so I don't think it's needed.

config/sewing-kit/index.ts
ismail-syed: I stand corrected. @b-ryu, sorry for the confusion. We'll need this.

.gitignore
```diff
@@ -44,3 +40,14 @@ tmp
...
 *.tsbuildinfo
+
+# Sewing Kit
+.sewing-kit
+build
+packages/*/*.esnext
+packages/*/*.node
+packages/*/*.mjs
+packages/*/*.d.ts
+packages/*/*.js
```
lemonmade, line 51: The thing to watch out for here is that we sometimes nest `.eslintrc.js` or similar in subdirectories, and we do generally want to include those files.

.travis.yml
```diff
@@ -15,12 +15,21 @@ matrix:
     - <<: *node_container
       name: 'Node - build, type-check, and end-to-end tests'
       script:
-        - yarn build --verbose
-        - yarn test:ci --testPathPattern react-server address
+        - yarn build
+        - yarn type-check
```
lemonmade, line 19: `build` effectively includes a `type-check`, because part of the build is generating the TypeScript outputs

config/sewing-kit/index.ts
```diff
@@ -0,0 +1,74 @@
...
+
+          hooks.babelConfig?.hook(addLegacyDecoratorSupport);
+        });
+      });
+    }),
+    createProjectTestPlugin('Quilt.PackageTest', ({hooks}) => {
+      hooks.configure.hook(hooks => {
+        hooks.jestEnvironment?.hook(() => jestEnv);
+
+        hooks.jestConfig?.hook(config => ({
```
lemonmade, line 58: There are more specific hooks for `jestTransforms` and the watch ignore patterns

package.json
```diff
@@ -1,65 +1,77 @@
...
+    "generate": "plop",
+    "generate:package": "plop package && yarn plop docs && prettier --write \"tsconfig.json\"",
+    "lint": "sk lint",
+    "lint:changelogs": "node scripts/lint-changelogs.js",
+    "publish": "lerna publish",
+    "release": "yarn lint:changelogs && lerna publish",
+    "release-beta": "lerna publish --dist-tag beta",
+    "pretest": "yarn build",
+    "test": "yarn _test",
```
lemonmade, line 33: I think this may be changing the default `test` command to be in watch mode, which may not be desirable

packages/address-consts/package.json
lemonmade: Yeah, I don't love littering the root of a package with all its entry files, but unfortunately there's not really any other way to do multi-entry packages consistently.

packages/react-cookie/package.json
```diff
@@ -3,8 +3,8 @@
   "version": "0.0.30",
   "license": "MIT",
   "description": "Cookies in React for the server and client",
-  "main": "dist/src/index.js",
-  "types": "dist/src/index.d.ts",
+  "main": "index.js",
+  "types": "index.d.ts",
```
lemonmade, line 7: For packages to get the full benefits of the different build outputs, you'll eventually need to also add `esnext`/ `module` entries to package.json (and, if you're interested, the future-looking `exports` map). Example: https://github.com/lemonmade/quilt/blob/main/packages/quilt/package.json#L10-L32

.vscode/settings.json
ismail-syed: I was able to repro this on first attempt. I tried a few hacks to see what would happen (eg. move the root `tsconfig.json` to `packages/tsconfig.json`). That seemed to help. I reverted my changes and cleaned out my git env and I'm not able to reproduce this anymore :/

b-ryu: It might have something to do with my own local setup then 🤔 @lemonmade @TheMallen is this an issue for you guys as well?

TheMallen: I'll try it out after our 1:1

packages/react-server/src/webpack-plugin/test/webpack-plugin.test.ts
```diff
@@ -371,6 +371,18 @@ const universal = {
   resolve: {
     modules: ['node_modules', path.resolve(__dirname, '${basePath}')],
   },
+  module: {
```
b-ryu, line 374: I had to modify this test webpack config in order to get these tests to pass. Otherwise they would error out and complain about not being able to import from `.mjs` or parse `node` modules. Not sure if this is something we should worry about, maybe someone with more webpack experience can weigh in?

Screen Shot 2020-07-28 at 12 11 56 PM

packages/address-consts/package.json
b-ryu: Did some digging. Apparently setting `outDir` in `tsconfig.json` only affects TS output. It seems that other outputs are [hardcoded](https://github.com/lemonmade/sewing-kit/blob/main/packages/model/src/base/fs.ts#L46) to go into the `build` folder (the generated entrypoints point to the proper respective folder).

packages/react-i18n/src/babel-plugin/test/generate-dictionaries.test.ts
```diff
@@ -27,14 +27,18 @@ describe('generate-dictionaries', () => {
...
-    await generateTranslationDictionaries(['de'], {
+    await generateTranslationDictionaries(['es', 'de'], {
       rootDir,
       fallbackLocale: 'en',
     });
 
     // eslint-disable-next-line @typescript-eslint/no-var-requires
     const dictionary = require(`${rootDir}/index`);
-    expect(Object.keys(dictionary.default)).toStrictEqual(['de']);
+    expect(Object.keys(dictionary.default)).toStrictEqual(['es', 'de']);
```
b-ryu, line 37: For some reason, Jest wasn't clearing the module cache properly, so this test kept re-importing the same generated dictionary as the test above, with the `es` translations. I modified this test so that it'd pass CI, but I can look deeper into this.

config/eslint.js
```diff
@@ -1,13 +0,0 @@
-// We need to specify the import/resolver setting in an extendable
```
b-ryu, line 1: This file was added in [this commit](https://github.com/Shopify/quilt/commit/6b3794df2de138c9aeadc64cb93d3c429beeeca0) but then became unused in [this commit](https://github.com/Shopify/quilt/commit/857890c91dda4fa9b36657dbdc8364bdb329c9c6). AFAIK it's no longer necessary?

packages/address-consts/package.json
ismail-syed: Thanks for digging into this. I'm ok with standardizing on our tooling to use `build` as the default output dir. We'll just need to make sure we update all our cases where we hard-coded `dist`. [`@shopify/polyfills`](https://github.com/Shopify/quilt/blob/master/packages/polyfills/src/index.ts#L35) comes to mind. It's hurt us in the [past](https://github.com/Shopify/quilt/commit/33f6eccb500ac576d42719260c31d095fbeabac7). 

We'll probably want an escape hatch out of this in the future though, unless it's already possible today (cc/ @lemonmade)

packages/react-server/src/webpack-plugin/test/webpack-plugin.test.ts
ismail-syed: I suspect this is because the `package.json`s include `"module": "index.mjs"`. Perhaps webpack takes precedence on the `module` over `main` resolution when it runs builds? 

I'm not an expert on `.mjs` yet, maybe @TheMallen can clarify.

ismail-syed: I'm thinking we should maybe run a [beta release](https://github.com/Shopify/quilt/blob/master/documentation/guides/release-and-deploy.md#step-1---publish-a-beta-release-for-testing) for these `sewing-kit-next` built packages. This can help verify that

* This doesn't break and dev/prod builds in sewing-kit apps. This seems [supported](https://github.com/Shopify/sewing-kit/blob/master/packages/sewing-kit/src/tools/webpack/config/rules.ts#L426-L430) in sewing-kit already but worth verifying things don't break. 
* Also helps validate tree-shaking benefits of esmodule packages. We can dogfood this in the shrink-ray app and use shrink-ray to verify the bundle reduction from treeshaking

packages/polyfills/src/index.ts
```diff
@@ -32,7 +32,7 @@ export const polyfills: {[polyfill: string]: PolyfillDescriptor} = {
 export function mappedPolyfillsForEnv(
   env: 'node' | 'jest' | string[],
 ): {[key: string]: string} {
-  const prefix = `@shopify/polyfills/dist/src`;
+  const prefix = `@shopify/polyfills/build/ts`;
```
b-ryu, line 35: I defaulted this to `build/ts` to match our previous behaviour with `dist/src`. However, this probably means we won't need to include the other build types for this package? I don't see how consumers would access them, unless we rework this package to add an entrypoint for each polyfill?

b-ryu: **Update:** changed this to `cjs` since we shouldn't be looking into the `ts` builds. 

**Question:** should we have multiple builds/entrypoints for this package and have this function map to the entrypoints? Or should we have it map to just the CommonJS builds? I feel like there's not a lot of value in trying to make it treeshakeable since `sideEffects` is `true` anyway, and most of the files are only a few lines at most, but maybe someone with more context can weigh in @lemonmade @ismail-syed

packages/polyfills/src/index.ts
```diff
@@ -32,7 +32,7 @@ export const polyfills: {[polyfill: string]: PolyfillDescriptor} = {
 export function mappedPolyfillsForEnv(
   env: 'node' | 'jest' | string[],
 ): {[key: string]: string} {
-  const prefix = `@shopify/polyfills/build/cjs`;
+  const prefix = `@shopify/polyfills`;
```
b-ryu, line 35: I'm wondering if it would be a terrible thing to just have the CommonJS builds and just have this map to the `build/cjs` instead. Not sure what the impact on treeshaking would be, since this package has `sideEffects` set to `true` anyway.

b-ryu:
Since the SK-next project is still in the exploration phase, we've decided to hold off on merging/releasing this until the build phase. Closing this PR for now.

PR #1543, opened 2020-07-06
Update README.md

ismail-syed:
I believe the docs meant to say this instead.


416serg:
Also 

https://github.com/Shopify/quilt/tree/master/packages/react-bugsnag#createbugsnagclient

it says `createClient`  when it should be `createBugsnagClient`

PR #1548, opened 2020-07-13
Add parseGidWithParams, composeGidFactory and update composeGid to support params

tleunen:
## Description

Following what was done in https://github.com/Shopify/quilt/pull/1524, I'm making a PR to fully support params in `parseGidWithParams` and also `composeGid`. I also added a factory function to create a custom `composeGid` with a specific namespace.

- [x] admin-graphql-api-utilities: Minor

### DIFF
--- a/packages/admin-graphql-api-utilities/src/index.ts
+++ b/packages/admin-graphql-api-utilities/src/index.ts
@@ -1,5 +1,5 @@
 const GID_TYPE_REGEXP = /^gid:\/\/[\w-]+\/([\w-]+)\//;
-const GID_REGEXP = /\/(\w[\w-]*)(?:\?([\w-]+=[\w-]+(&[\w-]+=[\w-]+)*))*$/;
+const GID_REGEXP = /\/(\w[\w-]*)(?:\?(.*))*$/;
 
 interface ParsedGID {
   id: string;
@@ -33,13 +33,7 @@ export function parseGidWithParams(gid: string): ParsedGID {
     const params =
       matches[2] === undefined
         ? {}
-        : matches[2]
-            .split('&')
-            .reduce<Record<string, string>>((acc, keyVal) => {
-              const [key, value] = keyVal.split('=');
-              acc[key] = value;
-              return acc;
-            }, {});
+        : fromEntries(new URLSearchParams(matches[2]).entries());
 
     return {
       id: matches[1],
@@ -49,7 +43,7 @@ export function parseGidWithParams(gid: string): ParsedGID {
   throw new Error(`Invalid gid: ${gid}`);
 }
 
-export function createComposeGidFunction(namescape: string) {
+export function composeGidFactory(namescape: string) {
   return function composeGid(
     key: string,
     id: number | string,
@@ -62,22 +56,12 @@ export function createComposeGidFunction(namescape: string) {
       return gid;
     }
 
-    const queryParams = paramKeys
-      .map(query => {
-        if (params[query] == null) {
-          return null;
-        }
-
-        return `${query}=${params[query]}`;
-      })
-      .filter(Boolean)
-      .join('&');
-
-    return `${gid}?${queryParams}`;
+    const paramString = new URLSearchParams(params).toString();
+    return `${gid}?${paramString}`;
   };
 }
 
-export const composeGid = createComposeGidFunction('shopify');
+export const composeGid = composeGidFactory('shopify');
 
 interface Edge<T> {
   node: T;
@@ -92,4 +76,13 @@ export function keyFromEdges<T, K extends keyof T>(
   key: K,
 ): T[K][] {
   return edges.map(({node}) => node[key]);
+}
+
+// Once we update to Node 12, we can drop this helper to use `Object.fromEntries` instead.
+function fromEntries<K extends string, T>(entries: IterableIterator<[K, T]>) {
+  const obj = {} as Record<K, T>;
+  for (const [key, val] of entries) {
+    obj[key] = val;
+  }
+  return obj;
 }
--- a/packages/admin-graphql-api-utilities/src/index.ts
+++ b/packages/admin-graphql-api-utilities/src/index.ts
@@ -1,4 +1,19 @@
-const GID_REGEXP = /\/(\w+(-\w+)*)(?:\?(?:[\w-_]+=[\w-_]+&)*(?:[\w-_]+=[\w-_]+))?$/;
+const GID_TYPE_REGEXP = /^gid:\/\/[\w-]+\/([\w-]+)\//;
+const GID_REGEXP = /\/(\w[\w-]*)(?:\?([\w-]+=[\w-]+(&[\w-]+=[\w-]+)*))*$/;
+
+interface ParsedGID {
+  id: string;
+  params: Record<string, string>;
+}
+
+export function parseGidType(gid: string): string {
+  const matches = GID_TYPE_REGEXP.exec(gid);
+
+  if (matches && matches[1] !== undefined) {
+    return matches[1];
+  }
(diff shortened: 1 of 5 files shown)


packages/admin-graphql-api-utilities/src/index.ts
```diff
@@ -1,4 +1,21 @@
-const GID_REGEXP = /\/(\w+(-\w+)*)(?:\?(?:[\w-_]+=[\w-_]+&)*(?:[\w-_]+=[\w-_]+))?$/;
+const GID_TYPE_REGEXP = /\/(\w+)\/(\w[\w-]*)(\?([\w-]+=[\w-]+(&[\w-]+=[\w-]+)*))*$/;
```
alexandcote, line 1: Do we need as much validation for an ID? Can we simplify the regex with something like : ```gid:\/\/[\w]+\/(\w*)\/.*```

packages/admin-graphql-api-utilities/src/index.ts
```diff
@@ -10,10 +27,60 @@ export function parseGid(gid: string): string {
   throw new Error(`Invalid gid: ${gid}`);
 }
 
-export function composeGid(key: string, id: number | string): string {
-  return `gid://shopify/${key}/${id}`;
+export function parseGidWithParams(gid: string): ParsedGID {
+  // appends forward slash to help identify invalid id
+  const id = `/${gid}`;
+  const matches = GID_REGEXP.exec(id);
```
alexandcote, line 33: Same question here? This GID is like an URL, can we use URLSearchParams instead? I have no idea of the cost tho.

packages/admin-graphql-api-utilities/src/index.ts
```diff
@@ -10,10 +27,60 @@ export function parseGid(gid: string): string {
...
+    params: Record<string, string> = {},
+  ): string {
+    const gid = `gid://${namescape}/${key}/${id}`;
+    const paramKeys = Object.keys(params);
+
+    if (paramKeys.length === 0) {
+      return gid;
+    }
+
+    const queryParams = paramKeys
```
alexandcote, line 67: Do we accept an array of query params? Something like `foo[]=hello&foo[]=world` if so, I think this will not work. Maybe it's overkill 🤷🏼‍♂️ Should we use URLSearchParams instead also here?

tleunen: Yeah, didn't want to support arrays at first because I don't think we should really be over-using these params 😅

I believe we could go with `URLSearchParams` to parse/compose the params, but not with the full URL object because it seems we should still support unprefixed gid for some reasons. Not so sure why we would even have unprefixed gids but it's in the tests lol

packages/tsconfig_base.json
```diff
@@ -25,6 +25,7 @@
       "es2016",
       "es2017",
       "es2018",
+      "es2019",
```
BPScott, line 28: @michenly IIRC we chatted about doing this in the global ts configs ages ago, did we ever reach a conclusion?

packages/admin-graphql-api-utilities/src/index.ts
```diff
@@ -10,10 +25,44 @@ export function parseGid(gid: string): string {
...
+
+    return {
+      id: matches[1],
+      params,
+    };
+  }
+  throw new Error(`Invalid gid: ${gid}`);
+}
+
+export function createComposeGidFunction(namescape: string) {
```
BPScott, line 46: `composeGidFactory` perhaps?

packages/tsconfig_base.json
michenly: it did not, in terms of global ts config the PR was [here](https://github.com/Shopify/web/pull/24496). I pause it since Gord had brought up some really good points but I did not have time to look into.

packages/admin-graphql-api-utilities/src/index.ts
BPScott: If we're using URLSearchParams, then could we go even further by using [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL_API) (and then use [url.searchParams](https://nodejs.org/docs/latest-v8.x/api/url.html#url_url_searchparams)) to parse the whole thing instead of writing regexes to do the first bit of parsing?

What's the browser support for URL / URLSearchParams like? at first glance it looks like [it's in everything we support](https://caniuse.com/#search=url)

.nvmrc
```diff
@@ -1 +1 @@
-v10.16.3
+v12.18.2
```
BPScott, line 1: How come this needs a version bump?

web is still running 10.x IIRC so we still need to support that

packages/admin-graphql-api-utilities/src/index.ts
tleunen: There's a test for "unprefixed gid" and forcing a gid to really be a full gid with protocol will be a breaking change :/

.nvmrc
tleunen: Because of `Object.fromEntries`. Same reason why TS needs ES2019.

How come web still uses node 10? 😁 Damn... 😅

packages/admin-graphql-api-utilities/src/index.ts
BPScott: oof. I'm guessing something like `if (!gid.startsWith('gid://') { id = `gid://${gid}` } is just gonna make this more confusing :(

.nvmrc
BPScott: I believe that'd be a question for @donnielrt. But the answer is probably something like "breaking changes in 12.x and folks not yet having the time to diagnose and address them".

End result, I think we still need to keep 10.x support for the time being, and the easiest way to do that is keep the tests running in 10.x, so we probably want to avoid fromEntries for the time being :(

(even though a version bump would [save me some headaches too](https://github.com/Shopify/quilt/pull/1454))

tleunen: Yeah, anyway I wanted to revert the version anyway following your first message since it wouldn't be wise to bump it here if web is not able to easily/quickly do it as well. Having said that, I still think the web (products) teams should spend some time to try and see if their code works in node 12. Node 14 LTS is just a few months away...

tleunen:
@BPScott @michenly @alexandcote - I've updated the PR to remove the usage of `Object.fromEntries` which requires TS to be enabled with ES2019 and the update to Node 12 since we're not ready yet to do these things here.

.nvmrc
donnielrt: > the answer is probably something like "breaking changes in 12.x

Yeah, it was mostly 100s of section tests that were failing, we've embarked on a big effort to fix and modernize tests, which should clear the path for 12.x 🤞

donnielrt: PR in question: https://github.com/Shopify/web/pull/28941. The Supersonic Web project is making a bunch of improvements to the dev experience, we might look at getting in Node 12 as part of that.

tleunen:
Can I get a new review on this please? :) @michenly @alexandcote @BPScott

packages/admin-graphql-api-utilities/src/index.ts
```diff
@@ -1,4 +1,19 @@
-const GID_REGEXP = /\/(\w+(-\w+)*)(?:\?(?:[\w-_]+=[\w-_]+&)*(?:[\w-_]+=[\w-_]+))?$/;
+const GID_TYPE_REGEXP = /^gid:\/\/[\w-]+\/([\w-]+)\//;
+const GID_REGEXP = /\/(\w[\w-]*)(?:\?(.*))*$/;
+
+interface ParsedGID {
```
BPScott, line 4: nitpick: We seem to say `Gid` rather than `GID` elsewhere so should this be `ParsedGid`?

Issue #1550, opened 2020-07-14
[Koa-shopify-auth] missing "/" in redirect url when passing prefix


barrypeng6:
## Overview
If passing prefix `/shopify`
```javascript
app.use(
  shopifyAuth({
    prefix: '/shopify',
    apiKey: SHOPIFY_API_KEY,
    secret: SHOPIFY_SECRET,
    scopes: ['write_orders, write_products'],
    accessMode: 'offline',
    afterAuth(ctx) {
      // do something...
    },
  }),
);
```
Expected redirect URL should be `/shopify/auth/enable_cookies?shop=test.myshopify.com`, but the result is `/shopifyauth/enable_cookies?shop=test.myshopify.com`

Should add `/` between `prefix` and `auth` of URL
```javascript
doesNotHaveStorageAccessUrl: "${prefix}/auth/enable_cookies?shop=${encodeURIComponent(
  shop,
)}",
```
https://github.com/Shopify/quilt/blob/master/packages/koa-shopify-auth/src/auth/client/request-storage-access.ts#L10

ismail-syed:
Thanks for the issue. Does passing in `/shopify/` work?

barrypeng6:
> Thanks for the issue. Does passing in `/shopify/` work?

No. It doesn't work.
I think if passing `/shopify/`, `oAuthStartPath` will be `/shopify//auth`.
So the following will go wrong
```javascript
if (
  ctx.path === oAuthStartPath &&
  !hasCookieAccess(ctx) &&
  !grantedStorageAccess(ctx)
) {
  await requestStorageAccess(ctx);
  return;
}
```
https://github.com/Shopify/quilt/blob/master/packages/koa-shopify-auth/src/auth/index.ts#L62

c5haw:
I have found the same issue when I add the `prefix` option to `shopifyAuth`. I have applied a very crufty fix until this is resolved

```
const prefix = '/my-prefix'
app.use((ctx, next) => {
  if (ctx.path === prefix + 'auth/inline') ctx.path = prefix + '/auth/inline'
  return next()
})
```

It works, but not ideal by any means.

seamusabshere:
We have confirmed that https://github.com/Shopify/quilt/pull/1498 broke this.

IMSupperkaka:
> Thanks for the issue. Does passing in `/shopify/` work?
This problem still exists, now I can only use nginx forwarding to deal with this problem, hope it can be resolved.

[...]

PR #1551, opened 2020-07-14
[Koa-shopify-auth] Add slash in redirect url

barrypeng6:
## Description

Fixes (issue #1550 )
Fix missing `/` between `prefix` and `auth` in redirect URL when passing prefix

## Type of change

- [x] <@shoppify/koa-shopify-auth> Patch: Bug (non-breaking change which fixes an issue)expected)

## Checklist

### DIFF
--- a/packages/koa-shopify-auth/src/auth/client/request-storage-access.ts
+++ b/packages/koa-shopify-auth/src/auth/client/request-storage-access.ts
@@ -4,10 +4,10 @@ const requestStorageAccess = (shop: string, prefix = '/') => {
       function redirect() {
         var targetInfo = {
           myshopifyUrl: "https://${encodeURIComponent(shop)}",
-          hasStorageAccessUrl: "${prefix}auth/inline?shop=${encodeURIComponent(
+          hasStorageAccessUrl: "${prefix}/auth/inline?shop=${encodeURIComponent(
     shop,
   )}",
-          doesNotHaveStorageAccessUrl: "${prefix}auth/enable_cookies?shop=${encodeURIComponent(
+          doesNotHaveStorageAccessUrl: "${prefix}/auth/enable_cookies?shop=${encodeURIComponent(
     shop,
   )}",
           appTargetUrl: "${prefix}?shop=${encodeURIComponent(shop)}"


packages/koa-shopify-auth/src/auth/client/request-storage-access.ts
```diff
@@ -4,10 +4,10 @@ const requestStorageAccess = (shop: string, prefix = '/') => {
       function redirect() {
         var targetInfo = {
           myshopifyUrl: "https://${encodeURIComponent(shop)}",
-          hasStorageAccessUrl: "${prefix}auth/inline?shop=${encodeURIComponent(
+          hasStorageAccessUrl: "${prefix}/auth/inline?shop=${encodeURIComponent(
```
ismail-syed, line 7: `prefix` defaults to `/` so in the case where a prefix isn't passed in, this would now become `//auth/incline...`
For this fix, we'll want to set the default `prefix = ''` in the function args. 

Also, we'll need to apply this change to `appTargetUrl` below. 

@dooman87, does this make sense given your previous PR?

barrypeng6:
> Thanks for the PR. Are we able to add some tests here as well? We've had a few back/forths with this fix so I'd like to maybe see some tests for this. 

Yes, sure.

packages/koa-shopify-auth/src/auth/client/request-storage-access.ts
dooman87: Sorry, missed this one @ismail-syed.

Looks like I messed it up in my previous one. And it works in my app cause I used my forked version before the latest change to default value of `prefix` :( 

I think you are right and we need to remove default value from prefix. `appTargetUrl` looks alright to me cause it's the last item in the path. 

I'm happy to add a unit test for this bit. What would I use to write it? I saw in the other parts jest was used. Is it the way to go?

ismail-syed: > What would I use to write it? I saw in the other parts jest was used. Is it the way to go?

Yes, jest is the way to go. Thanks!

evulse: @dooman87 @ismail-syed can we fix this? it makes using a prefix unusable, which is even worse as the example usage of using this package is with a prefix.

dooman87: @ismail-syed apologies, here in Melbourne, we went into stage 4 lockdown which means no childcare. It's going to finish in 2 weeks in best case scenario. Could we merge this one please and cut the release. I think functionality wise it's good to go. I have unit tests on my list and will add them.

ismail-syed:
Closing in favour of #1616.

packages/koa-shopify-auth/src/auth/client/request-storage-access.ts
ismail-syed: Follow-up: https://github.com/Shopify/quilt/issues/1550#issuecomment-683845036

PR #1552, opened 2020-07-14
Add a `--skip-yarn` option on the `ReactSetupGenerator`:

Edouard-chin:
- ### Context

  For the Gestalt project, each Rails application will be boostraped
  using a tool to generate Shopify's flavoured app. The application
  will be provisioned in the cloud, in this context installing JS
  dependencies of the app is useless. What we really need is to add
  the JS dependencies as part of the yarn package.json and lock
  the version in the yarn lockfile.

  ### Problem

  Yarn doesn't have a way to add new dependencies to a project
  without downloading them (i.e modify only the package.json/yarn.lock)
  files. Downloading those dependencies takes a realllly long time,
  basically 50% of the time to boostrap an app is dedicated to
  downloading JS dependencies.

  ### Solution

  Introduce a flag on this generator to not add/install JS
  dependencies.
  Having that flag will allow us to control programatically
  when JS dependencies get installed.

### DIFF
--- a/gems/quilt_rails/CHANGELOG.md
+++ b/gems/quilt_rails/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 <!-- ## [Unreleased] -->
 
+### Added
+
+- Added a `--skip-yarn` option when running the `bin/rails generate sewing_kit:install` generator. This option is mostly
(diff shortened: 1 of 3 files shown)


iaserrat:
Not sure I'm the right reviewer for this. What is this?

Edouard-chin:
My bad @iaserrat wrong ping

Edouard-chin:
Oops sorry I didn't see there was a CHANGELOG entry in this repo

PR #1554, opened 2020-07-15
Prepare to release v3.2.1

Edouard-chin:
Prepare for quilt_rails release v3.2.1


michenly:
same here, I will cut a release for you because the version change have to go with a tag and you will need master push permission. In the future you can request it in #web-foundations-tech channel when you would like a new release.

PR #1560, opened 2020-07-21
Remove dependency on javascript-utilities

BPScott:
## Description

Remove dependency on deprecated  javascript-utilities package

See https://github.com/Shopify/javascript-utilities/blob/master/src/events.ts to note that addEventListener / removeEventListener are effectively aliases of the built-in versions in the browsers that Shopify supports.

## Type of change

- [x] react-async Minor: New feature (non-breaking change which adds functionality)

## Checklist


michenly:
♥️♥️♥️♥️♥️♥️♥️♥️♥️ love this so much!!!

PR #1566, opened 2020-07-27
Use custom error page to wait react server boot

gmcgibbon:
## Description

Rescue from `Quilt::ReactRenderable::ReactServerNoResponseError` and retry via a refresh meta tag instead of surfacing the error in the host application. We should probably make the page look nicer, but all styles and JS have to be embedded onto the page.

## Type of change

Enhancement of error handling for `Quilt::ReactRenderable::ReactServerNoResponseError`.

## Checklist

### DIFF
--- a/gems/quilt_rails/app/controllers/quilt/ui_controller.rb
+++ b/gems/quilt_rails/app/controllers/quilt/ui_controller.rb
@@ -3,6 +3,11 @@
 module Quilt
   class UiController < ApplicationController
     include Quilt::ReactRenderable
+    layout(false)
+
+    rescue_from(Quilt::ReactRenderable::ReactServerNoResponseError) do
+      render(:react_render_error, status: :internal_server_error)
+    end
 
     def index
       render_react
--- a/gems/quilt_rails/app/views/quilt/ui/react_render_error.html
+++ b/gems/quilt_rails/app/views/quilt/ui/react_render_error.html
@@ -1,0 +1,21 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <title>React Render Error</title>
+  <meta name="viewport" content="width=device-width,initial-scale=1">
+  <meta http-equiv="refresh" content="3;URL='/'" />
+  <link rel="stylesheet" href="https://unpkg.com/@shopify/polaris@5.1.0/dist/styles.css" />
+</head>
+<body>
+  <div>
+    <div>
+      <h1>Waiting for React Sever to boot up.</h1>
+      <p>This page will refresh automatically.</p>
+    </div>
+    <p>
+      If this error persists, ensure <code>@shopify/react-server</code>
+      is running on <code>http://localhost:8081</code>.
+    </p>
+  </div>
+</body>
+</html>
--- a/gems/quilt_rails/test/controllers/quilt/ui_controller_test.rb
+++ b/gems/quilt_rails/test/controllers/quilt/ui_controller_test.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+require "test_helper"
+require "action_controller"
+
+module Quilt
+  class UiControllerTest < ActionDispatch::IntegrationTest
+    include ActiveSupport::Testing::Isolation
+
+    setup { boot_dummy }
+
+    test "react render error" do
+      get("/")
+
+      assert_select("h1", "Waiting for React Sever to boot up.")
+      assert_select("meta[http-equiv='refresh']") do |selector, *|
+        assert_equal("3;URL='/'", selector[:content])
+      end
+    end
+
+    private
+
+    def boot_dummy
+      Rails.env = "development"
+      require_relative "../../dummy/config/environment"
+    end
+  end
+end
--- a/gems/quilt_rails/test/dummy/config/environments/production.rb
+++ b/gems/quilt_rails/test/dummy/config/environments/production.rb
@@ -27,9 +27,6 @@ Rails.application.configure do
   # Compress CSS using a preprocessor.
   # config.assets.css_compressor = :sass
 
-  # Do not fallback to assets pipeline if a precompiled asset is missed.
-  config.assets.compile = false
-
   # Enable serving of images, stylesheets, and JavaScripts from an asset server.
   # config.action_controller.asset_host = 'http://assets.example.com'
 
--- a/gems/quilt_rails/.gitignore
+++ b/gems/quilt_rails/.gitignore
@@ -5,3 +5,5 @@ node_modules
 /test/**/public
 test/integration/fixtures/basic_app/yarn.lock
 test/integration/fixtures/basic_app/tmp
+test/dummy/log/*
+test/dummy/db/*
--- a/gems/quilt_rails/CHANGELOG.md
+++ b/gems/quilt_rails/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 <!-- ## [Unreleased] -->
 
+- Added a custom error page for `Quilt::ReactRenderable::ReactServerNoResponseError` that automatically refreshes until the react server has started.
+  ([#1566](https://github.com/Shopify/quilt/pull/1566))
+
 ## [3.2.1] - 2020-07-15
 
 ### Added
--- a/gems/quilt_rails/Gemfile
+++ b/gems/quilt_rails/Gemfile
@@ -8,4 +8,5 @@ group :test do
   gem 'minitest-hooks'
   gem 'mocha'
   gem 'pry'
+  gem 'sqlite3'
 end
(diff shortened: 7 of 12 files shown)


gems/quilt_rails/test/controllers/quilt/ui_controller_test.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+require 'rails/generators'
+require 'generators/quilt/rails_setup/rails_setup_generator'
+
+module Quilt
+  class UiControllerTest < ActionDispatch::IntegrationTest
+    def run(*)
+      Rails.env.stub(:test?, false) { super } # NOTE: Bypass DoNotIntegrationTestError
```
gmcgibbon, line 10: This is to bypass the guard we have to discourage integration testing on the route. If we don't want to test this I can remove it.

gems/quilt_rails/app/views/quilt/ui/react_render_error.html
```diff
@@ -0,0 +1,20 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <title>React Render Error</title>
+  <meta name="viewport" content="width=device-width,initial-scale=1">
+  <meta http-equiv="refresh" content="3;URL='/'" />
```
TheMallen, line 6: If we add 
```
<link
  rel="stylesheet"
  href="https://unpkg.com/@shopify/polaris@5.1.0/dist/styles.css"
/>
```

we can use Polaris CSS components and make the page at least a little bit nice :)

gems/quilt_rails/test/controllers/quilt/ui_controller_test.rb
TheMallen: this seems fine, do you think we should have a built in way to bypass this?

gems/quilt_rails/app/controllers/quilt/ui_controller.rb
```diff
@@ -4,6 +4,10 @@ module Quilt
   class UiController < ApplicationController
     include Quilt::ReactRenderable
 
+    rescue_from(Quilt::ReactRenderable::ReactServerNoResponseError) do
```
TheMallen, line 7: <3

gems/quilt_rails/test/controllers/quilt/ui_controller_test.rb
gmcgibbon: Yeah I can make it so we don't need to stub before merging.

gems/quilt_rails/test/controllers/quilt/ui_controller_test.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+require 'test_helper'
+require 'rails/generators'
+require 'generators/quilt/rails_setup/rails_setup_generator'
+
+module Quilt
+  class UiControllerTest < ActionDispatch::IntegrationTest
+    def run(*)
```
Edouard-chin, line 9: `include ActiveSupport::Testing::Isolation`

All tests needs in this file needs to run in a fork, otherwise as soon as someone adds another test and call `boot_dummy` subsequent tests would fail

gems/quilt_rails/test/controllers/quilt/ui_controller_test.rb
Edouard-chin: Better to actually boot the app in production/development rather than stubbing this.

gems/quilt_rails/test/controllers/quilt/ui_controller_test.rb
gmcgibbon: Ah! I was going to ask about what rails does to solve this problem. Thank you 🙇

gems/quilt_rails/app/views/quilt/ui/react_render_error.html
gmcgibbon: Stylesheet link added, but the page still needs some cosmetic changes. This can be done later though, I'll add a changelog for this and then merge once everything is 🟢

gems/quilt_rails/test/controllers/quilt/ui_controller_test.rb
gmcgibbon: I may be able to slash the configuration commit if that works. Let me check.

gems/quilt_rails/test/controllers/quilt/ui_controller_test.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+module Quilt
+  class UiControllerTest < ActionDispatch::IntegrationTest
+    include ActiveSupport::Testing::Isolation
+
+    def run(*)
+      with_integration_tests_allowed { super }
```
Edouard-chin, line 10: I don't think this is an elegant solution. Best would be to boot the app in production/development mode (whatever you prefer). You can achieve this by exporting the `RAILS_ENV` variable before doing the the call to `require 'environment'` https://github.com/Shopify/quilt/pull/1566/files#diff-af0adb8e661fc01cefcfafc3ee8efe4eR35

gmcgibbon: 🤦 This doesn't work because the dummy app is booted in the test helper. Let me try and sort this out...

gmcgibbon: Done. I had to keep this wrapped on `run` because I needed `Rails.env` changed only for this test case, it is used in others. Changing `RAILS_ENV` doesn't work because `Rails.env` gets memoized in other tests.

gems/quilt_rails/test/dummy/config/environments/production.rb
```diff
@@ -27,9 +27,6 @@
   # Compress CSS using a preprocessor.
   # config.assets.css_compressor = :sass
 
-  # Do not fallback to assets pipeline if a precompiled asset is missed.
-  config.assets.compile = false
```
gmcgibbon, line 31: Dummy app doesn't require sprockets so we don't expect asset config.

gems/quilt_rails/test/controllers/quilt/ui_controller_test.rb
```diff
@@ -0,0 +1,39 @@
...
+      assert_select("meta[http-equiv='refresh']") do |selector, *|
+        assert_equal("3;URL='/'", selector[:content])
+      end
+    end
+
+    private
+
+    def with_rails_env(environment)
+      previous_environment = Rails.env.to_s
+      Rails.env = environment
```
Edouard-chin, line 29: ENV['RAILS_ENV'] . In the forked process, Rails.env should have no values. This is also the normal way to boot an app with whatever environment (i.e. you don't do `Rails.env = 'test' bin/rails c`).

You also don't need to reset the ENV because when the test starts you are already in a forked process, the env will just "die" when the test ends (it won't bubble up to the main process)

gems/quilt_rails/test/controllers/quilt/ui_controller_test.rb
```diff
@@ -0,0 +1,39 @@
...
+
+module Quilt
+  class UiControllerTest < ActionDispatch::IntegrationTest
+    include ActiveSupport::Testing::Isolation
+
+    def run(*)
+      with_rails_env("development") { super }
+    end
+
+    setup { boot_dummy }
```
Edouard-chin, line 14: `setup { boot_dummy('development') } . No need to override Minitest#run

gems/quilt_rails/test/controllers/quilt/ui_controller_test.rb
Edouard-chin: > Changing RAILS_ENV doesn't work because Rails.env gets memoized in other tests.

Not sure I understand. Why do other tests matter ? Does changing `RAILS_ENV` breaks other tests ?

gems/quilt_rails/test/controllers/quilt/ui_controller_test.rb
gmcgibbon: I understand that, but if a previous test run memoizes `Rails.env` the fork seems to inherit the value, and then the env var has no effect. At least that's what I'm seeing.

gems/quilt_rails/test/controllers/quilt/ui_controller_test.rb
gmcgibbon: https://github.com/rails/rails/blob/c6f02cbb56b6291533a3756bb5016964c76f5c20/railties/lib/rails.rb#L73

The problem is memoization, changing `RAILS_ENV` does nothing once it is memoized.

gmcgibbon: Actually we can unset it in boot method, that will probably fix it if the fork _just dies_ as you say.

PR #1570, opened 2020-07-30
Fix bug because of precendence:

Edouard-chin:
Fix bug because of precendence:

- Calling `defined? Rails && Rails.logger.nil?` will always return
  true even if `Rails.logger` isn't nil.

  This is because `defined?` has a lower precedence that `&&`.
  List of precedence in ruby https://ruby-doc.org/core-2.6.2/doc/syntax/precedence_rdoc.html

  I have also modified the condition because it doesn't make sense.
  This gem is a Rails engine, `Rails` should always be defined.


Edouard-chin:
@michenly are you ok merging this ?

Issue #1574, opened 2020-08-04
Update dot-prop to resolved vulnerability warning


michenly:
Resolve this error https://github.com/Shopify/quilt/network/alert/yarn.lock/dot-prop/open
because of https://github.com/advisories/GHSA-ff7x-qrg7-qggm found using npm audit.

dot-prop is in our yarn.lock because of lerna.

dot-prop@^3.0.0 > 
compare-func@^1.3.1 >
conventional-changelog-angular@^1.6.6 | conventional-changelog-jshint@^0.3.8 | conventional-changelog-writer@^3.0.9 >
conventional-changelog-core@^2.0.11>
conventional-changelog@^1.1.24 >
conventional-changelog-cli@^1.3.13>
lerna@^2.9.0

------

I tried updating to lerna@^3.22.1 and and just ended up with dot-prop@^3.0.0 & dot-prop "^4.2.0"

Since it is only part of the release process and not included as part of any packages. I think this vulnerability is acceptable till lerna uses dor-prop > 5.1.1 and we can do a proper upgrade.

michenly:
To follow up, the alert is no longer showing up for me

PR #1575, opened 2020-08-04
lock dot-prop to 5.1.1

michenly:
Fixes (issue #) https://github.com/Shopify/quilt/issues/1574

Resolved by locking the resolution for dot-prop.
This lib is only being use for lerna and should not affect any of the packages.

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,6 +86,7 @@
   "dependencies": {},
   "resolutions": {
     "@types/react": "16.9.19",
-    "@types/react-dom": "16.9.5"
+    "@types/react-dom": "16.9.5",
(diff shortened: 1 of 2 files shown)


michenly:
mmm....I would expect these error to go away as well. Unless npm audit works on the library listed despite the resolution? 😕

keyfer:
@michenly As far as I know `npm audit` doesn't ignore resolutions.

TheMallen:
I think this might just be because we use yarn.lock instead of npm lock files :/

michenly:
should we merge and see if the warning on the repo goes away?

PR #1576, opened 2020-08-04
Append engine mount

gmcgibbon:
## Description

Adds automatic engine mounting via a prepend hook. This is to prevent other engine routes being overridden by the main engine mount. If the gem automatically prepends the quilt engine, then other engines can draw on the application routes without worrying about precedence.

## Type of change

- [x] quilt_rails Patch: Bug (non-breaking change)

## Checklist

### DIFF
--- a/gems/quilt_rails/lib/quilt_rails/engine.rb
+++ b/gems/quilt_rails/lib/quilt_rails/engine.rb
@@ -3,5 +3,11 @@
 module Quilt
   class Engine < ::Rails::Engine
     isolate_namespace Quilt
+
+    initializer(:mount_quilt, before: :add_builtin_route) do |app|
+      app.routes.append do
+        mount(Quilt::Engine, at: '/') unless has_named_route?(:quilt)
+      end
+    end
   end
 end
--- a/gems/quilt_rails/test/controllers/routing_test.rb
+++ b/gems/quilt_rails/test/controllers/routing_test.rb
@@ -1,0 +1,81 @@
+# frozen_string_literal: true
+
+require "test_helper"
+require "action_controller"
+
+module Quilt
+  class RoutingTest < ActionDispatch::IntegrationTest
+    include ActiveSupport::Testing::Isolation
+
+    class PerformanceReportController < ActionController::Base; end
+    class UiController < ActionController::Base; end
+
+    setup { boot_dummy }
+
+    test "routes on app without routes" do
+      assert_recognizes(
+        { controller: "quilt/performance_report", action: "create" },
+        { path: "/performance_report", method: "post" }
+      )
+      assert_recognizes(
+        { controller: "quilt/ui", action: "index", path: "anything" },
+        { path: "/anything" }
+      )
+      # Use after https://github.com/rails/rails/pull/39981 is released.
+      # assert_recognizes(
+      #   { controller: "quilt/ui", action: "index" },
+      #   { path: "/" }
+      # )
+    end
+
+    test "routes on app with engine mount" do
+      Rails.application.routes.draw do
+        mount(Quilt::Engine, at: '/quilt')
+      end
+
+      assert_recognizes(
+        { controller: "quilt/performance_report", action: "create" },
+        { path: "/quilt/performance_report", method: "post" }
+      )
+      assert_recognizes(
+        { controller: "quilt/ui", action: "index", path: "anything" },
+        { path: "/quilt/anything" }
+      )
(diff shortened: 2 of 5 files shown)


gems/quilt_rails/lib/quilt_rails/engine.rb
```diff
@@ -3,5 +3,11 @@
 module Quilt
   class Engine < ::Rails::Engine
     isolate_namespace Quilt
+
+    initializer(:mount_quilt, before: :add_builtin_route) do |app|
+      app.routes.append do
+        mount(Quilt::Engine, at: '/') unless has_named_route?(:quilt)
+      end
+    end
```
michenly, line 11: sorry my ruby skill is bad and I dont fully understand what this does.

how would this affect any application that want quilt on specific route 

```
# config/routes.rb
Rails.application.routes.draw do
  # ...
  mount Quilt::Engine, at: '/path/to/react'
end
```

or application that does not uses engine in routes file, but instead add Quilt to controller?
https://github.com/Shopify/quilt/blob/master/gems/quilt_rails/docs/manual-installation.md#option-2-add-a-react-controller-and-routes

gmcgibbon: The routes drawn in the application should take priority over the appended ones I'm defining in the gem. I should add more tests to cover this case. 👍

michenly: Thank you! and in the case of the controller, what you defined would get ignore?

gmcgibbon: Yes, but the engine still gets mounted. It will look like this at the bottom of the host app's `bin/rails routes` listing:
```
quilt        /                 Quilt::Engine

Routes for Quilt::Engine:

performance_report POST /performance_report(.:format) quilt/performance_report#create
                   GET  /*path(.:format)              quilt/ui#index
              root GET  /                             quilt/ui#index
```

We could add a config option to disable automatic mounting so the developer has the option to do everything themselves. That might be less annoying.

michenly: I am more so wondering if this would create any error? 
But honestly you will know that better than I do.

The very off pattern case I am think of is this.
They have both an engine from ShopifyApp & a catch all route to react controller https://github.com/Shopify/arrive-server/blob/master/engines/merchant/config/routes.rb#L10-L13

Then uses quilt in the controller https://github.com/Shopify/arrive-server/blob/master/engines/merchant/app/controllers/merchant/react_controller.rb#L7

Additionally, if the route is defined in `engines/something/config/routes.rb` instead of `config/routes.rb`. Would that create any issue?

gmcgibbon: > I am more so wondering if this would create any error?

It will not, it just mounts an additional engine that gets overridden.

> Additionally, if the route is defined in engines/something/config/routes.rb instead of config/routes.rb. Would that create any issue?

The engine case is slightly more complicated because it depends how it is mounted, and also how the quilt engine is mounted. In the example you linked, it should work because the engine mounted at `/` mount will take priority.

To put it simply, this change adds a default route that can be easily overridden by anything inside of an app's `Rails.routes.draw` block. If the app does not provide paths for `GET /`, `GET /*`, and `POST /performance_report `, the appended quilt engine mount (this change) will be routed to.

These are all great use-cases though, and we need to account for all of them. Thank you for bringing these to my attention. 😄

michenly: Thank you for explaining! Great education moment.

gems/quilt_rails/test/controllers/routing_test.rb
```diff
@@ -0,0 +1,81 @@
...
+      assert_recognizes(
+        { controller: "quilt/ui", action: "index", path: "anything" },
+        { path: "/quilt/anything" }
+      )
+      # Use after https://github.com/rails/rails/pull/39981 is released.
+      # assert_recognizes(
+      #   { controller: "quilt/ui", action: "index" },
+      #   { path: "/path" }
+      # )
+    end
```
michenly, line 49: can we add another test case for a engine mount but the path for the engine is not `/` ?

gmcgibbon:
> Look great! should we wait till rails/rails#39981 is in?

Since we aren't using edge rails in our tests we either have to wait for a release or use edge. Right now I think it is fine to merge as-is and worry about more granular testing later.

PR #1577, opened 2020-08-04
Add error link

michenly:
## Description

Part of resolving https://github.com/Shopify/shopify-rails/issues/282

Add default graphql error links. Mostly from [1](https://github.com/Shopify/web/blob/master/app/sections/Apps/AppList/graphql/appStore/link/error-handler.ts) & [2](https://github.com/Shopify/web/blob/master/packages/@web-utilities/graphql-client/link/parse-error.ts)

Before the error link
before error link

after error link
after error link

## Type of change

## Checklist

### DIFF
--- a/packages/react-graphql-universal-provider/src/network-error-link.ts
+++ b/packages/react-graphql-universal-provider/src/network-error-link.ts
@@ -1,0 +1,21 @@
+import {ApolloLink} from 'apollo-link';
+import {ErrorLink} from 'apollo-link-error';
+import {ServerParseError} from 'apollo-link-http-common';
+
+export function createNetworkErrorLink() {
+  // eslint-disable-next-line no-process-env
+  if (process.env.NODE_ENV === 'development') {
+    return new ErrorLink(({networkError}) => {
+      if (networkError && networkError.name === 'ServerParseError') {
+        networkError.message = (networkError as ServerParseError).bodyText;
+      }
+    });
+  } else {
+    return new ApolloLink((operation, nextLink) => {
+      if (nextLink == null) {
+        throw new Error('The parse error link must not be a terminating link');
+      }
+      return nextLink(operation);
+    });
+  }
+}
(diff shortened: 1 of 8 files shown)


michenly:
Thanks for the review @TheMallen 

would love to know if you have any feedback on how to make the code nicer.
Apollo tired really hard NOT to conflate network error with GraphQL error.

This implementation here actually take [a network error and pretend it is an GraphQL error](https://github.com/Shopify/quilt/pull/1577/files#diff-982936ff8f201f301e0e6540ae1b8b7d). Which mean this error would show up as an error in `const {error} = useQuery()`

I can also leave it as a network error with better message. But so far in my test I dont see the network error getting log in splunk.

packages/react-graphql-universal-provider/src/network-error-link.ts
```diff
@@ -0,0 +1,21 @@
+import {ApolloLink} from 'apollo-link';
+import {ErrorLink} from 'apollo-link-error';
+import {ServerParseError} from 'apollo-link-http-common';
+
+export function createNetworkErrorLink() {
+  // eslint-disable-next-line no-process-env
+  if (process.env.NODE_ENV === 'development') {
```
michenly, line 7: not sure if this needed to be bind to development. it is this way in web and maybe we are trying to avoid surfacing something?

TheMallen: In theory they probably want to build a custom error link for production, either error seems not the most useful thing to surface to users in prod. 

That said, I'd rather we surface the more useful error in prod just because that's where these usually happen since we do auth as infrastructure. At the very least I'd aways log the output so it's more debugabble via splunk  / bugsnag / whatever.

michenly: I am doing another test to see if the Network error show up in splunk. 
If that work I think I am going to remove the process.env.NODE_ENV === 'development' check then.

PR #1578, opened 2020-08-05
Fixed error when using chrome autofill

bhavathi:
## Description

Fixed error when using chrome autofill

`Uncaught TypeError: event.getModifierState is not a function`

When using native autofill feature in Chrome, keydown event is raised but event object is missing getModifierState function and causes this error.

Added an additional check on getModifierState.

### DIFF
--- a/packages/react-shortcuts/src/ShortcutManager/ShortcutManager.tsx
+++ b/packages/react-shortcuts/src/ShortcutManager/ShortcutManager.tsx
@@ -67,7 +67,10 @@ export default class ShortcutManager {
     }
 
     function keyGroupIsHeld(keyGroup: ModifierKey[]) {
-      return keyGroup.every((key: ModifierKey) => event.getModifierState(key));
+      return keyGroup.every(
+        (key: ModifierKey) =>
+          event.getModifierState && event.getModifierState(key),
+      );
     }
 
     return hasKeyGroups(heldKeys)


bhavathi:
@lemonmade I am not finding option to add reviewers.. can you please review this PR

lemonmade:
cc/ @Shopify/web-foundations for review

packages/react-shortcuts/src/ShortcutManager/ShortcutManager.tsx
```diff
@@ -67,7 +67,10 @@ export default class ShortcutManager {
     }
 
     function keyGroupIsHeld(keyGroup: ModifierKey[]) {
-      return keyGroup.every((key: ModifierKey) => event.getModifierState(key));
+      return keyGroup.every(
+        (key: ModifierKey) =>
+          event.getModifierState && event.getModifierState(key),
```
gmcgibbon, line 72: Since this is a bug fix, can you add a changelog to [this file](https://github.com/Shopify/quilt/blob/master/packages/react-shortcuts/CHANGELOG.md) and a test [here](https://github.com/Shopify/quilt/blob/8a9c68de5686d5f57827c7c3da7cfe375f58e837/packages/react-shortcuts/src/tests/ShortcutManager.test.tsx#L299)? I think this test should cover it:
```ts
    it('doesn’t fire on non-keyboard events', () => {
      const matchSpy = jest.fn();
      const heldGroup: ModifierKey[] = ['Control', 'Shift'];
      const heldToCheck: HeldKey = [[...heldGroup], ['Alt', 'Meta']];

      mount(
        <ShortcutProvider>
          <Shortcut held={heldToCheck} ordered={['/']} onMatch={matchSpy} />
        </ShortcutProvider>,
      );

      document.dispatchEvent(new UIEvent('keydown'));

      expect(matchSpy).not.toHaveBeenCalled();
    });
```

gmcgibbon (commented):
Our typing assumptions on the keydown event appear to be wrong, but this check is needed regardless, so 👍

Please add a changelog and a test to make sure we don't regress. You also need to sign our CLA before we can accept this change (see CLA check). Thanks!

bhavathi:
> Our typing assumptions on the keydown event appear to be wrong, but this check is needed regardless, so 👍
> 
> Please add a changelog and a test to make sure we don't regress. You also need to sign our CLA before we can accept this change (see CLA check). Thanks!

I have signed the CLA, still it is failing the check. Please let me know if I am missing anything.

packages/react-shortcuts/src/ShortcutManager/ShortcutManager.tsx
bhavathi: Updated test case and change log.

gmcgibbon (approved):
Thanks!

PR #1579, opened 2020-08-06
✨ add request id to logging

michenly:
## Description

`request_id` is missing in the logging which is essential during debugging in Splunk to be able to trace the chain of requests from Rails request to Node request to GraphQL request.

### @shopify/react-server
- Add request_id, hostname, ip to logging from header for debugging

## Type of change

- @shopify/react-server: minor

## Checklist

### DIFF
--- a/packages/network/CHANGELOG.md
+++ b/packages/network/CHANGELOG.md
@@ -5,11 +5,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## [Unreleased] -->
-
 ## [Unreleased]
 
-Added `ResponseType` function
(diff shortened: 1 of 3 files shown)


michenly:
@Edouard-chin no problem! Tagging you more so for an FYI on if the logic make sense to add requestID to both
1. Logging of any Node request (coming from Rails)
2. Send as part of server side GraphQL request (from Node to Rails /graphql end point)

My hope is that developer can then use the same requestID to see where the error occur when making request to Rails.
Maybe even use it as a learning opportunity to see the lifecycle of a request.

(ps. would you know if nginx/rails respect the requestID header I added for server side GraphQL request or assigned new value for this request?)

michenly:
Ok, after some discussion, I am going to remove the request id to graphql.
In the future we might add a tracing id but this PR is good for now.

michenly:
Add universal provider doc on whats the new default

michenly:
Add test for `ssrForceFetchDelay` & `connectToDevTools`

michenly:
included a test for customized `cache`

PR #1581, opened 2020-08-06
Explicitly set status code on successful response to 200

gmcgibbon:
## Description

When using react-server via the `quilt_rails` gem, I'm noticing proxied requests are return `404 Not Found`. The response doesn't seem to be set anywhere in the successful render path of `react-server`, so I've set it explicitly to `200 Ok` which seems to fix the issue. However, I'm not quite sure yet if this is a problem within `react-server`, `react-network` (`usesStatus`), or in the rails proxy somewhere.

The default status code for koa contexts is `404` as mentioned by [koa's docs](https://github.com/koajs/koa/blob/master/docs/api/response.md#responsestatus).

## Type of change

## Checklist

### DIFF
--- a/packages/react-server/src/render/render.tsx
+++ b/packages/react-server/src/render/render.tsx
@@ -141,6 +141,8 @@ export function createRender(render: RenderFunction, options: Options = {}) {
 
       ctx.set(Header.ContentType, 'text/html');
       ctx.body = response;
+
+      ctx.status = StatusCode.Ok;
     } catch (error) {
       const errorMessage = `React server-side rendering error:\n${
         error.stack || error.message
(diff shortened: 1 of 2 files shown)


michenly:
I try 🎩 it, and the problem with this method, is now the status code set by [`@shopify/react-network`](https://github.com/Shopify/gestalt-web-application-example/blob/master/web/components/NotFound/NotFound.tsx#L9) now wont propagate.

I also double checked the changelog for react-server, nothing seems too out of place that would cause this issue.

For `@shopify/react-network`, there are two PR recently that may be related to look into later [1](https://github.com/Shopify/quilt/pull/1515),  [2](https://github.com/Shopify/quilt/pull/1517)

gmcgibbon:
I tried reverting both of the PRs you mentioned locally, `useStatus` still seems to misbehave. 🤔

michenly:
Closing this in favor of #1589

PR #1582, opened 2020-08-06
Fix links in react-server README

gmcgibbon:
## Description

Update some broken links on the `react-server` package README.

## Type of change

## Checklist


michenly:
👍 LGTM!

PR #1585, opened 2020-08-11
🐛 update default fallback to react-server ip

michenly:
## Description

This [PR](https://github.com/Shopify/quilt/pull/1514) somehow breaks pinging the Node service for folks.
https://discourse.shopify.io/t/deploy-times-out-after-upgrading-sewing-kit-to-0-134-0/9870/7

While I have no clear explanation on why the above PR would break it (one possibility is that moving `process.env` from within sewing-kit runtime to Node runtime changed the value, this is un-confirmed thou). There seems to be a clear fix forward.

Changing the default value for ip from `localhost` to `0.0.0.0` will allow the server to ping using container IP , which is also the Koa/Node [default](https://nodejs.org/api/net.html#net_server_listen_port_host_backlog_callback).

## Type of change

## Checklist


TheMallen:
We might want a test for this btw, but seems good.

PR #1589, opened 2020-08-12
status code return 404 unexpectedly 

michenly:
## Description

Fixes (issue #)

Related https://github.com/Shopify/quilt/pull/1581, https://github.com/Shopify/quilt/pull/1567

## Type of change

## Checklist

### DIFF
--- a/packages/react-router/src/components/Router/Router.tsx
+++ b/packages/react-router/src/components/Router/Router.tsx
@@ -21,7 +21,7 @@ export default function Router({location, children}: Props) {
   }
 
   const locationString =
-    typeof location === 'object' ? location.href : location;
+    typeof location === 'object' ? location.pathname : location;
 
   return (
     <StaticRouter location={locationString} context={{}}>
--- a/packages/react-router/src/components/Router/Router.tsx
+++ b/packages/react-router/src/components/Router/Router.tsx
@@ -5,7 +5,7 @@ import {isClient} from './utilities';
 
 interface Props {
   children?: React.ReactNode;
-  location?: string | URL;
+  location?: string | {pathname: string; search: string};
 }
 
 export const NO_LOCATION_ERROR =
@@ -20,11 +20,21 @@ export default function Router({location, children}: Props) {
     throw new Error(NO_LOCATION_ERROR);
   }
 
-  const locationString =
-    typeof location === 'object' ? location.pathname : location;
+  // Internally react-router uses the spread (...) operator to clone the object we pass in.
+  // This means that we lose any properties which come from the prototype even if they are the ones we need.
+  // As a result we need to manually pull the properties off to build our object if we want this API to work with stuff like URLs or other instances.
+  const locationObject =
+    typeof location === 'object'
+      ? {
+          pathname: location.pathname,
+          search: location.search,
+          hash: '',
+          state: undefined,
+        }
+      : location;
 
   return (
-    <StaticRouter location={locationString} context={{}}>
+    <StaticRouter location={locationObject} context={{}}>
       {children}
     </StaticRouter>
   );
--- a/packages/react-server/src/webpack-plugin/webpack-plugin.ts
+++ b/packages/react-server/src/webpack-plugin/webpack-plugin.ts
@@ -97,7 +97,7 @@ function serverSource(options: Options, compiler: Compiler) {
         errorSSRComponentExists(options, compiler)
           ? `renderError: (ctx) => {
               return React.createElement(Error, {
-                url: ctx.request.URL,
+                url: ctx.request.url,
                 data: ctx.state.quiltData,
                 error: ctx.state.quiltError,
               });
--- a/packages/react-router/CHANGELOG.md
+++ b/packages/react-router/CHANGELOG.md
@@ -5,10 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<-- ## [Unreleased] -->
+## [Unreleased]
+
+### Changed
+
+- Pass in object with pathname (ie. `/test123`) and search to StaticRouter. The strange behaviour is cause by a react-router using spread operator to copy object. [1589](https://github.com/Shopify/quilt/pull/1589)
 
 ## [0.0.30] - 2020-07-28
 
+- ❗️ This version is broken and deprecated. Do not use ❗️
 - Fix bug where passing URL object would cause server router to page incorrectly [1567](https://github.com/Shopify/quilt/pull/1567)
 
 ## [0.0.25] - 2020-05-29
--- a/packages/react-router/package.json
+++ b/packages/react-router/package.json
@@ -25,8 +25,8 @@
   "homepage": "https://github.com/Shopify/quilt/blob/master/packages/react-router/README.md",
   "dependencies": {
     "@shopify/react-network": "^3.5.1",
-    "@types/react-router-dom": "^4.3.5",
-    "react-router-dom": "^5.0.1"
+    "@types/react-router-dom": "*",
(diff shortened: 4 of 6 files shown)


packages/react-server/src/webpack-plugin/webpack-plugin.ts
```diff
@@ -97,7 +97,7 @@ function serverSource(options: Options, compiler: Compiler) {
         errorSSRComponentExists(options, compiler)
           ? `renderError: (ctx) => {
               return React.createElement(Error, {
-                url: ctx.request.URL,
+                url: ctx.request.url,
```
michenly, line 100: dont think this does anything really

packages/react-router/src/components/Router/Router.tsx
```diff
@@ -21,7 +21,7 @@ export default function Router({location, children}: Props) {
   }
 
   const locationString =
-    typeof location === 'object' ? location.href : location;
+    typeof location === 'object' ? location.pathname : location;
```
michenly, line 24: whyyy, like whyyyy

michenly: In react router's doc. You are suppose to use [`req.url`](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/StaticRouter.md#location-string)

The example for the url being `In the request www.example.com/index.html?a=1&b=2, req.url will contain /index.html?a=1&b=2.` which does not explain why pathname works

michenly:
_so to summarize what I found so far_
using the following routing setup & react-router-dom@5.2.0
```
    <Switch>
      <Route path="/" exact component={() => <Home />} />
      <Route render={() => <NotFound />} />
    </Switch>
```

**[passing in object to StaticRouter](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/StaticRouter.md#location-object)**
the router renders differently on server & on client. but only slightly. 
As in the images is different but the text are the same.
Warning in browser `Text content did not match. Server: "Hello World!" Client: "The page you're looking for couldn't be found"`

This is currently what web uses with react-router-dom@5.1.2 
https://github.com/Shopify/web/blob/master/server/router/middleware/render-react-app/middleware/render-response/render-response.tsx#L122

**passing in location.href (ie. https://site-name.com/test123) to StaticRouter**
the router renders differently on server & on client fully.
When loading `/`, the NotFound component get loaded on server but Home get loaded on client.
Warning in browser `Warning: Did not expect server HTML to contain a <div> in <div>.`

This is currently what [beta buddy](https://github.com/Shopify/beta-buddy/blob/master/app/ui/foundation/App.tsx#L33) uses with react-router-dom@5.2.0

*[*passin in location.pathname (ie. /test123) to StaticRouter](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/StaticRouter.md#location-string)**
This is the current fix.
But like....WHYYYYYY

michenly:
@gmcgibbon did a little digging into react-router for me
https://github.com/ReactTraining/react-router/blob/master/packages/react-router/modules/StaticRouter.js#L72
https://github.com/ReactTraining/history/blob/v4.10.1/modules/LocationUtils.js#L10 
https://github.com/ReactTraining/history/blob/v4.10.1/modules/PathUtils.js#L24

So it looks like we do have to use either an URL object or a path string
Just need to figure out why the URL object wasn't wokring...

gmcgibbon:
https://github.com/ReactTraining/react-router/blob/master/packages/react-router/modules/StaticRouter.js#L72

```
import { createLocation } from "history";
const newLocation = createLocation(location)
```

Logging this [here](https://github.com/Shopify/quilt/blob/3723fcf75941f0d0090a9982d15463d91e19b12a/packages/react-router/src/components/Router/Router.tsx#L22) I see:
```
{ pathname: '/',
  search: '',
  hash: '',
  [Symbol(context)]:
   URLContext {
     flags: 400,
     scheme: 'http:',
     username: '',
     password: '',
     host: 'host.name',
     port: null,
     path: [ 'test' ],
     query: null,
     fragment: null },
  [Symbol(query)]: URLSearchParams {} }
```

The path name gets reset to `/` for the initial SSR render, maybe?

gmcgibbon:
> The path name gets reset to / for the initial SSR render, maybe?

We paired on this, and that is indeed the case. We tracked it down to https://github.com/ReactTraining/history/blob/v4.10.1/modules/LocationUtils.js#L6. Since this is an outdated version of the history package, the latest version doesn't include this function, and fixing this might take some time to get released, we can instead sidestep this by simply passing in the pathname as a string.

michenly:
What an amazing find by @gmcgibbon !!! 👏 👏 👏 

**To recap:**
Using react-router-dom@5.2.0 + history@4.10.1 and hitting the browser at `/test` produce the following:

```
import {createLocation} from 'history';

export default function Router({location, children}: Props) {
...
  console.log('old location', location);
  console.log('new location', createLocation(location));
...
}
```

```
old location URL {
  **href: 'http://gestalt-web-application-example.myshopify.io/test',**
  origin: 'http://gestalt-web-application-example.myshopify.io',
  protocol: 'http:',
  username: '',
  password: '',
  host: 'gestalt-web-application-example.myshopify.io',
  hostname: 'gestalt-web-application-example.myshopify.io',
  port: '',
  **pathname: '/test',**
  search: '',
  searchParams: URLSearchParams {},
  hash: '' }

new location { 
   **pathname: '/',**
  search: '',
  hash: '',
  [Symbol(context)]:
   URLContext {
     flags: 400,
     scheme: 'http:',
     username: '',
     password: '',
     host: 'gestalt-web-application-example.myshopify.io',
     port: null,
     **path: [ 'test' ],**
     query: null,
     fragment: null },
  [Symbol(query)]: URLSearchParams {} }
```

💯 something is happening in history@4.10.1

michenly:
I think what make the most sense right now would be 
- use `pathname` for StaticRouter, which is fix that require the least amount of consumer code change
- deprecate the last version of react-router as the previous fix is incorrect.
- I will follow up with react-router and see if there is a bug log

packages/react-router/src/components/Router/Router.tsx
```diff
@@ -20,11 +20,18 @@ export default function Router({location, children}: Props) {
...
-  const locationString =
-    typeof location === 'object' ? location.href : location;
+  const locationObject =
+    typeof location === 'object'
+      ? {
+          pathname: location.pathname,
+          search: location.search,
+          hash: '',
+          state: undefined,
+        }
```
michenly, line 30: Final thought:

The reason why this have to happen is this line 
[`location = { ...path }`](https://github.com/ReactTraining/history/blob/v4.10.1/modules/LocationUtils.js#L14)
where history is using spread operator to copy the object before using it, and we pass in URL object and copying it using spread operator means we lost all the property we actually do care about (like pathname & search).

Since the document recommend using [`{ pathname, search, hash, state }`](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/StaticRouter.md#location-object) instead of URL object. 

I am updating the type to reflect the only two pieces of information we need, and building an object to pass into StaticRouter that reflect this. 



This example below illustrate what it looks like using spread operator on URL object
```
> const url = new URL('https://gestalt-web-application-example.myshopify.io/test123');
undefined
> url
URL {
  href:
   'https://gestalt-web-application-example.myshopify.io/test123',
  origin: 'https://gestalt-web-application-example.myshopify.io',
  protocol: 'https:',
  username: '',
  password: '',
  host: 'gestalt-web-application-example.myshopify.io',
  hostname: 'gestalt-web-application-example.myshopify.io',
  port: '',
  pathname: '/test123',
  search: '',
  searchParams: URLSearchParams {},
  hash: '' }
> const foo = {...url};
undefined
> foo
{ [Symbol(context)]:
   URLContext {
     flags: 400,
     scheme: 'https:',
     username: '',
     password: '',
     host: 'gestalt-web-application-example.myshopify.io',
     port: null,
     path: [ 'test123' ],
     query: null,
     fragment: null },
  [Symbol(query)]: URLSearchParams {} }
> 
```

Issue #1586, opened 2020-08-11
Koa should be configured to trust the proxy header field


Edouard-chin:
## Overview

`ctx.request.url.origin` will always return a non https url. This is problematic because if a Rails application forces all request to be SSL, the Graphql request would fail.

`ctx.request.url` is used here https://github.com/Shopify/quilt/blob/master/packages/react-server/src/webpack-plugin/webpack-plugin.ts#L86 and later on forwarded to Graphql to make the request

## Solution

According to Koa's doc https://koajs.com/#settings , we should set `proxy = true` in order to trust the `X-FORWARDED-PROTO` header which is the header that determines what scheme was used to make the request.

cc/ @michenly

PR #1591, opened 2020-08-13
Add proxy and app configuration to createServer

gmcgibbon:
## Description

Part 1 of https://github.com/Shopify/quilt/issues/1586

Adds `proxy` option to `createServer`. I also added `app` to inject an already instantiated koa application (we might want to make this interface more generic) because I don't think its possible to mock a constructor in JS to return 1 object. The `new` keyword breaks inject possibilities AFAICS, but I'd be happy to be proven wrong here.

## Type of change

- [x] `@shopify/react-server` Minor: New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/packages/react-server/src/server/server.ts
+++ b/packages/react-server/src/server/server.ts
@@ -17,10 +17,12 @@ interface Options {
   ip?: string;
   port?: number;
   assetPrefix?: string;
+  proxy?: boolean;
   assetName?: string | ValueFromContext<string>;
   serverMiddleware?: compose.Middleware<Context>[];
   render: RenderFunction;
   renderError?: RenderFunction;
+  app?: Koa;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


packages/react-server/src/server/server.ts
```diff
@@ -17,10 +17,12 @@ interface Options {
   ip?: string;
   port?: number;
   assetPrefix?: string;
+  proxy?: boolean;
   assetName?: string | ValueFromContext<string>;
   serverMiddleware?: compose.Middleware<Context>[];
   render: RenderFunction;
   renderError?: RenderFunction;
+  app?: Koa;
```
michenly, line 25: I am not sure about exposing app in the API, it doesn't seems to be something we absolutely need other than for test?

michenly, line 25: It feels a little bit like we are only doing this for testing. But I also dont see anything too wrong with it. So 🙆‍♀️

PR #1594, opened 2020-08-17
Add eventbridge registration support for koa-shopify-webhooks

### LINKED ISSUE #1555: [koa-shopify-webhooks] - EventBridge Support
## Overview

The 2020-07 API release supports an [EventBridge integration](https://www.shopify.com/partners/blog/shopify-api-release-july-2020#AWS-support). You can create a webhook using the following query.

```
mutation {
  eventBridgeWebhookSubscriptionCreate(
    topic: PRODUCTS_CREATE
    webhookSubscription: {
      arn: "<ARN>"
      format: JSON
  })
  {
    webhookSubscription {
      id
    }
    userErrors {
      message
    }
  }
}
```

However, the `koa-shopify-webhooks` library does not expose the ability to change the mutation.

https://github.com/Shopify/quilt/blob/7539c544b0e336db7ab416435ace8a0f5437e008/packages/koa-shopify-webhooks/src/register.ts#L55-L69
...

## Motivation

I assume that quilt maintains feature parity with the API. If not, EventBridge solves a lot of problems that come with registering webhooks and it would be great to take advantage of the integration via quilt.
...

tbrenev:
Co-authored-by: James Langille <james.langille@shopify.com>

## Description

Closes #1555 

In the `koa-shopify-webhooks` package, adds a `deliveryMethod` option to `registerWebhook`. It is http by default, but can be set to eventbridge.

## Type of change

## Checklist


packages/koa-shopify-webhooks/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## [Unreleased] -->
 
+## [2.5.0] - 2020-08-17
```
james-langille, line 10: Minor: I take it we don't need to update the `version` in `package.json` since it's never been updated past `1.0.0`? Mostly asking to build my own context on this.

packages/koa-shopify-webhooks/src/register.ts
```diff
@@ -41,21 +48,46 @@ export async function registerWebhook({
...
+          result.data.webhookSubscriptionCreate &&
+          result.data.webhookSubscriptionCreate.webhookSubscription,
+      );
+    case DeliveryMethod.EventBridge:
+      return Boolean(
+        result.data &&
+          result.data.eventBridgeWebhookSubscriptionCreate &&
+          result.data.eventBridgeWebhookSubscriptionCreate.webhookSubscription,
+      );
   }
```
james-langille, line 68: Minor: Seems we can use [optional chaining](https://stackoverflow.com/questions/40238144/typescript-the-safe-navigation-operator-or-and-null-property-paths) to simplify this a bit if you're interested:
```suggestion
  switch (deliveryMethod) {
    case DeliveryMethod.Http:
      return Boolean(result.data?.webhookSubscriptionCreate?.webhookSubscription);
    case DeliveryMethod.EventBridge:
      return Boolean(result.data?.eventBridgeWebhookSubscriptionCreate?.webhookSubscription);
  }
```
What you have so far is also perfectly fine and valid.

packages/koa-shopify-webhooks/CHANGELOG.md
tbrenev: I recall reading somewhere in the docs in this repo, that we shouldn't touch `package.json`.

packages/koa-shopify-webhooks/src/register.ts
tbrenev: Yeah that's awesome, a lot cleaner!

tbrenev: The tests are giving a syntax error... we're probably behind on the versions of some things. Going to revert to the older code :/

packages/koa-shopify-webhooks/src/test/register.test.ts
```diff
@@ -96,4 +96,42 @@ describe('registerWebhook', () => {
...
   });
+
+  it('sends an eventbridge registration GraphQL query for an eventbridge webhook registration', async () => {
+    fetchMock.mock('*', successResponse);
+    const webhook: Options = {
+      address: 'myapp.com/webhooks',
+      topic: 'PRODUCTS_CREATE',
+      accessToken: 'some token',
+      shop: 'shop1.myshopify.io',
+      apiVersion: ApiVersion.Unstable,
```
mikedasilva, line 107: Might be worth using the api version that it was released in 2020-04. That's the version we're telling people in the docs so having a test here lining up is better than unstable.

packages/koa-shopify-webhooks/src/register.ts
james-langille: Darn. Well it was worth a shot, thanks for trying!

packages/koa-shopify-webhooks/src/test/register.test.ts
tbrenev: ✅

PR #1595, opened 2020-08-17
✨ Add useDelayedCallback hook

carolopolo:
## Description

- Adds a new `useDelayedCallback` hook to the `react-hooks` package. 

- It's useful for when you want to call a function after a given delay. For example, in `web`, we are currently looking to use it to call the `onClose()` of a `Modal` after a slight delay, in order to wait for the progress bar to become 100% filled visually. Without the delay, the `Modal` closes abruptly before the progress bar reaches 100% visually.

- This PR also introduces minor corrections in `README.md` to fix grammar errors/typos/inconsistent verb tenses. 

## Type of change

## Checklist

### DIFF
--- a/packages/react-hooks/README.md
+++ b/packages/react-hooks/README.md
@@ -14,6 +14,7 @@ $ yarn add @shopify/react-hooks
 ## List of hooks
 
 - [useDebouncedValue()](#usedebouncedvalue)
+- [useDelayedCallback()](#usedelayedcallback)
 - [useForceUpdate()](#useforceupdate)
 - [useInterval()](#useinterval)
 - [useLazyRef()](#uselazyref)
@@ -28,7 +29,7 @@ $ yarn add @shopify/react-hooks
 
 ### `useDebouncedValue()`
 
-This hook provide a debounced value.
+This hook provides a debounced value.
 
 ```tsx
 function MyComponent() {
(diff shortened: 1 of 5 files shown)


packages/react-hooks/src/hooks/delayed-callback.ts
```diff
@@ -0,0 +1,25 @@
+import {useEffect, useState} from 'react';
+
+export function useDelayedCallback(callback: () => void, delay: number) {
+  const [timeoutId, setTimeoutId] = useState<ReturnType<typeof setTimeout>>();
```
tleunen, line 4: Based on the list of dependencies of your `useEffect`, I guess this could be changed for a `useRef` instead. You don't really need to re-render when the timeout is set I believe.

loic-d, line 4: `ReturnType<typeof setTimeout>` 🎉

packages/react-hooks/src/hooks/delayed-callback.ts
```diff
@@ -0,0 +1,25 @@
+import {useEffect, useState} from 'react';
+
+export function useDelayedCallback(callback: () => void, delay: number) {
```
tleunen, line 3: `callback: Function` instead? Since it could literally be any type of function

packages/react-hooks/src/hooks/tests/delayed-callback.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+    const fakeComponent = await mount(
+      <FakeComponent callback={callbackSpy} delay={delay} />,
+    );
+
+    fakeComponent.find('button').trigger('onClick');
+
+    fakeComponent.unmount();
+
+    fakeComponent.act(() => {
+      clock.tick(delay - 1);
```
loic-d, line 93: Shouldn't this be `delay` instead of `delay - 1`?
If you haven't reached the delay the callback shouldn't be called wether the component unmounted or not.

carolopolo: you're right. we don't even need the clock tbh. just the unmount!

packages/react-hooks/README.md
```diff
@@ -48,9 +49,32 @@ function MyComponent() {
 }
 ```
 
+### `useDelayedCallback()`
+
+This hook provides a `callbackWithDelay` function. It takes a callback and a delay in milliseconds. This might be useful when you want to invoke a callback after a certain delay.
```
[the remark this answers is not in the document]
carolopolo: absolutely true!

PR #1596, opened 2020-08-17
✨ add more default values to graphql universal provider

michenly:
### @shopify/react-graphql-universal-provider
- Add default cache if user did not provided it
- Add default client options `ssrMode`, `ssrForceFetchDelay`, `connectToDevTools`
- Update apollo related library to use a range instead of a specific version

Todo:

- [x] Add universal provider doc on whats the new default
- [x] Add test for ssrForceFetchDelay & connectToDevTools
- [x] included a test for customized cache

## Type of change

## Checklist


packages/react-graphql-universal-provider/src/test/GraphQLUniversalProvider.test.tsx
```diff
@@ -1,31 +1,42 @@
...
 
 import {GraphQLUniversalProvider} from '../GraphQLUniversalProvider';
 
-jest.mock('@shopify/react-graphql', () => {
-  /* eslint-disable @typescript-eslint/no-var-requires */
-  const ApolloClient = require('apollo-client');
-  const InMemoryCache = require('apollo-cache-inmemory');
-  const ApolloLink = require('apollo-link');
-  /* eslint-enable @typescript-eslint/no-var-requires */
+jest.mock('apollo-client', () => {
```
michenly, line 11: Took me a bit of time to figure out he mock situation, but I am happy with the result.

This implementation mock ApolloClient while calling the actual for result.
It is important to note that `apollo-client` uses alias for both default and name export for ApolloClient. I mocked both to be more maintainable.

packages/react-graphql-universal-provider/src/test/GraphQLUniversalProvider.test.tsx
```diff
@@ -36,56 +47,192 @@ describe('<GraphQLUniversalProvider />', () => {
     );
 
     expect(graphQL).toContainReactComponent(ApolloProvider, {
-      client: expect.any(ApolloClient),
+      client: expect.any(require.requireActual('apollo-client').ApolloClient),
```
michenly, line 50: Have to check the actual value for type instead of the mock value.

packages/react-graphql-universal-provider/src/test/GraphQLUniversalProvider.test.tsx
```diff
@@ -36,56 +47,192 @@ describe('<GraphQLUniversalProvider />', () => {
...
+      const cache = new InMemoryCache({addTypename: true});
+      const clientOptions = {cache};
+
+      const graphQL = mount(
+        <GraphQLUniversalProvider createClientOptions={() => clientOptions} />,
+      );
+
+      expect(ApolloClient).toHaveBeenLastCalledWith(
+        expect.objectContaining({cache}),
+      );
```
michenly, line 91: The test check if the contractor was called with the correct options.

PR #1598, opened 2020-08-18
Allow proxy option in react-server webpack plugin

gmcgibbon:
## Description

Expose proxy option from https://github.com/Shopify/quilt/pull/1591 in the webpack plugin for `react-server`.

## Type of change

## Checklist

### DIFF
--- a/packages/react-server/src/webpack-plugin/test/webpack-plugin.test.ts
+++ b/packages/react-server/src/webpack-plugin/test/webpack-plugin.test.ts
@@ -172,6 +172,32 @@ describe('webpack-plugin', () => {
     );
 
     it(
+      'uses default server configuration options',
+      async () => {
+        const name = 'default-server-config';
+        const customConfig = {
+          port: 3000,
+          host: '127.0.0.1',
+          assetPrefix: 'https://localhost/webpack/assets',
+          basePath: '.',
+        };
+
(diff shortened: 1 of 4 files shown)


packages/react-server/src/webpack-plugin/webpack-plugin.ts
```diff
@@ -92,6 +94,7 @@ function serverSource(options: Options, compiler: Compiler) {
       port: ${port},
       ip: ${JSON.stringify(host)},
       assetPrefix: ${JSON.stringify(assetPrefix)},
+      proxy: ${JSON.stringify(proxy)},
```
michenly, line 97: dont think we need to JSON.stringify this option, otherwise we will always get string instead boolean value.

packages/react-server/src/webpack-plugin/test/webpack-plugin.test.ts
```diff
@@ -197,6 +198,7 @@ describe('webpack-plugin', () => {
           expect(serverModule.source).toMatch(
             `assetPrefix: "${customConfig.assetPrefix}"`,
           );
+          expect(serverModule.source).toMatch(`proxy: "${customConfig.proxy}"`);
```
michenly, line 201: can we add one more test case to test the default value

gmcgibbon:
Do webpack config changes need changelogs?

PR #1599, opened 2020-08-18
Add a document describing the React I18n schema

movermeyer:
## Description

React I18n is a schema fundamentally comprised of a structure of nested key-value pairs (KVPs). 
However, there are additional constraints placed on the contents of that structure. 

This document captures the nuances of both the structure and contents of the schema.
It also defines a pair of serialization formats (`JSON` and `JSON+Comments`) that consumers of React i18n files are expected to be able to parse.
Finally, it compares the React I18n schema to the Rails I18n schema, pointing out important differences

Ref: https://vault.shopify.io/projects/7976
Ref: https://docs.google.com/document/d/11wy2Sc3ovlu4Hv-tlv9Gm00zKY-dPg7EFYn60F9CKBA/edit

## Note to Reviewers

You probably want to review the [rendered version of this file](https://github.com/Shopify/quilt/blob/mo.react_i18n_json_format/packages/react-i18n/docs/react_i18n_schema.md), instead of the raw Markdown.

## Type of change

Documentation.

------

cc @sjahu @Shopify/web-foundations @TheMallen

### DIFF
--- a/packages/react-i18n/README.md
+++ b/packages/react-i18n/README.md
@@ -148,12 +148,14 @@ Most notably, you will frequently use `i18n`’s `translate()` method. This meth
 
 The most commonly-used feature of the `@shopify/react-i18n` library is looking up translations. In this library, translations are provided **for the component that need them**, and are **available for ancestors of the component**. This allows applications to grow while keeping translations manageable, makes it clearer where to add new translations, and follows Shopify’s principle of [isolation over integration](https://github.com/Shopify/web-foundations/blob/main/handbook/Principles/4%20-%20Isolation%20over%20integration.md) by collocating translations with all other component assets.
 
+Translations are serialized into files according to [the React I18n schema](./packages/react-i18n/docs/react_i18n_schema.md).
+
 Translations are provided using two keys in the `withI18n` decorator:
 
 - `fallback`: a translation file to use when translation keys are not found in the locale-specific translation files. These will usually be your English translations, as they are typically the most complete.
 - `translations`: a function which takes the locale and returns one of: nothing (no translations for the locale), a dictionary of key-value translation pairs, or a promise of one of the above. The `translations` function can also throw and `react-i18n` will handle the situation gracefully. Alternatively, you can pass an object where the keys are locales, and the values are either translation dictionaries, or promises for translation dictionaries.
 
-We recommend that you colocate your translation files in a `./translations` directory and that you include an `en.json` file in that directory as your fallback. We give preferential treatment to this structure via a [babel plugin](#Babel) that will automatically fill in the arguments to `useI18n`/ `withI18n` for you.
+We recommend that you co-locate your translation files in a `./translations` directory and that you include an `en.json` file ([schema specification](docs/react_i18n_schema.md)) in that directory as your fallback. We give preferential treatment to this structure via a [babel plugin](#Babel) that will automatically fill in the arguments to `useI18n`/ `withI18n` for you.
 
 If you provide any of the above options, you must also provide an `id` key, which gives the library a way to store the translation dictionary. If you're using the [babel plugin](#Babel), this `id` will the automatically generated based on the relative path to your component from your project's root directory.
 
@@ -262,7 +264,7 @@ if (keyExists) {
 
 ##### Pluralization
 
-`@shopify/react-i18n` handles pluralization similarly to [Rails’ default i18n utility](https://guides.rubyonrails.org/i18n.html#pluralization). The key is to provide the plural-dependent value as a `count` variable. `react-i18n` then looks up the plural form using [`Intl.PluralRules`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/PluralRules) and, within the keypath you have specified for the translation, will look up a nested translation matching the plural form:
+`@shopify/react-i18n` handles pluralization similarly to [Rails’ default i18n utility](https://guides.rubyonrails.org/i18n.html#pluralization) ([with some minor differences](docs/react_i18n_schema.md#pluralization-1)). The key is to provide the plural-dependent value as a `count` variable. `react-i18n` then looks up the plural form using [`Intl.PluralRules`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/PluralRules) and, within the keypath you have specified for the translation, will look up a nested translation matching the plural form:
 
 ```ts
 // Assuming a dictionary like:
--- a/packages/react-i18n/docs/react_i18n_schema.md
+++ b/packages/react-i18n/docs/react_i18n_schema.md
@@ -1,0 +1,291 @@
+# Introduction <!-- omit in toc -->
+
+React I18n is a schema comprised of a structure of nested key-value pairs (KVPs). This document captures the nuances of the constraints placed on the schema by the different levels of concern.
+
+<!-- Created using "Markdown All in One" extension for VS Code -->
+- [History](#history)
+  - [Version 1.0](#version-10)
+- [Versioning](#versioning)
+- [Basic Structure](#basic-structure)
+- [Pluralization](#pluralization)
+  - [Reserved pluralization keys](#reserved-pluralization-keys)
+  - [Invalid pluralization keys](#invalid-pluralization-keys)
+  - [Detecting pluralization contexts](#detecting-pluralization-contexts)
+- [Serialization](#serialization)
+  - [Serialization to JSON](#serialization-to-json)
+    - [Comments](#comments)
+  - [Serialization to JSON+Comments](#serialization-to-jsoncomments)
+- [Appendix A: Comparison with Rails I18n](#appendix-a-comparison-with-rails-i18n)
+    - [Root locale key](#root-locale-key)
+    - [Pluralization](#pluralization-1)
+    - [Interpolation](#interpolation)
+
+# History
+
+## Version 1.0
+
+Prior to the creation of this document, the React I18n schema was not documented.
+With the creation of this document, the schema was specified and given the version identifier `1.0`.
+
+# Versioning
+
+The version identifier for the schema is made of two parts, `<major>` and `<minor>`, separated by a `.`: `<major>.<minor>`.
+
+The `<major>` version number will be incremented any time a change is introduced that would mean that existing parsers would break while parsing the new schema. Put another way, software that supports parsing version `x.y` of the schema will always be able to parse any version `x.z`, where `z >= y`.
+
+The `<minor>` version number will be incremented for any other changes to the schema and/or serialization formats. For example, changes that make the schema "more strict" in what is accepted as valid can be done as increments to the `<minor>` version number, since parsers of older versions are still able to parse the newer versions, albeit "more loosely" than what a modern parser would.
+
+# Basic Structure
+
+React I18n is made of nested dictionaries of key-value pairs (KVPs). Each of the keys are strings, and the values are either strings, or a nested dictionary.
+
+In TypeScript, the type is a recursive definition:
+
+```typescript
+export interface TranslationDictionary {
+  [key: string]: string | TranslationDictionary;
+}
+```
+
+When [serialized to JSON](#serialization-to-json), each of the dictionaries become JSON `object` types, and the keys and leaf values are both JSON `string` types.
+
+**Example React I18n JSON file**
+
+```json
+{
+  "hello": "My name is {name}",
+  "parent": {
+    "child_1": "I am the first child of `parent`!",
+    "child_2": {
+      "grandchild_1": "I am the first grandchild of `parent`!",
+      "grandchild_2": "I am the second grandchild of `parent`!"
+    }
+  }
+}
+```
+
+# Pluralization
+
+Pluralization is handled by providing the pluralization keys required by the language as leaf KVPs, creating a "pluralization context".
+
+**Example English file with pluralization keys:**
+
+```jsonc
+{
+  "cars": {  // All the children of `cars` are in a "pluralization context"
+    "one": "I have {count} car",
+    "other": "I have {count} cars"
+  }
+}
+```
+
+Different languages require different sets of pluralization keys. For example, Polish (`pl`) has 4 required pluralization keys: `one`, `few`, `many`, and `other`. The Polish version of the previous example would be:
+
+**Example Polish file with pluralization keys:**
+
+```json
+{
+  "cars": {
+    "one": "Mam {count} samochód",
+    "few": "Mam {count} samochody",
+    "many": "Mam {count} samochodów",
+    "other": "Mam {count} samochodu"
+  }
+}
+```
+
+## Reserved pluralization keys
+
+In order to avoid key collisions, the following keys are reserved for use as pluralization keys and must not be used as keys of leaf KVPs outside of a pluralization context:
+
+* `few`
+* `many`
+* `one`
+* `other`
+* `two`
+* `zero`
+
+These keys are derived from the names given to the pluralization rules of all languages, as defined in the [Unicode Consortium's Common Locale Data Repository (CLDR)](https://github.com/unicode-org/cldr).
+
+This list is current as of version `37` of the CLDR. In the unlikely event that new pluralization rule names are added, they will be added to this reserved list in a future version of this specification.
+
+**Example INVALID English file:**
+```jsonc
+{
+  "foo": "bar",
+  // 🚫 INVALID: The use of `other` makes this a pluralization context,
+  // but it is missing the required `one` sibling key that English needs
+  // therefore it is not a valid pluralization context.
+  // It is either:
+  //   * not intended to be a pluralization context (in which case the reserved `other` key needs to be renamed)
+  //   * missing the `one` sibling key (in which case the `one` key needs to be added)
+  "other": "Other"
+}
+```
+
+## Invalid pluralization keys
+
+Furthermore, pluralization keys that are not required by the language must not be used, even within a pluralization context.
+
+For example, even though `zero` is a required pluralization key for some languages (eg. Arabic), it is not a required pluralization key in English. Therefore, the `zero` key must not be present in an English file alongside the pluralization KVPs required by English.
+
+**Example INVALID English file:**
+```jsonc
+{
+  "cars": {
+    "zero": "I have no cars", // 🚫 INVALID: `zero` is not a required pluralization key for English!
+    "one": "I have {count} car",
+    "other": "I have {count} cars"
+  }
+}
+```
+
+## Detecting pluralization contexts
+
+These constraints allow for a simple algorithm for identifying pluralization contexts:
+
+* If a node has child leaf KVPs for **any** of the pluralization keys required by the language, then its children are within a pluralization context.
+
+**Note**: in order to be a valid pluralization context, any other required pluralization keys must also be present within the pluralization context.
+
+# Serialization
+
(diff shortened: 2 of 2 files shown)


packages/react-i18n/README.md
```diff
@@ -148,12 +148,14 @@ Most notably, you will frequently use `i18n`’s `translate()` method. This meth
...
 
-We recommend that colocate your translations files in a `./translations` directory and that you include an `en.json` file in that directory as your fallback. We give preferential treatment to this structure via a [babel plugin](#Babel) that will automatically fill in the arguments to `useI18n`/ `withI18n` for you.
+We recommend that colocate your translations files in a `./translations` directory and that you include an `en.json` file ([schema specification](https://github.com/Shopify/quilt/blob/master/packages/react-i18n/docs/react_i18n_schema.md)) in that directory as your fallback. We give preferential treatment to this structure via a [babel plugin](#Babel) that will automatically fill in the arguments to `useI18n`/ `withI18n` for you.
```
sjahu, line 158:
```suggestion
We recommend that you co-locate your translations files in a `./translations` directory and that you include an `en.json` file ([schema specification](https://github.com/Shopify/quilt/blob/master/packages/react-i18n/docs/react_i18n_schema.md)) in that directory as your fallback. We give preferential treatment to this structure via a [babel plugin](#Babel) that will automatically fill in the arguments to `useI18n`/ `withI18n` for you.
```

packages/react-i18n/docs/react_i18n_schema.md
```diff
@@ -0,0 +1,295 @@
...
+
+If a node has child leaf KVPs for each of the pluralization keys required by the language, then its children are within a pluralization context.
+
+Any additional children within the context are invalid.
+
+# Serialization
+
+React I18n can be serialized to different formats. However, not all serialization formats are created equal.
+
+Two serializations are officialy defined. Consumers of React I18n `1.0` files are expected to be able to parse files serialized into each of these serialization formats.
```
gmcgibbon, line 150: Ambiguous version here.

packages/react-i18n/docs/react_i18n_schema.md
```diff
@@ -0,0 +1,295 @@
...
+
+| Serialization Format | JSON+Comments | JSON |
+| -------------------- | ------------- | ---- |
+| Comments             | ✅             | ❌    |
+
+## Serialization to JSON
```
gmcgibbon, line 156: I don't think this is worth repeating. Can we remove it?

packages/react-i18n/docs/react_i18n_schema.md
```diff
@@ -0,0 +1,295 @@
...
+        context comment for `grandchild_1`!
+      */
+      "grandchild_1": "I am the first grandchild of `parent`!",
+      "grandchild_2": "I am the second grandchild of `parent`!" /* This is a trailing context comment for `grandchild_2`! */
+    }
+  }
+}
+```
+
+Comments follow the [JSON5 syntax for comments](https://spec.json5.org/#comments): `//` signals the start of a comment that ends at the end of the line. `/*` signals the start of a comment that ends with the next `*/`.
```
gmcgibbon, line 211: This tangent about JSON5 and comments is a bit irrelevant to the schema definition don't you think? Can we move to another doc that talks about serialization for those interested and link it here?

packages/react-i18n/docs/react_i18n_schema.md
```diff
@@ -0,0 +1,295 @@
...
+
+**React I18n**
+
+```json
+{
+  "foo": "bar"
+}
+```
+
+Note the `en` (for English) locale key that is the root key for the Rails I18n file. This key is not present in the React I18n file.
```
gmcgibbon, line 245:
```suggestion
Note the `en` (for English) locale key that is the root key for the Rails I18n file. This key is not present in the React I18n file because it is denoted by filename.
```

packages/react-i18n/docs/react_i18n_schema.md
```diff
@@ -0,0 +1,295 @@
+# Introduction <!-- omit in toc -->
```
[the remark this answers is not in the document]
atesgoral: Seems to omit the section from the ToC:
image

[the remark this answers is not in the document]
movermeyer: Yes, that is the syntax that [Markdown All In One](https://marketplace.visualstudio.com/items?itemName=yzhang.markdown-all-in-one) uses to prevent the heading from getting added to the auto-generated table of content.

packages/react-i18n/docs/react_i18n_schema.md
```diff
@@ -0,0 +1,295 @@
...
+- [Comparison with Rails I18n](#comparison-with-rails-i18n)
+    - [Root locale key](#root-locale-key)
+    - [Pluralization](#pluralization-1)
+    - [Interpolation](#interpolation)
+
+# History
+
+## Version 1.0
+
+React I18n's schema was loosely based on the Rails I18n schema.
```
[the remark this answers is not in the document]
movermeyer: Done.

packages/react-i18n/docs/react_i18n_schema.md
movermeyer: Done.

packages/react-i18n/docs/react_i18n_schema.md
```diff
@@ -0,0 +1,295 @@
...
+
+Prior to the creation of this document, the React I18n schema was not documented.
+With the creation of this document, the schema was specified and given the version identifier `1.0`.
```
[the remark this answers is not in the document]
movermeyer: The `@shopify/react-i18n` library changes independently of the React I18n specification. So they have different version numbers.

React I18n is the specification, while `@shopify/react-i18n` is a client implementation that supports files in the React I18n schema.

This is the same as any other schema / format specification. For example, YAML is [version 1.2](https://yaml.org/spec/1.2/spec.html), but each implementation library has their own version numbers.

packages/react-i18n/docs/react_i18n_schema.md
movermeyer: Changed to `Consumers of React I18n `1.0` schema files` (added `schema`).

packages/react-i18n/docs/react_i18n_schema.md
movermeyer: Removed.

packages/react-i18n/README.md
```diff
@@ -148,12 +148,14 @@ Most notably, you will frequently use `i18n`’s `translate()` method. This meth
...
 - `translations`: a function which takes the locale and returns one of: nothing (no translations for the locale), a dictionary of key-value translation pairs, or a promise of one of the above. The `translations` function can also throw and `react-i18n` will handle the situation gracefully. Alternatively, you can pass an object where the keys are locales, and the values are either translation dictionaries, or promises for translation dictionaries.
 
-We recommend that you colocate your translation files in a `./translations` directory and that you include an `en.json` file in that directory as your fallback. We give preferential treatment to this structure via a [babel plugin](#Babel) that will automatically fill in the arguments to `useI18n`/ `withI18n` for you.
+We recommend that you co-locate your translation files in a `./translations` directory and that you include an `en.json` file ([schema specification](./packages/react-i18n/docs/react_i18n_schema.md)) in that directory as your fallback. We give preferential treatment to this structure via a [babel plugin](#Babel) that will automatically fill in the arguments to `useI18n`/ `withI18n` for you.
```
sjahu, line 158: Broken link:
```suggestion
We recommend that you co-locate your translation files in a `./translations` directory and that you include an `en.json` file ([schema specification](/packages/react-i18n/docs/react_i18n_schema.md)) in that directory as your fallback. We give preferential treatment to this structure via a [babel plugin](#Babel) that will automatically fill in the arguments to `useI18n`/ `withI18n` for you.
```
or
```suggestion
We recommend that you co-locate your translation files in a `./translations` directory and that you include an `en.json` file ([schema specification](docs/react_i18n_schema.md)) in that directory as your fallback. We give preferential treatment to this structure via a [babel plugin](#Babel) that will automatically fill in the arguments to `useI18n`/ `withI18n` for you.
```

packages/react-i18n/README.md
```diff
@@ -262,7 +264,7 @@ if (keyExists) {
 
 ##### Pluralization
 
-`@shopify/react-i18n` handles pluralization similarly to [Rails’ default i18n utility](https://guides.rubyonrails.org/i18n.html#pluralization). The key is to provide the plural-dependent value as a `count` variable. `react-i18n` then looks up the plural form using [`Intl.PluralRules`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/PluralRules) and, within the keypath you have specified for the translation, will look up a nested translation matching the plural form:
+`@shopify/react-i18n` handles pluralization similarly to [Rails’ default i18n utility](https://guides.rubyonrails.org/i18n.html#pluralization) ([with some minor differences](./packages/react-i18n/docs/react_i18n_schema.md#pluralization-1)). The key is to provide the plural-dependent value as a `count` variable. `react-i18n` then looks up the plural form using [`Intl.PluralRules`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/PluralRules) and, within the keypath you have specified for the translation, will look up a nested translation matching the plural form:
```
sjahu, line 267: as above
```suggestion
`@shopify/react-i18n` handles pluralization similarly to [Rails’ default i18n utility](https://guides.rubyonrails.org/i18n.html#pluralization) ([with some minor differences](/packages/react-i18n/docs/react_i18n_schema.md#pluralization-1)). The key is to provide the plural-dependent value as a `count` variable. `react-i18n` then looks up the plural form using [`Intl.PluralRules`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/PluralRules) and, within the keypath you have specified for the translation, will look up a nested translation matching the plural form:
```
or
```suggestion
`@shopify/react-i18n` handles pluralization similarly to [Rails’ default i18n utility](https://guides.rubyonrails.org/i18n.html#pluralization) ([with some minor differences](docs/react_i18n_schema.md#pluralization-1)). The key is to provide the plural-dependent value as a `count` variable. `react-i18n` then looks up the plural form using [`Intl.PluralRules`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/PluralRules) and, within the keypath you have specified for the translation, will look up a nested translation matching the plural form:
```

packages/react-i18n/docs/react_i18n_schema.md
```diff
@@ -0,0 +1,284 @@
...
+React I18n's schema is loosely based on the [Rails I18n](https://guides.rubyonrails.org/i18n.html) schema.
+
+Prior to the creation of this document, the React I18n schema was not documented.
+With the creation of this document, the schema was specified and given the version identifier `1.0`.
+
+# Versioning
+
+The version identifier for the schema is made of two parts, `<major>` and `<minor>`, separated by a `.`: `<major>.<minor>`.
+
+the `<major>` version number will be incremented any time a change is introduced that would mean that existing parsers would break while parsing the new schema. Put another way, software that supports parsing version `x.y` of the schema will always be able to parse any version `x.z`, where `z >= y`.
```
sjahu, line 36:
```suggestion
The `<major>` version number will be incremented any time a change is introduced that would mean that existing parsers would break while parsing the new schema. Put another way, software that supports parsing version `x.y` of the schema will always be able to parse any version `x.z`, where `z >= y`.
```

sjahu, line 36: > Put another way, software that supports parsing version `x.y` of the schema will always be able to parse any version `x.z`, where `z >= y`.

True, but should these two paragraphs be re-organized so that this is at the end? This example also demonstrates the meaning of a `<minor>` version change.

packages/react-i18n/docs/react_i18n_schema.md
```diff
@@ -0,0 +1,284 @@
...
+```jsonc
+{
+  "cars": {  // All the children of `cars` are in a "pluralization context"
+    "one": "I have {count} car",
+    "other": "I have {count} cars"
+  }
+}
+```
+
+Different languages require different sets of pluralization keys. For example, the Polish (`pl`) has 4 required pluralization keys: `one`, `few`, `many`, and `other`. The Polish version of the previous example would be:
```
sjahu, line 83:
```suggestion
Different languages require different sets of pluralization keys. For example, Polish (`pl`) has 4 required pluralization keys: `one`, `few`, `many`, and `other`. The Polish version of the previous example would be:
```
or
```suggestion
Different languages require different sets of pluralization keys. For example, the Polish language (`pl`) has 4 required pluralization keys: `one`, `few`, `many`, and `other`. The Polish version of the previous example would be:
```

packages/react-i18n/docs/react_i18n_schema.md
```diff
@@ -0,0 +1,284 @@
...
+    "few": "Mam {count} samochody",
+    "many": "Mam {count} samochodów",
+    "other": "Mam {count} samochodu",
+  }
+}
+```
+
+## Reserved pluralization keys
+
+In order to avoid collisions, the following keys are reserved for use as pluralization keys, and must not be used as keys of leaf KVPs outside of the pluralization context:
```
sjahu, line 99:
```suggestion
In order to avoid name collisions, the following keys are reserved for use as pluralization keys and must not be used as keys of leaf KVPs outside of a pluralization context:
```

sjahu: Can I use reserved key names for internal nodes?

packages/react-i18n/docs/react_i18n_schema.md
```diff
@@ -0,0 +1,284 @@
...
+```jsonc
+{
+  "foo": "bar",
+  "other": "Other" // INVALID: `other` is a keyword reserved for use in pluralization contexts!
+}
+```
+
+## Invalid pluralization keys
+
+Further, pluralization keys that are not required by the language must not be used, even within a pluralization context.
```
sjahu, line 122:
```suggestion
Furthermore, pluralization keys that are not required by the language must not be used, even within a pluralization context.
```

packages/react-i18n/docs/react_i18n_schema.md
```diff
@@ -0,0 +1,284 @@
...
+
+For example, even though `zero` is a required pluralization key for some languages (eg. Arabic), it is not a required pluralization key in English. Therefore, the `zero` key should not be present in an English file alongside the pluralization KVPs required by English.
```
sjahu, line 124:
```suggestion
For example, even though `zero` is a required pluralization key for some languages (eg. Arabic), it is not a required pluralization key in English. Therefore, the `zero` key must not be present in an English file alongside the pluralization KVPs required by English.
```

packages/react-i18n/docs/react_i18n_schema.md
```diff
@@ -0,0 +1,284 @@
...
+    "other": "I have {count} cars"
+  }
+}
+```
+
+## Detecting pluralization contexts
+
+These constraints allow for an algorithm for identifying pluralization contexts:
+
+If a node has child leaf KVPs for each of the pluralization keys required by the language, then its children are within a pluralization context.
```
sjahu, line 141: True, but could this be strengthened to simply "If a node has child leaf KVPs for any of the pluralization keys, then ..."? Since pluralization keys outside this context are invalid?

packages/react-i18n/docs/react_i18n_schema.md
```diff
@@ -0,0 +1,284 @@
...
+- [Comparison with Rails I18n](#comparison-with-rails-i18n)
+    - [Root locale key](#root-locale-key)
+    - [Pluralization](#pluralization-1)
+    - [Interpolation](#interpolation)
+
+# History
+
+## Version 1.0
+
+React I18n's schema is loosely based on the [Rails I18n](https://guides.rubyonrails.org/i18n.html) schema.
```
sjahu, line 27: Is this necessary with the comparison section in the TOC? Since that section starts with the same sentence.

packages/react-i18n/docs/react_i18n_schema.md
```diff
@@ -0,0 +1,284 @@
...
+
+**React I18n**
+
+```json
+{
+  "foo": "bar"
+}
+```
+
+Note the `en` (for English) locale key that is the root key for the Rails I18n file. This key is not present in the React I18n file because it is denoted by filename.
```
sjahu, line 234: > denoted by filename

If this part of the specification it should be more explicit. If not, we should add a sentence stating that React I18n files are locale-agnostic (not true, though, since the set of valid pluralization keys is determined by the locale).

packages/react-i18n/docs/react_i18n_schema.md
```diff
@@ -0,0 +1,284 @@
...
+    "one": "I have {count} car",
+    "other": "I have {count} cars"
+  },
+  "no_cars": "I don't have any cars"
+}
+```
+
+### Interpolation
+
+While neither schemas define an interpolation syntax, the most common syntax used in Rails I18n is `%{}`, while in React I18n, the most common syntax is `{}`.
```
sjahu, line 270:
```suggestion
While neither schema defines an interpolation syntax, the most common syntax used in Rails I18n is `%{}`, while in React I18n, the most common syntax is `{}`.
```

sjahu (approved):
Looks good! Added a couple questions and a bunch of pedantic suggestions

packages/react-i18n/docs/react_i18n_schema.md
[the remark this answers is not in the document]
gmcgibbon: If we define the "React I18n specification" and `@shopify/react-i18n` is the only library parsing a target schema version, I still don't see the reason for distinction. I might be the only person who thinks this, but I'd like to see what others think.

rafaelxy (approved):

packages/react-i18n/docs/react_i18n_schema.md
movermeyer: > Can I use reserved key names for internal (non-leaf) nodes?

@sjahu, yes. AFAIK, there is no need to restrict usage in non-leaf nodes.

packages/react-i18n/docs/react_i18n_schema.md
movermeyer: You are right. Fixed.

packages/react-i18n/docs/react_i18n_schema.md
movermeyer: Changed to make it clear that the locale doesn't have to be part of the filename, but typically is.

packages/react-i18n/docs/react_i18n_schema.md
movermeyer: Fixed.

packages/react-i18n/docs/react_i18n_schema.md
movermeyer: Removed.

packages/react-i18n/docs/react_i18n_schema.md
movermeyer: 🤷 ... I'm going to keep it this way.

packages/react-i18n/docs/react_i18n_schema.md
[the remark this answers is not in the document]
gmcgibbon: > @shopify/react-i18n isn't the only parser/client though.

I see, sorry I didn't realize that. Separate versioning seems good to me under those circumstances 👍

gmcgibbon (approved):

PR #1601, opened 2020-08-19
🎨 Refactor  `useDelayedCallback` unit test

carolopolo:
## Description

Just a small patch to improve a unit test

## Type of change

## Checklist


packages/react-hooks/src/hooks/tests/delayed-callback.test.tsx
```diff
@@ -89,6 +89,10 @@ describe('useDelayedCallback', () => {
 
     fakeComponent.unmount();
 
+    fakeComponent.act(() => {
+      clock.tick(1.5 * delay);
+    });
+
```
loic-d, line 95: No more false positive 😄 👍

PR #1603, opened 2020-08-19
Fix web-foundations link in react-testing decision

js-aff:
## Description

Change `master` to `main` branch in `react-testing` decision log. 

## Type of change

- [x] react-testing Patch: documentation 

## Checklist


416serg (approved):
🎩 ed, links work

PR #1608, opened 2020-08-24
✨ pass request id header to node automatically

michenly:
## Description

Part of https://github.com/Shopify/shopify-rails/issues/282

## Type of change

## Checklist

### DIFF
--- a/gems/quilt_rails/test/quilt_rails/react_renderable_test.rb
+++ b/gems/quilt_rails/test/quilt_rails/react_renderable_test.rb
@@ -11,7 +11,7 @@ module Quilt
         render_react,
         reverse_proxy(
           url,
-          headers: { 'X-Quilt-Data': {}.to_json }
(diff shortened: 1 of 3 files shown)


gems/quilt_rails/test/quilt_rails/react_renderable_test.rb
```diff
@@ -59,5 +66,8 @@ def reverse_proxy(url, headers: {})
     def form_authenticity_token
       'foo'
     end
+
+    # Stubbing request that exist in a controller
+    attr_reader :request
```
gmcgibbon, line 71:
```suggestion
    def request
      @request ||= ActionDispatch::TestRequest.create.tap do |request|
        request.request_id = SecureRandom.uuid
      end
    end
```
And then `request.request_id` in tests if you want to shorthand what the setup block does.

PR #1609, opened 2020-08-24
✨ automatically pass request id header to graphql request

michenly:
## Description

Part of https://github.com/Shopify/shopify-rails/issues/282

This PR aim to pass request-id header to any graphql request (the way `useRequestHeader` works, this will usually only happen with server side graphql requests).

After which we can use the unique request-id value to trace a request from Rails => Node => Graphql request to Rails and back to see if what actually cause the error during a browser request.

To 🎩 
You need this PR so request id will pass from Rails to Node https://github.com/Shopify/quilt/pull/1608
After I use `yarn tophat react-graphql-universal-provider ../consuming-repo` and log the request id in graphql controller to see the same value being pass.

## Type of change

## Checklist

### DIFF
--- a/packages/react-graphql-universal-provider/src/request-id-link.ts
+++ b/packages/react-graphql-universal-provider/src/request-id-link.ts
@@ -1,0 +1,10 @@
+import {setContext} from 'apollo-link-context';
+
+export function createRequestIdLink(requestId: string) {
+  return setContext((_, {headers}) => ({
+    headers: {
+      ...headers,
+      'X-Request-ID': requestId,
+    },
+  }));
+}
--- a/packages/react-graphql-universal-provider/src/test/GraphQLUniversalProvider.test.tsx
+++ b/packages/react-graphql-universal-provider/src/test/GraphQLUniversalProvider.test.tsx
@@ -5,6 +5,7 @@ import {extract} from '@shopify/react-effect/server';
 import {mount} from '@shopify/react-testing';
 import {HtmlManager, HtmlContext} from '@shopify/react-html';
 import {ApolloProvider} from '@shopify/react-graphql';
+import {NetworkContext, NetworkManager} from '@shopify/react-network';
 
 import {GraphQLUniversalProvider} from '../GraphQLUniversalProvider';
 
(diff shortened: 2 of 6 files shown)


packages/react-graphql-universal-provider/src/request-id-link.ts
```diff
@@ -0,0 +1,10 @@
+import {setContext} from 'apollo-link-context';
+
```
gmcgibbon, line 2: I don't understand why some source file names are hyphenized and some are camelized. How do you decide which one is correct?

packages/react-graphql-universal-provider/src/test/GraphQLUniversalProvider.test.tsx
```diff
@@ -20,12 +21,16 @@ jest.mock('apollo-client', () => {
...
 jest.mock('../utilities', () => ({
+  ...require.requireActual('../utilities'),
   isServer: jest.fn(),
 }));
+const {isServer} = require.requireMock('../utilities');
 
-const {isServer} = require.requireMock('../utilities') as {
-  isServer: jest.Mock;
-};
+jest.mock('../request-id-link', () => ({
```
gmcgibbon, line 29: Is this mock necessary if the apollo client is already mocked?

packages/react-graphql-universal-provider/src/test/request-id-link.test.ts
```diff
@@ -0,0 +1,25 @@
+import gql from 'graphql-tag';
+import {ApolloLink, Observable} from 'apollo-link';
+
+import {createRequestIdLink} from '../request-id-link';
+
+const query = gql`
```
gmcgibbon, line 6: I don't see this referenced anywhere, do we need it?

michenly: 🔥 ing!

packages/react-graphql-universal-provider/src/request-id-link.ts
michenly: React components => camel case
Rest of the javascript things => hypenized

packages/react-graphql-universal-provider/src/test/GraphQLUniversalProvider.test.tsx
michenly: I am mainly mocking it so I can do 
`expect(createRequestIdLink).toHaveBeenCalled()`

gmcgibbon: You also test this method individually so that seems reasonable 👍

PR #1610, opened 2020-08-25
Modified Address package to also query for Optional Labels.

hassaanali1:
## Description

The original problem i'm trying to solve is in [this issue](https://github.com/Shopify/intl-broken-experiences/issues/176)

The idea is to allow the `AddressForm` component in `web` to choose between optional labels and the required labels for its fields.

Ex: for address 2 I want to use the version:

`Apartment, suite, etc. (optional)` instead of just `Apartment, suite, etc.`

The different labels for each country and locale come from [country service.](https://github.com/Shopify/country-service) 

`web` interacts with `country service` through the `address` package here in quilt. `address` fetches the data from `country service` using a GraphQL call.

In order to make this change I need to do the following:

1. Allow Country Service to return optional labels in its GraphQL endpoint. [This has already been done in this PR.](https://github.com/Shopify/country-service/pull/520)
2. Modify the `address` package in quilt to query for the new optionalLabels field I added in `country service`. Thats what this PR is for. 
3. Modify `web` to make use of the newly returned optionalLabels and allow the `AddressForm` component to decide whether to display the optional label or the regular one. Another PR will be made after this one has been merged, but the work is already done and can be viewed in [this commit](https://github.com/Shopify/web/commit/7a9438da91d384d393b73789fb2a8c00ff46b620). 

## Changes

What I've done here is change the GraphQL call in `address` to include optionalLabels. I've chosen to only include `address2` in the query to avoid over-fetching. If in the future other fields are required then this package will need to be changed again.

The tests and their mocks have also been changed to include optionalLabels.

## Type of change

- [x] <--address--> Minor: New feature (Changed the GraphQL query)
- [x] <--address-consts--> Minor: New feature (Added optionalLabels to the `Country` interface)
- [x] <--address-mocks--> Minor: New feature (Changed the mocks to include the newly fetched optionalLabels)

## Checklist

## Tophatting

These changes can be tested locally by cherry-picking [this commit](https://github.com/Shopify/web/commit/7a9438da91d384d393b73789fb2a8c00ff46b620) from web.

You then need to build and tophat the address package using `yarn tophat`

You can then navigate to `admin/billing/settings` in your local setup and click on `Add Payment Method`. If all goes well the optional label for address2 will be there rather than the regular one.

## Other notes

* This is my first time contributing to quilt so I'm not entirely sure whether the changes are major or minor so some feedback would be appreciated there.

* The mocks for the `countries` call were created by doing a call to the live service.

### DIFF
--- a/packages/address-consts/CHANGELOG.md
+++ b/packages/address-consts/CHANGELOG.md
@@ -5,7 +5,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## [Unreleased] -->
+## [Unreleased]
+
+- Add `optionalLabels` to `Country` interface. [#1610](https://github.com/Shopify/quilt/pull/1610)
 
 ## [2.0.0] - 2020-03-10
 
--- a/packages/address/src/graphqlQuery.ts
+++ b/packages/address/src/graphqlQuery.ts
@@ -19,6 +19,9 @@ query countries($locale: SupportedLocale!) {
       postalCode
       zone
     }
+    optionalLabels {
+      address2
+    }
     formatting {
       edit
       show
@@ -49,6 +52,9 @@ query country($countryCode: SupportedCountry!, $locale: SupportedLocale!) {
       phone
       postalCode
       zone
+    }
+    optionalLabels {
+      address2
     }
     formatting {
       edit
--- a/packages/address-consts/src/index.ts
+++ b/packages/address-consts/src/index.ts
@@ -69,6 +69,9 @@ export interface Country {
     postalCode: string;
     zone: string;
   };
+  optionalLabels: {
+    address2: string;
+  };
   formatting: {
     edit: string;
     show: string;
--- a/packages/address-mocks/CHANGELOG.md
+++ b/packages/address-mocks/CHANGELOG.md
@@ -5,7 +5,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## [Unreleased] -->
+## [Unreleased]
+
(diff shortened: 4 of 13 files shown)


packages/address-consts/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## [Unreleased] -->
+## [2.1.0] - 2020-08-25
```
benoitzohar, line 9: Should be "Unreleased" (same for the others changelogs)
```suggestion
## [Unreleased]
```

hassaanali1: Ah makes sense, i've changed it.

packages/address-consts/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## [Unreleased] -->
 
+## [Unreleased]
```
benoitzohar, line 10: You can remove the commented `<!-- ## [Unreleased] -->` above.

hassaanali1: Good call, removed.

packages/address/src/graphqlQuery.ts
```diff
@@ -19,6 +19,9 @@ query countries($locale: SupportedLocale!) {
       postalCode
       zone
     }
+    optionalLabels {
+      address2
```
plentz, line 23: shouldn't this be querying all fields too?

hassaanali1: It could query everything but for the purposes right now it could be over-fetching. In the countries call, every single country is fetched. My worry was that it would be wasteful or slow things down to have all these fields if only address2 is being used. 

It could be handy though because then you wouldn't have to update this package if you want new optional fields in the future. 

What do you think?

plentz: my question was more in the likes of "why do we only care about address2 optionality?". IDK how much querying all fields would impact in performance, to be honest (server side wise, I would say it's negligible)

hassaanali1: I've only included address2 because thats the only field i'd like this functionality for in `web` right now

PR #1611, opened 2020-08-25
Config logger

gmcgibbon:
## Description

Deprecates `Quilt::Logger.log`, adds more standard `Quilt.logger`.

This allows us to switch out test logger easily so running tests doesn't log anything. A lot of work for a small change, but this is a more standard method of logging for gems.

## Type of change

- [ ] `quilt_rails` Patch: Bug (non-breaking change which fixes an issue)
- [x] `quilt_rails` Minor: New feature (non-breaking change which adds functionality)
- [ ] `quilt_rails` Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist


gems/quilt_rails/test/quilt_rails/controllers/routing_test.rb
```diff
@@ -32,7 +32,7 @@ class UiController < ActionController::Base; end
       boot_dummy
 
       Rails.application.routes.draw do
-        mount(Quilt::Engine, at: '/quilt')
+        mount(::Quilt::Engine, at: '/quilt')
```
michenly, line 35: what does the prefix `::` mean?

gmcgibbon: `QuiltRails::Quilt` gets defined by a controller test, so `::` makes sure the toplevel `Quilt` module gets referenced.

gems/quilt_rails/lib/quilt_rails/configuration.rb
```diff
@@ -11,6 +11,7 @@ def initialize
 
       self.react_server_host     = "#{react_server_ip}:#{react_server_port}"
       self.react_server_protocol = ENV['REACT_SERVER_PROTOCOL'] || "http"
+      self.logger                = ::Logger.new($stdout)
```
Edouard-chin, line 14: looks weird to me to set this logger inside the configuration. `Quilt::Configuration.logger.log` :/ 

Why not do like Rails and just set a class variable in the Quilt module directly ?

gmcgibbon: We already have a configuration object with a configuration pattern though. The new API is `Quilt.logger` which delegates to `Quilt.configuration.logger`. As long as the API is the same I don't see any problems.

Edouard-chin: No problem indeed but since you are delegating `Quilt.logger` to the configuration I don't see the point of not adding that directly in the logger. That being said you also can't use `delegate` because that gem don't depend on active support

gmcgibbon: > No problem indeed but since you are delegating Quilt.logger to the configuration I don't see the point of not adding that directly in the Quilt module. 

I mainly want to maintain the contract of the configuration object. We would have to deprecate `Quilt.configuration` too.

> That being said you also can't use delegate because that gem don't depend on active support

It depends on railties which depends on active support, but I could add a dependency in the gemspec to make that more apparent.

Edouard-chin (approved):

PR #1612, opened 2020-08-26
EncodeURI origin app-bridge

keyfer:
## Description

Fixes https://hackerone.shopifycloud.com/reports/966419

Encode the `origin` parameter to block XSS vulnerabilities

## Type of change

- [x] `koa-shopify-auth` Patch: Bug (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/packages/koa-shopify-auth/src/auth/redirection-page.ts
+++ b/packages/koa-shopify-auth/src/auth/redirection-page.ts
@@ -12,7 +12,7 @@ export default function redirectionScript({origin, redirectTo, apiKey}) {
           var Redirect = AppBridge.actions.Redirect;
           var app = createApp({
             apiKey: '${apiKey}',
-            shopOrigin: '${origin}',
+            shopOrigin: "${encodeURI(origin)}",
           });
           var redirect = Redirect.create(app);
           redirect.dispatch(Redirect.Action.REMOTE, '${redirectTo}');
--- a/packages/koa-shopify-auth/src/auth/test/redirection-page.test.ts
+++ b/packages/koa-shopify-auth/src/auth/test/redirection-page.test.ts
@@ -1,0 +1,15 @@
+import redirectionScript from '../redirection-page';
+
+const origin = 'https://shopify.com/?x=шеллы';
+const redirectTo = 'shop1.myshopify.io';
+const apiKey = 'fakekey';
+
+describe('redirectionScript', () => {
+  it('returns a script tag with formatted data', () => {
+    const script = redirectionScript({origin, redirectTo, apiKey});
+
+    expect(script).toContain(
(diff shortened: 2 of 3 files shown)


packages/koa-shopify-auth/src/auth/redirection-page.ts
```diff
@@ -12,7 +12,7 @@ export default function redirectionScript({origin, redirectTo, apiKey}) {
           var Redirect = AppBridge.actions.Redirect;
           var app = createApp({
             apiKey: '${apiKey}',
-            shopOrigin: '${origin}',
+            shopOrigin: '${encodeURI(origin)}',
```
ismail-syed, line 15: No idea how to 🎩 this but I think we can unit test this since it just returns a string that gets interpolated in JS.

dreamsInDigital, line 15:
```suggestion
            shopOrigin: "${encodeURI(origin)}",
```
Would suggest wrapping in double quotes here, since `encodeURI()` doesn't encode single quotes. So getting an input like `shop='syntaxerror` could still cause some brokenness.

TheMallen: Unless I'm missing something here I think wrapping in anything other than backticks will break this templating

packages/koa-shopify-auth/src/auth/redirection-page.ts
```diff
@@ -12,7 +12,7 @@ export default function redirectionScript({origin, redirectTo, apiKey}) {
           var Redirect = AppBridge.actions.Redirect;
           var app = createApp({
             apiKey: '${apiKey}',
-            shopOrigin: '${origin}',
+            shopOrigin: "${encodeURI(origin)}",
```
TheMallen, line 15: does this work?  How are we templating here without using backticks?

packages/koa-shopify-auth/src/auth/test/redirection-page.test.ts
```diff
@@ -0,0 +1,15 @@
...
+
+const origin = 'https://shopify.com/?x=шеллы';
+const redirectTo = 'shop1.myshopify.io';
+const apiKey = 'fakekey';
+
+describe('redirectionScript', () => {
+  it('returns a script tag with formatted data', () => {
+    const script = redirectionScript({origin, redirectTo, apiKey});
+
+    expect(script).toContain(
```
TheMallen, line 11: 👍  for test existing

packages/koa-shopify-auth/src/auth/redirection-page.ts
TheMallen: Oh wait I see it now this is in a much larger template. This makes sense.

PR #1615, opened 2020-08-29
[react-form] improving asChoiceField to support multi-value base fields

patsissons:
## Description

This PR is addressing an issue with how ChoiceFields work with `<RadioButton />` components. They were originally designed to function as a simple wrapper to a `Field<boolean>` that would provide simple expansion into a choice component. This neglected the common use case where a single multi-value base field is driving multiple radio buttons in a group. This improved API allows a single multi-value base field to be projected into a choice field with a discrete checked value predicate.

By providing the `checkedValue` with `asChoiceField` we can create unique choice field derivatives that are discriminated by their unique choice value. Additionally, when a choice field is _checked_ it will mutate the base field with that checked value, effectively causing a render sync event on all linked choice components.

### Example

```tsx
function Example() {
  const selectedOption = useField<'A' | 'B' | 'C'>('A');

  return (
    <Stack vertical>
      <RadioButton label="A" {...asChoiceField(selectedOption, 'A')} />
      <RadioButton label="B" {...asChoiceField(selectedOption, 'B')} />
      <RadioButton label="C" {...asChoiceField(selectedOption, 'C')} />
    </Stack>
  );
}
```

<details>
  <summary>
    some sample Polaris code to demonstrate the changes.
  </summary>

```tsx
import React from 'react';
import {
  AppProvider,
  Button,
  Checkbox,
  Form,
  FormLayout,
  RadioButton,
} from '@shopify/polaris';
import translations from '@shopify/polaris/locales/en.json';
import {
  asChoiceField,
  submitSuccess,
  useChoiceField,
  useField,
  useForm,
} from '@shopify/react-form';

export default function Playground() {
  const bool = useChoiceField(false);
  const {
    fields: {multi, simple},
    submit,
  } = useForm({
    fields: {
      multi: useField<'A' | 'B' | 'C'>('A'),
      simple: useField(true),
    },
    async onSubmit(fieldValues) {
      console.log({...fieldValues, bool: bool.checked});
      return submitSuccess();
    },
  });

  return (
    <AppProvider i18n={translations}>
      <div
        style={{display: 'flex', justifyContent: 'center', minHeight: '100vh', alignItems: 'center'}}
      >
        <Form onSubmit={submit}>
          <FormLayout>
            <Checkbox label="Boolean" {...bool} />
            <Checkbox label="Simple" {...asChoiceField(simple)} />
            {['A', 'B', 'C'].map((checkedValue) => (
              <RadioButton
                key={checkedValue}
                label={`Multi: ${checkedValue}`}
                {...asChoiceField(multi, checkedValue)}
              />
            ))}
            <Button primary submit>
              Submit
            </Button>
          </FormLayout>
        </Form>
      </div>
    </AppProvider>
  );
}
```

</details>

## Type of change

## Checklist

### DIFF
--- a/packages/react-form/CHANGELOG.md
+++ b/packages/react-form/CHANGELOG.md
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## [Unreleased] -->
+## [Unreleased]
+
+### Added
+
+- New multi-value base field api for `asChoiceField` utility to support `<RadioButton />` groups [#1615](https://github.com/Shopify/quilt/pull/1615)
 
 ## [0.7.0] - 2020-06-02
 
--- a/packages/react-form/README.md
+++ b/packages/react-form/README.md
@@ -427,7 +427,7 @@ return <TextField label="Title" {...title} />;
 
 #### `useChoiceField()`
 
-An extension to `useField()` that produces a new field compatible with `<Checkbox />` and `<RadioButton />` from `@shopify/polaris`.
+An extension to `useField()` that produces a new field compatible with `<Checkbox />` and `<RadioButton />` from `@shopify/polaris`. Note that this hook cannot be used within a `useForm`, `useChoiceField` can only be used for standalone fields.
 
 ##### Signature
 
@@ -454,6 +454,57 @@ For fields that need to be compatible with choice components on the fly, the `as
 const enabled = useField(false);
 
 return <Checkbox label="Enabled" {...asChoiceField(enabled)} />;
+```
+
+#### `asChoiceField()`
+
+A utility to convert a `Field<Value>` into a derivative that is compatible with `<Checkbox />` and `<RadioButton />` from `@shopify/polaris`. The `value` member will be replaced by a new `checked` member, and the `onChange` will be replaced with a choice component compatible callback.
+
+##### Signature
+
+<!-- The signature is identical to `useField()` for `boolean` fields. -->
+
+`asChoiceField` consumes an existing `Field<Value>`, and optionally a `checkedValue` predicate when dealing with multi-value base fields.
+
+```tsx
+const simpleField = useField(false);
+const simple = asChoiceField(simpleField);
+
+const multiField = useField<'A' | 'B'>('A');
+const multiA = asChoiceField(multiField, 'A');
+const multiB = asChoiceField(multiField, 'B');
+```
+
+##### Examples
+
+`asChoiceField` is used as a helper to expand an existing field into a choice component (`<Checkbox />` and `<RadioButton />`).
+
+```tsx
+const enabled = useChoiceField(false);
+
+return <Checkbox label="Enabled" {...enabled} />;
+```
+
+You can also expand an existing field directly into a choice component by wrapping the field in `asChoiceField` if you want to retain the original field's shape.
+
+```tsx
+const enabled = useField(false);
+
+return <Checkbox label="Enabled" {...asChoiceField(enabled)} />;
+```
+
+For multi-value base fields, we expand the same field into multiple `<RadioButton>` components.
+
+```tsx
+const selectedOption = useField<'A' | 'B'>('A');
+
+return (
+  <Stack vertical>
+    <RadioButton label="A" {...asChoiceField(selectedOption, 'A')} />
+    <RadioButton label="B" {...asChoiceField(selectedOption, 'B')} />
+    <RadioButton label="C" {...asChoiceField(selectedOption, 'C')} />
+  </Stack>
+);
 ```
 
 #### `useList()`
--- a/packages/react-form/src/hooks/field/field.ts
+++ b/packages/react-form/src/hooks/field/field.ts
@@ -198,31 +198,56 @@ export function useField<Value = string>(
   return field as Field<Value>;
 }
 
+export type ChoiceField<Value = boolean> = Omit<
+  Field<Value>,
+  'value' | 'onChange'
+> & {
+  checked: boolean;
+  onChange(checked: boolean): void;
+};
+
 /**
- * Converts a standard `Field<boolean>` into a `ChoiceField` that is compatible
+ * Converts a standard `Field<Value>` into a `ChoiceField` that is compatible
  * with `<Checkbox />` and `<RadioButton />` components in `@shopify/polaris`.
  *
  * For fields that are used by both a choice components and other components, it
- * can be beneficial to retain the original `Field<boolean>` shape and convert
+ * can be beneficial to retain the original `Field<Value>` shape and convert
  * the field on the fly for the choice component.
+ *
+ * For multi-value base fields (not simple boolean fields), you can provide a
+ * checkedValue predicate to project the base field's value into the boolean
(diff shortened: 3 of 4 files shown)


packages/react-form/src/hooks/field/field.ts
```diff
@@ -198,31 +198,54 @@ export function useField<Value = string>(
...
- * Converts a standard `Field<boolean>` into a `ChoiceField` that is compatible
+ * Converts a standard `Field<Value>` into a `ChoiceField` that is compatible
  * with `<Checkbox />` and `<RadioButton />` components in `@shopify/polaris`.
  *
  * For fields that are used by both a choice components and other components, it
- * can be beneficial to retain the original `Field<boolean>` shape and convert
+ * can be beneficial to retain the original `Field<Value>` shape and convert
  * the field on the fly for the choice component.
  *
+ * For multi-value base fields (not simple boolean fields), you can provide an
```
patsissons, line 217:
```suggestion
 * For multi-value base fields (not simple boolean fields), you can provide a
```

packages/react-form/src/hooks/field/field.ts
```diff
@@ -198,31 +198,54 @@ export function useField<Value = string>(
...
-export function asChoiceField({value: checked, ...fieldData}: Field<boolean>) {
+export function asChoiceField<Value>(
+  {value, ...fieldData}: Field<Value>,
+  checkedValue: Value = true as any,
+): ChoiceField<Value> {
   return {
-    checked,
     ...fieldData,
+    checked: value === checkedValue,
+    onChange(checked: boolean) {
```
patsissons, line 238: need a test for this wrapper

packages/react-form/src/hooks/field/test/field.test.tsx
```diff
@@ -762,6 +762,18 @@ describe('asChoiceField', () => {
   it('replaces value with checked', () => {
     expect(asChoiceField({value: true} as any)).toMatchObject({checked: true});
   });
```
michaelmelanson, line 764: Since this is modifying an existing hook, we need to be careful about breakage. It looks like this suite didn't have a test that it works properly when _boolean_ values don't match. 

What would you think about that as a regression test? It's a bit redundant with your "projects an unchecked value from a predicate" test, but all the same it would probably be valuable since any consumers would be using it for boolean values to ensure there's no breakage.

packages/react-form/README.md
```diff
@@ -456,6 +456,57 @@ const enabled = useField(false);
...
+For multi-value base fields, we expand the same field into multiple `<RadioButton>` components.
+
+```tsx
+const selectedOption = useField<'A' | 'B'>('A');
+
+return (
+  <Stack vertical>
+    <RadioButton label="A" {...asChoiceField(selectedOption, 'A')} />
+    <RadioButton label="B" {...asChoiceField(selectedOption, 'B')} />
+    <RadioButton label="C" {...asChoiceField(selectedOption, 'C')} />
```
michaelmelanson, line 505: This API looks great!

packages/react-form/src/hooks/field/field.ts
```diff
@@ -198,31 +198,54 @@ export function useField<Value = string>(
...
+      if (checked) {
+        fieldData.onChange(checkedValue);
+      }
+    },
   };
 }
 
-export type ChoiceField = ReturnType<typeof asChoiceField>;
```
ismail-syed, line 221: Is there a reason why this export is being removed? Would this constitute a breaking change in-case users are consuming this type?

ismail-syed: Oh, nvm. It's in the top of the file.

packages/react-form/README.md
```diff
@@ -456,6 +456,57 @@ const enabled = useField(false);
 return <Checkbox label="Enabled" {...asChoiceField(enabled)} />;
 ```
 
+#### `asChoiceField()`
```
ismail-syed, line 459: Does it make more sense to just create an `asMutliChoiceField()` and re-use w/e parts of `asChoiceField` we need?

packages/react-form/src/hooks/field/field.ts
```diff
@@ -198,31 +198,54 @@ export function useField<Value = string>(
   return field as Field<Value>;
 }
 
+export type ChoiceField<Value = boolean> = Omit<
+  Field<Value>,
+  'value' | 'onChange'
```
ismail-syed, line 203: Would users ever rely on these omitted types existing in `ChoiceField`? Would exporting a `MultiChoiceField` type make more sense?

packages/react-form/CHANGELOG.md
```diff
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## [Unreleased] -->
+## [Unreleased]
+
+### Added
+
+- new multi-value base field api for `asChoiceField` utility to support `<RadioButton />` groups [#1615](https://github.com/Shopify/quilt/pull/1615)
```
416serg, line 12:
```suggestion
- New multi-value base field api for `asChoiceField` utility to support `<RadioButton />` groups [#1615](https://github.com/Shopify/quilt/pull/1615)
```

packages/react-form/src/hooks/field/field.ts
patsissons: it was just replaced with a more expressive version above

packages/react-form/src/hooks/field/field.ts
patsissons: they shouldn't be relying on these, both are non-sensical to use in a choice component context (and `value` has never been visible)

packages/react-form/src/hooks/field/field.ts
```diff
@@ -198,31 +198,54 @@ export function useField<Value = string>(
...
+  {value, ...fieldData}: Field<Value>,
+  checkedValue: Value = true as any,
+): ChoiceField<Value> {
   return {
-    checked,
     ...fieldData,
+    checked: value === checkedValue,
+    onChange(checked: boolean) {
+      if (checked) {
+        fieldData.onChange(checkedValue);
```
patsissons, line 240: I think i may need to also call this for simple (non-multi) choice fields when unchecked. i'm going to add a new test to cover this case.

patsissons: the latest change will now push changes for simple boolean fields, which was the original behaviour. thank you folks for poking me to make sure it was fully backwards compatible.

packages/react-form/README.md
```diff
@@ -427,7 +427,7 @@ return <TextField label="Title" {...title} />;
 
 #### `useChoiceField()`
 
-An extension to `useField()` that produces a new field compatible with `<Checkbox />` and `<RadioButton />` from `@shopify/polaris`.
+An extension to `useField()` that produces a new field compatible with `<Checkbox />` and `<RadioButton />` from `@shopify/polaris`. Note that this hook cannot be used within a `useForm`, `useChoiceField` can only be used for standalone fields.
```
patsissons, line 430: upon topatting it was discovered that `useChoiceField` is no longer compatible with `useForm` compositions, so I am adding this extra note to clarify that case. i.e.,

```
useForm({
  fields: {
    // this field will not be compatible with the FieldBag type for fields above
    flag: useChoiceField(true),
  },
});
```

sanjanadesai27 (approved):
Reviewed / Tophatted over pair

PR #1617, opened 2020-09-01
Remove codecov

keyfer:
## Description

Removes codecov from Quilt as we're no longer using it across Shopify.


keyfer:
Nothing needed there, @atesgoral. AFAIK, those aren't related to **codecov**

PR #1620, opened 2020-09-08
fixing-prefix

### LINKED ISSUE #1590: App using createShopifyAuth from @shopify/koa-shopify-auth cannot be installed when app is using prefix
This line(and probably others): 

https://github.com/Shopify/quilt/blob/3723fcf75941f0d0090a9982d15463d91e19b12a/packages/koa-shopify-auth/src/auth/client/request-storage-access.ts#L7

Can cause infinite redirect when prefix does not  end with a trailing `/`, however in other code default prefix is `""` and forces user to configure prefix without trailing `/`

Basically this makes it impossible to install from path like `/a/b/app` or alike

sirrodgepodge:
## Description

Fixes #1590 

making `prefix` work

## Type of change

- [@shopify/koa-shopify-auth]  Patch: Bug (non-breaking change which fixes an issue)


sirrodgepodge:
I signed CLA, idk how to rerun

Issue #1618, opened 2020-09-03
Migrate koa-shopify-auth from Quilt into a standalone repo


atesgoral:
- [x] Clone package into new repo
- [x] Remove package from Quilt
- [x] Remove docs/references
- [x] Transfer issues and PRs
- [x] Fix README badges in new repo

atesgoral:
There were no PRs with the label `Packages: koa-shopify-auth`

PR #1621, opened 2020-09-08
Remove packages/koa-shopify-auth and update READMEs

atesgoral:
## Description

Part of #1618

This package now lives at https://github.com/Shopify/koa-shopify-auth. Removing the code and updating references.

## Checklist


TheMallen (approved):
🔥  🔥  🔥  ✅

PR #1622, opened 2020-09-08
allow host to be overridden by env

sirrodgepodge:
## Description

Allow overriding of auth callback host via environmental variable `SHOPIFY_CALLBACK_HOST`.

In proxied environments for local development, not having the ability to do this caused issues for me.

## Type of change

- [@shopify/koa-shopify-auth] Patch


mkevinosullivan:

Thank you!

PR #1623, opened 2020-09-09
[react-i18n] Fixes time zone handling for humanized date style

mmargel:
## Description

This PR fixes an issue with the way that `formatDate` handles time zones when using the `Humanize` date style.

Currently, it works as expected for any dates that are not "yesterday", "today", and "tomorrow". When the date is between yesterday and tomorrow, it uses the time zone to generate the time (e.g. "11:00 AM") but it does not use the time zone when generating the yesterday/today/tomorrow label. From the nature of the bug, it looks like it falls back to the user's local time zone.

For example: Given a timestamp of "Sept 9, 2020, 2:00 am, UTC" and a user located in EST, `formatDate` (with UTC as the time zone) will currently humanize the timestamp as "Yesterday at 2:00 am", since the date _in the user's time zone_ is Sept 8th. This PR changes this behaviour so it humanizes the timestamp as "Today at 2:00 am", since the date _in the supplied time zone_ is Sept 9th.

Also, the test changes are a bit... verbose and could probably be cleaned up. I left the existing tests intact because I wanted to ensure that this didn't introduce any regressions. The drawback is that the "today" / "tomorrow" / "yesterday" tests now all have "when a locale is specified" and "when a locale is not specified" counterparts, which feels a bit awkward.

cc @Shopify/web-foundations 

## Type of change

## Checklist


packages/react-i18n/src/i18n.ts
```diff
@@ -469,17 +469,18 @@ export class I18n {
       });
     }
 
+    const timeZone = options ? options.timeZone : undefined;
```
cejaekl, line 472: I'm not sure whether our style guidance allows [optional chaining](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining) in `Shopify/quilt`, and hope that the @Shopify/web-foundations  team will chime in if I need to be corrected... but I'm wondering whether
```
  const timeZone = options?.timeZone;
```
would be appropriate here?

packages/react-i18n/src/test/i18n.test.ts
```diff
@@ -1236,38 +1236,94 @@ describe('I18n', () => {
...
-        it('formats a date from today', () => {
-          const today = new Date('2012-12-20T23:00:00-00:00');
-          const moreThanOneHourAgo = new Date(today.getTime());
-          const hoursAgo = 5;
-          moreThanOneHourAgo.setHours(today.getHours() - hoursAgo);
-          clock.mock(today);
-          const i18n = new I18n(defaultTranslations, {
-            ...defaultDetails,
-            timezone,
+        describe('when the locale time zone is not specified', () => {
```
cejaekl, line 1239:
```suggestion
        describe('when the local time zone is not specified', () => {
```

packages/react-i18n/src/i18n.ts
mmargel: I had tried that originally, but it causes Jest to throw an error:
> Jest encountered an unexpected token

alexandcote: Can you open an issue for that @mmargel? I think we should solve that. Probably we use an old ts-parser for jest.

alexandcote (commented):
`isFutureDate` doesn't check the timezone. 

```ts
export function isFutureDate(date: Date, now = new Date()) {
  return now < date;
}
```

Should we also pass the timezone to this function 🤔 

The reason why I'm asking that is I'm not sure this case will work

```
              12 am ||
    7pm  |                             
                       2h  ||
        9pm |

|| Change day
| Doesn't change day
```

If so, maybe we should add a test for both edge cases like that.

atesgoral:
@alexandcote `now < date` should be a time-zone-agnostic comparison that just compares in the internal representations (unix timestamp). Where are you suggesting we do the TZ manipulation?

packages/react-i18n/src/i18n.ts
atesgoral: 🎉 https://github.com/Shopify/quilt/pull/1643

atesgoral (approved):

packages/react-i18n/src/i18n.ts
atesgoral, line 472:
```suggestion
    const timeZone = options?.timeZone;
```

packages/react-i18n/src/i18n.ts
```diff
@@ -513,17 +514,18 @@ export class I18n {
   }
 
   private humanizeFutureDate(date: Date, options?: Intl.DateTimeFormatOptions) {
+    const timeZone = options ? options.timeZone : undefined;
```
atesgoral, line 517:
```suggestion
    const timeZone = options?.timeZone;
```

Issue #1631, opened 2020-09-10
Currency list is out of sync with `Shopify/shopify`


cejaekl:
## Overview

The list of currencies in `Shopify/quilt` has not been kept updated as we changed the list in `Shopify/shopify`.

The following changes are needed to bring them into sync again:

    New currencies:
      - `BYN` Belarusia New Ruble
      - `BMD` Bermuda Dollar (pegged 1:1 to USD)
      - `DJF` Djibouti Franc
      - `ERN` Eritrea Nakfa
      - `FKP` Falkland Islands Pound (pegged 1:1 to GBP)
      - `GIP` Gibraltar Pound (pegged 1:1 to GBP)
      - `GNF' Guinea Franc
      - `IRR` Iran Rial
      - `KID` Kiribati Dollar (pegged 1:1 to AUD)
      - `LYD` Libya Dinar
      - `MRU` Mauritania New (Second) Ouguiya
      - `PAB` Panama Balboa
      - `SHP` Saint Helena Pound (pegged 1:1 to GBP)
      - `SLL` Sierra Leone Leone
      - `SOS` Somalia Shilling (a fiat currency that is unique because it's no longer backed by a central bank)
      - `TJS` Tajikistan Simoni
      - `TOP` Tonga Pa'anga
      - `VES` Venezuelan Sovereign Bolivar

    Historical currencies that are being removed:
      - `BYR` Belarusia (Old) Ruble (replaced by `BYN`)
      - `VEF` Venezuealan Strong Bolivar (replaced by `VES`)

## Consuming repo

`Shopify/web`

### Area

- [ ] Add any relevant `Area: <area>` labels to this issue

### Scope

- Is this issue related to a specific package?

  - [ ] Tag it with the `Package: <package_name>` label.

---

## Checklist

- [X] Please delete the labels section before submitting your issue
- [X] I have described this issue in a way that is actionable (if possible)

atesgoral:
Assuming this is where all the currencies are defined in shopify: https://github.com/Shopify/shopify/blob/master/db/data/currencies.yml

PR #1632, opened 2020-09-10
[react-i18n] Update currencies

cejaekl:
## Description

Fixes: https://github.com/Shopify/quilt/issues/1631

The list of currencies in `Shopify/quilt` has not been kept in sync with the list in `Shopify/shopify`.  After reviewing the two lists, I've identified a handful of codes that need to be included so that we have the same version of "the truth" in both places.

New currencies:
  - `BYN` Belarusia New Ruble
  - `BMD` Bermuda Dollar (pegged 1:1 to USD)
  - `DJF` Djibouti Franc
  - `ERN` Eritrea Nakfa
  - `FKP` Falkland Islands Pound (pegged 1:1 to GBP)
  - `GIP` Gibraltar Pound (pegged 1:1 to GBP)
  - `GNF' Guinea Franc
  - `IRR` Iran Rial
  - `KID` Kiribati Dollar (pegged 1:1 to AUD)
  - `LYD` Libya Dinar
  - `MRU` Mauritania New (Second) Ouguiya
  - `PAB` Panama Balboa
  - `SHP` Saint Helena Pound (pegged 1:1 to GBP)
  - `SLL` Sierra Leone Leone
  - `SOS` Somalia Shilling (a fiat currency that is unique because it's no longer backed by a central bank)
  - `TJS` Tajikistan Simoni
  - `TOP` Tonga Pa'anga
  - `VES` Venezuelan Sovereign Bolivar

Historical currencies that are being removed:
  - `BYR` Belarusia (Old) Ruble (replaced by `BYN`)
  - `VEF` Venezuealan Strong Bolivar (replaced by `VES`)

## Type of change

- [X] `react-i18n` Patch: Bug (non-breaking change which fixes an issue)

## Checklist


atesgoral (approved):

plentz (commented):
Do we know what's the impact of removing `BYR`, `VES`, and `GIP` from this list? any backfill that we need to run?

alexandcote:
GIP is still in the list, but I have the same question for `BYR` and `VES`. My other question is can we share this list somewhere so both shopify and quilt can use the share values?

atesgoral:
@alexandcote Great idea. Let's discuss different ways that could be achieved. I'm thinking:

a. We spin up a shopify-currencies repo that publishes both a gem and an npm package
b. Both shopify and quilt use an already-existing source of truth (like an ISO currencies package -- but then we need both a gem and an npm package, unless shopify can pull from npm during build)

cejaekl:
> @alexandcote Great idea. Let's discuss different ways that could be achieved. I'm thinking:
> 
> a. We spin up a shopify-currencies repo that publishes both a gem and an npm package
> b. Both shopify and quilt use an already-existing source of truth (like an ISO currencies package -- but then we need both a gem and an npm package, unless shopify can pull from npm during build)

The `shopify-i18n` gem already has a [list of currencies](https://github.com/shopify/shopify-i18n/#--currency-support).
This list is accessible via `country-service` ([PR](https://github.com/Shopify/country-service/pull/459)).

Things still on our [to-do list](https://github.com/Shopify/intl-globalization/issues/222):
  - allow selection of different currency lists (all known, only currencies that are currently in use, only currencies that can be used with Shopify Pay, top "most common" currencies).
  - update country-service to allow passing that sub-list selection through

plentz (approved):

rafaelxy (approved):

PR #1633, opened 2020-09-11
Add useProgressState hook

NickLewanowicz:
## Description
There is a common use case when you want to provide feedback to the user that an action is happening and usually that requires defining some state as false and wrapping your method in a couple setters. This can be simplified with a hook. 
Fixes (issue #)

## Type of change

## Checklist

### DIFF
--- a/packages/react-hooks/src/hooks/progress-state.ts
+++ b/packages/react-hooks/src/hooks/progress-state.ts
@@ -1,0 +1,16 @@
+import {useState} from 'react';
+
+export function useProgressState<T>(
+  event: () => Promise<T>,
+): [boolean, () => Promise<T>] {
+  const [inProgress, setInProgress] = useState(false);
+  async function handleEvent() {
+    try {
+      setInProgress(true);
+      return await event();
+    } finally {
+      setInProgress(false);
+    }
+  }
+  return [inProgress, handleEvent];
+}
--- a/packages/react-hooks/src/hooks/tests/progress-state.test.ts
+++ b/packages/react-hooks/src/hooks/tests/progress-state.test.ts
@@ -1,0 +1,35 @@
+import React from 'react';
+import {mount} from '@shopify/react-testing';
+
+import {useProgressState} from '../progress-state';
+
+describe('useProgressState', () => {
+  function MockComponent({method = (() => new Promise(null)) {
+    const [value, methodHandler] = useProgressState(method);
+
+    const inProgress = value ? 'true' : 'false';
+
+    return (
+      <>
+        <p>Progress: {inProgress}</p>
+        <button type="button" id="method" onClick={methodHandler} />
+      </>
+    );
+  }
+
+  it('sets progress to false', () => {
+    const wrapper = mount(<MockComponent />);
+    expect(wrapper).toContainReactText('Progress: false');
+  });
+
+  it('sets progress to true when method is called', () => {
+    const wrapper = mount(<MockComponent />);
+    expect(wrapper).toContainReactText('Value: false');
+
+    wrapper.find('button', {id: 'method'}).trigger('onClick');
+    expect(wrapper).toContainReactText('Progress: true');
+
+    wrapper.find('button', {id: 'toggle'}).trigger('onClick');
+    expect(wrapper).toContainReactText('Progress: false');
+  });
+});


packages/react-hooks/src/hooks/progress-state.ts
```diff
@@ -0,0 +1,16 @@
+import {useState} from 'react';
+
+export function useProgressState<T>(
```
gmcgibbon, line 3: Is `progress` is the right word here? Typically progress bar APIs use numbers to track changes. Maybe `switch` or `toggle` is more specific to what this is doing?

gmcgibbon (commented):
Needs a changelog and an example in the readme please 🙏  cc @Shopify/web-foundations if anyone has opinions on this.

packages/react-hooks/src/hooks/progress-state.ts
NickLewanowicz: yeah progress can be misleading. `Toggle` or `Switch` might work, how do you feel about `useToggleMethodState`/`useMethodToggleState`? I feel like`useToggleState` could be confused with `useToggle`?

patsissons: `useAsyncActionState`?

michenly:
Can you provided a few example where this pattern is being use currently?

I personally would prefer to use `useState` myself, mostly because I can see all the possibilities of having to break out of this pattern as a project change and using a base functionality unsure the easiest understanding / maintenance in the future.

NickLewanowicz:
> Can you provided a few example where this pattern is being use currently?
> 
> I personally would prefer to use `useState` myself, mostly because I can see all the possibilities of having to break out of this pattern as a project change and using a base functionality unsure the easiest understanding / maintenance in the future.

Folks use many names for their state properties so this isnt exhaustive but grepping for `setLoading` shows some of the uses within web, here are some of them [link 1](https://github.com/Shopify/web/blob/6fafc81911519e682fa9210efa95dd30cc8cacaa/app/sections/Analytics/components/SaveReportModal/SaveReportModal.tsx#L44-L149), [link2](https://github.com/Shopify/web/blob/0512fc73f68f730f6788a256bcd5546b8cd40401/app/sections/Banking/BalanceCardDetails/components/LockCardModal/LockCardModal.tsx#L75-L125), [link3](https://github.com/Shopify/web/blob/77720e7cfd87dc8b6b04479d74db44d501659d75/app/sections/Billing/BillingSettings/BillDetails/BillDetails.tsx#L287-L327), [link 4](https://github.com/Shopify/web/blob/92c00cbfe87cc6ddd6a24b4124a9a519f66e78b7/app/sections/Capital/components/CapitalAcceptance/acceptanceFlow/CapitalAcceptNonSpUs/CapitalAcceptNonSpUsOnboarding.tsx#L277-L297) (and dozens more). 

I agree there is places where you want more control but the general pattern of clicking a button that triggers an async action and wanting to disable that button until it resolves is common enough to warrant a hook like this. In fact some of the examples I link are more complicated then needed because they use `useState` [for example this](https://github.com/Shopify/web/blob/6fafc81911519e682fa9210efa95dd30cc8cacaa/app/sections/Analytics/components/SaveReportModal/SaveReportModal.tsx#L44-L149) calls `setLoading(false);` 4 times due to error handling but if they used `useAsyncActionState` it would have eliminated those 5 lines of code and the chance of a bug if somebody forgot to `setLoading(false)` in one of the conditionals.

packages/react-hooks/src/hooks/progress-state.ts
NickLewanowicz: Thanks for the input at Pat I think `useAsyncActionState` is fairly descriptive, updated the PR!

alexandcote:
Another point is if you have 2 different callbacks that need to toggle the loading, you will have to use this hook twice.
This will create two "loading" state. Every check for the loading state will have to contain those 2 states.

packages/react-hooks/src/hooks/progress-state.ts
```diff
@@ -0,0 +1,16 @@
+import {useState} from 'react';
+
+export function useAsyncActionState<T>(
+  event: () => Promise<T>,
```
atesgoral, line 4: Call this `action`?

NickLewanowicz:
> Another point is if you have 2 different callbacks that need to toggle the loading, you will have to use this hook twice.
> This will create two "loading" state. Every check for the loading state will have to contain those 2 states.

is its a common case where you have different actions that toggle the same loading property? for example if you had an element on a page with a delete action and a save action I would imagine you would want the loading feedback to exist either one each button so you would need `deleteLoading` and `saveLoading` anyways. Maybe Im missing some use cases though.

packages/react-hooks/src/hooks/progress-state.ts
NickLewanowicz: Thanks for this! Ill update it now :)

alexandcote:
>is its a common case where you have different actions that toggle the same loading property? for example if you had an element on a page with a delete action and a save action I would imagine you would want the loading feedback to exist either one each button so you would need deleteLoading and saveLoading anyways. Maybe Im missing some use cases though.

Why you would need `deleteLoading` and `saveLoading`? I think most of the time we disable other buttons when an action is in progress. I think just `loading` is fine here no? You will toggle loading to true or increment the loading count by one if one of those actions (save/delete) happens. I'm not against this hook if a lot of folks think it's useful but I personally will rarely use it.

PR #1634, opened 2020-09-14
[use-import-remote] exporting useImportRemote hook

dleroux:
## Description

This exports the `useImportRemote` hook wasn't exported from the package. 

## Type of change

## Checklist


dleroux:
Not sure if this should be considered a bug or a patch?

packages/react-import-remote/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## [Unreleased] -->
 
+- Exported `useImportRemote` hook and `Status`([1634](https://github.com/Shopify/quilt/pull/1634))
```
dleroux, line 10: Everything has been released not sure if this is the proper way to add an unreleased change?

cartogram (approved):

packages/react-import-remote/CHANGELOG.md
cartogram: looks right 👍

cartogram:
@dleroux I think patch

PR #1635, opened 2020-09-15
Update validators.md

ishanray:
I was confused over how to actually run the validation so adding it here in the docs.

## Description

Fixes (issue #)

## Type of change

## Checklist

### DIFF
--- a/packages/react-form-state/docs/validators.md
+++ b/packages/react-form-state/docs/validators.md
@@ -23,7 +23,7 @@ import {validators} from '@shopify/react-form-state';
 
 We can pass these directly into the `validators` prop of `<FormState />`.
 
-To validate the form automatically on form submit set the `validateOnSubmit` prop of `FormState` to be `true`.
+To use validators when submitting your form, set the `validateOnSubmit` prop of `<FormState />` to `true`. Here's an example:
 
 ```typescript
 import {TextField} from '@shopify/polaris';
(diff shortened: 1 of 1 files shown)


packages/react-form-state/docs/validators.md
```diff
@@ -23,6 +23,8 @@ import {validators} from '@shopify/react-form-state';
 
 We can pass these directly into the `validators` prop of `<FormState />`.
 
+To validate the form automatically on form submit set the `validateOnSubmit` prop of `FormState` to be `true`.
```
gmcgibbon, line 26:
```suggestion
To use validators when submitting your form, set the `validateOnSubmit` prop of `<FormState />` to `true`. Here's an example:
```

I would also remove the line break to make a paragraph out of the sentence above, and then edit the example below to use this property.

packages/react-form-state/docs/validators.md
```diff
@@ -35,6 +36,7 @@ function MyComponent() {
         description: 'Cool product',
         quantity: 0,
       }}
+      validateOnSubmit
```
gmcgibbon, line 39:
```suggestion
      validateOnSubmit={true}
```
or is just specifying the property magic syntax for the same thing?

ishanray: Yup it is for the same thing, just some syntactical sugar.

gmcgibbon (approved):
:shipit:

PR #1637, opened 2020-09-16
Export referenceTime from performance package

MLDimitry:
## Description

Yesterday I encountered an issue where Safari didn't support the performance.timeOrigin Web API. After posting on Slack I was pointed to https://github.com/Shopify/quilt/blob/d3720fd2380feac38c575fc42edf8271992e885b/packages/performance/src/utilities.ts#L8 which is a workaround used in `@shopify/performance`.

I've since copied this solution into my own PR but I think it would make sense to have this exported from this package.

My related PR where we're using this https://github.com/Shopify/online-store-web/pull/8080#discussion_r489412172

We're using this in an embedded Shopify admin app to get the `timeOrigin` of when the Shopify/web browser context begins to measure the full loading time of our new React app.

## Type of change

Minor/non-breaking

## Checklist

### DIFF
--- a/packages/performance/CHANGELOG.md
+++ b/packages/performance/CHANGELOG.md
@@ -7,6 +7,12 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## [Unreleased] -->
 
+## [1.2.3] - 2020-9-16
+
+### Changed
+
+- Change `@shopify/performance` package to export the `referenceTime` util
+
 ## [1.2.2] - 2019-10-16
 
 ### Fixed
--- a/packages/performance/CHANGELOG.md
+++ b/packages/performance/CHANGELOG.md
@@ -7,11 +7,9 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## [Unreleased] -->
 
-## [1.2.3] - 2020-9-16
+### Added
 
-### Changed
-
-- Change `@shopify/performance` package to export the `referenceTime` util
+- Added `timeOrigin` util as an export of `@shopify/performance` so it can be reused outside this package
 
 ## [1.2.2] - 2019-10-16
 
--- a/packages/performance/src/performance.ts
+++ b/packages/performance/src/performance.ts
@@ -6,7 +6,7 @@ import {
   withNavigation,
   withTiming,
   supportsPerformanceObserver,
-  referenceTime,
+  timeOrigin as referenceTime,
   hasGlobal,
 } from './utilities';
 import {
--- a/packages/performance/README.md
+++ b/packages/performance/README.md
@@ -78,6 +78,8 @@ performance.event({
 performance.finish();
 ```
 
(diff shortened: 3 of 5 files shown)


utkarshsaxena93 (approved):

packages/performance/CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## [Unreleased] -->
 
+## [1.2.3] - 2020-9-16
+
```
atesgoral, line 11:
```suggestion
```

We'll add the version + date when we're cutting a new tag.

packages/performance/CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
...
+### Changed
```
atesgoral, line 12:
```suggestion
### Added
```

Exposing a method that already existed could be counted as adding of new functionality.

atesgoral:
If we're exposing this method:

1. Need to decide on the name. Instead of the internal name `referenceTime`, we could call it `timeOrigin`, since we're wrapping that.
2. Need to document the new method in the new README, somewhere, by also mentioning a motivation for using it. (See the mention of `now()` in the README.

MLDimitry:
All comments have been addressed now, let me know if there's anything I can further improve on.

packages/performance/src/performance.ts
```diff
@@ -6,7 +6,7 @@ import {
   withNavigation,
   withTiming,
   supportsPerformanceObserver,
-  referenceTime,
+  timeOrigin as referenceTime,
```
keyfer, line 9: Because I don't think we ever exported `referenceTime` from `@shopify/performance`, you can probably just export `timeOrigin` here.

keyfer (approved):
This is great, thanks for this! Just one small naming change.

packages/performance/src/performance.ts
MLDimitry: This is an import, not the export of the util. The `timeOrigin` util was previously named `referenceTime` but we're renaming it in this PR. The variable in this file that uses `referenceTime` is `const timeOrigin = referenceTime()`. 

So I used the `as referenceTime` renaming to preserve the other naming in this file, but maybe we could improve it? Would something like `crossBrowserTimeOrigin` or `browserSafeTimeOrigin` be better a better name alias? Or maybe we'd prefer to rename the `timeOrigin` variable?

MLDimitry:
@atesgoral and I had a bit further of a chat about this on Slack.

Our use case is very unique to comparing `Shopify/web` performance timings against `Shopify/shopify`. Since our app is only going to be using this for the duration of our ThemeEditor rollout, until our old Rails code is removed, we felt it doesn't make sense to expose such a fringe feature of the package that may not get reused much. 

Instead we will just copy and paste the few lines of code we need into online-store-web.

PR #1641, opened 2020-09-27
Update [koa-shopify-webhooks] ApiVersion

frankleng:
# Description
add support for `July20` 

per https://github.com/Shopify/quilt/blob/e0d52556170e26a8bd8a1d5f42c5c4247fe72482/packages/koa-shopify-graphql-proxy/src/shopify-graphql-proxy.ts

## Checklist

- [ x ] I have added a changelog entry, prefixed by the type of change noted above (Documentation fix and Test update does not need a changelog as we do not publish new version)

### DIFF
--- a/packages/koa-shopify-webhooks/CHANGELOG.md
+++ b/packages/koa-shopify-webhooks/CHANGELOG.md
@@ -9,7 +9,7 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [2.6.0] - 2020-09-27
 
-- The `ApiVersion` enum now has an `July20` option
+- The `ApiVersion` enum now has an `July20` option.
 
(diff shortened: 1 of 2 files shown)


packages/koa-shopify-webhooks/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## [Unreleased] -->
 
+## [2.6.0] - 2020-09-27
+
+- The `ApiVersion` enum now has an `July20` option.
```
atesgoral, line 12:
```suggestion

### Added

- The `ApiVersion` enum now has an `July20` option.
```

atesgoral (approved):

frankleng:
👍

PR #1645, opened 2020-10-06
Improve types and exports of react-router

BPScott:
##  Description

When attempting to migrate a codebase from using react-router to using @shopify/react-router I found two shortcomings:

- The Props on `Link` did not inherit from the props on react-router's `Link` even though we passed everything through. This meant that `<Link className="foo">` did what was expected at runtime but typescript claimed className was not a valid prop.
- I used MemoryRouter for some of my tests, and I want it to be exposed via @shopify/react-router so that I don't have to reference a transitive dependency.

This PR does the following:

- Update Link's Props so can handle passing in props defined on
  react-router-dom's link props
- Tweak test so typescript will check validity of Link props
- Reexport MemoryRouter for usage in tests

## Type of change

- [x] @shopify/react-router Patch: Bug (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/packages/react-router/src/components/Link/tests/Link.test.tsx
+++ b/packages/react-router/src/components/Link/tests/Link.test.tsx
@@ -43,10 +43,9 @@ describe('<Link />', () => {
 
     it('sets the `to` prop and passes along other props to ReactRouterLink', () => {
       const url = '/home';
-      const props = {className: 'foo'};
-      const link = mount(<Link url={url} {...props} />);
+      const link = mount(<Link url={url} className="foo" />);
       expect(link).toContainReactComponent(ReactRouterLink, {
-        ...props,
+        className: 'foo',
         to: url,
       });
     });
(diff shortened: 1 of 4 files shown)


packages/react-router/src/components/Link/tests/Link.test.tsx
```diff
@@ -43,10 +43,9 @@ describe('<Link />', () => {
 
     it('sets the `to` prop and passes along other props to ReactRouterLink', () => {
       const url = '/home';
-      const props = {className: 'foo'};
```
BPScott, line 46: This refactoring surfaces the problem where it complains "className is not a valid prop" if you don't make the change to Link.tsx

atesgoral (approved):

michenly (approved):
Looks great!

vsumner (approved):

PR #1648, opened 2020-10-07
[@shopify/react-html] Let consumer add inline style in the head

alexandcote:
## Description
You can now add inline style in the head of the document with `@shopify/react-html`.

The `<Style />` and `<Script />` component should never be used on the client so we move the component in the `server` folder. You can now access them via `@shopify/react-html/server`. `<Style>` has been renamed for `<Stylesheet>`.

#### This will now render a `<style />` element.
```
<InlineStyle>.foo {color: red;}</InlineStyle>
```

You can also use the  `useInlineStyle` hook.

## Type of change
- [x] `@shopify/react-html` Major: Breaking change in the import of `Script` and `Style`

## Checklist


packages/react-html/src/components/tests/Style.test.tsx
```diff
@@ -4,17 +4,32 @@ import {mount} from '@shopify/react-testing';
...
+        integrity: '00000000',
+        crossOrigin: 'anonymous',
+      });
+    });
+  });
+
+  describe('<style />', () => {
+    it('renders attributes', () => {
+      const children = 'hello world';
+      const style = mount(<Style id="foo">{children}</Style>);
```
michenly, line 26: I would prefer for the test to reflect how something should be use.
maybe use the inline style you had mentioned in the description?

alexandcote:
@michenly you can stop reviewing this PR. @clauderic found a mistake in my code. I want the <style> tag to be in the head so I will need to change how we do this.

packages/react-html/src/components/HtmlUpdater.tsx
```diff
@@ -43,59 +43,46 @@ function updateOnClient(state: State) {
...
-    } else {
-      fragment.appendChild(element);
-    }
-  }
+  document.head.appendChild(fragment);
+}
 
-  const oldLinks = Array.from(
-    document.head.querySelectorAll(`link[${MANAGED_ATTRIBUTE}]`),
+function updateElement<T extends HTMLProps<HTMLElement>>(
```
alexandcote, line 53: I found a lot of duplication here. I will highlight the only part I changed.

packages/react-html/src/components/HtmlUpdater.tsx
```diff
@@ -43,59 +43,46 @@ function updateOnClient(state: State) {
...
-  for (const link of links) {
-    const element = document.createElement('link');
+  for (const item of items) {
+    const element = document.createElement(selector);
     element.setAttribute(MANAGED_ATTRIBUTE, 'true');
 
-    for (const [attribute, value] of Object.entries(link)) {
-      element.setAttribute(attribute, value);
+    for (const [attribute, value] of Object.entries(item)) {
+      if (attribute === 'children') {
```
alexandcote, line 67: This is the only part I changed. Instead of calling `setAttribute` when you pass children, we set the value to the `innerHTML`.

Any XSS issue with that 🤔 ?

packages/react-html/src/components/Stylesheet.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+interface StyleProps extends React.HTMLProps<HTMLLinkElement> {
+  href?: never;
+  children: string;
+}
+
+type Props = LinkProps | StyleProps;
+
+export function Stylesheet(props: Props) {
+  const hook = isLink(props) ? useLink : useStyle;
```
alexandcote, line 18: I guess I'm calling conditional hook here. Any idea how I can solve this?

alexandcote (commented):
I 🎩 the code in one of our app and everything work 💯

packages/react-html/src/components/HtmlUpdater.tsx
alexandcote: Decided to use `innerText` instead.

alexandcote:
After some discussion with @lemonmade, I changed the approach a little bit. The PR is now ready to be reviewed.

packages/react-html/CHANGELOG.md
```diff
@@ -5,7 +5,17 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## [Unreleased] -->
+## [Unreleased]
+
+### Changed
+
+- The `<Script />` and `<Style />` components are now exported from `@shopify/react-html/server` ([#1648](https://github.com/Shopify/quilt/pull/1648))
```
michenly, line 12: I understand why you would want to do this (to be more consistent with Link I presume) but this is actually going to be a breaking change since the<link/> that is meant to be inline will now to to the head and the css specificity will change.

alexandcote: Yep, it's a breaking change but for the good of the library. Consumers should not use `Style` or `Script` directly.

michenly: my worry is that this will be too easily missed when consumer upgrade to the next version of react-html.

I think we should at least add a carefully documented migration guide and also mentioned in the README how the exported version of <Style /> component had changed.

alexandcote: I will add the migration doc for sure. The API of the component change so Typescript will complain also.

alexandcote: Here is the new migration steps: https://github.com/Shopify/quilt/pull/1648/files#diff-2c1ecb54d6ceef3255296a43f4ddf9c8

sanjanadesai27 (approved):
Everything looks logical to me - haven't top hatted myself though. I'm also not the most familiar with this package. 

cc. @michenly - do you have any follow up feedback?

packages/react-html/CHANGELOG.md
clauderic: Agreed, this change may be really easy for consumers to miss. Hopefully the migration steps are clear enough. I understand the distinction though, makes sense to me.

packages/react-html/src/server/components/Style.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+export interface Props extends React.LinkHTMLAttributes<HTMLLinkElement> {
+  href: string;
+}
+
+export function Style(props: Props) {
```
clauderic, line 7: While we're making breaking changes, could this be renamed to `StyleSheet`? Would avoid the issue of having two components with the same name

packages/react-html/src/components/Style.tsx
```diff
@@ -1,9 +1,10 @@
 import React from 'react';
 
-export interface Props extends React.LinkHTMLAttributes<HTMLLinkElement> {
-  href: string;
-}
+import {useStyle} from '../hooks';
+
+type Props = React.HTMLProps<HTMLStyleElement>;
 
 export function Style(props: Props) {
```
clauderic, line 7: Should this also be called `InlineStyle`? All of the other references in the manager refer to `inlineStyles`

```suggestion
export function InlineStyle(props: Props) {
```

This would probably help with the migration as if a consumer tries to upgrade they will immediately see that the `Style` component no longer exists

clauderic (approved):

packages/react-html/src/components/Style.tsx
alexandcote: I can if we prefer but like I said, typescript will also complain. I like the component and hook to match the native HTML element.

packages/react-html/CHANGELOG.md
```diff
@@ -5,7 +5,18 @@ All notable changes to this project will be documented in this file.
...
+## [Unreleased]
+
+### Changed
+
+- The `<Script />` components is now exported from `@shopify/react-html/server` ([#1648](https://github.com/Shopify/quilt/pull/1648))
+- The `<Style>` has been renamed for `<Stylesheet>` and is now exported from `@shopify/react-html/server` ([#1648](https://github.com/Shopify/quilt/pull/1648))
+
+### Added
+
+- New `useInlineStyle` hook add `<style>` element in the head of the document. ([#1648](https://github.com/Shopify/quilt/pull/1648))
```
clauderic, line 17:
```suggestion
- New `useInlineStyle` hook to add inline `<style>` elements in the head of the document. ([#1648](https://github.com/Shopify/quilt/pull/1648))
```

packages/react-html/CHANGELOG.md
```diff
@@ -5,7 +5,18 @@ All notable changes to this project will be documented in this file.
...
+- New `<InlineStyle />` component so you can add a `<style>` element in the head of the document. ([#1648](https://github.com/Shopify/quilt/pull/1648))
```
clauderic, line 18:
```suggestion
- New `<InlineStyle />` component so you can add inline `<style>` elements in the head of the document. ([#1648](https://github.com/Shopify/quilt/pull/1648))
```

packages/react-html/CHANGELOG.md
```diff
@@ -5,7 +5,18 @@ All notable changes to this project will be documented in this file.
...
+- New `inlineStyles` prop in the `<Html />` component to render `<style>` elements in the head. ([#1648](https://github.com/Shopify/quilt/pull/1648))
```
clauderic, line 19:
```suggestion
- Introduced new `inlineStyles` prop on the `<Html />` component to render inline`<style>` elements in the head of the document. ([#1648](https://github.com/Shopify/quilt/pull/1648))
```

packages/react-html/README.md
```diff
@@ -166,10 +166,11 @@ The `<Html>` component serves as a top level wrapper for a React application, al
...
 - `children`: the application. It will be rendered to a string and placed inside a `div` with an ID of `app`.
 - `locale`: the language to use for the HTML `lang` attribute.
-- `styles`: descriptors for any style tags you want to include in the HEAD of the document.
+- `styles`: descriptors for any stylesheet link tags you want to include in the HEAD of the document.
+- `inlineStyles`: descriptors for any style tags you want to include in the HEAD of the document.
 - `scripts`: descriptors for any script tags you want to include in your document. All scripts passed to this property will be deferred by appending them to the end of the document. We encourage this as a default because it improves the initial rendering performance of your page.
 - `blockingScripts`: descriptors for any script tags you want to include in the HEAD of the document. These will block HTML parsing until they are evaluated, so use them carefully.
-- `headMarkup`: additional JSX to be embedded in the head of the document (after styles, but before blocking scripts).
+- `headMarkup`: additional JSX to be embedded in the head of the document (after inlineStyles, but before blocking scripts).
```
clauderic, line 173: I think this should be:

```suggestion
- `headMarkup`: additional JSX to be embedded in the head of the document (after stylesheets and inline styles, but before blocking scripts).
```

packages/react-html/README.md
```diff
@@ -186,26 +187,37 @@ const html = (
...
+<Stylesheet
   href="./some-style.css"
   integrity="some-integrity-hash"
   crossOrigin="anonymous"
 />;
 ```
 
+### `<InlineStyle />`
+
+The `<InlineStyle />` component lets you render `<style>` tags in your document dynamically as part of your react app. It supports all of the props of a basic `style` tag, but forces some properties to be the values needed for a stylesheet. In general, prefer the `inlineStyles` prop of the `Html` component instead of using this component explicitly.
```
clauderic, line 206:
```suggestion
The `<InlineStyle />` component lets you render `<style>` tags in your document dynamically as part of your react app. It supports all of the props of a basic `style` tag, but forces some properties to be the values needed for an inline style. In general, prefer the `inlineStyles` prop of the `Html` component instead of using this component explicitly.
```

packages/react-html/README.md
```diff
@@ -216,14 +228,20 @@ import {Script} from '@shopify/react-html';
...
 
 ### `useLink()` and `<Link />`
 
 Renders a `<link />` tag in the head with the specified attributes. On the server, links are recorded in the `Manager` and automatically applied to the `Html` component. On the client, the `<link />` tags are updated in a deferred callback to minimize DOM manipulations.
 
 Both the hook and component versions accept any properties you would supply to a `<link />` tag. If you are using this component to create a favicon, use the [`useFavicon()`/ `<Favicon />` component](#usefavicon-and-favicon-) instead.
 
+### `useInlineStyle()` and `<InlineStyle />`
+
+Renders a `<style />` tag in the head with the specified attributes. On the server, links are recorded in the `Manager` and automatically applied to the `Html` component. On the client, the `<style />` tags are updated in a deferred callback to minimize DOM manipulations.
```
clauderic, line 241:
```suggestion
Renders an inline `<style />` tag in the head of the document with the specified attributes. On the server, links are recorded in the `Manager` and automatically applied to the `Html` component. On the client, the `<style />` tags are updated in a deferred callback to minimize DOM manipulations.
```

packages/react-html/README.md
```diff
@@ -216,14 +228,20 @@ import {Script} from '@shopify/react-html';
...
+
+Both the hook and component versions accept any properties you would supply to a `<style />` tag.
```
clauderic, line 243:
```suggestion
Both the hook and component versions accept any properties you would supply to an inline `<style />` tag.
```

packages/react-html/documentation/migration-version-9-to-10.md
```diff
@@ -0,0 +1,21 @@
...
++import {Stylesheet} from '@shopify/react-html/server';
+```
+
+## `Script`
+
+This component has been moved to `@shopify/react-html/server`.
+
+```diff
+-import {Script} from '@shopify/react-html';
++import {Stylesheet} from '@shopify/react-html/server';
```
clauderic, line 20:
```suggestion
+import {Script} from '@shopify/react-html/server';
```

PR #1649, opened 2020-10-07
Add helper function to wrap content with language disclaimers

gregoriokusowski:
## Description

In some scenarios we show users links with text like `Learn more`, and those links are managed by third-party and often don't provide content in the users language. For those cases, we would like to show a disclaimer to inform users that the content is not on their language, like `Learn more (in German)` or similar. The example in `utilities.test.tsx` shows this example.

I'm first adding this PR as a spike/draft to discuss if/how we should proceed with this approach.

## Type of change

- [ ] react-i18n Minor: New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/packages/react-i18n/src/test/utilities.test.tsx
+++ b/packages/react-i18n/src/test/utilities.test.tsx
@@ -7,6 +7,7 @@ import {
   getCurrencySymbol,
   memoizedNumberFormatter,
   memoizedPluralRules,
+  wrapWithDisclaimer,
 } from '../utilities';
 
 const {pseudotranslate} = require.requireMock('@shopify/i18n') as {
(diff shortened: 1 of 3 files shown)


packages/react-i18n/src/test/utilities.test.tsx
```diff
@@ -398,4 +399,18 @@ describe('translate()', () => {
...
+  describe('wrapWithDisclaimer', () => {
+    const translations = {
+      disclaimer: {language: {en: '{content}', de: '{content} (in German)'}},
+    };
+    it('wraps the content in the translation', () => {
+      expect(wrapWithDisclaimer('Learn more', 'en', translations, locale)).toBe(
+        'Learn more',
+      );
+      expect(wrapWithDisclaimer('Learn more', 'de', translations, locale)).toBe(
+        'Learn more (in German)',
```
dahukish, line 412: If we can access the translation to add `In German` could we not update the string itself?

PR #1650, opened 2020-10-08
[statsd] Update hot-shots dependencies

alexandcote:
## Description
- Bumping the `hot-shots` dependency on the last version  from [`5.9.1`](https://github.com/brightcove/hot-shots/blob/master/CHANGES.md#591-2018-9-18) to [`8.2.0`](https://github.com/brightcove/hot-shots/blob/master/CHANGES.md#820-2020-9-30)
- Bumping the `change-case` dependency.
- Adding a new `timing` function.

## Type of change
- [x] `@shopify/statsd` Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist


atesgoral:
What is breaking with this change? If there's a migration path, could you document it in a migration guide?

alexandcote:
The only breaking change is required Node V8 and more.

atesgoral (approved):

PR #1651, opened 2020-10-09
Don't use importHelpers in react-effect

9at8:
## Description

Using import helpers requires us to depend on the `tslib` package. Since react-effect is transpiled to cjs, it's not tree shakable. The `tslib` package is about 2 kb, which is roughly twice the size of the actual library. This PR removes that dependency to half the package size.

|Before|After|
|---|---|
|~ 3.6 kb |~ 1.8 kb |
|Screen Shot 2020-10-08 at 20 44 54|Screen Shot 2020-10-08 at 20 45 10|

The helpers that were imported from the tslib package:

|What|Where|
|---|---|
|`__importDefault`|`./src/context`|
|`__awaiter`|`./src/manager`|
|`__generator`|`./src/manager`|
|`__read`|`./src/manager`|
|`__spread`|`./src/manager`|
|`__awaiter`|`./src/server`|
|`__generator`|`./src/server`|
|`__importDefault`|`./src/server`|

## Type of change

- [x] `@shopify/react-effect` Patch: Bug (non-breaking change which fixes an issue)

## Checklist


9at8:
What about the previous changelog entries? For 3.2.2 -> 3.2.12?

packages/react-effect/package.json
```diff
@@ -23,9 +23,6 @@
     "url": "https://github.com/Shopify/quilt/issues"
   },
   "homepage": "https://github.com/Shopify/quilt/blob/master/packages/react-effect/README.md",
-  "dependencies": {
-    "tslib": "^1.9.3"
```
ismail-syed, line 27: Where were we using this before? Was it just accidentally there?

ismail-syed: I'm a bit confused. The `tslib` [docs](https://github.com/Microsoft/tslib) mention 

> this means delivering users smaller files on average, as well as less runtime overhead.

Shouldn't having this be more beneficial? 🤔

BPScott:
In #1657 @alexandcote bumped tslib to be > 1.14 - which [added esmodule support](https://github.com/microsoft/tslib/releases/tag/1.14.0).

Update your app to the latest version of tslib and this should be treeshaken away.

---

Even without this new esmodules support I don't think this PR is the right approach - sure this size difference is not an ideal look for react-effect on its own but the wins of using tslib comes when several modules all use it together. With this PR as soon as you include any other package from this library (and shopify apps depends upon several) you depend upon the helpers in tslib AND the helpers inlined into the react-effect package.  And because of this duplication you end up with a net bundle size increase.

9at8:
@BPScott This makes sense. I planned on closing this earlier. Sorry for the delay!

PR #1653, opened 2020-10-14
Load async webpack assets after sync ones

### LINKED ISSUE #1652: Revise sewing-kit-koa production asset loading order
## Overview

The current asset load ordering in sewing-kit koa feels back to front. It load async assets before synchronous entrypoint assets. I would expect it to load assets the other way around - load synchronous assets from the entrypoint then load any async assets.

I don't believe this will have any affect on JS assets - as webpack marshals the execution order of its bundles so load order doesn't really matter. 

For CSS assets however the load order impacts the what selector is applied if two of equal precedence are discovered. Most of the time this isn't an issue thank to us leaning heavily on CSS modules but it can crop up. This gets particularly annoying as this asset order in production  does not seem to match what occurs in development, where synchronous CSS is included before async CSS. That gets us into the unfortunate situation where production and development output differ.

Consider an app with a main entrypoint, that imports some css to generate a `main.css` and a react-async component route (which in turn imports some CSS) that generates a `myRoute.js` and `myRoute.css`.

### Current loading order:

- async CSS - myRoute.css
- sync CSS - main.js
- async JS - myRoute.js (and it's various vendor files before it)
- sync JS - main.js (and it's various vendor files before it)

## Expected behaviour 

- sync CSS - main.js
- async CSS - myRoute.css
- sync JS - main.js (and it's various vendor files before it)
- async JS - myRoute.js (and it's various vendor files before it)

----

For Slack talk: https://shopify.slack.com/archives/CCNRSKLTH/p1602621868239300

@lemonmade :
>  I assume I know where you are going with this, but IMO the existing (A) behavior is correct because the main files are typically in charge of holding all the shared dependencies of the routes, which means it needs to go first for JS (load in the JS dependencies like React the route JS needs) and CSS (load in the resets and custom properties the route CSS needs)
>
> err
>
> now I think I might have the current thing wrong :P
>
> JS it gets kind of weird, CSS I think I would expect it to be main then route which I think is the opposite of now

@BPScott:
> you're right that  A is the current behaviour.

@lemonmade: 
> yeah for CSS I think that's definitely incorrect

BPScott:
## Description

This mimics how assets are loaded in development. This shouldn't have
any effect for JS files which are marshaled by the webpack runtime, but
CSS content will now be loaded in a similar order to development.

Fixes #1652

I think this does the job, but I've not currently got the time to thoroughly investigate the ordering in complex applications. It'd be amazing if somebody from web foundations could finish this investigation and take this over the line.

I'm not quite sure of the effect of the `nonVendorEntrypointIndex` stuff and I think it might be buggy anyway as in the event of no result `nonVendorEntrypointIndex` will be set to -1 and you'd end up going into that if clause which doesn't seem right.

## Type of change

- [] @shopify/sewing-kit-koa Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

### DIFF
--- a/packages/sewing-kit-koa/src/assets.ts
+++ b/packages/sewing-kit-koa/src/assets.ts
@@ -167,18 +167,7 @@ function getAssetsFromManifest(
     return entrypointAssets;
   }
 
-  const bundleTester = new RegExp(`\\b${name}[^\\.]*\\.${kind}`);
-
-  const nonVendorEntrypointIndex = entrypointAssets.findIndex(bundle =>
-    bundleTester.test(bundle.path),
-  );
-
-  if (nonVendorEntrypointIndex) {
-    entrypointAssets.splice(nonVendorEntrypointIndex, 0, ...asyncAssets);
-    return entrypointAssets;
-  }
-
-  return [...asyncAssets, ...entrypointAssets];
+  return [...entrypointAssets, ...asyncAssets];
 }
 
 function getAsyncAssetsFromManifest(
--- a/packages/sewing-kit-koa/src/tests/assets.test.ts
+++ b/packages/sewing-kit-koa/src/tests/assets.test.ts
@@ -78,7 +78,7 @@ describe('Assets', () => {
       ]);
     });
 
-    it('prefixes async assets matching the passed IDs', async () => {
+    it('appends async assets matching the passed IDs', async () => {
       const js = '/custom.js';
       const asyncJs = '/used.js';
 
@@ -100,7 +100,7 @@ describe('Assets', () => {
 
       expect(
         await assets.scripts({name: 'custom', asyncAssets: ['used']}),
-      ).toStrictEqual([{path: asyncJs}, {path: js}]);
+      ).toStrictEqual([{path: js}, {path: asyncJs}]);
     });
 
     it('throws an error when no scripts exist for the passed entrypoint', async () => {
@@ -185,7 +185,7 @@ describe('Assets', () => {
       ]);
     });
 
-    it('prefixes async assets matching the passed IDs', async () => {
+    it('appends async assets matching the passed IDs', async () => {
       const css = '/custom.css';
       const asyncCss = '/used.css';
 
@@ -207,7 +207,7 @@ describe('Assets', () => {
 
       expect(
         await assets.styles({name: 'custom', asyncAssets: ['used']}),
-      ).toStrictEqual([{path: asyncCss}, {path: css}]);
+      ).toStrictEqual([{path: css}, {path: asyncCss}]);
     });
 
     it('throws an error when no styles exist for the passed entrypoint', async () => {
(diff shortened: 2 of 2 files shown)


packages/sewing-kit-koa/src/assets.ts
```diff
@@ -167,18 +167,7 @@ function getAssetsFromManifest(
     return entrypointAssets;
   }
 
-  const bundleTester = new RegExp(`\\b${name}[^\\.]*\\.${kind}`);
```
dahukish, line 170: What's the context around this removal? @BPScott to allow any resource?

BPScott: I'm not really sure what this code was supposed to do (@lemonmade may have better insight)

It looks like it tries to say "inject the async assets before the first non vendor entrypoint" - so it goes VendorEntrypointAssets, AsnycAssets,  OtherEntryPointAssets" but there's a bug in it because findIndex returns -1 when not found so I think actually happens is

- When no vendor items found (findIndex returns -1) it injects the async assets at the last-but-one position in the EntryPoint Array (so SomeOtherEntrypointAssets, AsyncAssets, TheFinalEntrypointAsset) 
- When a vendor item is found as the first entrypointAsset (findIndex returns 0) it returns the async assets and then the entrypoint assets (so AsyncAssets, VendorEntryPointAssets, OtherEntryPointAssets)
- When greater than one vendor item is found it injects the Asnync assets before the first non vendor entrypoint asset (so VendorEntrypointAssets AsyncAssets, OtherEntrypointAssets).

With this PR we always put the async assets after the entrypoint ones (regardless of if those are vendor entrypoints or not) rather than trying to inject them in the middle so there's no need to do any of that slicing or dicing.

- Add the

lemonmade: My assumption is that this would not work for JS, but I unfortunately don't have time to verify. My understanding was that webpack needed all of the async scripts to be loaded before the entrypoint that consumes them in order for those async assets to be resolvable synchronously from the entrypoint. The ability to resolve them synchronously is key since that is typically necessary to render the same screen as what was rendered during SSR.

lemonmade: (Agreed that there is a bug here, though probably one that is effectively never hit given that vendor bundles are created fairly consistently; also agreed that this logic is correct for styles)

BPScott:
https://github.com/Shopify/shopify/pull/268492 suggests that @lemonmade's thoughts are well founded - async JS after sync JS does work (As that's what shopify/shopify did for ages) but results in a bit of visual jank that it sounds like Checkout One (and thus probably the rest of the company) want to avoid:

- Load sync JS
- Partial page render containing just content from sync JS. So probably page chrome but nothing route-specific.
- Load async JS
- Page rerenders now that it can contain all the async stuff.

By loading async JS before sync JS it avoids that rerender

- Load async JS
- Load sync JS
- CompletePage can render both sync and async route specific stuff in one pass

---


Sounds like we want to leave the JS ordering as is, but adjust the CSS load order.

BPScott:
Going to close this. It's old and stale, and web doesn't use sewing-kit-koa anymore

PR #1654, opened 2020-10-14
Add latest API Version (October 2020)

gil--:
## Description

Adds the `October20` ApiVersion as an option.

## Type of change

- [X]  Minor: New feature (non-breaking change which adds functionality)

## Checklist


gil--:
Do we leave in old supported API versions? Not sure how API Version deprecation works with this library.

dahukish:
> Do we leave in old supported API versions? Not sure how API Version deprecation works with this library.

Going to approve this for now, we can circle back and prune old versions. Could look into a github action that compares latest and triggers a message possibly?

dahukish (approved):
LGTM

packages/koa-shopify-graphql-proxy/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## [Unreleased] -->
 
+
+## [4.1.0] - 2020-10-14
+
```
atesgoral, line 12: We should add versions and dates while publishing

```suggestion
```

atesgoral (approved):

PR #1655, opened 2020-10-16
Export StringMapper interface

helloneele:
## Description

While adding project references in Web, I ran into this error

```
app/sections/Products/utilities/validators.ts:25:17 - error TS4058: Return type of exported function has or is using name 'StringMapper' from external module "/app/node_modules/@shopify/react-form-state/dist/validators" but cannot be named.
```

## Type of change

- [x] `@shopify/react-form-state` Patch: Bug (non-breaking change which fixes an issue)

## Checklist


dahukish (approved):
:shipit:

atesgoral (approved):

PR #1658, opened 2020-10-21
Add Origin header to Header enum

JackMc:
## Description

We need the `Origin` header to detect if we should perform server side rendering.  This adds it to the `Header` enum in `@shopify/network`.

## Type of change

## Checklist


alexandcote (approved):

JackMc:
We can get this via `ctx.request.origin`, closing

PR #1660, opened 2020-10-22
Replace deprecated `babel-polyfills` in favor of `core-js` and `regenerator-runtime`

tleunen:
## Description

Babel-polyfill is deprecated and shouldn't be used anymore. We should use core-js directly now.
https://babeljs.io/docs/en/babel-polyfill

I'm aware of [this issue](https://github.com/Shopify/quilt/pull/1518) that cause the revert somehow. But I think it was caused by a misconfiguration maybe? By using `core-js` v3. We should also update our babel preset to configure preset-env with corejs 3, and might prevent the size increase?

Currently, we still default on v2, but consumers can customize it. Once this PR is shipped, we should default on v3. https://github.com/Shopify/web-configs/blob/ba2c02cac3fc150c5603493506e7115774335e59/packages/babel-preset/web.js#L6

## Type of change

## Checklist

### DIFF
--- a/packages/polyfills/CHANGELOG.md
+++ b/packages/polyfills/CHANGELOG.md
@@ -7,6 +7,8 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## [Unreleased] -->
 
+- Replace deprecated `babel-polyfills` in favor of `core-js` and `regenerator-runtime`. [#1660](https://github.com/Shopify/quilt/pull/1660)
+
 ## [1.2.4] - 2020-10-20
 
 - Updated `tslib` dependency to `^1.14.1`. [#1657](https://github.com/Shopify/quilt/pull/1657)
--- a/packages/polyfills/package.json
+++ b/packages/polyfills/package.json
@@ -24,15 +24,16 @@
   },
   "homepage": "https://github.com/Shopify/quilt/blob/master/packages/polyfills/README.md",
   "dependencies": {
-    "@babel/polyfill": "^7.10.1",
     "@shopify/useful-types": "^2.2.1",
     "browser-unhandled-rejection": "^1.0.2",
     "caniuse-api": "^3.0.0",
+    "core-js": "^3.6.5",
     "formdata-polyfill": "^3.0.18",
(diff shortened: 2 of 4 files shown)


packages/polyfills/package.json
```diff
@@ -24,15 +24,16 @@
   },
   "homepage": "https://github.com/Shopify/quilt/blob/master/packages/polyfills/README.md",
   "dependencies": {
-    "@babel/polyfill": "^7.10.1",
     "@shopify/useful-types": "^2.2.1",
     "browser-unhandled-rejection": "^1.0.2",
     "caniuse-api": "^3.0.0",
+    "core-js": "^3.6.5",
```
ismail-syed, line 30: Why isn't v3 begin reflected in the yarn.lock 🤔

packages/polyfills/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## [Unreleased] -->
 
+- Replace deprecated `babel-polyfills` in favor of `core-js` and `regenerator-runtime`. [#1660](https://github.com/Shopify/quilt/pull/1660)
```
ismail-syed, line 10: Would this be a breaking change or minor?

packages/polyfills/package.json
tleunen: Probably because other packages use v2? I haven't really looked much into that as it seems upgrading to v3 is more problematic than I thought

packages/polyfills/CHANGELOG.md
tleunen: I think it would actually be a major since it requires consumers to run babel with the corejs v3 flag. But @alexandcote shared worries with such an upgrade as he tried to do a the same upgrade in sewing kit https://github.com/Shopify/sewing-kit/pull/1899 So I'm not exactly sure what the foundation team decided to do with this core-js upgrade

alexandcote: I'm planning to do the full core-js bump during the next hack day so after BFCM most of the project will be able to use core-js 3 and new ES features.

PR #1661, opened 2020-10-22
Support configuring Html props from react-server

BPScott:
## Description

react-html's `<Html>` component supports a bunch of props for configuring
content to inject into the head, however previously that was not exposed
from react-server's createRender and createServer functions.

This commit lets you pass a htmlProps option into createRender and
createServer that will then be applied when createRender calls `<Html>`.

This means you can use createServer, and inject arbitary content into
the head using `<Html>`'s headMarkup prop (and use any of `<Html>`'s other
props too):

```js
createServer({
  render: () => <MockApp />,
  htmlProps: {
    scripts: [{path: '/extraScript.js'}],
    styles: [{path: '/extraStyle.css'}],
    headMarkup:  <>
      <script>let ScriptFromHeadMarkup = 1;</script>
      <script>let ScriptFromSecondHeadMarkup = 1;</script>
    </>,
  },
}),
```

## To test

An hand-cranked end-to-end test of this might be hard as there probably aren't many demo apps lying around. 

Find an application that uses `createServer` within it's server entrypoint ([brand-registry](https://github.com/Shopify/brand-registry/blob/master/app/ui/server.ts#L28) does this but I'm otherwise unfamiliar with the app). Add 'htmlProps' to its list of options, and then note that the markup is injected into the head:

```
htmlProps: {
  headMarkup:  <>
    <script>let ScriptFromHeadMarkup = 1;</script>
  </>,
}
```

Fortunately the createServer tests act as automated end-to-end test by configuring the server and asserting what html gets returned. [this new unit test](https://github.com/Shopify/quilt/pull/1661/files#diff-0e3f16674e62102134dc2abefc47aeb1282f0d23213f96c2cada5c2886e763ddR56) demonstrates adding `htmlProps` to the call to `createServer` and asserting that the returned html includes that head markup.

So the real tophat instructions are "check tests pass, check the added unit tests showcase e2e behaviour" :D

## Type of change

- [x] @shopify/react-html Patch: Bug (non-breaking change which fixes an issue)
- [x] @shopify/react-server Minor: New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/packages/react-html/CHANGELOG.md
+++ b/packages/react-html/CHANGELOG.md
@@ -7,6 +7,8 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## [Unreleased] -->
 
+- Update `HtmlProps` to mark children as optional (same as any React component) and export it ([#1661](https://github.com/Shopify/quilt/pull/1661))
+
 ## [10.0.1] - 2020-10-20
 
 - Updated `tslib` dependency to `^1.14.1`. [#1657](https://github.com/Shopify/quilt/pull/1657)
--- a/packages/react-server/CHANGELOG.md
+++ b/packages/react-server/CHANGELOG.md
@@ -7,7 +7,7 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## [Unreleased] -->
 
-- Add `htmlProps` to the options for `createRender` and `createServer`, these props will be passed into the call to `@shopify/react-html`'s `<Html>` component ([#1661](https://github.com/Shopify/quilt/pull/1661))
+- Added `htmlProps` to the options for `createRender` and `createServer`, these props will be passed into the call to `@shopify/react-html`'s `<Html>` component ([#1661](https://github.com/Shopify/quilt/pull/1661))
 
 ## [0.18.4] - 2020-10-20
 
--- a/packages/react-server/CHANGELOG.md
+++ b/packages/react-server/CHANGELOG.md
@@ -7,6 +7,8 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## [Unreleased] -->
 
+- Add `htmlProps` to the options for `createRender` and `createServer`, these props will be passed into the call to `@shopify/react-html`'s `<Html>` component ([#1661](https://github.com/Shopify/quilt/pull/1661))
+
 ## [0.18.4] - 2020-10-20
 
 - Added `tslib@^1.14.1` in the list of dependencies. [#1657](https://github.com/Shopify/quilt/pull/1657)
--- a/packages/react-server/README.md
+++ b/packages/react-server/README.md
@@ -87,6 +87,9 @@ interface Options {
   debug?: boolean;
   // a function similar to the render option but specifically used to render error pages for production SSR errors
   renderError: RenderFunction;
+  // additional props to pass into the Html component, or a function that takes a Koa.Context and returns a props object.
+  // See https://github.com/Shopify/quilt/blob/master/packages/react-html/README.md#html-
+  htmlProps?: HtmlProps | (ctx: Context) => HtmlProps
 }
 ```
 
--- a/packages/react-server/README.md
+++ b/packages/react-server/README.md
@@ -87,7 +87,7 @@ interface Options {
   debug?: boolean;
   // a function similar to the render option but specifically used to render error pages for production SSR errors
   renderError: RenderFunction;
-  // additional props to pass into the Html component, or a function that takes a Koa.Context and returns a props object.
+  // additional props to pass into the Html component, or a function that takes a Koa.Context and returns a props object
   // See https://github.com/Shopify/quilt/blob/master/packages/react-html/README.md#html-
   htmlProps?: HtmlProps | (ctx: Context) => HtmlProps
 }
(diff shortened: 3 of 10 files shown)


packages/react-server/src/render/render.tsx
```diff
@@ -133,8 +148,16 @@ export function createRender(render: RenderFunction, options: Options = {}) {
         assets.scripts({name: assetName, asyncAssets: immediateAsyncAssets}),
       ]);
 
+      styles.push(...additionalStyles);
```
BPScott, line 151: Making sure that if somebody decides to use styles/scripts then they don't clobber the assets provided by `getAssets`. I've added a test case for this.

packages/react-server/src/server/server.ts
```diff
@@ -18,10 +17,11 @@ interface Options {
   port?: number;
   assetPrefix?: string;
   proxy?: boolean;
-  assetName?: string | ValueFromContext<string>;
```
BPScott, line 21: Strictly speaking the change to assetName and renderError isn't needed as this is functionally identical, but I figure reusing the existing type is nicer than having to repeat the declarations.

packages/react-html/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## [Unreleased] -->
 
+- Update `HtmlProps` to mark children as optional (same as any React component) and export it ([#1661](https://github.com/Shopify/quilt/pull/1661))
```
kaelig, line 10:
```suggestion
- Updated `HtmlProps` to mark children as optional (same as any React component) and export it ([#1661](https://github.com/Shopify/quilt/pull/1661))
```

packages/react-html/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## [Unreleased] -->
```
kaelig, line 8:
```suggestion
## [Unreleased]
```

kaelig:
Would it make sense to add an example in the README, showing what's now possible to do?

BPScott:
@kaelig I added to the README section for what createServer accepts:

https://github.com/Shopify/quilt/pull/1661/files#diff-1557c7c2c28cba8b284d1990ba4ad33ddf799098943c401e2cc307886d5c5ff3R90

packages/react-html/CHANGELOG.md
BPScott: quilt convention tends towards not uncommenting / recommending this heading constantly

packages/react-server/src/render/render.tsx
```diff
@@ -74,6 +80,15 @@ export function createRender(render: RenderFunction, options: Options = {}) {
         ? assetNameInput(ctx)
         : assetNameInput;
 
+    const {
+      scripts: additionalScripts = [],
+      styles: additionalStyles = [],
+      ...additionalHtmlProps
+    } =
+      typeof htmlPropsInput === 'function'
```
ismail-syed, line 88: A bit nitpicky but I found this a bit difficult to parse at first glance because of the line break. 
Maybe split this up into 2 expressions by pulling out this ternary into a  `const props` or something.

ismail-syed (approved):
> might be hard as there probably aren't many demo apps lying around.

A majority of our `createServer` use cases are abtracted behind the [react-server webpack plugin](https://github.com/Shopify/quilt/blob/master/packages/react-server/src/webpack-plugin/webpack-plugin.ts#L93). I don't foresee any impact on those cases with this PR (I didn't tophat).  🎩ing this is challenging without a [beta release](https://github.com/Shopify/quilt/blob/master/documentation/guides/release-and-deploy.md#step-1---publish-a-beta-release-for-testing) of some sort. If you feel confident with these changes, that's 👍  with me. If you want to run beta release, I can give this a 🎩 .

packages/react-server/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## [Unreleased] -->
 
+- Add `htmlProps` to the options for `createRender` and `createServer`, these props will be passed into the call to `@shopify/react-html`'s `<Html>` component ([#1661](https://github.com/Shopify/quilt/pull/1661))
```
kaelig, line 10:
```suggestion
- Added `htmlProps` to the options for `createRender` and `createServer`, these props will be passed into the call to `@shopify/react-html`'s `<Html>` component ([#1661](https://github.com/Shopify/quilt/pull/1661))
```

packages/react-server/src/render/render.tsx
kaelig: I agree, the linebreak after the `=` is quite unconventional. Not sure it's worth adding a level of indirection to fix this, though.

packages/react-server/README.md
```diff
@@ -87,6 +87,9 @@ interface Options {
   debug?: boolean;
   // a function similar to the render option but specifically used to render error pages for production SSR errors
   renderError: RenderFunction;
+  // additional props to pass into the Html component, or a function that takes a Koa.Context and returns a props object.
```
kaelig, line 90:
```suggestion
  // additional props to pass into the Html component, or a function that takes a Koa.Context and returns a props object
```

(for consistency with the previous comments)

kaelig (approved):
Thank you for the README update, I approve the README side of things! 👍

Issue #1322, opened 2020-03-12
i18n.translationKeyExists() does not work for root translations


larouxn:
## Overview
I seem to have come across a broken use case for `i18n.translationKeyExists()`. In my experience, a key that certainly exists and returns translations properly can cause `i18n.translationKeyExists()` to return false when passed.
![custom_packaging](https://user-images.githubusercontent.com/1557529/76478311-926a8280-644b-11ea-8a69-bf04e3e7ba0e.png)

One thing to note, the translations I'm trying this with are in a root/base translations directory, `app/locales/`, as opposed to being in some sort of component level `translations` folder, which seems to be the preferred location these days.

## Reproduction
1. Open a project using quilt
2. Place some `console.log` to check the return of `i18n.translationKeyExists()` for root level keys versus deeper level keys (not sure how to get into actual in-context `debugger` session), ex:
```js
console.log(
  `root level translation key exists? (ShippingSettings) ${this.props.i18n.translationKeyExists(
    'ShippingSettings',
  )}`,
);
console.log(
  `deeper level translation key exists? (ShippingSettings.Packages.CustomPackaging.custom_box) ${this.props.i18n.translationKeyExists(
    'ShippingSettings.Packages.CustomPackaging.custom_box',
  )}`,
);
```
3. Boot said project, visit the page in which you places the `console.log`
4. Compare the printed logs

## Checklist

- [x] Please delete the labels section before submitting your issue
- [x] I have described this issue in a way that is actionable (if possible)

gmcgibbon:
#1662 was merged and released in https://github.com/Shopify/quilt/commit/29e5390e1515b6a8db1fb6e994546eb1b37d65f8. Closing this for now, but feel free to reopen if we missed something. Thanks!

larouxn:
👋 @gmcgibbon, works for me. Was actually planning on looking at the fix and closing this today anyway. 😅

PR #1662, opened 2020-10-27
Fix getTranslationTree when having multiple dictionaries

gregoriokusowski:
Co-authored-by: Shalini Mekala <shalini.mekala@shopify.com>

## Description

When having multiple dictionaries, getTranslationTree was failing to retrieve results for a given key.
I guess it's also related to #1322

## Type of change

## Checklist


msterle (approved):
good catch, lgtm

It would be interesting to check if this actually fixes #1322 entirely - it looks like it might

gmcgibbon (approved):

PR #1663, opened 2020-10-27
Add ordinal pluralization to React I18n schema specification document

movermeyer:
## Description

The original React i18n schema document had a major omission: ordinal pluralization.

This was already implicitly defined by `@shopify/react-i18n`'s [`ordinal`](https://github.com/Shopify/quilt/blob/ea57a3d8cc077143e04ff193ea8ec00a215dd6ba/packages/react-i18n/src/i18n.ts#L312-L316) method

## Type of change

A new version of the React I18n schema specification document.

Since the change technically makes the specification "looser" in what it accepts, this is a `2.0` change, and not a `1.1` change (per the versioning rules laid out in the `1.0` version). 

(`1.0` parsers would consider ordinal pluralization contexts to be invalid cardinal pluralization contexts)

# Note on nested pluralization contexts

While unusual, the spec currently allows pluralization contexts can be nested within themselves, so long as they obey the rules for valid pluralization contexts.

* **Cardinal** pluralization contexts can be nested inside **ordinal** pluralization contexts, and vice versa.
* **Cardinal** pluralization contexts can be nested inside **cardinal** pluralization contexts
* **Ordinal** pluralization contexts can be nested inside **ordinal** pluralization contexts

**Example English file with nested pluralization contexts:**

```jsonc
{
  // Note: these are strange, and probably not how you'd typically structure your files
  "cars": {
    "ordinal": {
      "one": "This is my {amount}st car",
      "two": "This is my {amount}nd car",
      "few": "This is my {amount}rd car",
      "other": "This is my {amount}th car",
      "cardinal": { // Cardinal pluralization context nested within an ordinal pluralization context
        "one": "I have {count} car",
        "other": "I have {count} cars"
      },
    }
  },
  "boats": {
    "one": "I have {count} boat",
    "other": "I have {count} boat",
    "ordinal": { // Ordinal pluralization context nested within a cardinal pluralization context
      "one": "This is my {amount}st boat",
      "two": "This is my {amount}nd boat",
      "few": "This is my {amount}rd boat",
      "other": "This is my {amount}th car",
    }
  }
}
```

I'm not going to mention this in the specification, as it's a bit of a grey area that I don't think that we need to address yet. A future version of the spec may decide to officially support/restrict this.

I include it as a note in this PR description, because usually the inability to have arbitrary nesting of constructs is an indicator of a bad specification (ie. a smell).

## Checklist

- [x] I have added a changelog entry, within the "History" section of the document.

### DIFF
--- a/packages/react-i18n/docs/react_i18n_schema.md
+++ b/packages/react-i18n/docs/react_i18n_schema.md
@@ -4,13 +4,18 @@ React I18n is a schema comprised of a structure of nested key-value pairs (KVPs)
 
 <!-- Created using "Markdown All in One" extension for VS Code -->
 - [History](#history)
+  - [Version 2.0](#version-20)
   - [Version 1.0](#version-10)
 - [Versioning](#versioning)
 - [Basic Structure](#basic-structure)
 - [Pluralization](#pluralization)
-  - [Reserved pluralization keys](#reserved-pluralization-keys)
-  - [Invalid pluralization keys](#invalid-pluralization-keys)
-  - [Detecting pluralization contexts](#detecting-pluralization-contexts)
+  - [Cardinal Pluralization](#cardinal-pluralization)
+    - [Reserved cardinal pluralization keys](#reserved-cardinal-pluralization-keys)
+    - [Invalid cardinal pluralization keys](#invalid-cardinal-pluralization-keys)
+    - [Detecting cardinal pluralization contexts](#detecting-cardinal-pluralization-contexts)
+  - [Ordinal Pluralization](#ordinal-pluralization)
+    - [Invalid ordinal pluralization keys](#invalid-ordinal-pluralization-keys)
+    - [Detecting ordinal pluralization contexts](#detecting-ordinal-pluralization-contexts)
 - [Serialization](#serialization)
   - [Serialization to JSON](#serialization-to-json)
     - [Comments](#comments)
@@ -21,6 +26,14 @@ React I18n is a schema comprised of a structure of nested key-value pairs (KVPs)
     - [Interpolation](#interpolation)
 
 # History
+
+## Version 2.0
+
+* Use the more precise language: `"Pluralization" -> "Cardinal Pluralization"`
+* Clarify that a cardinal pluralization context only includes immediate children
+* Add a description of how ordinal pluralization works.
+    * This was an omission in the version `1.0` specification.
+* Adjust the rules for reserved cardinal pluralization keys to account for their usage in ordinal pluralization contexts
 
 ## Version 1.0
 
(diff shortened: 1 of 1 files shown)


packages/react-i18n/docs/react_i18n_schema.md
```diff
@@ -105,48 +120,116 @@ In order to avoid key collisions, the following keys are reserved for use as plu
...
+## Ordinal Pluralization
+
+Pluralization of ordinal numbers is handled by providing the ordinal pluralization keys required by the language as leaf KVPs, under an `ordinal` parent, creating an "ordinal pluralization context"
+
+**Example English file with ordinal pluralization keys:**
+
+```jsonc
+{
+  "cars": {
+    "ordinal": { // The special key "ordinals" creates an ordinalization context
```
sjahu, line 181:
```suggestion
    "ordinal": { // The special key "ordinal" creates an ordinalization context
```

sjahu: I don't mind `ordinalization context`, but everywhere else you call it an `ordinal pluralization context`

sjahu (commented):
Makes sense to me

packages/react-i18n/docs/react_i18n_schema.md
movermeyer: Great spot. Fixed on both counts.

packages/react-i18n/docs/react_i18n_schema.md
```diff
@@ -105,48 +120,116 @@ In order to avoid key collisions, the following keys are reserved for use as plu
...
   }
 }
 ```
 
-## Detecting pluralization contexts
+### Detecting cardinal pluralization contexts
+
+These constraints allow for a simple algorithm for identifying cardinal pluralization contexts:
+
+* If a node has child leaf KVPs for **any** of the cardinal pluralization keys required by the language, outside of a then its **immediate** children are within a cardinal pluralization context.
```
gmcgibbon, line 168: `outside of a then its` doesn't make sense to me, I think you're missing a word here.

packages/react-i18n/docs/react_i18n_schema.md
```diff
@@ -105,48 +120,116 @@ In order to avoid key collisions, the following keys are reserved for use as plu
...
+
+Different languages require different sets of ordinal pluralization keys. For example, Polish (`pl`) only has 1 required ordinal pluralization key: `other`. The Polish version of the previous example would be:
+
+**Example Polish file with ordinal pluralization keys:**
+
+```json
+{
+  "cars": {
+    "ordinal": {
+      "other": "To mój {amount}. samochód"
```
gmcgibbon, line 199: Is the period after the number intentional?

packages/react-i18n/docs/react_i18n_schema.md
```diff
@@ -22,6 +27,14 @@ React I18n is a schema comprised of a structure of nested key-value pairs (KVPs)
 
 # History
 
+## Version 2.0
```
gmcgibbon, line 30: It is unclear to me if you simply didn't mention this support in all of our current parsers or if there are any parsers actually implemented with the 1.0 spec that _aren't_ compatible with this new assumption? We only need to define a new spec if anyone is actually affected by this amendment.

packages/react-i18n/docs/react_i18n_schema.md
movermeyer: Yes, that is how Polish does it: https://en.wikipedia.org/wiki/Ordinal_indicator#Ordinal_dot
I also confirmed it with a native Polish speaker.

packages/react-i18n/docs/react_i18n_schema.md
movermeyer: Very true. Fixed.

packages/react-i18n/docs/react_i18n_schema.md
movermeyer: 🤷 You're right. There probably aren't any.

However, I like to treat specification documents as "append only". This could have been considered errata, or perhaps the `1.0` document should have been in some form of "draft" mode until we had confidence that all the details had been ironed out. 

In the end, versions numbers are cheap. 🤷

gmcgibbon: A major version jump seems excessive to me when there are no breaking changes. I would rather use `1.1`, but I'll leave that up to you.

gmcgibbon (approved):

sjahu (approved):

rafaelxy (approved):

PR #1664, opened 2020-11-02
Refresh the react server on current page

beckygowland:
## Description

When the node react server is re-starting and not yet booted up you'll see the react_render_error page, which tries to refresh until the server is up. The refresh currently always tries to re-direct to '/', which throws you out of the page you were just developing. This removes the hardcoded path to refresh to and instead refreshes the current page.

## Type of change

- [x] Quilt-rails Patch: Bug (non-breaking change which fixes an issue)

## Checklist


ismail-syed (approved):
🎩 Checkouts. Thank you for the fix! 🙏 ❤️

PR #1665, opened 2020-11-02
Add dynamic list functionality to react-form package [useDynamicList]

oluwatimio:
## Description

Adds useDynamicList to `react-form` package. This is a hook that adds on the useList hook. It's main feature is to dynamically add and remove list items.

A code sandbox can be seen [here](https://codesandbox.io/s/hungry-rubin-exrkz?fontsize=14&hidenavigation=1&theme=dark) 

This will allow us create unique user interfaces. For example, adding bulk payment information at once, adding bulk addresses, adding bulk customers in one go. 

This will help merchants save time. For example, if they have more than 1 address/payment method for a customer they can add these at once and save it in one go.

PS - While reviewing the code, please note that the `dynamiclist.test.tsx` file includes similar tests from `list.test.tsx` this is to ensure that all previous functionality also works in `useDynamicList`. The `dynamiclist.test.tsx` file also adds a new test suite called `add and remove fields`

## Tophat

A code sandbox can be seen [here](https://codesandbox.io/s/hungry-rubin-exrkz?fontsize=14&hidenavigation=1&theme=dark) 

## Type of change

`useList` this has been modified to use a new `useHandler` hook.
`useDynamicList` a new hook has been added.

- [X]  Minor: New feature (non-breaking change which adds functionality)

## Checklist


packages/react-form/src/hooks/list/reducer.ts
```diff
@@ -21,6 +23,16 @@ interface ReinitializeAction<Item> {
   payload: {list: Item[]};
 }
 
+interface AddFieldsAction<Item> {
+  type: 'addFields';
+  payload: {list: Item[]};
+}
+
+interface RemoveFieldsAction {
```
attila-berki, line 31: It might be better to use a singular form `RemoveFieldAction` as it only removes one field at a time.

packages/react-form/src/hooks/list/dynamiclist.ts
```diff
@@ -0,0 +1,180 @@
...
+} from './reducer';
+
+interface DynamicList<Item extends object> {
+  fields: FieldDictionary<Item>[];
+  addField(): void;
+  removeField(index: number): void;
+}
+
+export function useDynamicList<Item extends object>(
+  initialList: Item[],
```
attila-berki, line 34: Would it be possible to keep the same signature as `useList`, where you can either pass in the list of fields of a config object? 
`listOrConfig: FieldListConfig<Item> | Item[],`

What was the reason to separate them?

packages/react-form/src/hooks/list/dynamiclist.ts
```diff
@@ -0,0 +1,180 @@
...
+          ...field,
+          ...(handlers[index][key] as any),
+        };
+      });
+    });
+  }, [state.list, handlers]);
+  return {fields, addField, removeField};
+}
+
+function runValidation<Value, Record extends object>(
```
attila-berki, line 153: Would it be possible to move this function into a utility and share it between the `list` and `dynamiclist`?

packages/react-form/src/hooks/list/dynamiclist.ts
```diff
@@ -0,0 +1,180 @@
...
+
+  function addField() {
+    dispatch(addFieldsAction([factory()]));
+  }
+
+  function removeField(index: number) {
+    dispatch(removeFieldsAction(index));
+  }
+
+  const handlers = useMemo(() => {
```
attila-berki, line 70: I think this would be a good candidate for a custom hook which is shared between the `list` and `dynamic list`. What do you think?

packages/react-form/src/hooks/list/dynamiclist.ts
oluwatimio: Ahh yes that's fair! I think I separated them for simplicity when testing but will put them back together

packages/react-form/src/hooks/list/dynamiclist.ts
```diff
@@ -0,0 +1,180 @@
...
+
+interface DynamicList<Item extends object> {
+  fields: FieldDictionary<Item>[];
+  addField(): void;
+  removeField(index: number): void;
+}
+
+export function useDynamicList<Item extends object>(
+  initialList: Item[],
+  factory: Function,
```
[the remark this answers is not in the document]
oluwatimio: Will look into that! I think it's a possibility

packages/react-form/src/hooks/list/dynamiclist.ts
oluwatimio: I think it's good! Will try to move it to a hook

michenly (commented):
The first initial thoughts. 
* 1. if `useList` & `useDynamicList` are so similar, can we either try doing method overloading to determined what return in the hook, or have `useDynamicList` use `useList` under the hood. The current code for the two hooks are so similar that I imagine future bug fix for one will also go into another and they should share as much logic as possible.

* 2. I think `addField` and `removeField` should be renamed to `addItem` & `removeItem`. The functionality I saw is that these method work over a list item and not to a single field.

3. The Code Sandbox  works really well in demonstrating usage example! You gotta teach me how you included code that have not been published into Code Sandbox . Also I wonder if we should consider adding link to Code Sandbox into README ❤️

packages/react-form/src/hooks/list/dynamiclist.ts
```diff
@@ -0,0 +1,74 @@
...
+
+interface DynamicList<Item extends object> {
+  fields: FieldDictionary<Item>[];
+  addField(): void;
+  removeField(index: number): void;
+}
+
+export function useDynamicList<Item extends object>(
+  listOrConfig: FieldListConfig<Item> | Item[],
+  fieldFactory: Function,
```
michenly, line 24: Would like to see this function more tightly typed.
Also, should this be itemFactory? You use it to create an Item in the list that has multiple fields, correct?

oluwatimio:
> 1. n the hook, or have `useDynamicList` use `us`

For `i` what I did was to extract the handler to a reusable hook also some utils and the reducer are also extracted as well. I think using useList under the hood may be difficult to implement, not sure how that will work. Modifying the return of useList might be out of the question considering the return type of `useList` is a `FieldDictionary`. The return type of `DynamicList` is of type `DynamicList<Item>`. There are a lot of places in web that expect `useList` to be returning a `FieldDictionary`. In order to not make this a breaking change, I decided to separate them but take out a lot of the logic that is repeated in both into a reusable hook such as `useHandlers` and the util file. 

for `ii` I can change the naming to Item since technically we are adding an Item not just one field. Thanks!

packages/react-form/src/hooks/list/dynamiclist.ts
oluwatimio: Yes, will change to itemFactory :) thanks!

packages/react-form/CHANGELOG.md
```diff
@@ -5,7 +5,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## [Unreleased] -->
+## [Unreleased]
+
+- Added `useDynamicList` functionality.
```
michenly, line 10: can you described a little bit on what `useDynamicList` does here?

michenly:
@oluwatimio 🙆‍♀️ agree on not introducing breaking change, however, I still see a lot of coy & paste code in `useDynamicList` & `useList` so we can defin' refactor further.

 What I am thinking:


```ts
export function useDynamicList<Item extends object>(
  listOrConfig: FieldListConfig<Item> | Item[],
  fieldFactory: Function,
  validationDependencies: unknown[] = [],
): DynamicList<Item> {
  const {fields, dispatch} = useBaseList(listOrConfig, validationDependencies);

  function addField() {
    dispatch(addFieldsAction([fieldFactory()]));
  }

  function removeField(index: number) {
    dispatch(removeFieldsAction(index));
  }

  return {fields, addField, removeField};
}
```

```ts
export function useList<Item extends object>(
  listOrConfig: FieldListConfig<Item> | Item[],
  validationDependencies: unknown[] = [],
): FieldDictionary<Item>[] {
  const {fields} = useBaseList(listOrConfig, validationDependencies);

  return fields;
}
```

```ts
export function useBaseList<Item extends object>(
  listOrConfig: FieldListConfig<Item> | Item[],
  validationDependencies: unknown[] = [],
): {fields: FieldDictionary<Item>[]; action: React.Dispatch<ListAction<Item>>} {

 /* current useList implementation */

  return {fields, dispatch};
```

same can be say about the test, where there is no need for duplication, and most functionalities can be tested in `useBaseList`

---

Also some smaller nitpick
- `useHandlers.ts` should be renamed to `handlers.ts`
- both `handlers` and `reducer` function as utilities for list, they should be treated the same. ie. both file located in the same folder.

michenly (approved):
Small last nitpick. The comments in `list.ts` mostly applied to `baselist.ts`. Pleas move / edit accordingly.

Otherwise this is 💯 🎉 🥂

packages/react-form/src/hooks/list/test/dynamiclist.test.tsx
```diff
@@ -0,0 +1,105 @@
...
+            </li>
+          ))}
+          <button type="button" onClick={() => addItem()}>
+            Add Variant
+          </button>
+        </ul>
+      );
+    }
+
+    it('can remove field', () => {
```
attila-berki, line 43: To make sure nothing breaks it might worth adding another test case which handles the remove when the list is already empty.

attila-berki (approved):
It looks really good 💯 🎉  and the code-sandbox looks great ❤️

shameelabd:
Quick question about this. Our use case is that we have an initial page of say 10 items, then a `Load remaining items` button. In this scenario would we have to call `addItem` for each new item returned from the api?

oluwatimio:
> Quick question about this. Our use case is that we have an initial page of say 10 items, then a `Load remaining items` button. In this scenario would we have to call `addItem` for each new item returned from the api?

Yes, I think this is possible with addItem. Basically, for each item, that you want to load you will have to call addItem. Also note that your factory function will be important in this case because that is the function that gets called for every time you call addItem.

shameelabd:
> > Quick question about this. Our use case is that we have an initial page of say 10 items, then a `Load remaining items` button. In this scenario would we have to call `addItem` for each new item returned from the api?
> 
> Yes, I think this is possible with addItem. Basically, for each item, that you want to load you will have to call addItem. Also note that your factory function will be important in this case because that is the function that gets called for every time you call addItem.

Yeah I would assume I would make my factory take in some parameters that would potentially override the default values. For my use-case I guess it would be nice if I could add a bunch of items in one function call, maybe something like `addItems` or something

oluwatimio:
> > > Quick question about this. Our use case is that we have an initial page of say 10 items, then a `Load remaining items` button. In this scenario would we have to call `addItem` for each new item returned from the api?
> > 
> > 
> > Yes, I think this is possible with addItem. Basically, for each item, that you want to load you will have to call addItem. Also note that your factory function will be important in this case because that is the function that gets called for every time you call addItem.
> 
> Yeah I would assume I would make my factory take in some parameters that would potentially override the default values. For my use-case I guess it would be nice if I could add a bunch of items in one function call, maybe something like `addItems` or something

When building this, I think the main functionality we were thinking about was to add items one at a time after the initial list load. The are some ways you can get around this. One way would be to have the factory be able to return an array of Items not just one Item at a time.

I can try adding this in a follow up PR

shameelabd:
> > > > Quick question about this. Our use case is that we have an initial page of say 10 items, then a `Load remaining items` button. In this scenario would we have to call `addItem` for each new item returned from the api?
> > > 
> > > 
> > > Yes, I think this is possible with addItem. Basically, for each item, that you want to load you will have to call addItem. Also note that your factory function will be important in this case because that is the function that gets called for every time you call addItem.
> > 
> > 
> > Yeah I would assume I would make my factory take in some parameters that would potentially override the default values. For my use-case I guess it would be nice if I could add a bunch of items in one function call, maybe something like `addItems` or something
> 
> When building this, I think the main functionality we were thinking about was to add items one at a time through `addItem` after the initial list load. The are some ways you can get around this. One way would be to have the factory be able to return an array of Items not just one Item at a time.
> 
> I can try adding this in a follow up PR

dope ty, even this is a huge step compared to what we're currently doing. Thanks a lot!

oluwatimio:
> > > > > Quick question about this. Our use case is that we have an initial page of say 10 items, then a `Load remaining items` button. In this scenario would we have to call `addItem` for each new item returned from the api?
> > > > 
> > > > 
> > > > Yes, I think this is possible with addItem. Basically, for each item, that you want to load you will have to call addItem. Also note that your factory function will be important in this case because that is the function that gets called for every time you call addItem.
> > > 
> > > 
> > > Yeah I would assume I would make my factory take in some parameters that would potentially override the default values. For my use-case I guess it would be nice if I could add a bunch of items in one function call, maybe something like `addItems` or something
> > 
> > 
> > When building this, I think the main functionality we were thinking about was to add items one at a time through `addItem` after the initial list load. The are some ways you can get around this. One way would be to have the factory be able to return an array of Items not just one Item at a time.
> > I can try adding this in a follow up PR
> 
> dope ty, even this is a huge step compared to what we're currently doing. Thanks a lot!

Turns out it's fairly straightforward to do this. Here's a draft PR https://github.com/Shopify/quilt/pull/1679. I will need to add a few more tests and add some more documentation

Issue #1666, opened 2020-11-02
Error entry is missing props on client


beckygowland:
## Overview

Adding a custom renderError component has props {url, data, error} which are populated on the server-side, but empty on the client-side. Tested with adding a console.log to print the props inside the custom error component. Looks like it's missing the hydration step.

## Consuming repo
services-db

PR #1667, opened 2020-11-02
Pass caught server error to custom error component

beckygowland:
## Description

For custom error components, there's the server error that is supposed to be passed in as a [prop](https://github.com/Shopify/quilt/blob/master/packages/react-server/src/webpack-plugin/webpack-plugin.ts#L105) that's currently undefined at the moment. This sets that value to the same error as what's caught and logged in the react server.

## Type of change

- [x] React-server Patch: Bug (non-breaking change which fixes an issue)

## Checklist


ismail-syed (approved):
👍 💯 
Can you please open an issue for the bug we caught with client-side hydration issues? Thanks!

beckygowland:
> 👍 💯
> Can you please open an issue for the bug we caught with client-side hydration issues? Thanks!

🎉  opened https://github.com/Shopify/quilt/issues/1666

PR #1669, opened 2020-11-04
Remove noop util and export in react-i18n

tleunen:
## Description

When working in a project with react-i18n and [`@shopify/javascript-utilities`](https://github.com/Shopify/javascript-utilities), it happens very often that the `noop` function from react-i18n is suggested instead of the one from the Js util repo and it's a bit annoying. Even if the noop is not exported from the root of the project.

Removing it is easy as it wasn't even used anyway :)

## Checklist

### DIFF
--- a/packages/react-i18n/CHANGELOG.md
+++ b/packages/react-i18n/CHANGELOG.md
@@ -8,6 +8,7 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 <!-- ## [Unreleased] -->
 
 - Fix bug in `getTranslationTree` when having multiple translation dictionaries. [#1662](https://github.com/Shopify/quilt/pull/1662)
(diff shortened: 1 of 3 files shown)


cejaekl (approved):
OK by me.

BPScott:
Removing a public export is a breaking change innit? Even if nobody in their right mind should be using that export

tleunen:
I think it depends how you see it. For me it's not a breaking change because because the root exports don't change. `noop` has never been exported from `react-i18n`. It's only exported from `src/utilities`, which for some reasons is picked up by VSCode and it suggests to import noop from `@shopify/react-i18n/dist/src/utilities` :/

BPScott:
Ah I misread the paths, sorry. I thought it was exported from the main index not utilities/index. In that case, yep ignore me  and :shipit: !

BPScott (approved):

rafaelxy (approved):

PR #1671, opened 2020-11-09
[Dates] Calculate DST offset for days where we jump forward or back

KyleKilbride:
## Description

Fixes https://github.com/Shopify/store/issues/18493

Currently, there is an issue where if you are in a different timezone locally than your shops timezone, you cannot select the days where the time jumps forward or falls back. 

This PR adds a value for the offset of the DST changes made on those days. Any other day will have an offset of 0 but days where the time changes will get an offset of however much that change is, and will be applied to the target date, allowing you to select it properly.

This change exists only in `applyTimeZoneOffset()` and only updates the `Dates` package.

## Type of change

Adds a field to check for DST change in a given day

- [x] Dates - Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

### DIFF
--- a/packages/dates/src/apply-time-zone-offset.ts
+++ b/packages/dates/src/apply-time-zone-offset.ts
@@ -12,7 +12,7 @@ export function applyTimeZoneOffset(
   const initialOffset = getTimeZoneOffset(date, timeZone1, timeZone2);
   const adjustedDate = new Date(date.valueOf() - initialOffset * 60 * 1000);
   const targetOffset = getTimeZoneOffset(adjustedDate, timeZone1, timeZone2);
-  const offsetDiff = targetOffset - initialOffset - Math.abs(dstOffsetDiff);
+  const offsetDiff = targetOffset - initialOffset + dstOffsetDiff;
 
   return new Date(adjustedDate.valueOf() - offsetDiff * 60 * 1000);
 }
--- a/packages/dates/src/apply-time-zone-offset.ts
+++ b/packages/dates/src/apply-time-zone-offset.ts
@@ -7,12 +7,13 @@ export function applyTimeZoneOffset(
 ) {
   const dateToCheckForDSTOffset = new Date(date);
   dateToCheckForDSTOffset.setHours(2);
+  const needsDSTOffset = date.getTimezoneOffset() !== dateToCheckForDSTOffset.getTimezoneOffset();
+  const dstOffset = needsDSTOffset ? 60 : 0;
 
-  const dstOffsetDiff = date.getTimezoneOffset() - dateToCheckForDSTOffset.getTimezoneOffset();
   const initialOffset = getTimeZoneOffset(date, timeZone1, timeZone2);
   const adjustedDate = new Date(date.valueOf() - initialOffset * 60 * 1000);
   const targetOffset = getTimeZoneOffset(adjustedDate, timeZone1, timeZone2);
-  const offsetDiff = targetOffset - initialOffset + dstOffsetDiff;
+  const offsetDiff = targetOffset - initialOffset - dstOffset;
 
   return new Date(adjustedDate.valueOf() - offsetDiff * 60 * 1000);
 }
--- a/packages/dates/src/apply-time-zone-offset.ts
+++ b/packages/dates/src/apply-time-zone-offset.ts
@@ -5,10 +5,14 @@ export function applyTimeZoneOffset(
   timeZone1?: string,
   timeZone2?: string,
 ) {
+  const dateToCheckForDSTOffset = new Date(date);
+  dateToCheckForDSTOffset.setHours(2);
+
+  const dstOffsetDiff = date.getTimezoneOffset() - dateToCheckForDSTOffset.getTimezoneOffset();
   const initialOffset = getTimeZoneOffset(date, timeZone1, timeZone2);
   const adjustedDate = new Date(date.valueOf() - initialOffset * 60 * 1000);
   const targetOffset = getTimeZoneOffset(adjustedDate, timeZone1, timeZone2);
-  const offsetDiff = targetOffset - initialOffset;
+  const offsetDiff = targetOffset - initialOffset - Math.abs(dstOffsetDiff);
 
   return new Date(adjustedDate.valueOf() - offsetDiff * 60 * 1000);
 }
--- a/packages/dates/src/tests/apply-time-zone-offset.test.ts
+++ b/packages/dates/src/tests/apply-time-zone-offset.test.ts
@@ -58,9 +58,9 @@ describe('applyTimeZoneOffset()', () => {
       ).toStrictEqual(expected);
     });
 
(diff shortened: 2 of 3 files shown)


packages/dates/src/apply-time-zone-offset.ts
```diff
@@ -5,10 +5,14 @@ export function applyTimeZoneOffset(
...
 ) {
+  const dateToCheckForDSTOffset = new Date(date);
+  dateToCheckForDSTOffset.setHours(2);
+
+  const dstOffsetDiff = date.getTimezoneOffset() - dateToCheckForDSTOffset.getTimezoneOffset();
   const initialOffset = getTimeZoneOffset(date, timeZone1, timeZone2);
   const adjustedDate = new Date(date.valueOf() - initialOffset * 60 * 1000);
   const targetOffset = getTimeZoneOffset(adjustedDate, timeZone1, timeZone2);
-  const offsetDiff = targetOffset - initialOffset;
+  const offsetDiff = targetOffset - initialOffset - Math.abs(dstOffsetDiff);
```
atesgoral, line 15: Why are we always subtracting the absolute value of the offset?

KyleKilbride: I couldn't really think of a better way of doing this, but in order to have it properly calculate the jump forward and back, we need to subtract the absolute value of the difference otherwise the calculation will be wrong on the jump forward.

atesgoral: This feels weird. An offset is an offset. You should be able to just add (or subtract it) based on how the offset is calculated. Could you maybe illustrate that problem by having two unit tests, one for the positive offset case and one for the negative offset case?

KyleKilbride: You know what, I totally goofed. It can be written like so: `const offsetDiff = targetOffset - initialOffset  + dstOffsetDiff;`

KyleKilbride: Ok sorry for stringing this along. I have refactored it to check if it needs a DST offset of `60`. DST will only ever be an offset of an hour.

packages/dates/src/tests/apply-time-zone-offset.test.ts
```diff
@@ -57,5 +57,14 @@ describe('applyTimeZoneOffset()', () => {
         applyTimeZoneOffset(date, 'Australia/Sydney', 'UTC'),
       ).toStrictEqual(expected);
     });
+
+    it('applies extra dst offset for days where DST is applied or reverted', () => {
+      const date = new Date('2020-11-01T04:00:00.000Z');
+      const expected = new Date('2020-11-01T06:00:00.000Z');
+
+      expect(
+        applyTimeZoneOffset(date, 'America/Denver', undefined),
```
thetrevorharmon, line 66: Does this test partially rely on the timezone of the test runner? If so, do we need to explicitly set it with jest?

KyleKilbride: no, I don't believe so, I am however going to add an explicit second timezone to this function call. 👀

thetrevorharmon (approved):
Well done. This was a tough bug to squash but you killed it!

packages/dates/src/apply-time-zone-offset.ts
cejaekl: > DST will only ever be an offset of an hour.

Nope.  😢 

Granted, such edge cases are rare, but they [do](https://www.zainrizvi.io/blog/falsehoods-programmers-believe-about-time-zones/) [happen](https://en.wikipedia.org/wiki/Time_in_Russia).

packages/dates/src/apply-time-zone-offset.ts
```diff
@@ -5,10 +5,15 @@ export function applyTimeZoneOffset(
   timeZone1?: string,
   timeZone2?: string,
 ) {
+  const dateToCheckForDSTOffset = new Date(date);
+  dateToCheckForDSTOffset.setHours(2);
+  const needsDSTOffset = date.getTimezoneOffset() !== dateToCheckForDSTOffset.getTimezoneOffset();
+  const dstOffset = needsDSTOffset ? 60 : 0;
```
cejaekl, line 11: DST offsets are not always 60 minutes.
 - Lord Howe Island [currently uses](https://www.zainrizvi.io/blog/falsehoods-programmers-believe-about-time-zones/) a 30-minute offset
 - England used a 2-hour offset in the past
 - Russia effectively [applied](https://en.wikipedia.org/wiki/Time_in_Russia) a 2-hour offset when they changed their DST rules in 2011

I'm OK with shipping this as is for now, but we should open an issue reminding us to circle back and handle such cases in the future.

cejaekl (commented):
Does this make things better than they were before?  Definite yes.
Is the change totally correct?  Not quite.

Up to you whether you want to ship what you have and then make things better later, or revise the PR before it goes out.

KyleKilbride:
> Does this make things better than they were before? Definite yes.
> Is the change totally correct? Not quite.
> 
> Up to you whether you want to ship what you have and then make things better later, or revise the PR before it goes out.

Really appreciate the context and review @cejaekl! I think I will revise this PR and ship it fully fixed, that would be ideal. Ideas are also welcome, this dst / timezone stuff has been breaking my brain!

packages/dates/src/apply-time-zone-offset.ts
KyleKilbride: So I have looked into these timezones and others, and it seems this solution still works fine. I haven't been able to observe any issues. It also looks like as of 2014 Russia isn't participating in DST? 
>  Daylight saving time is not used in Russia (since 26 October 2014)

KyleKilbride: I think with this I am ok moving forward with this solution, just need to figure out the tests.

atesgoral:
We had a chat with @KyleKilbride about the root cause of the issue. Sure, we can apply a bandaid fix, but I *strongly* think we should fix the root cause.

If the merchant impact it not high, let's hold off on just patching things up and find a longer-term solution by spring 2021, when we'll have to deal with daylight saving shenanigans again.

The root cause is that the data picker component deals with JavaScript Date objects for no reason. If you think about the user intent of picking a date, the date picker should be dealing with *logical* dates. So, if as a user I'm seeing and picking 2020-11-01, I'd better get 2020-11-01 as the output (could even be the ISO8601 string "2020-11-01"). It's the consumer of the date picker that should deal with translating that logical date into a JavaScript Date object on a specific time zone (shop, browser, etc.)

We seem to have ~100 instances in Shopify/web where we're juggling time zone offsets. I think it's a smell. Let's have a UX + I18n + general software principle discussion around cleaning up our act instead of patching something that is already broken.

cc @Shopify/web-foundations

atesgoral (changes requested):
Let's fix the issue at the source -- which is asking why we're doing this much time zone juggling in the first place

PR #1673, opened 2020-11-11
Support external urls with Link

### LINKED ISSUE #1672: Better Link default behaviour from react-router package
## Overview

See this Discourse link for full details: https://discourse.shopify.io/t/default-behaviour-of-linkcomponent-is-wrong-for-external-links/11940

TLDR: The `Link` component exported by `@shopify/react-router` should support external links.

## Type

- [ ] New feature

## Motivation

We pass through the React-Router `Link` component in our `@shopify/react-router` package, but the Polaris documentation recommends a [code snippet](https://github.com/Shopify/polaris-react/blob/72748e3ddb7d60433ba6ed9d9895b390ffa7a283/src/components/AppProvider/README.md#using-linkcomponent) that each application implement that wraps this `Link` component if they want to support links to external URLs. Here's an example of where we had to manually implement this behaviour in ServicesDB: https://github.com/Shopify/services-db/pull/6459/

I'd argue that links to external URLs is the type of behaviour that we would want to support by default (and instead give you the ability to opt-out if desired) and that this behaviour should live in `@shopify/react-router`. This proposal would change the `Link` component to provide the behaviour mentioned in the snippet, and export the original `Link` as `ReactRouterLink`. I believe this would be a breaking change.

When generating a new app with `sewing-kit`, the default `web/foundation/Polaris.tsx` module looks like this:

```ts
import React from 'react';
import {Link} from '@shopify/react-router';
import {AppProvider} from '@shopify/polaris';
import {useI18n} from '@shopify/react-i18n';

interface Props {
  children?: React.ReactNode;
}

export function Polaris({children}: Props) {
  const [i18n] = useI18n({
    id: 'Polaris',
    translations: async (code) => {
      return import(`@shopify/polaris/locales/${code}.json`);
    },
  });

  return (
    <AppProvider i18n={i18n.translations} linkComponent={Link}>
      {children}
    </AppProvider>
  );
}
```

By providing this change, the included `linkComponent` prop would "just work" for external links, without requiring any override code.

## Checklist

ryanwilsonperkin:
## Description

Fixes #1672 

Introduces support for external (ie. not relative) URLs to the `<Link>` component exported by the `@shopify/react-router` package.

When an external URL is rendered with the Link component, render it as a plain `<a>` tag rather than a `<ReactRouterLink>` if the URL is external (ie. not relative).

This ensures that we can continue to use the `<Link>` component without needing to provide workarounds in each service to first check the URL and resolve it differently.

This logic is taken from the [Polaris documentation](https://github.com/Shopify/polaris-react/blob/72748e3ddb7d60433ba6ed9d9895b390ffa7a283/src/components/AppProvider/README.md#using-linkcomponent) where it is a recommended code snippet, but I believe that it should be upstream.

## Type of change

## Checklist

### DIFF
--- a/packages/react-router/src/components/Link/Link.tsx
+++ b/packages/react-router/src/components/Link/Link.tsx
@@ -8,16 +8,20 @@ export interface Props extends ReactRouterLinkProps {
   url: string;
 }
 
-export default function Link({url, external, children, ...rest}: Props) {
-  let target: string | undefined;
-  let rel: string | undefined;
-  if (external) {
-    target = '_blank';
-    rel = 'noopener noreferrer';
+const IS_EXTERNAL_LINK_REGEX = /^(?:[a-z][a-z\d+.-]*:|\/\/)/;
+
+export default function Link(props: Props) {
+  const {url, external, children, ...rest} = props;
+  if (external || IS_EXTERNAL_LINK_REGEX.test(url)) {
+    return (
+      <a href={url} {...rest} target="_blank" rel="noopener noreferrer">
+        {children}
+      </a>
+    );
   }
 
   return (
-    <ReactRouterLink target={target} to={url} rel={rel} {...rest}>
+    <ReactRouterLink to={url} {...props}>
       {children}
     </ReactRouterLink>
   );
--- a/packages/react-router/src/components/Link/tests/Link.test.tsx
+++ b/packages/react-router/src/components/Link/tests/Link.test.tsx
@@ -36,9 +36,20 @@ describe('<Link />', () => {
   });
 
   describe('external', () => {
-    it('renders a react router link', () => {
+    it('does not render a react router link for external URL', () => {
       const link = mount(<Link url="https://shopify.com" />);
-      expect(link).toContainReactComponent(ReactRouterLink);
+      expect(link).not.toContainReactComponent(ReactRouterLink);
+      expect(link).toContainReactComponent('a', {
+        href: 'https://shopify.com',
+      });
+    });
+
+    it('does not render a react router link when external prop is provided', () => {
+      const link = mount(<Link external url="something-else" />);
+      expect(link).not.toContainReactComponent(ReactRouterLink);
+      expect(link).toContainReactComponent('a', {
+        href: 'something-else',
+      });
     });
 
     it('sets the `to` prop and passes along other props to ReactRouterLink', () => {
@@ -52,7 +63,7 @@ describe('<Link />', () => {
 
     it('sets default external attributes', () => {
       const link = mount(<Link url="https://shopify.com" external />);
-      expect(link).toContainReactComponent(ReactRouterLink, {
+      expect(link).toContainReactComponent('a', {
         target: '_blank',
         rel: 'noopener noreferrer',
       });


ismail-syed (approved):
I can't recall why we didn't have this supported from the beginning. Given how regularly this logic is copy/pasted into apps, I'm glad to see this supported 🤔 

Also, would love to get a changelog note added to `@shopify/react-router`.

packages/react-router/src/components/Link/Link.tsx
```diff
@@ -8,16 +8,20 @@ export interface Props extends ReactRouterLinkProps {
   url: string;
 }
 
-export default function Link({url, external, children, ...rest}: Props) {
-  let target: string | undefined;
-  let rel: string | undefined;
-  if (external) {
-    target = '_blank';
-    rel = 'noopener noreferrer';
+const IS_EXTERNAL_LINK_REGEX = /^(?:[a-z][a-z\d+.-]*:|\/\/)/;
```
atesgoral, line 11: Can we make the regex simpler by just looking for `/^https?:\/\/`? I mean do we really need to support any other protocols?

atesgoral, line 11: Nitpick: Drop the `IS_`

ryanwilsonperkin: @atesgoral I'd expect so yeah, we're looking at using some other schemes that apps can register handlers for like `slack://` which we wouldn't want react-router to try and intercept

atesgoral (approved):

ryanwilsonperkin:
cc @michenly from https://discourse.shopify.io/t/default-behaviour-of-linkcomponent-is-wrong-for-external-links/11940

michenly (changes requested):
I don't think this change belongs in react-router as it make a numbers of UX assumption that is polaris specific. 

This belongs in the default gestalt sacffolding where we can be sure the developer is using polaris.

ryanwilsonperkin:
> I don't think this change belongs in react-router as it make a numbers of UX assumption that is polaris specific.

@michenly Which part of this makes a UX assumption about Polaris? The tests provided for this aren't rendering the URL within Polaris, which I think shows that this can be used (and provides benefit) outside of a Polaris context. That's certainly the benefit that we're shooting for, but I think this can stand on its own as an incremental improvement on the default behaviour of `react-router`

ryanwilsonperkin:
See for example https://github.com/ReactTraining/react-router/issues/1147 where the maintainers of React-Router suggest creating your own wrapper component that performs this behaviour

michenly:
It forces any non relative url to be open in a new tab, which is an UX behaviour specific to the application and does not belong in general tooling. For example, an embedded app created at Shopify with non relative but the link is to Shopify admin, they may want those url to be open with app bridge instead of a link. 

The example you provided said to `creating your own wrapper component that performs this behaviour` which would be specific wrapper to your application.

ryanwilsonperkin:
> It forces any non relative url to be open in a new tab, which is an UX behaviour specific to the application and does not belong in general tooling. For example, an embedded app created at Shopify with non relative url that goes to Shopify admin, they may want those url to be open with app bridge instead of a link.


That's a really good point, the way I've structured this right now actually conflates two things and maybe we could improve it by separating them: the `external` prop and the url test.

What if we said "if the url matches this regex, then its an "external" URL and requires an `<a>` tag, but we _only_ want to include the new tab behaviour _if_ the user specifies the `external` prop.

So here's how the behaviour would break down:

|Scenario|What gets rendered|What happens|
|---|---|---|
|`<Link url="/example" />`|`<ReactRouterLink to="/example" />`|Opens in current tab, redirects without page load to current domain|
|`<Link url="https://example.com" />`|`<a href="https://example.com" />`|Opens in current tab, redirects with page load to external domain|
|`<Link external url="/example" />`|`<ReactRouterLink to="/example" target="_blank" rel="noopener noreferrer" />`|Opens in new tab, redirects without page load to current domain|
|`<Link external url="https://example.com" />`|`<a href="https://example.com" target="_blank" rel="noopener noreferrer" />`|Opens in new tab, redirects with page load to external domain|

What do you think of that approach?

michenly:
What you described are still application behaviours. And those behaviours simply don't belong in an over arching tool because tooling should not dictate application behaviours. Please remember that the entire quilt lib are open source and can be use by 3rd party developers. I have some more internal context which I will share in Discourse.

BPScott:
👋 from the Polaris team.

I'm with @michenly on this. I don't think these behaviours should be encoded into the Link component - whose role is basically "use this instead of react-router's Link because we've decided to rename pass links in a `url` prop instead of a `to` prop". Thus it should be very unopinionated and is not designed as a drop-in thing to use in Polaris's AppProvider's linkComponent prop.


There is however a space for an opinionated thing that can be an a drop in "just use this" for AppProvider's linkComponent as having to repeat that logic in every app is rather tedious.

I like something similar to Michelle's approach in https://github.com/Shopify/quilt/compare/add-polaris-link. I think we should leave `Link` as is, and ship a new component whose goal is to be an opinionated default for dropping into AppProvider's linkComponent.

ryanwilsonperkin:
Just popping in to say I haven't forgotten about this! BFCM/Hackdays just has us all pretty busy so this is in my personal backlog to come back around to when things cool off a bit 😄

BPScott:
I'd like to propose the following breaking change:

`@shopify/react-router`'s Link export should be a boring reexport of `react-router'` link. https://github.com/Shopify/quilt/blob/master/packages/react-router/src/components/Link/Link.tsx should cease to exist

There should be a new export from `@shopify/react-router` - `PolarisLink` that acts as a drop-in opinionated component that can be passed into Polaris AppProvider's linkComponent:

```
import {PolarisLink} from '@shopify/react-router';
import {AppProvider} from '@shopify/polaris';

export function MyAppWrapper({children}) {
  return (
    <AppProvider linkComponent={PolarisLink}>
      {children}
    </AppProvider>
  );
}
```

PolarisLink shall handle mapping the props that [polaris provides](https://github.com/Shopify/polaris-react/blob/00808ec3ab893a39af5ebc9b52a525be8916d227/src/utilities/link/types.ts#L1-L12) into stuff that react-router understands and any other custom shopify opinionated logic .

- mapping the `url` prop to `to`
- if the `external` prop is present then adding ` target = '_blank'` and `rel = 'noopener noreferrer';` to the props
- If the `external` prop is present or the url contains a hostname then use a plain `<a>` tag.

BPScott:
Going to close this.

`@shopify/react-router` reexporting everything from react-router-dom is confusing to users. The [Polaris docs](https://polaris.shopify.com/components/app-provider#using-link-component) show that it is trivial to implement a Link that handles Polaris conventions, external links and target="_blank" stuff all in one go.

I think the cleanest approach is to outright remove Link in #2422, not provide a replacement and direct people to Polaris's docs for instructions on how have a good Link component.

PR #1675, opened 2020-11-16
Create Phone Field Data Package

ashwinnarayanan2001:
## Description
The PhoneField component on Polaris requires country information such as flag, country name, population, and phone number formatting (https://github.com/Shopify/polaris-react/pull/3320). This package has been built in order to efficiently provide users of the component with the data when they require it. Users can simply provide an import statement to provide all of the data (the most common usage). Another usage of this package is across other teams in Shopify. 

This package pulls from other NPM packages, and provides an array of country data inside of index.json. 

In order to tophat this package, pull down this branch and go into the 'Shopify/quilt/packages/phone-field-data' directory. Once inside of the directory, enter a 'node src/build.js', and inside of index.json, all of the country data can be shown.

## Checklist


atesgoral:
Thank you! This looks like a great feature to have. But unfortunately we're no longer accepting new packages into Quilt. Please see the note at the top of the [main README](https://github.com/Shopify/quilt):

> ⚠️ Over the past few years, this repo has become a dumping ground for a variety of packages unrelated to the core problems Quilt aims to solve. For this reason, we are no longer accepting new packages.

It will be best to publish this as a standalone package in a new repo.

PR #1679, opened 2020-11-18
[dynamicList] add ability to add multiple items with factory

oluwatimio:
## Description

Adds ability to add multiple items dynamically and also pass in an argument

## Tophat

https://codesandbox.io/s/fervent-chaum-bogfr?fontsize=14&hidenavigation=1&theme=dark

You should be able to add more than one item through the factory.

You should be able to pass in a `factoryArgument` into addItem which is passed to the factory. This way the factory can be used for something like pagination. There are a lot of possibilities for this.

## Type of change

`react-form` - DynamicList

## Checklist

### DIFF
--- a/packages/react-form/CHANGELOG.md
+++ b/packages/react-form/CHANGELOG.md
@@ -9,6 +9,7 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [0.9.0] - 2020-12-03
 
+- Added new functionality to `useDynamicList`. Added the ability to dynamically add more than one list item and the ability to pass in an argument into the dynamic list factory.
 - Added `useDynamicList` functionality. `useDynamicList` adds the ability to dynamically add and remove list items.
 
 ## [0.8.1] - 2020-10-20
--- a/packages/react-form/src/hooks/list/test/dynamiclist.test.tsx
+++ b/packages/react-form/src/hooks/list/test/dynamiclist.test.tsx
@@ -7,12 +7,11 @@ import {FieldListConfig} from '../baselist';
 
 import {Variant, randomVariants, clickEvent, TextField} from './utils';
 
-const factory = () => {
-  return {price: '', optionName: '', optionValue: ''};
-};
-
 describe('useDynamicList', () => {
   describe('add and remove fields', () => {
+    const factory = () => {
+      return {price: '', optionName: '', optionValue: ''};
+    };
     function DynamicListComponent(config: FieldListConfig<Variant>) {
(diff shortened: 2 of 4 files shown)


shameelabd (approved):
This is sick! Looks really good to me. Will be super useful in our use-case where we have a load more button that loads the remaining results of a list via an api call. Can't wait to refactor it!

Thanks a lot for this, I really appreciate it!

The only potential thing I could bring up is the naming. It could be a bit weird that when calling a function called `addItem` that multiple items get added to the list (based on the factory that was implemented). Not sure if im looking to much into this and its a non-issue or if that makes sense lol

oluwatimio:
> This is sick! Looks really good to me. Will be super useful in our use-case where we have a load more button that loads the remaining results of a list via an api call. Can't wait to refactor it!
> 
> Thanks a lot for this, I really appreciate it!
> 
> The only potential thing I could bring up is the naming. It could be a bit weird that when calling a function called `addItem` that multiple items get added to the list (based on the factory that was implemented). Not sure if im looking to much into this and its a non-issue or if that makes sense lol

I think it makes sense. I didn't want to overcomplicate it by adding another function `addItems`since the result of `addItem` is really based on the factory. I personally don't think it's an issue for me but i'm curious what other's think

packages/react-form/src/hooks/list/test/dynamiclist.test.tsx
```diff
@@ -101,5 +100,78 @@ describe('useDynamicList', () => {
...
+
+      const wrapper = mount(<DynamicListComponent list={variants} />);
+
+      expect(
+        wrapper.findAll('button', {children: 'Remove Variant'}),
+      ).toHaveLength(0);
+    });
+  });
+
+  describe('add mulitple items with payload', () => {
```
atesgoral, line 115:
```suggestion
  describe('add multiple items with payload', () => {
```

packages/react-form/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [0.9.0] - 2020-12-03
 
+- Added new functionality to `useDynamicList`. Added the ability to dynamically add more than one list item and the ability to pass in an argument into the dynamic list factory.
```
gmcgibbon, line 12: This is listed as already released in 0.9.0, but I think you meant to place it above. I'll move the changelog ☝️ when releasing 0.10.0.

oluwatimio: thanks! @gmcgibbon

PR #1687, opened 2020-11-28
Order asynchronous assets _after_ normal entrypoints

gmcgibbon:
## Description

Orders asynchronous assets after normal entrypoints.

Asset order for requests to a react server via sewing-kit is currently incorrect. The first request works because there are no async assets. Subsequent reloads result in `webpackHotUpdate` not defined in the console because async hot reload assets are being loaded first. This patch makes sure the react server is ordering asset entrypoints correctly, by placing async ones last.

## Type of change

- [x] `@shopify/sewing-kit-koa` Patch: Bug (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/packages/sewing-kit-koa/src/assets.ts
+++ b/packages/sewing-kit-koa/src/assets.ts
@@ -178,7 +178,7 @@ function getAssetsFromManifest(
     return entrypointAssets;
   }
 
-  return [...asyncAssets, ...entrypointAssets];
+  return [...entrypointAssets, ...asyncAssets];
 }
 
 function getAsyncAssetsFromManifest(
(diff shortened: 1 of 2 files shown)


lemonmade (changes requested):
I'm pretty sure this change breaks the production behaviour of async components. We recently had to make the opposite change being done here for Checkout's assets so that async components were able to synchronously resolve on initial render: https://github.com/Shopify/shopify/pull/268492

gmcgibbon:
🤔 It would seem async assets is primarily a production concern, but webpack hot updates are being lumped into this bucket erroneously. Production is most important, but we should be able to group webpack updates into a separate group and make sure they come last.

BPScott:
Very related: https://github.com/Shopify/quilt/pull/1653

PR #1688, opened 2020-11-28
Wait for react server recompile synchronously

gmcgibbon:
## Description

Makes react server wait time synchronous. Without this, the current behaviour renders a page that automatically refreshes with a wait message. This is great for initial loading but for webpack hot reloads that require a full refresh, it confuses webpack. If the server is not ready, we need to wait for it synchronously and hold onto the request so that webpack can get the data it expects back from the server.

As a safety precaution, I've made it wait ~10 seconds and then fail with a 500 error (that would be surfaced by the debug middleware in dev mode) because we don't want it waiting forever.

It is either we make the proxy requests synchronous, or we change our webpack client connection code, and I think this makes more sense. Thoughts?

## Type of change

- [x] `quilt_rails` gem Minor: New feature (non-breaking change which adds functionality)

## Checklist


TheMallen (approved):

TheMallen:
Seems nicer than how it was.

atesgoral (approved):

gmcgibbon:
Was originally on top of https://github.com/Shopify/quilt/pull/1687 but this will need further work. Merging this change by itself for now.

PR #1689, opened 2020-12-02
✨ Add support for getting example postal code format

Bringer128:
## Description

The Address package supports queries to https://country-service.shopifycloud.com/graphql to pull down the list of countries. However a chunk of people who fill out the forms get the postal code format subtly wrong and get validation errors from the backend.

Today one of these validations returns the example zip in the error message from the backend. It would be great if we could include that on the front end and give the user a hint before they enter the value.

## Type of change

- [ ] `address-consts` Minor: New feature (now has a type that maps to the returned value from the API)
- [ ] `address` Minor: New feature (now queries the value from the API)

## Checklist

### DIFF
--- a/packages/address-consts/src/index.ts
+++ b/packages/address-consts/src/index.ts
@@ -77,6 +77,7 @@ export interface Country {
     show: string;
   };
   zones: Zone[];
+  zipExample: string;
 }
 
 export interface ResponseError {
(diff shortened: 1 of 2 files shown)


plentz (approved):
LGTM (I just updated the field to use the non-deprecated one)

Bringer128:
Don't have the time to follow up on this

PR #1702, opened 2020-12-09
Transpile createAsyncQuery from alpaql

9at8:
## Description

We need the babel plugin to transpile alpaql's `createAsyncQuery` to make it work with SSR.

## Type of change

- [x] async: Minor: New feature (non-breaking change which adds functionality)

## Checklist


packages/async/src/babel-plugin.ts
```diff
@@ -2,6 +2,7 @@ import Types from '@babel/types';
 import {NodePath, Binding} from '@babel/traverse';
 
 const DEFAULT_PACKAGES_TO_PROCESS = {
+  '@shopify/alpaql/async': ['createAsyncQuery'],
```
ismail-syed, line 5: Since `alpaql` is an external package to the `quilt` codebase, does it make more sense to add this into sewing-kit via the `packages` option?

michenly: Agree, this took a while to find because I didnt expect @shopify/async to have these implicit connection to other packages. Ideally `react-async` and `@shopify/react-graphql` are move out as well, but since that will be a breaking change, we can leave it for now.

michenly: Although we might have a problem splitting it out now since the logic below is OR
`(state.opts && state.opts.packages) || DEFAULT_PACKAGES_TO_PROCESS`

sk wil have to included all the `DEFAULT_PACKAGES_TO_PROCESS` in order to achieve backward compatibility.
so either we keep the `alpaql` logic here, or we make a breaking change for the babel-plugin options to be additive.

michenly (approved):
While I agree that this implicit dependency to continue, I think a better time for refactoring this separation of concern will be when we do sewing-kit-next

packages/async/src/babel-plugin.ts
9at8: I agree with Michelle https://github.com/Shopify/quilt/pull/1702#pullrequestreview-548682728

Yes, it should probably live in sewing-kit, and probably should be done when we wanna move out react-async and react-graphql as well.

gmcgibbon (approved):
As was already pointed out, we should encourage consumers to include the packages and method names they wish to process, but this should be done in a separate, more breaking change (including `'@shopify/react-graphql': ['createAsyncQueryComponent', 'createAsyncQuery']`).

9at8:
I'm not sure what the PR is waiting for, but I'm good to merge this.

michenly:
@9at8 you've got two ✅ , I say merge away!

PR #1705, opened 2020-12-10
Parallel CI jobs

ismail-syed:
## Description

This PR splits up our CI checks into parallel jobs to boost CI times (shaves of a couple mins). This is also helps with CI reporting errors sooner than later since it's parallel. 

### Before

image

### After

image

### DIFF
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -1,15 +1,46 @@
 name: Node-CI
 
-on:
-  push:
-    branches: [master]
-  pull_request:
-    branches: [master]
+on: push
 
 jobs:
-  node-tests:
+  node-lint:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [10.x, 12.x, 14.x]
 
+    steps:
+      - uses: actions/checkout@v2
+        name: Checkout
+
+      - uses: actions/setup-node@v1
+        name: Use Node.js ${{ matrix.node-version }}
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: Get yarn cache directory
+        id: yarn-cache-get-dir
+        run: echo "::set-output name=dir::$(yarn cache dir)"
+
+      - uses: actions/cache@v2
+        id: yarn-cache
+        name: Restore yarn cache
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/node-ci.yml
```diff
@@ -1,15 +1,10 @@
 name: Node-CI
 
-on:
-  push:
-    branches: [master]
-  pull_request:
-    branches: [master]
+on: push
```
ismail-syed, line 3: This was previously setup just to run CI on pushes to master, not sure why. It now runs CI on all pushes to this repo.

.github/workflows/node-ci.yml
```diff
@@ -39,16 +34,113 @@ jobs:
...
 
-      - name: 🔨 Build
-        run: yarn build --verbose
-
       - name: 💅🏼 Lint
         run: |
           yarn lint
           yarn ci:lint-docs
 
+  node-build:
```
ismail-syed, line 42: I wonder if there's a way to re-use all of this duplication :/

ismail-syed:
Hmm, wonder why these are hanging.

image

.github/workflows/node-ci.yml
ismail-syed: Doesn't seem so - https://github.com/actions/starter-workflows/issues/245. 

I stumbled across the `needs` field while looking through their [docs](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idneeds). Couldn't get it to work properly and it arguably ends up creating more duplication - https://github.com/Shopify/quilt/pull/1705/commits/89013b8e737a1e0dd7d1b0f777ddf67aca589456.

atesgoral: Not yet, it seems: https://github.community/t/reusing-sharing-inheriting-steps-between-jobs-declarations/16851/15

atesgoral:
Grr we have a flaky test that got flakier after the switch to GitHub Actions somehow.

ismail-syed:
Actually, it could be  a Node 14 perf thing /shrug. 

Were we running node 14 in Travis?

image

atesgoral (approved):

PR #1710, opened 2020-12-15
[react-form, react-form-state] Update `fast-deep-equal` to `^3.1.3`

kaelig:
## Description

Keeps packages relying on `fast-deep-equal` ([CHANGELOG](https://github.com/epoberezkin/fast-deep-equal/releases)) up to date with the latest version.

This is useful because I'm introducing a new devDependency in Shopify/web that relies on `fast-deep-equal@3`, and this will prevent having to add this dependency to the list of ignored packages in `failOnDuplicateDependencies` ([as set up in Shopify/web’s `sewing-kit.config.js`](https://github.com/Shopify/web/blob/4b536816071de3db29f31af72dcc47e1a541c7e2/sewing-kit.config.ts#L341-L344)).

## Type of change

- [ ] react-form Patch: dependency update
- [ ] react-form-state Minor: dependency update

## Checklist

### DIFF
--- a/packages/react-form-state/CHANGELOG.md
+++ b/packages/react-form-state/CHANGELOG.md
@@ -5,7 +5,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## [Unreleased] -->
+## [Unreleased]
+
+- Updated `fast-deep-equal` dependency to `^3.1.3` [#1710](https://github.com/Shopify/quilt/pull/1710)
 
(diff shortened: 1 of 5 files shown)


ismail-syed (approved):

BPScott:
This is a major bump of fast-deep-equal, which offers up a breaking change per https://github.com/epoberezkin/fast-deep-equal/releases/tag/v3.0.0-beta.0: **"instances of different classes are now considered unequal"**

Taking a quick look at what we pass into isEqual in quilt libs it looks like we never pass in classes, so sounds like we're fine.

BPScott (approved):

gmcgibbon (approved):

gmcgibbon:
No `/shipit` here, you have to press the button

kaelig:
Awesome, I love pressing buttons! So old school 🥸

PR #1724, opened 2021-01-19
Allow custom id graphql universal provider

vsumner:
## Description

This PR adds the ability to customize the data-id of the serialized apollo cache object. Today it is hardcoded to `apollo` in the case where you might have multiple graphql clients, this would result in collisions. This PR adds an `id` prop that defaults to `apollo` so there is no change in behavior.

## Type of change

- [X] `@shopify/react-graphql-universal-provider` Minor: New feature (non-breaking change which adds functionality)

## Checklist


michenly (approved):
LGTM

PR #1725, opened 2021-01-20
i18n: fix pseudotranslate

dlech:
Using the default options `psueduotranslate` was dropping the last part of a translation string after the last matched token (e.g. `{xyz}`). For strings without replacements, this was causing the entire string to be dropped instead of being added to the `parts` array.

## Description

Fixes (issue #)

## Type of change

## Checklist

### DIFF
--- a/packages/i18n/CHANGELOG.md
+++ b/packages/i18n/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
+- Fixed pseudotranslate leaving out final part.
+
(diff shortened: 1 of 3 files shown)


dlech:
I have signed the CLA now.

dlech:
Added changelog and test.

packages/i18n/CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Fixed
+
+- Fixed pseudotranslate leaving out final part.
```
gmcgibbon, line 12: Next time please add the PR number to the changelog (see below 👇)

gmcgibbon (approved):
CI is broken, but it appears to not conflict with other behaviour/tests. Makes sense to me 👍

PR #1726, opened 2021-01-21
✨ <Html> now accepts `type` for scripts

GoodForOneFare:
## Description
With various bundlers/frameworks providing much improved support for ES modules, it'd be nice if code like:
```ts
<Html scripts={[{path: '/foo.js', type: 'module'}] />
```

generated content like
```
<script src="/foo.js" type="module" />
```

## Type of change
- [x] `@shopify/react-html` Minor
- [x] `@shopify/sewing-kit-koa` Minor

## Checklist

### DIFF
--- a/packages/react-html/src/server/components/Html.tsx
+++ b/packages/react-html/src/server/components/Html.tsx
@@ -16,6 +16,10 @@ export interface Asset {
   integrity?: string;
 }
 
+export interface ScriptAsset extends Asset {
+  type?: 'module' | 'text/javascript';
+}
+
 export interface InlineStyle {
   content: string;
 }
@@ -27,8 +31,8 @@ export interface HtmlProps {
   locale?: string;
   styles?: Asset[];
   inlineStyles?: InlineStyle[];
-  scripts?: Asset[];
-  blockingScripts?: Asset[];
+  scripts?: ScriptAsset[];
+  blockingScripts?: ScriptAsset[];
   preloadAssets?: Asset[];
   headMarkup?: React.ReactNode;
   bodyMarkup?: React.ReactNode;
@@ -108,6 +112,7 @@ export default function Html({
         key={script.path}
         src={script.path}
         integrity={script.integrity}
+        type={script.type}
         crossOrigin="anonymous"
       />
     );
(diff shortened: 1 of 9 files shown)


packages/react-html/src/server/components/Html.tsx
```diff
@@ -16,6 +16,10 @@ export interface Asset {
   integrity?: string;
 }
 
+export interface ScriptAsset extends Asset {
+  type?: 'module' | 'text/javascript';
```
GoodForOneFare, line 20: I was tempted to do this instead:
```
  module?: boolean;
```

The [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#attr-type) say that arbitrary values are accepted, but just result in an unprocessed script.  So not really in the spirit of a explicitly scripty properties.

Interested in your thoughts either way.

lemonmade: I know this just matches up to the HTML spec where type is either a mime type or this magic 'module' string, but it feels like such an awkward API that I'd prefer almost anything else. My initial intuition was to do `type?: 'module' | 'script';`, which is kind of how I've seen these "execution modes" described in other tools like Babel.

As another piece to the puzzle, how do you think we will specify `nomodule` variants for a bundle? I know you are probably not worried about that for your specific case where you just care about a dev mode, but using this feature in a production setting would almost certainly require the use of the `nomodule` pattern for older browsers. Maybe that can help guide us to a nicer API here?

GoodForOneFare: 🤔 That's an interesting wrinkle. I'm on board with no leaking spec details into this API, so how about:

```
type?: 'module' | 'nomodule' | 'script';

// type: 'module'
<script src="/foo.js" type="module" crossorigin="anonymous">

// type: 'nomodule'
<script src="/foo.js" type="text/javascript" nomodule crossorigin="anonymous">

// type: 'script' / default behaviour
<script src="./foo.js" type="text/javascript" crossorigin="anonymous">
```

lemonmade: I like this API 👍

GoodForOneFare:
@Shopify/web-foundations I've updated the API based on my discussion with Chris, and added some `nomodule` tests.  Please take another 👀

atesgoral (approved):

PR #1728, opened 2021-01-21
Adds January 2021 API version

milancermak:
## Description

Adds support for using the latest API version.


gmcgibbon (commented):
Thanks, can you please sign the CLA and update `packages/koa-shopify-webhooks/src/register.ts` too? They will also both need changelogs.

milancermak:
I've updated the `register.ts` as well and signed the CLA. I've also update the appropriate changelogs, thought not 100% sure if I did it right. Please let me know if I need to change or update something.

BPScott:
Fixed by #2001

PR #1732, opened 2021-01-29
call useDomEffect cleanup and filter metas

sylvhama:
## Description

This PR fixes two issues:
- the `useDomEffect` cleanup function was never called. So for example imagine you render a `<Meta />` for a specific page. If you unmount the component it won't remove the meta tag from the DOM. The logic was always there (see return statement of `addDescriptor`) but it was just not called.
- It renders the last `meta` tag if some have same `name` or `property`. Imagine you have a website with a global description but you want to render a different one within a very specific page. This is similar to how `useTitle` works at the moment. https://github.com/Shopify/polaris-styleguide/pull/3776 is a real life example where this would be useful.

## Type of change

- [x] `react-html` Patch: Bug (non-breaking change which fixes an issue)
- [x] `react-html` Minor: New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/packages/react-html/CHANGELOG.md
+++ b/packages/react-html/CHANGELOG.md
@@ -7,6 +7,9 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+- Minor: Only render the last meta tag if they have same name or property []()
+- Minor: Cleanup tags from DOM when unmounting []()
+
 ### Added
 
 - Better ES module support via `<Html />`'s `script` / `blockingScripts` accepting a `type` property
--- a/packages/react-html/CHANGELOG.md
+++ b/packages/react-html/CHANGELOG.md
@@ -7,8 +7,13 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
-- Minor: Only render the last meta tag if they have same name or property [#1732](https://github.com/Shopify/quilt/pull/1732)
-- Patch: Cleanup tags from DOM when unmounting [#1732](https://github.com/Shopify/quilt/pull/1732)
+### Changed
+
+- Only render the last meta tag if they have same name or property [#1732](https://github.com/Shopify/quilt/pull/1732)
+
+### Fixed
+
+- Cleanup tags from DOM when unmounting [#1732](https://github.com/Shopify/quilt/pull/1732)
 
 ### Added
 
--- a/packages/react-html/CHANGELOG.md
+++ b/packages/react-html/CHANGELOG.md
@@ -7,8 +7,8 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
-- Minor: Only render the last meta tag if they have same name or property []()
-- Minor: Cleanup tags from DOM when unmounting []()
+- Minor: Only render the last meta tag if they have same name or property [#1732](https://github.com/Shopify/quilt/pull/1732)
+- Patch: Cleanup tags from DOM when unmounting [#1732](https://github.com/Shopify/quilt/pull/1732)
 
 ### Added
 
--- a/packages/react-html/src/hooks.ts
+++ b/packages/react-html/src/hooks.ts
@@ -10,9 +10,7 @@ export function useDomEffect(
   inputs: unknown[] = [],
 ) {
   const manager = useContext(HtmlContext);
-  const effect = () => {
-    perform(manager);
-  };
+  const effect = () => perform(manager);
 
   useServerEffect(effect, manager.effect);
   useEffect(effect, [manager, ...inputs]);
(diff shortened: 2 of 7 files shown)


packages/react-html/src/hooks.ts
```diff
@@ -10,9 +10,7 @@ export function useDomEffect(
   inputs: unknown[] = [],
 ) {
   const manager = useContext(HtmlContext);
-  const effect = () => {
-    perform(manager);
-  };
+  const effect = () => perform(manager);
```
sylvhama, line 13: The cleanup function will be correctly returned now.

packages/react-html/src/manager.ts
```diff
@@ -41,7 +42,7 @@ export class HtmlManager {
 
     return {
       title: lastTitle && lastTitle.title,
-      metas: this.metas,
+      metas: uniqby(uniqby(this.metas.reverse(), 'name'), 'property'),
```
sylvhama, line 45: I am opened to suggestions if we can improve the logic here.
I am reversing the array to make sure we keep the latest entry in the array.

sylvhama: I made a [codesandbox](https://codesandbox.io/s/unruffled-moon-hihfq?file=/src/index.js) to illustrate the logic here.

packages/react-html/src/manager.ts
```diff
@@ -41,7 +42,7 @@ export class HtmlManager {
 
     return {
       title: lastTitle && lastTitle.title,
-      metas: this.metas,
+      metas: uniqwith(this.metas.reverse(), metaComparator),
```
sylvhama, line 45: I am opened to suggestions if we can improve the logic here.
I am reversing the array to make sure we keep the latest entry in the array. I made a [codesandbox](https://codesandbox.io/s/unruffled-moon-hihfq?file=/src/index.js) to illustrate the logic here.

packages/react-html/src/server/components/tests/Html.test.tsx
```diff
@@ -264,8 +264,34 @@ describe('<Html />', () => {
         .filter(meta => meta.data(MANAGED_ATTRIBUTE));
 
       expect(metas).toHaveLength(2);
-      expect(metas[0]).toHaveReactProps(metaOne);
-      expect(metas[1]).toHaveReactProps(metaTwo);
+      expect(metas[0]).toHaveReactProps(metaTwo);
+      expect(metas[1]).toHaveReactProps(metaOne);
```
sylvhama, line 268: Order is reversed now, this won't have any impact on SEO or performance.

packages/react-html/CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
+- Minor: Only render the last meta tag if they have same name or property [#1732](https://github.com/Shopify/quilt/pull/1732)
+- Patch: Cleanup tags from DOM when unmounting [#1732](https://github.com/Shopify/quilt/pull/1732)
+
```
atesgoral, line 12:
```suggestion
### Changed

- Only render the last meta tag if they have same name or property [#1732](https://github.com/Shopify/quilt/pull/1732)

### Fixed

- Cleanup tags from DOM when unmounting [#1732](https://github.com/Shopify/quilt/pull/1732)

```

packages/react-html/src/server/components/tests/Html.test.tsx
```diff
@@ -264,8 +264,34 @@ describe('<Html />', () => {
...
+      const pageDescription = 'page description';
+
+      const latestDescription = {name: 'desciption', content: pageDescription};
+      const latestOgDescription = {
+        property: 'og:description',
+        content: pageDescription,
+      };
+
+      const manager = new HtmlManager();
+      manager.addMeta({name: 'desciption', content: globalDescription});
```
atesgoral, line 282:
```suggestion
      manager.addMeta({name: 'description', content: globalDescription});
```

atesgoral:
While it does the job in a pragmatic way, I wonder if we could roll our own small uniqeness function instead of bringing in lodash.uniq, which is adding some K to our bundle: https://bundlephobia.com/result?p=lodash.uniq@4.5.0

packages/react-html/src/manager.ts
atesgoral: Reversing makes a lot of sense

atesgoral (approved):

packages/react-html/src/manager.ts
```diff
@@ -130,3 +130,30 @@ export class HtmlManager {
     }
   }
 }
+
+function filterMetas(metas: React.HTMLProps<HTMLMetaElement>[]) {
```
michenly, line 134: nitpik: maybe call this `removeDuplicate` ?

also since the order here is important, maybe reverse it again before return so order maintain the same.

michenly (approved):
Just a bit of nitpick otherwise 👍

sylvhama:
Thanks @atesgoral and @michenly for your feedback, I learned few stuff on this PR, it was nice 😃

PR #1733, opened 2021-01-29
[React Form] Correct validation examples used regarding title length being longer than three characters

hnaeem96:
## Description

This PR will edit the validation examples used regarding the title length in the `react-form` to correctly match the validation description.

## Type of change

## Checklist

### DIFF
--- a/packages/react-form/README.md
+++ b/packages/react-form/README.md
@@ -362,7 +362,7 @@ You can also pass a more complex configuration object specifying a validation fu
 const field = useField({
   value: someRemoteData.title,
   validates: title => {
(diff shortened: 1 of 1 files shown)


packages/react-form/README.md
```diff
@@ -362,7 +362,7 @@ You can also pass a more complex configuration object specifying a validation fu
 const field = useField({
   value: someRemoteData.title,
   validates: title => {
-    if (title.length > 3) {
+    if (title.length <= 3) {
```
dahukish, line 365: good catch

PR #1736, opened 2021-02-04
move removeDuplicate call

sylvhama:
## Description

This PR is a follow-up of https://github.com/Shopify/quilt/pull/1732

It seems that removing duplicated metas in the state directly causes weird effects. I am not really sure what's happening, I have made this little [demo](https://codesandbox.io/s/gracious-bash-lb1jd?file=/src/App.js) and you can tell that the latest description added by `Child` is ignored. In order to fix that I created a new utility that I re-used in both places where we render HTML meta tags.

## 🎩 instructions

So i published a new beta: `10.2.1-react-html-fix-beta.1`

- [Demo with current version](https://codesandbox.io/s/gracious-bash-lb1jd?file=/src/App.js)
- [Demo with beta version](https://codesandbox.io/s/stupefied-lamport-se75k?file=/src/App.js)

It seems to work as expected.

1. For each demo open the preview in a new window (top right button);
2. Open your dev tools and look at the `<head>` content
3. Click on the button, you should see the Child description when Child is rendered (only with beta version)

## Type of change

- [x] `react-html` Patch: Bug (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/packages/react-html/src/components/HtmlUpdater.tsx
+++ b/packages/react-html/src/components/HtmlUpdater.tsx
@@ -3,7 +3,7 @@ import {HydrationTracker} from '@shopify/react-hydrate';
 
 import {State} from '../manager';
 import {useClientDomEffect} from '../hooks';
-import {MANAGED_ATTRIBUTE} from '../utilities';
+import {MANAGED_ATTRIBUTE, removeDuplicate} from '../utilities';
 
 export function HtmlUpdater() {
   const queuedUpdate = React.useRef<number | null>(null);
@@ -43,7 +43,7 @@ function updateOnClient(state: State) {
 
   const fragment = document.createDocumentFragment();
 
-  updateElement('meta', metas, fragment);
+  updateElement('meta', removeDuplicate(metas), fragment);
   updateElement('link', links, fragment);
   updateElement('style', inlineStyles, fragment);
 
--- a/packages/react-html/src/components/tests/HtmlUpdater.test.tsx
+++ b/packages/react-html/src/components/tests/HtmlUpdater.test.tsx
@@ -142,4 +142,38 @@ describe('<Provider />', () => {
     expect(metaElement).toBe(allMetas[0]);
     expect(allMetas[1].getAttribute('content')).toBe(metaTwo.content);
   });
+
+  it('only keeps the last added meta based on name or property', () => {
+    const globalDescription = 'global description';
+    const pageDescription = 'page description';
+
+    const latestDescription = {name: 'description', content: pageDescription};
+    const latestOgDescription = {
+      property: 'og:description',
+      content: pageDescription,
+    };
+
+    const manager = new HtmlManager();
+    mountWithManager(<HtmlUpdater />, manager);
+
+    manager.addMeta({name: 'description', content: globalDescription});
+    manager.addMeta({property: 'og:description', content: globalDescription});
+    manager.addMeta(latestDescription);
+    manager.addMeta(latestOgDescription);
+    animationFrame.runFrame();
+
+    const allMetas = Array.from(
+      document.querySelectorAll(`meta[${MANAGED_ATTRIBUTE}]`),
+    );
+
+    expect(allMetas).toHaveLength(2);
+    expect(allMetas[0].getAttribute('name')).toBe(latestDescription.name);
+    expect(allMetas[0].getAttribute('content')).toBe(latestDescription.content);
+    expect(allMetas[1].getAttribute('property')).toBe(
+      latestOgDescription.property,
+    );
+    expect(allMetas[1].getAttribute('content')).toBe(
+      latestOgDescription.content,
+    );
+  });
 });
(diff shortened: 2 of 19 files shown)


packages/react-html/src/components/tests/HtmlUpdater.test.tsx
```diff
@@ -142,4 +142,38 @@ describe('<Provider />', () => {
     expect(metaElement).toBe(allMetas[0]);
     expect(allMetas[1].getAttribute('content')).toBe(metaTwo.content);
   });
+
+  it('only keeps the last added meta based on name or property', () => {
```
sylvhama, line 146: In my previous PR I forgot to test that part... if you run this test in master it will fail.

packages/react-html/src/components/HtmlUpdater.tsx
```diff
@@ -43,7 +43,7 @@ function updateOnClient(state: State) {
 
   const fragment = document.createDocumentFragment();
 
-  updateElement('meta', metas, fragment);
+  updateElement('meta', removeDuplicate(metas), fragment);
```
dahukish, line 46: Why are duplicate metas being added in the first place? Could we not use a Set or Map to get around this?

michenly:
can you provided a test with beta release to show case this is working, I really don't want to publish a major package just for test and figure out it doesn't work.

packages/react-html/src/utilities.ts
```diff
@@ -53,3 +53,32 @@ export function showPage(): Promise<void> {
       })
     : Promise.resolve();
 }
+
+export function removeDuplicate(metas: React.HTMLProps<HTMLMetaElement>[]) {
+  const names = new Set();
```
dahukish, line 58: ❤️

packages/react-html/src/components/tests/HtmlUpdater.test.tsx
```diff
@@ -142,4 +142,38 @@ describe('<Provider />', () => {
...
+    const latestDescription = {name: 'description', content: pageDescription};
+    const latestOgDescription = {
+      property: 'og:description',
+      content: pageDescription,
+    };
+
+    const manager = new HtmlManager();
+    mountWithManager(<HtmlUpdater />, manager);
+
+    manager.addMeta({name: 'description', content: globalDescription});
```
dahukish, line 159: Is there any reason why `HtmlManager` can't handle the dedupe instead of doing it afterwards?

packages/react-html/src/components/HtmlUpdater.tsx
sylvhama: We need to keep track of every metas added.

Imagine I have a `Layout` component that wraps my app and that will never unmount. It contains default meta for the description. Sometimes I will render a child component (e.g. AboutUsPage) that will want to override the description. If I go back to the homepage then the description added by `Layout`  should be back.

See https://github.com/Shopify/polaris-styleguide/pull/3776 for a real example

packages/react-html/src/components/tests/HtmlUpdater.test.tsx
sylvhama: I tried that in my previous PR, see https://github.com/Shopify/quilt/pull/1732
`Html` and `HtmlUpdater` will both uses the `manager.state` so I originally thought it would be a nice idea to filter duplicates here but it does not work as expected.

sylvhama:
> can you provided a test with beta release to show case this is working, I really don't want to publish a major package just for test and figure out it doesn't work.

I did not know we could do that... sorry again. I guess only admin can do it? I can test as soon as possible (tonight or tomorrow morning EST).

michenly:
Publishing beta package process would be similar to [this](https://github.com/Shopify/sewing-kit/blob/63f83a8b179d92a64d3f4c4e2128f081bb50c0ea/docs/contributing/README.md#step-1---publish-a-betas)

Since you are pushing into this branch and using your own shipit stack you shouldn't need admin permission. If this doesnt work let me know and I will figure it out with you.

sylvhama:
## 🎩 instructions

So i published a new beta: `10.2.1-react-html-fix-beta.1`

- [Demo with current version](https://codesandbox.io/s/gracious-bash-lb1jd?file=/src/App.js)
- [Demo with beta version](https://codesandbox.io/s/stupefied-lamport-se75k?file=/src/App.js)

It seems to work as expected.

1. For each demo open the preview in a new window (top right button);
2. Open your dev tools and look at the `<head>` content
3. Click on the button, you should see the Child description when Child is rendered

atesgoral:
🎩 checks out. Thanks for the demos!

atesgoral (approved):

sylvhama:
I have never released a beta before, I noticed it changed the package.json of a lot of packages, can I merge all of this?

BPScott (approved):
Thanks for the demos! Gave the codepens a poke and they look good.


Per request from @ismail-syed: Can you revert the package version changes prior to merging? That'll result in a cleaner main branch.

Once that's done merge away!

PR #1738, opened 2021-02-09
✨ add options to disabled some of the links in GraphQLUniversalProvider

michenly:
## Description

1. breaking change: changing the header from `X-Request-ID` to `X-Initiated-By-Request-ID`, this is what web uses and actually does make more sense.
2. add in two options: quiltRails and addRequestId, both can be use to disabled certain link if not needed. Planning on using `quiltRails=false` in web to turn off csrf header. 

## Type of change

## Checklist


dahukish:
@michenly This should be a major change then if breaking no?

TheMallen (approved):

Issue #1740, opened 2021-02-11
@shopify/react-i18n doesn't set react-dom as a peer dependency


jasonwilliams:
Hello!

We've been getting these errors in our installs.
I think `react-dom` needs to be added here:
https://github.com/Shopify/quilt/blob/master/packages/react-i18n/package.json#L54

```
➤ YN0000: @shopify/react-i18n@npm:5.1.4 [42b1e] doesn't provide react-dom, breaking the following requirements:
➤ YN0000: @shopify/react-effect@npm:3.3.0 [ad6a5] → >=16.8.0 <17.0.0 ✘
```

gmcgibbon:
The thing that references `react-dom` is a test: `src/test/e2e/server.test.tsx`.

Something might be wrong with how we build the package. Relaxing the React constraint makes sense to me.

 cc @Shopify/web-foundations

PR #1742, opened 2021-02-12
[React I18n] Allow React 16+

gmcgibbon:
## Description

https://github.com/Shopify/quilt/issues/1740

Relaxes react constraint for `react-i18n`. Is this safe, or do we not support React 17 yet?

## Type of change

## Checklist

### DIFF
--- a/packages/react-i18n/package.json
+++ b/packages/react-i18n/package.json
@@ -51,7 +51,7 @@
   },
   "sideEffects": false,
   "peerDependencies": {
-    "react": "^16.0.0"
+    "react": ">= 16.0.0"
   },
   "optionalDependencies": {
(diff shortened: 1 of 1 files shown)


packages/react-i18n/package.json
```diff
@@ -51,7 +51,7 @@
   },
   "sideEffects": false,
   "peerDependencies": {
-    "react": "^16.0.0"
+    "react": ">= 16.0.0"
```
dahukish, line 54: I think if we lock in the ranges we should be fine. Web usually is the main focus of what we use and support. Re: https://github.com/Shopify/quilt/issues/1740

michenly:
We dont officially support 17 yet. Closing this PR for now.

PR #1743, opened 2021-02-12
Fix invalid peer dep range on magic-entries-webpack-plugin

gmcgibbon:
## Description

I think the version constraint on the `webpack` peer dependency of `magic-entries-webpack-plugin` is incorrect. According to https://semver.npmjs.com/, there should be no comma.

## Type of change

- [x] `@shopify/magic-entries-webpack-plugin` Patch: Bug (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/packages/magic-entries-webpack-plugin/package.json
+++ b/packages/magic-entries-webpack-plugin/package.json
@@ -29,7 +29,7 @@
     "tslib": "^1.14.1"
   },
   "peerDependencies": {
-    "webpack": ">4.25.1, <=4.5.0"
+    "webpack": "^4.25.1"
   },
   "files": [
     "build/*",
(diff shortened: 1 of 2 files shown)


packages/magic-entries-webpack-plugin/package.json
```diff
@@ -29,7 +29,7 @@
     "tslib": "^1.14.1"
   },
   "peerDependencies": {
-    "webpack": ">4.25.1, <=4.5.0"
+    "webpack": ">4.25.1 <=4.5.0"
```
vsumner, line 32: I think this needs to be `^4.25.1` The current webpack version on sewing-kit is `4.44.1`.
According to the tool you pointed out. https://semver.npmjs.com/ `^4.25.1`  will match the versions that are supported.

gmcgibbon: Hm, I think you're right. It doesn't like the range.

vsumner (approved):

PR #1748, opened 2021-02-18
Update docs and URLs to reference main

gmcgibbon:
## Description

Docs and URLs update to mention main instead of master.

## Type of change

All packages. I believe we can run a patch release for each after this is merged and treat it similarly to dependency bumps.

## Checklist

- [ ] ~I have added a changelog entry, prefixed by the type of change noted above (Documentation fix and Test update does not need a changelog as we do not publish new version)~


atesgoral (approved):
Fast-scrolled through.

PR #1752, opened 2021-02-23
Improve quilt error messages

gmcgibbon:
## Description

Improves error messages for react renderable errors.

## Type of change

`quilt_rails` gem

## Checklist

- [ ] ~I have added a changelog entry, prefixed by the type of change noted above (Documentation fix and Test update does not need a changelog as we do not publish new version)~


michenly (approved):
LGTM!

Issue #1756, opened 2021-02-25
Add basename prop to @shopify/react-router Router


ryanwilsonperkin:
## Overview

Add support for the `basename` prop from React Router https://reactrouter.com/web/api/BrowserRouter/basename-string

## Type

- [x] New feature
- ~[ ] Changes to existing features~

## Motivation

When migrating existing Rails apps to a modern stack using Sewing Kit and Quilt, it's often necessary to run the new React routes under a particular path prefix for a period of time while updating views from being served by Rails to being rendered by React.

In several services, we pick a path prefix (eg. `/next`) and nest all of the React routes under that prefix. This results in two issues within the React app:
1. All routes need to include this prefix (eg. `/next/create`, `/next/view`)
2. When rendering a `Link` we need to apply additional steps to see if the target URL is served by one of our Rails routes (ie. no leading `/next`) and render a formatted `<a>` tag instead.

The `basename` prop in React-Router seems to tackle both of these issues. It would only need to be specified once in the props of the Router component, and it would allow our Links to automatically be passed to the browser to resolve as a new GET request rather than resolving the route change using the History API.

PR #1757, opened 2021-02-25
✨ Add basename prop to Router

ryanwilsonperkin:
## Description

Fixes (issue #1756)

Support the `basename` prop from React Router to make it easier to support mixed React and server routes.

## Type of change

## Checklist


packages/react-router/src/components/Router/test/Router.test.tsx
```diff
@@ -40,6 +47,14 @@ describe('Router', () => {
     expect(wrapper).toContainReactComponent(StaticRouter, {location});
   });
 
+  it('mounts a StaticRouter on the server with basename if provided', () => {
+    isClient.mockReturnValue(false);
+    const basename = '/next';
+    const wrapper = mount(<Router basename={basename} />);
```
dahukish, line 53: Need to pass location here: https://github.com/Shopify/quilt/pull/1757/checks?check_run_id=1981190998#step:8:3421

dahukish (approved):
LGTM! :shipit:

PR #1758, opened 2021-02-25
🐛 ✨ [react-i18n][i18n.unformatNumber] handle negative numbers

danloomer:
## Description

Fixes bug with `i18n.unformatNumber()` stripping the negative sign `-` from negative numbers.

Exposes `i18n.numberSymbols()` as a public method.

## Type of change

- [ ] Patch: Bug (non-breaking change which fixes an issue)
- [x] `react-i18n` Minor: New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/packages/react-i18n/src/i18n.ts
+++ b/packages/react-i18n/src/i18n.ts
@@ -371,6 +371,23 @@ export class I18n {
     return Boolean(easternNameOrderFormatter);
   }
 
+  @memoize()
+  numberSymbols() {
+    const formattedNumber = this.formatNumber(123456.7, {
+      maximumFractionDigits: 1,
+      minimumFractionDigits: 1,
+    });
+    let thousandSymbol;
+    let decimalSymbol;
+    for (const char of formattedNumber) {
+      if (isNaN(parseInt(char, 10))) {
+        if (thousandSymbol) decimalSymbol = char;
+        else thousandSymbol = char;
+      }
+    }
+    return {thousandSymbol, decimalSymbol};
+  }
+
   private formatCurrencyAuto(
     amount: number,
     options: Intl.NumberFormatOptions = {},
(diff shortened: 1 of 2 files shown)


packages/react-i18n/src/i18n.ts
```diff
@@ -615,8 +615,11 @@ function normalizedNumber(
     .substring(lastDecimalIndex + 1)
     .replace(nonDigits, '');
 
+  const isNegative = input.trim().startsWith('-');
```
cejaekl, line 618: I'm not sure how much we care about this, but some locales may use a different character as their "negative sign".  Possibilities include:
  - U+002D `hyphen-minus`
  - U+2212 `minus`
  - U+FE63 `small hyphen-minus`
  - U+FF0D `full-width hyphen-minus`

Note, also, that some locales (e.g. `ja-JP`) may use either narrow or wide versions of the minus sign.

cejaekl (approved):

packages/react-i18n/src/i18n.ts
danloomer: Oh interesting, I'll experiment with that a bit.

Regarding `ja-JP` and the narrow/wide characters, I don't think that `unformatNumber` handles wide versions of 0-9 since they don't fall into the non-digit `/\D/g` regex. [Example](https://flems.io/#0=N4IgZglgNgpgziAXAbVAOwIYFsZJAOgAsAXLKEAGhAGMB7NYmBvEAXwvW10QICsEqdBk2J4ADgAoA5ICR-wAj-U-ACcYYqBmowJAegA6ugCLaA5hQAEUqQEorutJKkAmAAyKVajVr2GT5yzcoQOBhYamIIegQeAEZERwA2Ng4QTBw8fGo4ARp6RmYeNgBdViA)

Along the same lines I was reading this number formatting guide from MS: https://docs.microsoft.com/en-us/globalization/locale/number-formatting

> The way negative numbers are displayed.
The negative sign can be used at the beginning of the number, but it can also be used at the end of the number. Alternatively, the number can be displayed with parentheses around it or even in a color such as red. Thus a negative five hundred and twenty-seven could be displayed as:  
-527
527-
(527)
[527]

It mentions how the negative sign could be at the start or end of the string as well as different ways of representing negative numbers.. but I don't feel like it wouldn't be practical (or reasonable maybe) to handle all those.

Thoughts?

cejaekl: I'm a big fan of incremental improvements; what you have here now is better than what we had before, so it's worth shipping (assuming @ismail-syed is OK with the style).  

Personally, I'd be a fan of circling back with a second PR to handle alternative minus signs at the beginning of the number.  
IMHO postfix minus signs, and parentheses to indicate negativity, are too esoteric and not worth supporting.
But others may have other opinions.  :)

danloomer: Sounds good to me! Thanks @cejaekl

BPScott (approved):
Fixup the failing test, but other than that, looks good to me :shipit:

PR #1759, opened 2021-03-01
Mandatory extension support for builds

ismail-syed:
## Description

This PR update our build system to support [mandatory file extensions](https://nodejs.org/api/esm.html#esm_mandatory_file_extensions). While exploring snowpack and swc, we hit issues where those respective bundlers couldn't parts our `mjs` and `esnext` builds. This PR fixes that. 

This PR fixes the following errors we get when running quilt packages in a webpack 5 react app: 
```
Module not found: Error: Can't resolve './build/esm/index' in 'node_modules/@shopify/react-form'"
```

## 🎩 

- `yarn run clean && yarn run build` to run a fresh build
- Peek into a `esm` build to verify the `.mjs` extensions (eg. `packages/react-html/build/esm/common.mjs`)
image

- Peek into an `esnext` build to verify the `.esnext` extensions (eg. `packages/react-html/build/esnext/common.esnext`)
image

- Peek into a `cjs` build to verify the relative `.js` extensions  (eg. `quilt/packages/react-html/build/cjs/common.js`)
image

## Checklist


BPScott (approved):
Haven't thumbed through every package but it looks like sk is now indeed providing full paths!

PR #1762, opened 2021-03-09
🐛Fix behaviour of makeCleanAfterSubmit

### LINKED ISSUE #1717: useForm "undirty" itself after a failed submission when makeCleanAfterSubmit is true.
## Overview

The [`useForm`](https://github.com/Shopify/quilt/tree/a7594add91ec3fc3ddc73ec12865ba71c3d863f6/packages/react-form#useform) document state that:

> makeCleanAfterSubmit, A boolean flag (defaults to false) indicating whether the form should "undirty" itself after a **successful submission**. ...

I suppose that the return value of the `onSubmit` indicate whether or not the submission is successful. So when it return a `{ status: 'sucess' }` the subsmission is successful , and `{ status : 'fail', errors: [...] }` mean  it failed.

The problem is when `makeCleanAfterSubmit` is set to `true`, the failed submission also "undirty" the form itself.

### Steps to reproduce

* [makeCleanAfterSubmit: false](https://codesandbox.io/s/shopifyreact-form-2-forked-43mr4?file=/src/MyForm.tsx)
    1. fill in the name field with "error" text then submit the form
    2. the "error message" is shown 

* [makeCleanAfterSubmit: true](https://codesandbox.io/s/shopifyreact-form-2-forked-d8cyv?file=/src/MyForm.tsx)
    1. fill in the name field with "error" text then submit the form
    2. the "error message" is **not** shown (its supposed to be shown)

### Potential Solution

Either my assumption about "successful submission" is wrong or `makeCleanAfterSubmit` logic is not correct. So either:

1. Update the document to make it clearer.
2. Maybe we should return after `setErrors(result.errors); `

https://github.com/Shopify/quilt/blob/a7594add91ec3fc3ddc73ec12865ba71c3d863f6/packages/react-form/src/hooks/submit.ts#L57-L65

### Scope

- Package: react-form

rox163:
## Description

Fixes #1717
Fixing the code so that `makeCleanAfterSubmit`, if true, should only "undirty" the form after a successful submission, not on a failed submission. 
Update the tests to correctly test this condition. 

## Type of change

## Checklist

### DIFF
--- a/packages/react-form/CHANGELOG.md
+++ b/packages/react-form/CHANGELOG.md
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## [Unreleased] -->
+## [Unreleased]
+
+### Fixed
+
+- Fixed a bug in `makeCleanAfterSubmit` for `useForm` [#1762](https://github.com/Shopify/quilt/pull/1762)
 
 ## [0.12.1] - 2021-03-30
 
--- a/packages/react-form/src/hooks/submit.ts
+++ b/packages/react-form/src/hooks/submit.ts
@@ -58,10 +58,9 @@ export function useSubmit<T extends FieldBag>(
         setErrors(result.errors);
       } else {
         setSubmitErrors([]);
-      }
-
-      if (makeCleanAfterSubmit) {
-        makeCleanFields(fields);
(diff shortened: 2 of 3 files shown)


dleroux (approved):

vsumner (approved):

rox163:
@vsumner Any ideas why one o f the node tests either get stuck forever or randomly fail? been trying to get this all green all day

vsumner:
> @vsumner Any ideas why one o f the node tests either get stuck forever or randomly fail? been trying to get this all green all day

I'm not sure. I have seen it happen, but I haven't dug into why yet.

Issue #1765, opened 2021-03-11
[react-form] enable sorting of Dynamic list fields


jonarnaldo:
## Overview
Currently, if a user wants to rearrange a list they must manually remove items and/or add items.  I would like to propose a `moveItem` method which allow an item to be moved from `original index` to `new index` which would perhaps involve calling `removeItem` which takes the current index as an argument and a new `addItemAtIndex` which would take the desired index as an argument.  Additionally, moving an item should also trigger a 'dirty' state in the form.

Adding this functionality would also play nicely with UI packages like https://github.com/clauderic/react-sortable-hoc, where `moveItem` could be called within `onSortEnd`, for example.  
...

## Type

- [x] New feature
- [ ] Changes to existing features

## Motivation

> What inspired this feature request? What problems were you facing,
> or what else makes you think this should be in `quilt`?

Currently working on https://vault.shopify.io/projects/16738 which uses [dynamic lists](https://github.com/Shopify/quilt/tree/main/packages/react-form#using-usedynamiclist) to display a series of items a user has added.  Updated UX Designs require the ability for user to sort items via drag handles.

Additionally, I think this would contribute to the overall user experience of any application that requires dynamic sortable lists as well as validation.

---

## Checklist

- [x] Please delete the labels section before submitting your issue
- [x] I have described this issue in a way that is actionable (if possible)

dahukish:
Thanks @jonarnaldo! It looks like `moveItem` is already part of the api:
```
const {fields, addItem, removeItem, moveItem} = useDynamicList(
  [{cardNumber: '4242 4242 4242 4242', cvv: '000'}],
  emptyCardFactory,
);
```

dahukish:
Closing as this has been merged.

PR #1785, opened 2021-03-13
[React Form] Enable reordering of Dynamic Lists 

jonarnaldo:
## Description
This PR will enable sorting of [dynamic lists](https://www.npmjs.com/package/@shopify/react-form#usedynamiclist) via an additional `moveItem` method

Fixes (issue #1765)

## Type of change

## Checklist

### DIFF
--- a/packages/react-form/CHANGELOG.md
+++ b/packages/react-form/CHANGELOG.md
@@ -9,7 +9,7 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ### Added
 
-- Add ability to sort dynamic lists [#1785](https://github.com/Shopify/quilt/pull/1785)
+- Add ability to reorder dynamic lists [#1785](https://github.com/Shopify/quilt/pull/1785)
 
 ## [0.11.2] - 2021-03-03
 
--- a/packages/react-form/CHANGELOG.md
+++ b/packages/react-form/CHANGELOG.md
@@ -7,6 +7,10 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## [Unreleased] -->
 
+### Added
+
+- Add ability to sort dynamic lists [#1785](https://github.com/Shopify/quilt/pull/1785)
+
 ## [0.11.2] - 2021-03-03
 
 ### Fixed
--- a/packages/react-form/README.md
+++ b/packages/react-form/README.md
@@ -652,7 +652,7 @@ const emptyCardFactory = (): Card => ({
   cvv: '',
 });
 
-const {fields, addItem, removeItem} = useDynamicList(
+const {fields, addItem, removeItem, moveItem} = useDynamicList(
   [{cardNumber: '4242 4242 4242 4242', cvv: '000'}],
   emptyCardFactory,
 );
@@ -672,7 +672,7 @@ const emptyCardFactory = (): Card[] => [
   },
 ];
 
-const {fields, addItem, removeItem} = useDynamicList(
+const {fields, addItem, removeItem, moveItem} = useDynamicList(
   [{cardNumber: '4242 4242 4242 4242', cvv: '000'}],
   emptyCardFactory,
 );
@@ -700,6 +700,16 @@ You can choose to initialize the list with an existing number of cards or no car
       />
       <div>
         <Button onClick={() => removeItem(index)}>Remove</Button>
+      </div>
+      <div>
+        <Button onClick={() => moveItem(field, index, index - 1)}>
+          Move Item Up
+        </Button>
+      </div>
+      <div>
+        <Button onClick={() => moveItem(field, index, index + 1)}>
+          Move Item Down
+        </Button>
       </div>
     </FormLayout.Group>
   ));
--- a/packages/react-form/README.md
+++ b/packages/react-form/README.md
@@ -702,12 +702,12 @@ You can choose to initialize the list with an existing number of cards or no car
         <Button onClick={() => removeItem(index)}>Remove</Button>
       </div>
       <div>
-        <Button onClick={() => moveItem(field, index, index - 1)}>
+        <Button onClick={() => moveItem(index, index - 1)}>
           Move Item Up
         </Button>
       </div>
       <div>
-        <Button onClick={() => moveItem(field, index, index + 1)}>
+        <Button onClick={() => moveItem(index, index + 1)}>
           Move Item Down
         </Button>
       </div>
--- a/packages/react-form/README.md
+++ b/packages/react-form/README.md
@@ -702,12 +702,18 @@ You can choose to initialize the list with an existing number of cards or no car
         <Button onClick={() => removeItem(index)}>Remove</Button>
       </div>
       <div>
-        <Button onClick={() => moveItem(index, index - 1)}>
+        <Button
+          disabled={index === 0}
+          onClick={() => moveItem(index, index - 1)}
+        >
           Move Item Up
         </Button>
       </div>
       <div>
-        <Button onClick={() => moveItem(index, index + 1)}>
+        <Button
+          disabled={index === fields.length}
+          onClick={() => moveItem(index, index + 1)}
+        >
           Move Item Down
         </Button>
       </div>
--- a/packages/react-form/src/hooks/list/dynamiclist.ts
+++ b/packages/react-form/src/hooks/list/dynamiclist.ts
@@ -1,12 +1,13 @@
 import {FieldDictionary} from '../../types';
 
-import {addFieldItemAction, removeFieldItemAction} from './hooks';
+import {addFieldItemAction, removeFieldItemAction, moveFieldItemAction} from './hooks';
 import {useBaseList, FieldListConfig} from './baselist';
 
 interface DynamicList<Item extends object> {
   fields: FieldDictionary<Item>[];
   addItem(factoryArgument?: any): void;
   removeItem(index: number): void;
+  moveItem(item: Item, oldIndex: number, newIndex: number): void;
 }
 
 type FactoryFunction<Item extends object> = (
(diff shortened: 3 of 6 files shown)


packages/react-form/src/hooks/list/hooks/reducer.ts
```diff
@@ -28,6 +29,11 @@ interface AddFieldItemAction<Item> {
   payload: {list: Item[]};
 }
 
+interface MoveFieldItemAction<Item> {
+  type: 'moveFieldItem';
+  payload: {item: Item; oldIndex: number; newIndex: number};
```
atesgoral, line 34: Why do we need to pass the item? Wouldn't just `oldIndex` and `newIndex` suffice?

I also suggest calling these `fromIndex` and `toIndex`.

oluwatimio: Yes, it should be sufficient to just pass the indexes and grab the item from the list with `oldIndex`

packages/react-form/src/hooks/list/hooks/reducer.ts
```diff
@@ -158,6 +175,20 @@ function reduceList<Item extends object>(
         list: action.payload.list.map(initialListItemState),
       };
     }
+    case 'moveFieldItem': {
+      const newList = [...state.list];
+      newList.splice(action.payload.oldIndex, 1);
+      newList.splice(
+        action.payload.newIndex,
+        0,
+        action.payload.item as FieldStates<Item>,
```
oluwatimio, line 184: I believe if we just grab the item from the list, there might be no need to typecast to FieldStates<Item> when adding it to the newIndex

oluwatimio (changes requested):

atesgoral:
I wouldn't use the term "sort" to describe this feature. The way this is implemented, it allows moving an item between arbitrary indices. "Enable reordering"? "Enable relocating an item"?

packages/react-form/src/hooks/list/hooks/reducer.ts
jonarnaldo: good point, I suppose we can just look up item in `state.list`

packages/react-form/src/hooks/list/hooks/reducer.ts
jonarnaldo: 👍

packages/react-form/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## [Unreleased] -->
 
+### Added
+
+- Add ability to sort dynamic lists [#1785](https://github.com/Shopify/quilt/pull/1785)
```
atesgoral, line 12: Maybe:

```suggestion
- Add ability to reorder dynamic lists [#1785](https://github.com/Shopify/quilt/pull/1785)
```

packages/react-form/src/hooks/list/dynamiclist.ts
```diff
@@ -1,12 +1,17 @@
...
+  removeFieldItemAction,
+  moveFieldItemAction,
+} from './hooks';
 import {useBaseList, FieldListConfig} from './baselist';
 
 interface DynamicList<Item extends object> {
   fields: FieldDictionary<Item>[];
   addItem(factoryArgument?: any): void;
   removeItem(index: number): void;
+  moveItem(oldIndex: number, newIndex: number): void;
```
atesgoral, line 14: Just "index" or "existingIndex" could be better than "oldIndex". Have you considered calling these properties "fromIndex"+"toIndex"? (Don't know if you saw my earlier comment about that.)

packages/react-form/src/hooks/list/test/dynamiclist.test.tsx
```diff
@@ -39,6 +57,36 @@ describe('useDynamicList', () => {
       );
     }
 
+    it('can move field to a new position', () => {
```
oluwatimio, line 60: Would it be possible to separate this test into move up and move down?

so a describe('can move field to a new position') which will then have two tests (moving up and moving down) ?

packages/react-form/src/hooks/list/hooks/reducer.ts
```diff
@@ -158,6 +174,17 @@ function reduceList<Item extends object>(
         list: action.payload.list.map(initialListItemState),
       };
     }
+    case 'moveFieldItem': {
+      const newList = [...state.list];
+      const item = newList[action.payload.oldIndex];
+      newList.splice(action.payload.oldIndex, 1);
```
atesgoral, line 180:
```suggestion
      const [item] = newList.splice(action.payload.oldIndex, 1);
```

packages/react-form/src/hooks/list/hooks/reducer.ts
```diff
@@ -158,6 +174,17 @@ function reduceList<Item extends object>(
         list: action.payload.list.map(initialListItemState),
       };
     }
+    case 'moveFieldItem': {
```
atesgoral, line 177: Shouldn't we do range checking here? Should that be responsibility of the app?

packages/react-form/src/hooks/list/test/dynamiclist.test.tsx
jonarnaldo: yes, that's a good suggestion.

packages/react-form/src/hooks/list/dynamiclist.ts
jonarnaldo: no, I didn't. I do like the `fromIndex` and `toIndex` because it's pretty informative for both args - will update

packages/react-form/src/hooks/list/test/dynamiclist.test.tsx
atesgoral: But do we really need to test the up/down case? The way this is implemented, it's completely agnostic of what the move is. I personally wouldn't clutter tests with redundant cases.

atesgoral: i.e. I would just test one move and be done with it.

packages/react-form/src/hooks/list/hooks/reducer.ts
```diff
@@ -176,8 +176,19 @@ function reduceList<Item extends object>(
     }
     case 'moveFieldItem': {
       const newList = [...state.list];
-      const [item] = newList.splice(action.payload.oldIndex, 1);
-      newList.splice(action.payload.newIndex, 0, item);
+
+      if (
+        action.payload.fromIndex < state.list.length ||
```
oluwatimio, line 181: This might still pass if the index is greater than the length since it will satisfy `action.payload.fromIndex >= 0`

jonarnaldo: yes, sorry, just noticed that

oluwatimio: No problem! 🙂

packages/react-form/src/hooks/list/hooks/reducer.ts
```diff
@@ -158,6 +174,29 @@ function reduceList<Item extends object>(
         list: action.payload.list.map(initialListItemState),
       };
     }
+    case 'moveFieldItem': {
+      const newList = [...state.list];
+
+      if (
```
atesgoral, line 180: It would be better throw an exception instead of silently ignoring bad values.

packages/react-form/src/hooks/list/hooks/reducer.ts
```diff
@@ -158,6 +174,29 @@ function reduceList<Item extends object>(
         list: action.payload.list.map(initialListItemState),
       };
     }
+    case 'moveFieldItem': {
+      const newList = [...state.list];
```
atesgoral, line 178: We shouldn't make an unnecessary copy when an index is invalid and we don't do a move.

packages/react-form/src/hooks/list/hooks/reducer.ts
jonarnaldo: 👍

oluwatimio (approved):
This LGTM! Thanks for adding this functionality ❤️

packages/react-form/README.md
```diff
@@ -701,6 +701,16 @@ You can choose to initialize the list with an existing number of cards or no car
       <div>
         <Button onClick={() => removeItem(index)}>Remove</Button>
       </div>
+      <div>
+        <Button onClick={() => moveItem(index, index - 1)}>
```
atesgoral, line 705: Since react-form will now blow up with an exception, it's the responsibility of the app to do the range checking. Untested suggestion:

```suggestion
        <Button disabled={index === 0} onClick={() => moveItem(index, index - 1)}>
```

packages/react-form/README.md
```diff
@@ -701,6 +701,16 @@ You can choose to initialize the list with an existing number of cards or no car
...
+          Move Item Up
+        </Button>
+      </div>
+      <div>
+        <Button onClick={() => moveItem(index, index + 1)}>
```
atesgoral, line 710:
```suggestion
        <Button disabled={index === fields.length} onClick={() => moveItem(index, index + 1)}>
```

packages/react-form/src/hooks/list/test/dynamiclist.test.tsx
```diff
@@ -39,6 +48,26 @@ describe('useDynamicList', () => {
...
```
atesgoral, line 51: Would be nice to have a test to verify that we're throwing an exception on invalid indices. (A fully satisfactory test would test all 4 cases, but we can keep it simple and just do one case for now.)

packages/react-form/src/hooks/list/test/dynamiclist.test.tsx
```diff
@@ -31,6 +31,7 @@ describe('useDynamicList', () => {
                 Remove Variant
               </button>
               <button
+                // disabled={index === 0}
```
atesgoral, line 34: `\o/`

atesgoral (approved):

PR #1788, opened 2021-03-16
Use named exports in jest transforms

BPScott:
## Description

Rather than doing strange cjs-only things in jest transforms, lets use named exports.

This looks scary but it's 99% whitespace changes, review with suppressed whitespace.

Still need to confirm this works in apps.

## Type of change

- [x] graphql-mini-transforms Patch: Bug (non-breaking change which fixes an issue)

## Checklist


vsumner (approved):
I tested the build output against another project and it works.

PR #1789, opened 2021-03-16
Metafile polishing

BPScott:
## Description

- Remove dist from gitignore as it is no longer used now we're building
  with sk1
- Add .editorconfig file
- Update vscode settings to ignore sk1 build output and sundry polishing

## Type of change

Repo meta improvements, no changes relating to single packages.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,20 @@
+# editorconfig.org
+root = true
+
+[*]
+charset = utf-8
+indent_size = 2
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+# Markdown syntax specifies that trailing whitespaces can be meaningful,
+# so let’s not trim those. e.g. 2 trailing spaces = linebreak (<br />)
+# See https://daringfireball.net/projects/markdown/syntax#p
+[*.md]
+trim_trailing_whitespace = false
+
+# Disable trailing whitespace removal in diff files,
+# where whitespace is meaningful
+[*.diff]
+trim_trailing_whitespace = false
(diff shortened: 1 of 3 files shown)


.vscode/settings.json
```diff
@@ -4,30 +4,21 @@
...
     "**/node_modules": true,
     "**/.DS_Store": true,
-    "build": true,
-    ".dev": true
+    "packages/*/{.sewing-kit,build}": true,
+    ".{dev,shadowenv.d,sewing-kit}": true
   },
   "files.insertFinalNewline": true,
   "files.trimTrailingWhitespace": true,
-  "search.exclude": {
```
BPScott, line 20: search.exclude grandfathers in all the patterns in files.exclude so we don't need to duplicate them.

.editorconfig
```diff
@@ -0,0 +1,20 @@
...
+
+# Markdown syntax specifies that trailing whitespaces can be meaningful,
+# so let’s not trim those. e.g. 2 trailing spaces = linebreak (<br />)
+# See https://daringfireball.net/projects/markdown/syntax#p
+[*.md]
+trim_trailing_whitespace = false
+
+# Disable trailing whitespace removal in diff files,
+# where whitespace is meaningful
+[*.diff]
```
atesgoral, line 19: What diff files though?

atesgoral (approved):

.editorconfig
BPScott: Nothing in the actual repo, but this is is my standard config that I copy about everywhere, so it's easier to leave in :)

The diff stuff only really crops up when I'm creating temporary files when testing build output stuff

PR #1791, opened 2021-03-17
Add react hook useSteps

mateus:
## Description

Adds `useSteps` to `react-hooks` package.

This hook provides an object that contains a numeric state value for the current step and two memoized callbacks to manipulate it (nextStep, previousStep). It accepts two arguments: the total number of steps possible, and the initial step (optional - defaults to 0). This is useful for paged modals or anything that has a finite set of steps to walkthrough.

```tsx
function MyComponent() {
  const {currentStep, nextStep, previousStep} = useSteps(5, 2);

  return (
    <>
      <p>Step: {currentStep}</p>
      <button onClick={nextStep}>Go to next step</button>
      <button onClick={previousStep}>Go to previous step</button>
    </>
  );
}
```

## Type of change

## Checklist

### DIFF
--- a/packages/react-hooks/README.md
+++ b/packages/react-hooks/README.md
@@ -25,6 +25,7 @@ $ yarn add @shopify/react-hooks
 - [usePrevious()](#useprevious)
 - [useTimeout()](#usetimeout)
 - [useToggle()](#usetoggle)
+- [useSteps()](#usesteps)
 
 ## Usage
 
@@ -253,6 +254,24 @@ function MyComponent() {
       <button onClick={toggleIsActive}>Toggle isActive state</button>
       <button onClick={setIsActiveTrue}>Set isActive state to true</button>
       <button onClick={setIsActiveFalse}>Set isActive state to false</button>
+    </>
+  );
+}
+```
+
+### `useSteps()`
+
+This hook provides an object that contains a numeric state value for the current step and two memoized callbacks to manipulate it (nextStep, previousStep). It accepts two arguments: the total number of steps possible, and the initial step (optional - defaults to 1). This is useful for paged modals or anything that has a finite set of steps to walkthrough.
+
+```tsx
+function MyComponent() {
+  const {currentStep, nextStep, previousStep} = useSteps(5, 2);
+
+  return (
+    <>
+      <p>Step: {currentStep}</p>
+      <button onClick={nextStep}>Go to next step</button>
+      <button onClick={previousStep}>Go to previous step</button>
     </>
   );
 }
--- a/packages/react-hooks/src/hooks/steps.ts
+++ b/packages/react-hooks/src/hooks/steps.ts
@@ -1,0 +1,19 @@
+import {useState, useCallback} from 'react';
+
+export function useSteps(steps: number, initialStep = 1) {
+  const [currentStep, setCurrentStep] = useState(initialStep);
+
+  return {
+    currentStep,
+    nextStep: useCallback(() => {
+      if (currentStep < steps) {
+        setCurrentStep(currentStep + 1);
+      }
+    }, [currentStep, steps]),
+    previousStep: useCallback(() => {
+      if (currentStep > 1) {
+        setCurrentStep(currentStep - 1);
+      }
+    }, [currentStep]),
+  };
+}
--- a/packages/react-hooks/CHANGELOG.md
+++ b/packages/react-hooks/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [1.13.0] - 2021-03-18
+
+### Added
+
+- Added `useSteps` hook ([#1791](https://github.com/Shopify/quilt/pull/1791))
+
 ## [1.12.2] - 2021-03-03
 
 ### Fixed
--- a/packages/react-hooks/src/hooks/tests/steps.test.tsx
+++ b/packages/react-hooks/src/hooks/tests/steps.test.tsx
@@ -1,0 +1,91 @@
+import React from 'react';
+import {mount} from '@shopify/react-testing';
+
+import {useSteps} from '../steps';
+
+describe('useSteps', () => {
+  function TestMyProps(_props: any) {
+    return null;
+  }
+
+  function MockComponent({
+    steps,
+    initialStep,
+  }: {
+    steps: number;
+    initialStep?: number;
+  }) {
+    const props = useSteps(steps, initialStep);
+    return <TestMyProps {...props} />;
+  }
+
+  it('starts with current step set to 1', () => {
+    const wrapper = mount(<MockComponent steps={3} />);
+
+    expect(wrapper).toContainReactComponent(TestMyProps, {
+      currentStep: 1,
+    });
+  });
+
+  it('sets custom initial step when initialStep is provided', () => {
+    const wrapper = mount(<MockComponent steps={3} initialStep={2} />);
+
+    expect(wrapper).toContainReactComponent(TestMyProps, {
+      currentStep: 2,
+    });
+  });
+
+  it('sets the next step when nextStep is called', () => {
+    const wrapper = mount(<MockComponent steps={3} />);
+
+    wrapper.find(TestMyProps).trigger('nextStep');
+    expect(wrapper).toContainReactComponent(TestMyProps, {
+      currentStep: 2,
+    });
+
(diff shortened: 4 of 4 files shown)


packages/react-hooks/src/hooks/steps.ts
```diff
@@ -0,0 +1,19 @@
+import {useState, useCallback} from 'react';
+
+export function useSteps(steps: number, initialStep = 0) {
```
JaKXz, line 3: Using 0 as the default initialStep might cause off by one errors for someone who is counting steps "naturally" as opposed to index-based.

```suggestion
export function useSteps(steps: number, initialStep = 1) {
```

packages/react-hooks/src/hooks/steps.ts
```diff
@@ -0,0 +1,19 @@
...
+  return {
+    currentStep,
+    nextStep: useCallback(() => {
+      if (currentStep < steps - 1) {
+        return setStep(currentStep + 1);
+      }
+    }, [currentStep, steps]),
+    previousStep: useCallback(() => {
+      if (currentStep > 0) {
+        return setStep(currentStep - 1);
```
JaKXz, line 15: The return here is not doing much for you but is kind of confusing to read because it implies a value is being returned. I think this is clearer:

```suggestion
        setStep(currentStep + 1);
      }
    }, [currentStep, steps]),
    previousStep: useCallback(() => {
      if (currentStep > 0) {
        setStep(currentStep - 1);
```

packages/react-hooks/README.md
```diff
@@ -257,3 +258,21 @@ function MyComponent() {
   );
 }
 ```
+
+### `useSteps()`
+
+This hook provides an object that contains a numeric state value for the current step and two memoized callbacks to manipulate it (nextStep, previousStep). It accepts two arguments: the total number of steps possible, and the initial step (optional - defaults to 0). This is useful for paged modals or anything that has a finite set of steps to walkthrough.
```
JaKXz, line 264:
```suggestion
This hook provides an object that contains a numeric state value for the current step and two memoized callbacks to manipulate it (nextStep, previousStep). It accepts two arguments: the total number of steps possible, and the initial step (optional - defaults to 1). This is useful for paged modals or anything that has a finite set of steps to walkthrough.
```

JaKXz (commented):
I think this is a reasonable hook to add :) I had some minor thoughts but I'll let the owners make the call

I also appreciate the Arrange Act Assert spacing in your tests 👍🏽

packages/react-hooks/src/hooks/steps.ts
mateus: I was on the fence about this one too and I decided to go with 0 as the default. I think it's better to keep 0-based indexing because we can leverage them to walkthrough arrays or even objects. Maybe I could add to the documentation that it is 0-based indexed.

packages/react-hooks/src/hooks/steps.ts
```diff
@@ -0,0 +1,19 @@
+import {useState, useCallback} from 'react';
+
+export function useSteps(steps: number, initialStep = 0) {
+  const [currentStep, setStep] = useState(initialStep);
+
+  return {
```
JaKXz, line 6: I would recommend returning an array here so that people can name the variable results conveniently with array destructuring, as `useState` does.

packages/react-hooks/src/hooks/steps.ts
```diff
@@ -0,0 +1,19 @@
+import {useState, useCallback} from 'react';
+
+export function useSteps(steps: number, initialStep = 0) {
+  const [currentStep, setStep] = useState(initialStep);
```
JaKXz, line 4: super nitpicky suggestion about the useState naming conventions: 
```suggestion
  const [currentStep, setCurrentStep] = useState(initialStep);
```

totally inconsequential here but reads a little clearer :)

packages/react-hooks/src/hooks/steps.ts
JaKXz: I'm thinking of what would be the most common case, where people write:

```jsx
// this gives me 5 steps, not 4 as this might imply at a glance.
const {currentStep, nextStep, previousStep} = useSteps(4)

// to get exactly 4 steps I have to write
const {currentStep, nextStep, previousStep} = useSteps(4, 1)
```

mateus: I'm convinced. Just updated it

packages/react-hooks/src/hooks/steps.ts
mateus: You should be able to rename them as an object as well 

```tsx
const {
  currentStep: step,
  nextStep: next,
  previousStep: previous,
} = useSteps(5, 2);
```

packages/react-hooks/src/hooks/steps.ts
JaKXz: @mateus 
I think another win is when you are using an array of steps, it would force you to be more explicit about it:

```jsx
const {currentStep, nextStep, previousStep} = useSteps(data.length - 1, 0)
```
vs just
```jsx
const {currentStep, nextStep, previousStep} = useSteps(data.length - 1)
```

What do you think of handling arrays automatically and allowing the user to just pass them in as the first argument:

```suggestion
export function useSteps(steps: number | Array<T>, initialStep = 1) {
  let count = steps;
  if (Array.isArray(steps)) {
    count = steps.length - 1;
  }
```

So that you can write
```jsx
const {currentStep, nextStep, previousStep} = useSteps([{key: "value"}], 0)
```

packages/react-hooks/src/hooks/steps.ts
JaKXz: For sure, it's just more convenient as an array IMO but I don't feel strongly.

packages/react-hooks/src/hooks/steps.ts
mateus: 🤔   I don't see a lot of benefit in it if we only use the array to determine the number of steps. If we pass down an array of steps I would likely expect that `currentStep` isn't only a number representing the current step but also that it includes the value from the array. 

Something like this:
```tsx
const {currentStep, nextStep, previousStep} = useSteps(['potato', 'apple', 'banana'])

> currentStep 
{index: 0, value: 'potato', step: 1}

> nextStep()
> currentStep
{index: 1, value: 'apple', step: 2}
```

The scope of this hook might be growing a bit if we choose this route and I don't know if that's the right level of abstraction here. Unless I am missing something from your suggestion?

mateus: I wanna play around with this idea a bit more though. I don't wanna discard it right away. Let me write a couple of demos using this approach and I will get back to you.

mateus: What do you think of this:

```tsx
export function useSteps(steps: number | any[], initialStep = 1) {
  const [currentStep, setCurrentStep] = useState(initialStep);
  const stepsArr = useMemo(
    () => (Array.isArray(steps) ? steps : [...Array(steps).keys()]),
    [steps],
  );

  return {
    currentStep,
    currentValue: stepsArr[currentStep - 1],
    nextStep: useCallback(() => {
      if (currentStep < stepsArr.length) {
        setCurrentStep(currentStep + 1);
      }
    }, [currentStep, stepsArr]),
    previousStep: useCallback(() => {
      if (currentStep > 1) {
        setCurrentStep(currentStep - 1);
      }
    }, [currentStep]),
  };
}
``` 

Usage cases:

```tsx
const {currentStep, currentValue, nextStep, previousStep} = useSteps(3);

> currentStep, currentValue
1, 0

> nextStep
> currentStep, currentValue
2, 1

> nextStep
> currentStep, currentValue
3, 2
```

or

```tsx
const {currentStep, currentValue, nextStep, previousStep} = useSteps(['apple', 'banana', 'corn']);

> currentStep, currentValue
1, apple

> nextStep
> currentStep, currentValue
2, banana

> nextStep
> currentStep, currentValue
3, corn
```

or 

```tsx
const {currentStep, currentValue, nextStep, previousStep} = useSteps(['apple', 'banana', 'corn'], 2);

> currentStep, currentValue
2, banana

> nextStep
> currentStep, currentValue
3, corn
```

JaKXz: I thought of the currentValue idea as well but the user of this will have access to the currentStep index so try can just do `data[currentStep]` as they need, which is why I didn’t suggest it.

dahukish:
I love the though and effort that is in this PR. I wonder though if this is just another version of this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Iterators_and_Generators

Most time I have used `current` `previous` or `next` have been to page through graphql data which usually has its own pointer etc. Would love some real world examples. 🙇

mateus:
Closing it, for now, to explore more real use cases to validate my assumptions.

PR #1792, opened 2021-03-19
[koa-performance] Include custom tags for navigation metrics

darrenhebner:
## Description
This PR adds the ability to add custom tags to additional navigation metrics. We currently tag these metrics with the same navigation tags as the base navigation metrics (complete, usable). There are some cases where tags specific to the type of additional metric could be useful.

Here's an example. In Checkout One, we want to track that our GraphQL client is effectively hydrating. To do this, we want to add an additional `failed_hydration` metric. In order to know which query is failing, we will use a custom tag.

```ts
navigation.eventsByType(EventType.graphql).filter(({metadata}) =>
  serverSideQueries.includes(metadata.name),
).forEach(({metadata}) => {
  metrics.push({
    name: 'failed_hydration',
    value: 1,
    tags: {
      name: metadata.name
    }
  })
})
```

## Type of change

## Checklist

### DIFF
--- a/packages/koa-performance/CHANGELOG.md
+++ b/packages/koa-performance/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-## [1.4.0] - 2021-03-19
+## [Unreleased]
 
 ### Added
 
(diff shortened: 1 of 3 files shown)


packages/koa-performance/CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [1.4.0] - 2021-03-19
```
BPScott, line 8: This should be `## Unreleased`. We change to be an actual version number and date when we perform the release

```suggestion
## Unreleased
```

BPScott (commented):
Code looks good, thanks!

One changelog formatting correction inline. Fix that up then merge away!

atesgoral (approved):

atesgoral:
@darrenhebner https://www.npmjs.com/package/@shopify/koa-performance v1.4.0 published.

Issue #1794, opened 2021-03-22
[react-form] TS bug in useList submit handler


debchn:
## Overview

TypeScript bugs crop up in `useForm`’s submit handler when combined with `useList`:

1. When the [submit handler is defined in a function declaration](https://github.com/Shopify/web/pull/40084/files) outside of the useForm config object, a TS error appears stating that the array `is missing the following properties ..: [Symbol.iterator], [Symbol.unscopables]`

2. When the [submit handler is defined as a method on the useForm config object](https://github.com/Shopify/web/pull/40085/files), the Array methods (e.g. `map`) expect a `Field<T>` rather than `T`. Note that the property at `[x: number]` is correctly typed; only the types on the Array methods are incorrect:

![image](https://user-images.githubusercontent.com/2354121/112056901-f6c82380-8b2e-11eb-990b-7ee0f2032b62.png)

## Consuming repo

shopify/web

Used `as any` as a workaround for now. https://github.com/Shopify/web/blob/d3024ba4c9a1cf58f193be5ceb0921a7a29746b6/app/sections/Legal/components/LegalPage/components/LegalPoliciesForm/useLegalPoliciesForm.ts#L41

dahukish:
Thanks @debchn 🥇

PR #1795, opened 2021-03-22
Fix FormMapping for array of FieldDictionaries

debchn:
## Description

Fixes https://github.com/Shopify/quilt/issues/1794

See description in linked issue above.

The changed line is responsible for mapping a list of FieldDictionaries (e.g. `FieldDictionary<{body: string}>[]`) to a list of plain dictionary values (`{body: string}[]`). I think that the original type resulted in bugs because every key (including method names) was mapped over, resulting in incorrect argument types for Array methods such as `map`.

I've updated the type with explicit array syntax, and replaced `Index` with the more generic `number` type.

🎩 locally using `yalc` on my original branch in shopify/web showed no more TS errors, although I needed to add an explicit return type of `Promise<SubmitResult>` to the function declaration version.

## Type of change

- [X] react-form Patch: Bug (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/packages/react-form/src/types.ts
+++ b/packages/react-form/src/types.ts
@@ -115,6 +115,6 @@ type FieldProp<T, K extends keyof Field<any>> = T extends Field<any>
 */
 export type FormMapping<Bag, FieldKey extends keyof Field<any>> = {
   [Key in keyof Bag]: Bag[Key] extends any[]
-    ? {[Index in keyof Bag[Key]]: FieldProp<Bag[Key][Index], FieldKey>}
+    ? FieldProp<Bag[Key][number], FieldKey>[]
     : FieldProp<Bag[Key], FieldKey>;
 };


kaelig:
👋 I can confirm that this fix works as expected in `web`, however I couldn't make it work out of the box in services-db (I only tried a few small changes, as I'm missing context).

What would it look like to update https://github.com/Shopify/services-db/blob/0e51a4a4d9efaaa0b339f51cd482cfeb02ba04e4/app/ui/components/ServiceLinks/components/EditLinksModal.tsx#L33-L51 to test if this fix works there too?

cc @ryanwilsonperkin

packages/react-form/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## [Unreleased] -->
 
+### Fixed
+
+- Fixed a TypeScript bug in FormMapping for FieldDictionary arrays [#1795](https://github.com/Shopify/quilt/pull/1795)
```
kaelig, line 12: Would this be a good place to tell people that `Promise<SubmitResult>` is still required when encountering this issue (if I'm following correctly?)?

kaelig:
I'd love to see a test that reproduces this pattern, to avoid future regressions. Having a test can also serve as an example for other devs, but if you'd like to add an actual example in the README or in the block, I won't stop you :)

https://github.com/Shopify/quilt/blob/7c6de6198cafdb914813e4b391025b8dc2ada1e7/packages/react-form/src/hooks/form.ts#L10-L58

ryanwilsonperkin:
> 👋 I can confirm that this fix works as expected in `web`, however I couldn't make it work out of the box in services-db (I only tried a few small changes, as I'm missing context).
> 
> What would it look like to update [Shopify/services-db@`0e51a4a`/app/ui/components/ServiceLinks/components/EditLinksModal.tsx#L33-L51](https://github.com/Shopify/services-db/blob/0e51a4a4d9efaaa0b339f51cd482cfeb02ba04e4/app/ui/components/ServiceLinks/components/EditLinksModal.tsx#L33-L51) to test if this fix works there too?
> 
> cc @ryanwilsonperkin

👋🏻 I can confirm that this works for our use-case in ServicesDB, to 🎩 I just edited the compiled .d.ts directly to use this approach and then confirmed that typechecking compiled properly.

What we did previously: https://github.com/Shopify/services-db/blob/46789d4/app/ui/components/ServiceLinks/components/EditLinksModal.tsx#L33-L41

Compared with what we could do now:
https://github.com/Shopify/services-db/blob/2102e64515f822804e57629ba25017add998370e/app/ui/components/ServiceLinks/components/EditLinksModal.tsx#L33-L38

However one thing still seems odd, our `id` parameter from the field value doesn't seem to be typed correctly. It should be `string | undefined` but it's type in the `fieldValues` array ends up typed as `unknown`.

ryanwilsonperkin:
Here's a minimal reproduction, it seems that it's specifically when we use the "optional" type rather than an explicit `string | undefined` that the unknown shows up.

```tsx
interface Widget {
  id?: string;
  name: string;
}

function widgetFactory() {
  return {id: undefined, name: 'placeholder'};
}

function Example() {
  const widgets: Widget[] = [];
  const {fields} = useDynamicList(widgets, widgetFactory);
  useForm({
    fields: {fields},
    onSubmit: async (fieldValues) => {
      const ids: (string | undefined)[] = fieldValues.fields.map(
        (field) => field.id,
      );
      return {status: 'success'};
    },
  });
}
```

In this case we pass a list of "widgets" into the dynamic list and `Widget`'s "id" prop is an optional string. I'd expect this to result in `fieldValues.fields[number].id` to be typed as `string | undefined` but it ends up typed as `unknown`

packages/react-form/CHANGELOG.md
debchn: Explicitly adding `Promise<SubmitResult>` is only necessary if you define the submit handler outside of the config object. I've just realized that, because the `SubmitHandler` needs to return an object containing a string literal (`'success'` or ``'fail'`), it should really always be defined within the config object, and that is how the documentation examples work, too.

Sorry for including that red herring in my PR description 😅

sylvhama (approved):
I have tophatted those changes in `web`. It works great thanks!

kaelig (approved):
>I can confirm that this works for our use-case in ServicesDB

That's great, thank you for testing!

PR #1797, opened 2021-03-23
Pass time zone to humanizeDate()

### LINKED ISSUE #1796: [i18n] Humanized date that's within today can be reported as tomorrow
## Overview

`i18n.formatDate()` with `style: DateStyle.Humanize` ignores time zone.

It causes this test to be flaky and to fail locally (EST): https://github.com/Shopify/quilt/blob/main/packages/react-i18n/src/test/i18n.test.ts#L1428

```
yarn test packages/react-i18n/src/test/i18n.test.ts

  ● I18n › #formatDate() › with DateStyle.Humanize › future dates › when the locale time zone is not specified › formats a date for today

    expect(jest.fn()).toHaveBeenCalledWith(...expected)

    - Expected
    + Received

    - "date.humanize.today",
    + "date.humanize.tomorrow",
```

## Consuming repo

An existing unit test

atesgoral:
## Description

Fixes #1796

## Type of change

## Checklist

### DIFF
--- a/packages/react-i18n/CHANGELOG.md
+++ b/packages/react-i18n/CHANGELOG.md
@@ -7,6 +7,10 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## [Unreleased] -->
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


TheMallen (approved):
checks out

PR #1801, opened 2021-03-25
Update the bundle so that we can get mimemagic 0.3.7

vsumner:
## Description
Update bundle dependencies so that `mimemagic` can be updated to `0.3.7` after `0.3.6` was yanked.


gems/quilt_rails/test/quilt/header_csrf_strategy_test.rb
```diff
@@ -7,7 +7,7 @@ def test_raises_an_exception_if_the_samesite_header_is_missing
       DummyRequest.any_instance.stubs(:headers).returns({})
       strategy = HeaderCsrfStrategy.new(DummyController.new)
 
-      assert_raises HeaderCsrfStrategy::NoSameSiteHeaderError do
+      assert_raises(HeaderCsrfStrategy::NoSameSiteHeaderError) do
```
dahukish, line 10: was it complaining about missing parens?

vsumner: Yes

dahukish (approved):
LTGM! no 🎩

PR #1804, opened 2021-03-30
🛠 Fix react-router exporting functions as types

NicholasEllul:
## Description

Fixes 

Currently I am unable to import `useLocation` and a handful of other `react-router-dom` utilities from `@shopify/react-router` due to the fact that they are being exported as types. This PR moves the exports which shouldn't be types to being exported normally.

## Type of change
- [X]react-router Patch: Bug (non-breaking change which fixes an issue)

## Checklist


packages/react-router/CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [0.2.6] - 2021-03-30
+
+### Fixed
+
+- Updated exports which were mistakenly exporting functions as types [#1804](https://github.com/Shopify/quilt/pull/1804)
```
BPScott, line 12:
```suggestion
- Export value for `MemoryRouter`, `useRouteMatch`,  `useParams`, `useLocation` and `useHistory`, which were previously only exporting their types [#1804](https://github.com/Shopify/quilt/pull/1804)
```

BPScott (approved):
Thanks for the fix!

Merge once CI is green :)

PR #1811, opened 2021-04-01
Use @shopify/typescript-configs as a base

BPScott:
## Description

Simplify our to tsconfig.base.json by inheriting from the pan-org ts configs.

The pan org config enables strict mode but we're not compliant yet, so opt out of the currently problematic sub-checks for now. We can come back to enabling these items in follow up PRs

### DIFF
--- a/config/typescript/tsconfig.base.json
+++ b/config/typescript/tsconfig.base.json
@@ -1,33 +1,15 @@
 {
-  "compileOnSave": false,
+  "extends": "@shopify/typescript-configs/library.json",
   "compilerOptions": {
     "composite": true,
-    "declaration": true,
-    "declarationMap": true,
-    "pretty": true,
-    "moduleResolution": "node",
-    "target": "es5",
-    "downlevelIteration": true,
-    "esModuleInterop": true,
-    "jsx": "react",
-    "isolatedModules": true,
-    "strictNullChecks": true,
+    // Strict mode enables these strict* and noImplicit* options.
+    // Disable them temporarily while we fix up problematic behaviour from
+    // the time before we wanted strict mode
+    "strictBindCallApply": false,
+    "strictFunctionTypes": false,
+    "strictPropertyInitialization": false,
     "noImplicitAny": false,
-    "sourceMap": false,
-    "noEmitOnError": false,
-    "experimentalDecorators": true,
-    "noUnusedParameters": true,
-    "noUnusedLocals": true,
-    "importHelpers": true,
-    "lib": [
-      "dom",
-      "dom.iterable",
-      "es2015",
-      "es2016",
-      "es2017",
-      "es2018",
-      "esnext.asynciterable"
-    ],
+    "noImplicitThis": false,
     "typeRoots": ["../../node_modules/@types"]
   }
 }
--- a/config/typescript/tsconfig.base.json
+++ b/config/typescript/tsconfig.base.json
@@ -2,9 +2,8 @@
   "extends": "@shopify/typescript-configs/library.json",
   "compilerOptions": {
     "composite": true,
-    // Strict mode enables these strict* and noImplicit* options.
-    // Disable them temporarily while we fix up problematic behaviour from
-    // the time before we wanted strict mode
+    "__TODOCOMMENT1": "Strict mode enables these strict* and noImplicit* options.",
+    "__TODOCOMMENT2": "Disable them temporarily while we fix up problematic behaviour from the time before we wanted strict mode",
     "strictBindCallApply": false,
     "strictFunctionTypes": false,
     "strictPropertyInitialization": false,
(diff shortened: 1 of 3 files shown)


config/typescript/tsconfig.base.json
```diff
@@ -1,33 +1,15 @@
...
-    "declarationMap": true,
-    "pretty": true,
-    "moduleResolution": "node",
-    "target": "es5",
-    "downlevelIteration": true,
-    "esModuleInterop": true,
-    "jsx": "react",
-    "isolatedModules": true,
-    "strictNullChecks": true,
+    // Strict mode enables these strict* and noImplicit* options.
```
atesgoral, line 5: While this obviously seems to work fine, could we perhaps use a hack like a `"_NOTICE": "...."` to keep this as valid JSON?

BPScott: Updated!

config/typescript/tsconfig.base.json
```diff
@@ -1,33 +1,14 @@
...
-    "declarationMap": true,
-    "pretty": true,
-    "moduleResolution": "node",
-    "target": "es5",
-    "downlevelIteration": true,
-    "esModuleInterop": true,
-    "jsx": "react",
-    "isolatedModules": true,
-    "strictNullChecks": true,
+    "__TODOCOMMENT1": "Strict mode enables these strict* and noImplicit* options.",
```
ismail-syed, line 5: Interesting, what are these?

BPScott: Comments are valid in tsconfig files, but @atesgoral asked about keeping valid json; https://github.com/Shopify/quilt/pull/1811#discussion_r605777199 


So I made up some keys that do nothing.

Personally I'm not a fan and would prefer actual comments but it's not a hill i'm gonna die on

config/typescript/tsconfig.base.json
BPScott: Ah nope. TS type check says no:

```
config/typescript/tsconfig.base.json:6:5 - error TS5023: Unknown compiler option '__TODOCOMMENT2'.

6     "__TODOCOMMENT2": "Disable them temporarily while we fix up problematic behaviour from the time before we wanted strict mode",
      ~~~~~~~~~~~~~~~~
```

config/typescript/tsconfig.base.json
BPScott: Ah well, turns out arbitary strings cause type check to fail. so back to real comments.

atesgoral:
I followed the config trail to do a diligent review:

https://github.com/Shopify/web-configs/blob/main/packages/typescript-configs/library.json:

```json
{
  "extends": "./dom.json",
  "compilerOptions": {
    "declaration": true,
    "declarationMap": true,
    "noEmit": false,
    "noEmitHelpers": true
  }
}
```

./dom.json:

```json
{
  "extends": "./base.json",
  "compilerOptions": {
    "jsx": "react",
    "lib": [
      "dom",
      "dom.iterable",
      "es2017",
      "esnext.asynciterable",
      "scripthost"
    ],
    "module": "esnext",
    "target": "es2017"
  },
  "files": ["./definitions/images.d.ts", "./definitions/styles.d.ts"]
}
```

./base.json:

```json
{
  "compileOnSave": false,
  "compilerOptions": {
    "esModuleInterop": true,
    "isolatedModules": true,
    "experimentalDecorators": true,
    "lib": ["dom", "es2018", "esnext.asynciterable"],
    "moduleResolution": "node",
    "noUnusedLocals": true,
    "noUnusedParameters": true,
    "resolveJsonModule": true,
    "skipLibCheck": true,
    "strict": true
  }
}
```

Squashed:

```json
{
  "compileOnSave": false,
  "compilerOptions": {
    "esModuleInterop": true,
    "isolatedModules": true,
    "experimentalDecorators": true,
    "lib": ["dom", "es2018", "esnext.asynciterable"],
    "moduleResolution": "node",
    "noUnusedLocals": true,
    "noUnusedParameters": true,
    ?"resolveJsonModule": true,
    ?"skipLibCheck": true,
    ?"strict": true,
    "jsx": "react",
    ?"lib": [
      "dom",
      "dom.iterable",
      "es2017",
      "esnext.asynciterable",
      "scripthost"
    ],
    ?"module": "esnext",
    ?"target": "es2017",
    "declaration": true,
    "declarationMap": true,
    ?"noEmit": false,
    ?"noEmitHelpers": true
  },
  "files": ["./definitions/images.d.ts", "./definitions/styles.d.ts"]
}
```

These are the properties that are different/missing from the removed bits:

- [ ]  ?"resolveJsonModule": true,
- [ ]  ?"skipLibCheck": true,
- [ ]  ?"strict": true,
-  ?"lib": [
- [ ]    "dom",
- [ ]    "dom.iterable",
- [ ]    "es2017",
- [ ]    "esnext.asynciterable",
- [ ]    "scripthost"
-   ],
- [ ]  ?"module": "esnext",
- [ ]  ?"target": "es2017",
- [ ]  ?"noEmit": false,
- [ ]  ?"noEmitHelpers": true

Just want to ensure all of these are justified ✅

atesgoral:
LGTM! Thanks for the detailed explanations!

atesgoral (approved):

PR #1812, opened 2021-04-01
[react-testing] improved performance by making descendants calculations lazy for element children

melnikov-s:
## Description

Fixes (issue #)

N/A

## Type of change

I've been looking into a flaky slow test and doing some profiling and noticed that one of the significant bottlenecks is found in `react-testing` specifically when  attempting to flatten a react fiber linked list into an array of elements. It seems like determining the descendants of each child is an exponential time algorithm. This has a pretty big impact on tests with big fiber trees. I've changed the descendants to be lazily evaluated and derived from the children though grabbing the root descendants still goes through the entire tree which is slow. Additionally I hand optimized some code, for example the spread operator when compiled with typescript to es5 into a `spread` call seems to be a rather big culprit of performance issues in this computational heavy process.  

Performing the above optimization made a significant improvement to the test in question (which had a really big fiber tree ~80 in depth) :

Before:
Screen Shot 2021-04-01 at 9 18 39 AM

After:

Screen Shot 2021-04-01 at 9 19 47 AM

For other tests with smaller fiber trees the improvement is a more modest 10-20%.

Before:
Screen Shot 2021-04-01 at 9 35 06 AM

After:
Screen Shot 2021-04-01 at 9 36 02 AM

This should improve unit tests times across the board.

Note: A change in descendants order is a breaking one. I sketched out the order in this PR and it seems to be the same as before, but something worth verifying for the reviewers.

## Checklist

### DIFF
--- a/packages/react-testing/src/element.ts
+++ b/packages/react-testing/src/element.ts
@@ -45,6 +45,24 @@ export class Element<Props> implements Node<Props> {
     return this.elementChildren;
   }
 
+  private get elementDescendants() {
+    if (!this.descendantsCache) {
+      this.descendantsCache = getDescendants(this);
+    }
+
+    return this.descendantsCache;
+  }
+
+  private get elementChildren() {
+    if (!this.elementChildrenCache) {
+      this.elementChildrenCache = this.allChildren.filter(
+        element => typeof element !== 'string',
+      ) as Element<unknown>[];
+    }
+
+    return this.elementChildrenCache;
+  }
+
   get descendants() {
     return this.elementDescendants;
   }
@@ -71,23 +89,14 @@ export class Element<Props> implements Node<Props> {
     return domNodes[0] || null;
   }
 
-  private readonly elementChildren: Element<unknown>[];
-  private readonly elementDescendants: Element<unknown>[];
+  private descendantsCache: Element<unknown>[] | null = null;
+  private elementChildrenCache: Element<unknown>[] | null = null;
 
   constructor(
     private readonly tree: Tree<Props>,
     private readonly allChildren: (Element<unknown> | string)[],
-    allDescendants: (Element<unknown> | string)[],
     public readonly root: Root,
-  ) {
-    this.elementChildren = allChildren.filter(
-      element => typeof element !== 'string',
-    ) as Element<unknown>[];
-
-    this.elementDescendants = allDescendants.filter(
-      element => typeof element !== 'string',
-    ) as Element<unknown>[];
-  }
+  ) {}
 
   data(key: string): string | undefined {
     return this.props[key.startsWith('data-') ? key : `data-${key}`];
@@ -251,4 +260,19 @@ function equalSubset(subset: object, full: object) {
   return Object.keys(subset).every(
     key => key in full && full[key] === subset[key],
   );
+}
+
+function getDescendants(element: any) {
+  const descendants: Element<unknown>[] = [];
+  // eslint-disable-next-line @typescript-eslint/prefer-for-of
+  for (let i = 0; i < element.allChildren.length; i++) {
+    const child = element.allChildren[i];
+    if (typeof child !== 'string') {
+      descendants.push(child);
+      // eslint-disable-next-line prefer-spread
+      descendants.push.apply(descendants, child.elementDescendants);
+    }
+  }
+
+  return descendants;
 }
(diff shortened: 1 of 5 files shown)


packages/react-testing/src/element.ts
```diff
@@ -252,3 +258,14 @@ function equalSubset(subset: object, full: object) {
     key => key in full && full[key] === subset[key],
   );
 }
+
+function getDescendants(element) {
+  const descendants: Element<unknown>[] = [];
+  // eslint-disable-next-line @typescript-eslint/prefer-for-of
```
atesgoral, line 264: Could we actually use `for..of` here?

melnikov-s: I'm going to leave the for loop, the iterators are too much of a hit. Wonder if it would be faster if not compiled to es5 🤔 

for loop
Screen Shot 2021-04-01 at 2 31 13 PM

for of
Screen Shot 2021-04-01 at 2 30 37 PM

atesgoral: Oof.

packages/react-testing/src/root.tsx
```diff
@@ -284,52 +286,40 @@ function defaultRender(element: React.ReactElement<unknown>) {
   return element;
 }
 
-function flatten(
-  element: Fiber,
+function fiberToElement(
```
TheMallen, line 289: +1 for rename, the preact version of the library I did similar renames :)

TheMallen (commented):
The code changes look good to me and the perf improvements are very exciting, we should definitely test this in web (or another big codebase) so we can be sure it doesn't secretly break anything though.

packages/react-testing/src/element.ts
```diff
@@ -252,3 +261,18 @@ function equalSubset(subset: object, full: object) {
     key => key in full && full[key] === subset[key],
   );
 }
+
+function getDescendants(element: any) {
+  const descendants: Element<unknown>[] = [];
+  // eslint-disable-next-line @typescript-eslint/prefer-for-of
+  for (let i = 0; i < element.allChildren.length; i++) {
```
TheMallen, line 268: I wonder if we could change our compilation settings for this package to have the foreach version of this not get transpiled, this is a node library, and node definitely supports foreach.

melnikov-s: I've tried targeting `ES2015` and changing that function to a more simplified: 

```
function getDescendants(element: any) {
  const descendants: Element<unknown>[] = [];
  for (const child of element.allChildren) {
    if (typeof child !== 'string') {
      descendants.push(child, ...child.elementDescendants);
    }
  }

  return descendants;
}
```

but it is unfortunately considerably slower :(

`for of` with spread operator: 

Screen Shot 2021-04-05 at 11 42 26 AM

for loop with no spread:

Screen Shot 2021-04-05 at 11 42 05 AM

(There's a slight variance between runs. I run these tests several times, wait for them to stabilize and take the lowest run time)

melnikov-s:
Things are looking stable. I've ran all `tsx` tests in web and here are the results:

Before:
Screen Shot 2021-04-05 at 2 55 58 PM

After:
Screen Shot 2021-04-05 at 4 54 58 PM

The failures are the same for both and unrelated to this change, the run on the `main` branch has an additional timeout result which is why it has one more failure. 

One thing I've noticed is that there's a test `Globe.test.tsx` that's consistently passing on `main` and has intermittent failures on this branch. It has something to do with timing as the behavior should be the same between the two branches. I've added an artificial delay to the `fiberToElement` function and sure enough got the tests to consistently pass. `Globe.tsx` has a `setTimeout(fn, 0)` to flip a loading state for whatever reason so I think using jest fake timers and advancing that timer manually should fix the issue. Other then that, it's looking good.

atesgoral:
@melnikov-s HUGE!

TheMallen (commented):
@vsumner confirmed with a beta that the lazy evaluation change doesn't cause regressions in web, so that is good by me.

If you like you could pull the new er optimization into another PR and we can test that one with a beta too, and get the original one shipped ASAP

melnikov-s:
> If you like you could pull the new er optimization into another PR and we can test that one with a beta too, and get the original one shipped ASAP

Thanks, https://github.com/Shopify/quilt/pull/1821

TheMallen (approved):

melnikov-s:
> @melnikov-s could you please add a changelog entry? And then feel free to merge and I can cut a new release.

Will do, is this a breaking change that requires a major version bump? I'm noticing that `Element` is exported from the module and I've changed it's constructor signature.

TheMallen:
@melnikov-s I don't think its intended that folks manually construct elements, rather they can just use `Element` for type annotations in stuff like custom matchers if they need to, so my bias would be towards not doing a major bump

that said it is *technically correct* to do a major one I suppose

PR #1813, opened 2021-04-01
[react-hooks] Fix mutating refs during render phase or asynchronously

clauderic:
### Updating mutable references to callbacks

The `useEffect` hook runs asynchronously, after the rendering phase of a component is complete. When storing values in a mutable ref that should point to the latest argument passed to a component or hook, there can be a window of time between when the commit phase is kicked off and when the async `useEffect` function is called where it points to a stale callback.

To avoid this, we need to use `useLayoutEffect` to make sure to synchronously update the ref to the latest value during the commit phase.

`useLayoutEffect` cannot be used in a server-side environment, so we need to also introduce the `useIsomorphicLayoutEffect` hook to safely be able to use this strategy in a server-side rendered application.

We need to do this inside `useLayoutEffect` rather than directly in the body of the component for reasons described below.

### Mutating refs in the render phase

Refs should never be mutated while rendering. The `useLazyRef` component has been updated to a different strategy that does not involve mutating refs during the render phase of a component.

This is currently more of a technicality, as with current versions of React it is technically safe to do this, but this will no longer be safe with some of the upcoming changes that are coming to React.

As a rule of thumb, React components should be pure. They should only calculate the output of the next render or invoke hooks. Updating a ref (as well as updating state) shouldn’t be performed inside the immediate scope of a component’s render cycle for that reason.

In the future, React could decide to pause rendering a component, and potentially abort completely and re-start from scratch. Other examples include the `useTransition` hook of the Suspense API and future work around built-in animations.

Rendering should be treated as something that needs to be sandboxed and have no side-effects. You don't want the fact that React tried rendering something to leak out and have side-effects on mutable refs, for the same reason why you don't want to write to mutable state in a multithreaded environment. Exactly because it's hard to think about what can go wrong.

## Next steps

We should create a linting rule against mutating refs in the body of functional components.

## Type of change

- [x] `@shopify/react-hooks` Patch: Bug (non-breaking change which fixes an issue)
- [x]  `@shopify/react-hooks` Minor: New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/packages/react-hooks/src/hooks/isomorphic-layout-effect.ts
+++ b/packages/react-hooks/src/hooks/isomorphic-layout-effect.ts
@@ -1,0 +1,14 @@
+import {useEffect, useLayoutEffect} from 'react';
+
+// https://github.com/facebook/react/blob/master/packages/shared/ExecutionEnvironment.js
+const canUseDOM =
+  typeof window !== 'undefined' &&
+  typeof window.document !== 'undefined' &&
+  typeof window.document.createElement !== 'undefined';
+
+/**
+ * A hook that resolves to useEffect on the server and useLayoutEffect on the client
+ */
+export const useIsomorphicLayoutEffect = canUseDOM
+  ? useLayoutEffect
+  : useEffect;
--- a/packages/react-hooks/CHANGELOG.md
+++ b/packages/react-hooks/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+- Added `useIsomorphicLayoutEffect` hook [#1813](https://github.com/Shopify/quilt/pull/1813).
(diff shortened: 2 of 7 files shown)


alexandcote (approved):

vsumner (changes requested):
Great work! Could you please update the readme to include `useIsomorphicLayoutEffect`.

packages/react-hooks/src/hooks/isomorphic-layout-effect.ts
```diff
@@ -0,0 +1,14 @@
+import {useEffect, useLayoutEffect} from 'react';
+
+// https://github.com/facebook/react/blob/master/packages/shared/ExecutionEnvironment.js
+const canUseDOM =
+  typeof window !== 'undefined' &&
+  typeof window.document !== 'undefined' &&
+  typeof window.document.createElement !== 'undefined';
+
+/**
+ * A hook that resolves to useEffect on the server and useLayoutEffect on the client
```
dahukish, line 10: shouldn't this be a server effect for the server?

clauderic: I'm not sure what you're referring to. Are you talking about [`react-effect`](https://github.com/Shopify/quilt/blob/main/packages/react-effect/src/hook.ts#L6)? 

This change isn't related to running effects on the server. `useEffect` does not run in a server-side environment. `useLayoutEffect` throws a warning when used in a server-side environment, which is why we only invoke it when on the client.

vsumner (approved):

PR #1814, opened 2021-04-02
Enable strictPropertyInitialization and noImplicitThis checks in tsconfig

BPScott:
## Description

 Enable the strictPropertyInitialization and noImplicitThis checks in tsconfig.

Update code to allow for stricter typings.

## Type of change

- [x] address Patch: Bug (non-breaking change which fixes an issue)
- [x] function-enhancers Patch: Bug (non-breaking change which fixes an issue)
- [x] jest-dom-mocks Patch: Bug (non-breaking change which fixes an issue)
- [x] polyfills Patch: Bug (non-breaking change which fixes an issue)

## Checklist


packages/jest-dom-mocks/src/match-media.ts
```diff
@@ -3,35 +3,32 @@ export interface MediaMatching {
 }
 
 export default class MatchMedia {
-  originalMatchMedia: (mediaQuery: string) => MediaQueryList;
-  private isUsingMockMatchMedia = false;
+  originalMatchMedia: ((mediaQuery: string) => MediaQueryList) | null = null;
```
BPScott, line 6: By correctly typing `originalMatchMedia` as either a function or null (and setting its initial value to null) we can remove the concept of the `isUsingMockMatchMedia` property.

This is because we can consider that a mock is being used if `originalMatchMedia` is not null

packages/react-i18n/src/babel-plugin/index.ts
```diff
@@ -80,6 +81,7 @@ export default function injectWithI18nArguments({
         state.program = nodePath;
       },
       ImportDeclaration(
+        this: {file: BabelFile},
```
BPScott, line 84: This isn't complete, but it's all we need for now.

packages/react-i18n/src/babel-plugin/index.ts
```diff
@@ -115,6 +117,12 @@ export default function injectWithI18nArguments({
           ).name;
           const {filename} = this.file.opts;
 
+          if (typeof filename !== 'string') {
```
BPScott, line 120: According to `BabelFile` a filename may be undefined/null - such as when you get babel to parse content from stdin.  Here we add a typeguard to ensure that the filename is a string.

packages/jest-dom-mocks/src/match-media.ts
```diff
@@ -3,35 +3,32 @@ export interface MediaMatching {
...
 
   mock(media: MediaMatching = defaultMatcher) {
-    if (this.isUsingMockMatchMedia) {
+    if (this.originalMatchMedia !== null) {
```
TheMallen, line 9: fyi we usually just use `== null` and `!= null` over their stricter versions since generally speaking an `undefined` is fine too, but I don't feel that strongly.

PR #1818, opened 2021-04-06
export ErrorContent type from react-form

debchn:
## Description

Makes `ErrorContent` available for import directly from `@shopify/react-form`.

It's a useful type to have when building custom validators. In one case in web, it is already being imported by [reaching into dist](https://github.com/Shopify/web/blob/cf34d0b/app/sections/Payments/PaymentsSettings/ShopifyPaymentsDetails/formValidators.ts#L8)

## Type of change

- [X] react-form Patch: Bug (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/packages/react-form/src/validation/index.ts
+++ b/packages/react-form/src/validation/index.ts
@@ -1,2 +1,3 @@
+export type {ErrorContent} from './validator';
 export {validator} from './validator';
 export * from './validators';
--- a/packages/react-form/CHANGELOG.md
+++ b/packages/react-form/CHANGELOG.md
@@ -5,7 +5,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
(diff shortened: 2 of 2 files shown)


atesgoral (approved):

packages/react-form/src/validation/index.ts
```diff
@@ -1,2 +1,3 @@
+export type {ErrorContent} from './validator';
```
dahukish, line 1: ❤️

dahukish (approved):
:shipit:

sylvhama (approved):

packages/react-form/src/types.ts
```diff
@@ -9,6 +9,7 @@ export type DirtyStateComparator<Value> = (
 export interface Validator<Value, Context> {
   (value: Value, context: Context): ErrorValue;
 }
+export type {ErrorContent} from './validation';
```
BPScott, line 12: I don't think we need this?

The index file contains `export * from './types'; export * from './validation';` and you're already reexporting ErrorContent from validation.

debchn: Oops. Good point! I'll update.

PR #1820, opened 2021-04-06
Disable flaky test that takes a long time to run/fail

atesgoral:
## Description

Skip a flaky e2e test that takes a long time to run & fail.

## Type of change

Skipping a test

## Checklist


TheMallen (approved):
Maybe we should just 🔥  it entirely, or at least add a comment like "disabled because its flaky"

PR #1824, opened 2021-04-08
✨ custom watch/glob behaviour for graphql-typescript-definitions

GoodForOneFare:
## Description
In Spin environments, graphql-typescript-definitions's chokidar / glob setup causes out of memory errors, and is extremely slow even when it does succeed. Bumping memory limits is off the table for me, so I'm experimenting with other tools. This PR gets the package into a state where I can switch out tools without forking quilt packages (currently looking into watchman and vscode's filesystem).

### Did you figure out why these changes are even necessary?
I believe the slowness / memory pressure isn't happening in local development because ([per chokidar's docs](https://github.com/paulmillr/chokidar#how)):

> On MacOS, chokidar by default uses a native extension exposing the Darwin FSEvents API. This provides very efficient recursive watching... [o]n most other platforms... chokidar will initiate watchers recursively for everything within scope of the paths that have been specified, so be judicious about not wasting system resources by watching much more than needed.

Native FS support aside, we can't assume ultra-fast SSDs are available in cloud environments.

## Type of change

- [x] graphql-typescript-definitions Minor: New feature

## Checklist

### DIFF
--- a/packages/graphql-typescript-definitions/src/filesystem/default-graphql-filesystem.ts
+++ b/packages/graphql-typescript-definitions/src/filesystem/default-graphql-filesystem.ts
@@ -1,0 +1,74 @@
+import {FSWatcher, watch} from 'chokidar';
+import {GraphQLProjectConfig, GraphQLConfig} from 'graphql-config';
+import {
+  getGraphQLProjectIncludedFilePaths,
+  getGraphQLProjects,
+  getGraphQLSchemaPaths,
+  resolvePathRelativeToConfig,
+} from 'graphql-config-utilities';
+
+import {AbstractGraphQLFilesystem} from './graphql-filesystem';
+
+export class DefaultGraphQLFilesystem extends AbstractGraphQLFilesystem {
+  private readonly watchers: FSWatcher[] = [];
+
+  async watch(config: GraphQLConfig) {
+    this.watchers.push(
+      ...this.setupDocumentWatchers(config).concat(
+        this.setupSchemaWatcher(config),
+      ),
+    );
+
+    await Promise.all(
+      this.watchers.map(
+        watcher =>
+          new Promise<void>(resolve => watcher.on('ready', () => resolve())),
+      ),
+    );
+  }
+
+  dispose() {
+    this.watchers.forEach(watcher => {
+      watcher.close();
+    });
+
+    this.watchers.length = 0;
+  }
+
+  getGraphQLProjectIncludedFilePaths(projectConfig: GraphQLProjectConfig) {
+    return getGraphQLProjectIncludedFilePaths(projectConfig);
+  }
+
+  private setupDocumentWatchers(config: GraphQLConfig) {
+    return getGraphQLProjects(config)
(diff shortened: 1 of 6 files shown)


packages/graphql-typescript-definitions/src/filesystem/default-graphql-filesystem.ts
```diff
@@ -0,0 +1,74 @@
...
+  getGraphQLSchemaPaths,
+  resolvePathRelativeToConfig,
+} from 'graphql-config-utilities';
+
+import {AbstractGraphQLFilesystem} from './graphql-filesystem';
+
+export class DefaultGraphQLFilesystem extends AbstractGraphQLFilesystem {
+  private readonly watchers: FSWatcher[] = [];
+
+  async watch(config: GraphQLConfig) {
```
GoodForOneFare, line 15: Just migrating code from https://github.com/Shopify/quilt/blob/6488657a0287d6b9de9e8811f623e92de5b7f92c/packages/graphql-typescript-definitions/src/index.ts#L145-L157

packages/graphql-typescript-definitions/src/filesystem/default-graphql-filesystem.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    await Promise.all(
+      this.watchers.map(
+        watcher =>
+          new Promise<void>(resolve => watcher.on('ready', () => resolve())),
+      ),
+    );
+  }
+
+  dispose() {
```
GoodForOneFare, line 30: Original code https://github.com/Shopify/quilt/blob/6488657a0287d6b9de9e8811f623e92de5b7f92c/packages/graphql-typescript-definitions/src/index.ts#L184-L190

packages/graphql-typescript-definitions/src/filesystem/default-graphql-filesystem.ts
```diff
@@ -0,0 +1,74 @@
...
+    });
+
+    this.watchers.length = 0;
+  }
+
+  getGraphQLProjectIncludedFilePaths(projectConfig: GraphQLProjectConfig) {
+    return getGraphQLProjectIncludedFilePaths(projectConfig);
+  }
+
+  private setupDocumentWatchers(config: GraphQLConfig) {
```
GoodForOneFare, line 42: Adapting migrated code to trigger `change` / `delete` events https://github.com/Shopify/quilt/blob/6488657a0287d6b9de9e8811f623e92de5b7f92c/packages/graphql-typescript-definitions/src/index.ts#L207-L235

packages/graphql-typescript-definitions/src/filesystem/graphql-filesystem.ts
```diff
@@ -0,0 +1,69 @@
+import {EventEmitter} from 'events';
```
GoodForOneFare, line 1: So much boilerplate 😅

atesgoral (approved):
So much cleaner! <3

PR #1825, opened 2021-04-08
New import from lemonmade/quilt/packages/sewing-kit-plugins

dahukish:
https://github.com/lemonmade/quilt/tree/main/packages/sewing-kit-plugins

This PR imports the latest version of @lemonmade's work with sewing-kit-plugins in @quilted.

We essentially need to bring these back into the Shopify fold so that we can grow these in the direction needed for skn.

Quilted is being consumed by Checkout One and other projects owned by @lemonmade.
This was blocking our update path as Chris would be the one to make updates in the aforementioned repo and
is not scalable while we grow the next iteration of sewing-kit.

This was last started September of last year by @gmcgibbon but had not been fully completed.

🎩

install yalc
run yarn run build
then yalc publish (from within the packages/plugin-quilt folder)

next clone down alpaql
run yalc add @shopify/sewing-kit-plugins from within packages/alpaql-playground
swap import {quiltWebApp} from '@quilted/sewing-kit-plugins' for import {quiltWebApp}
from '@sewing-kit/plugin-quilt' in packages/alpaql-playground/sewing-kit.config.ts
then run yarn run dev at the root of the repo.
visit localhost:8081
💰

## Description

Fixes (issue #)

## Type of change

## Checklist


kaelig:
👋  Please run:

```
git pull origin main --rebase && git push origin import-latest-external-quilt-plugin --force
```

packages/sewing-kit-plugin-quilt/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "@shopify/sewing-kit-plugin-quilt",
+  "version": "0.0.1-import-latest-external-quilt-plugin-beta.2",
```
vsumner, line 3: From a test? I think this needs to be fixed.

vsumner (approved):
Minor version issue, but LGTM

packages/sewing-kit-plugin-quilt/src/web-app-convenience-aliases.ts
```diff
@@ -0,0 +1,41 @@
...
+            '^utilities/(.*)': project.fs.resolvePath('utilities/$1'),
+            '^tests/(.*)': project.fs.resolvePath('tests/$1'),
+          }));
+        });
+      });
+
+      function addWebpackAliases(aliases: {[key: string]: string}) {
+        return {
+          ...aliases,
+          components$: project.fs.resolvePath('components'),
```
BPScott, line 35: aliases, aliases everywhere!

Not a dealbreaker for this PR but having to add these aliases for jest,webpack,snowpack etc feels like a sign that we're doing something weird and it makes config of new tools more annoying. It's a bigger convo but what if we start telling people: your absolute imports must be in the form `web/components/.....` instead of `components/.....` and then that way we wouldn't need to write alias configs in N tools

packages/sewing-kit-plugin-quilt/src/web-app-multi-build.ts
```diff
@@ -0,0 +1,188 @@
...
+            ];
+          });
+
+          configuration.quiltBrowserslist?.hook(async () => {
+            if (browserGroupMap.has(browsers)) {
+              return browserGroupMap.get(browsers);
+            }
+
+            const {default: browserslist} = await import('browserslist');
+            const browserslistResult = browserslist(undefined, {
```
BPScott, line 146: Why do we need to resolve this here? I think most tools (babel certainly) can target a browsers list expression rather than needing an explicit list of browsers

BPScott (approved):
All looks good for now. 

I've probably got some thoughts on that alias wrangling, and how we can specify browsers lists but I don't think they need to block you.

packages/sewing-kit-plugin-quilt/src/web-app-convenience-aliases.ts
lemonmade: How would making the alias `web/components` instead of just `components` change the problem of aliases-per-tool? I recognize that there are some alternative ways of specifying "special roots" that the community has adopted (`@/...`, `$foo/...`, etc), though none are really close to standards IMO. I also think it would be nice for sewing-kit to have a more generic hook, maybe available by default, that is just `aliases`, and each tool could use those as the basis for their own aliases. But I don't think the aliases themselves are the problem.

BPScott: > How would making the alias web/components instead of just components change the problem of aliases-per-tool?

At the very least it would mean we'd need to configure a single alias for the "web" folder instead of having to configure aliases for every subfolder within it (which may change from project-to-project).

It would also reduce an issue I've recently found confusing where there's not a 1:1 mapping between import paths and files on disk and differing paths might get chosen depending on what files exist on disk. e.g.

-The import path `components/BLAH` could refer the file on disk `components/BLAH.ts` or  `web/components/BLAH.ts` depending on which exists
-The import path `web/components/BLAH` could also refer to the file `web/components/BLAH.ts`.

I've saw a couple of cases where importing from `app/components` and `components` are used interchangably as the both work. It'd be good to have a single approach, which is treating the root of the project as the root for imports and thus only `app/components` is valid and we don't need to write aliases for "components" etc.



 single import path has the potential to resolve to multiple locations. For instance in web `app/components/BLAH` and `components/

lemonmade: Everyone is certainly entitled to their own style, and I never documented the structure of what "acceptable" aliases in the project structure we promote through web and related projects, so ultimately this is on me, but it was not designed as a generic "everything at the top level is available", and it was never intended that you would import from `web/X` (unless maybe that's a result of a monorepo with multiple apps? I haven't been in the admin in awhile). In my original design, only the following aliases were supposed to work at all:

1. `components` is an alias to `app/components` (I am strongly against `components/X`, but that change was made in the admin without me ever being able to complain about how terrible the ergonomics are so I guess it's a lost cause :P)
2. `utilities/x` is an alias to `app/utilities/x` (ideally code should never reach "deeper" than that but there is no real way to prevent it in most cases :/)
3. `tests/utilities` is an alias to wherever the test utilities folder is
4. Assorted other ones for one-off things you need all over the app, like GraphQL types

I think it's important to keep aliases pretty tight because otherwise it is trivially easy to break the implicit boundaries part of the app have with one another, like making it very easy for a shared component to reach over into `foundation`, or having sections depend on one another's implementation details directly.

PR #1828, opened 2021-04-09
[react-form | DynamicList] add reset function and dirty state

oluwatimio:
## Description

Fixes https://github.com/Shopify/quilt/issues/1830 and part of https://github.com/Shopify/core-issues/issues/23975

There is no reset ability for dynamic list or flag to know if the dynamic list is dirty. This means if for example we are using a context bar and click discard, the dynamic list new field will still persist.

This PR adds the ability to reset to the initial list state and to know if the dynamic list is dirty.

It also adds the ability to add multiple dynamic lists

co-authored with @sylvhama 

## Tophat

Code sandbox [here](https://codesandbox.io/s/quirky-haze-9tewn) .

## Type of change

- [x] react-form/dynamic-list Minor: New feature (non-breaking change which adds functionality)

## Checklist


kaelig:
👋 Thank you for the PR!

Please run:

```
git pull origin main --rebase && git push origin dynamic-lists/add-reset-and-dirty-state --force
```

oluwatimio:
This PR will just be adding the reset functionality and dirty state to the `DynamicList` itself. I would like to keep this PR small so this PR wont focus on adding dynamic list reset to the `useForm` reset function as that will take some modifications to `FieldBag` and other files. I am thinking of exploring that part after this PR is merged. Consuming apps that need the reset or dirty state of the dynamic list can use them from the dynamic list directly

packages/react-form/CHANGELOG.md
```diff
@@ -5,7 +5,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+- Adds reset functionality and dirty state for dynamic list
```
attila-berki, line 10: It would be nice to link the PR as done below.

packages/react-form/src/hooks/list/baselist.ts
```diff
@@ -86,5 +93,13 @@ export function useBaseList<Item extends object>(
     });
   }, [state.list, handlers]);
 
-  return {fields, dispatch};
+  const fieldsDirty = fields.some(field => {
+    const keys: [keyof Item] = Object.keys(field) as [keyof Item];
```
sylvhama, line 97: Why do we need type casting? 
Could we replace by this?
```ts
return Object.entries(field).some(
  ([key, value]) => key === 'dirty' && value,
);
```

packages/react-form/src/hooks/list/baselist.ts
```diff
@@ -86,5 +93,13 @@ export function useBaseList<Item extends object>(
...
+
+    return keys.some(key => field[key].dirty);
+  });
+
+  const dirty = fieldsDirty || state.list.length !== state.initial.length;
```
sylvhama, line 102: We could have an early return if `state.list.length !== state.initial.length`, this would avoid to loop though each item for no purpose.

sylvhama (changes requested):
I have tested locally via `yalc` and it does not fix https://github.com/Shopify/quilt/issues/1830
It seems `useForm` is not aware that the dynamic list is dirty. If I add a new item **the form** remains not dirty. Plus when I press reset I would expect that the list becomes empty again but it does not.

oluwatimio:
> s not aware that the dyn

It fixes the issue if you use the dirty state from the dynamic list. Check https://github.com/Shopify/quilt/pull/1828#issuecomment-816845090

oluwatimio:
With this commit https://github.com/Shopify/quilt/pull/1828/commits/8a4cfa76f8fe6560c37fcd1bdbde47b9ba7c3e8b we can now reset and get the dirty state of the dynamic list from the forms reset and dirty state

packages/react-form/src/hooks/listreset.ts
```diff
@@ -0,0 +1,16 @@
+import {useCallback, useRef} from 'react';
+
+import {ListBag} from '../types';
+
+export function useListReset(lists?: ListBag) {
+  const listBagRef = useRef(lists);
```
oluwatimio, line 6: used a ref in other to prevent re-creating a new reset function on each render.

oluwatimio:
PR now ready for review. Resetting lists in a listbag with the form's reset function should now be possible with dynamic list or any type of list created in the future

packages/react-form/src/hooks/listdirty.ts
```diff
@@ -0,0 +1,5 @@
+import {ListBag} from '../types';
+
+export function useListDirty(lists?: ListBag) {
+  return lists ? lists.dynamicList.dirty : false;
+}
```
sylvhama, line 5:
```suggestion
export function useListDirty(lists: ListBag) {
  return lists.dynamicList.dirty;
}
```
I think we can assume lists is defined here otherwise there is no need to seek if it's dirty.
But this hook seems a bit overkilled no? do we really need it?

packages/react-form/src/types.ts
```diff
@@ -97,6 +100,10 @@ export interface FieldBag {
   [key: string]: FieldOutput<any>;
 }
 
+export interface ListBag {
+  dynamicList: DynamicList<any>;
```
sylvhama, line 104: What if I want to handle several dynamic lists? I think we need to find inspiration on how `FieldBag` works to allows that.

packages/react-form/src/hooks/listdirty.ts
sylvhama: Ah I guess this would be useful if we allow several dynamic lists. We would have to check each dirty state.

sylvhama (commented):
My biggest concern is the comment about `export interface ListBag`.
Otherwise it looks good, I have not checked all the tests since we might update several stuff first.

packages/react-form/src/types.ts
```diff
@@ -72,6 +74,7 @@ export interface Form<T extends FieldBag> {
   reset(): void;
   submit(event?: React.FormEvent): void;
   makeClean(): void;
+  lists?: ListBag;
```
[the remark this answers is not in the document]
oluwatimio: I am not sure if we should rename to `DynamicListBag` because we can possibly in future create a new type of list and want to do the exact same thing for that list. The way it is now, you can possibly add another type of list in the bag.

packages/react-form/src/hooks/form.ts
```diff
@@ -21,6 +23,9 @@ import {useSubmit} from './submit';
  *     fields: {
  *       title: useField('some default title'),
  *     },
+ *     lists: {
+ *       dynamicList: useDynamicList<Card>([{cardNumber: '4242 4242 4242 4242', cvv: '422'}], emptyCardFactory)
```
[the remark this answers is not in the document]
oluwatimio: Currently, the way we get the `dynamicList` is because the `ListBag` takes a dynamicList property. Hence, if you are passing in a dynamicList in the form it needs to be 

```typescript
lists: {dynamicList: useDynamicList}
```

packages/react-form/src/hooks/form.ts
```diff
@@ -50,36 +55,48 @@ import {useSubmit} from './submit';
  * @param fields - A dictionary of `Field` objects, dictionaries of `Field` objects, and lists of dictionaries of `Field` objects. Generally, you'll want these to be generated by the other hooks in this package, either `useField` or `useList`. This will be returned back out as the `fields` property of the return value.
  *
  * @param onSubmit - An async function to handle submission of the form. If this function returns an object of `{status: 'fail', error: FormError[]}` then the submission is considered a failure. Otherwise, it should return an object with `{status: 'success'}` and the submission will be considered a success. `useForm` will also call all client-side validation methods for the fields passed to it. The `onSubmit` handler will not be called if client validations fails.
+ * @param lists - Pass in list objects like dynamic list into the form
+ * @param makeCleanAfterSubmit
```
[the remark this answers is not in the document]
oluwatimio: Its not new. I just added the lists param above it so it makes it look new

packages/react-form/src/hooks/list/baselist.ts
```diff
@@ -86,5 +93,19 @@ export function useBaseList<Item extends object>(
...
+    return state.list.map(item => {
+      return mapObject(item, field => field.value);
+    });
+  }, [state.list]);
+
+  const fieldsDirty = fields.some(field => {
+    return Object.entries(field).some(
+      ([key, value]) => key === 'dirty' && value,
+    );
+  });
```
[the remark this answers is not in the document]
oluwatimio: Will try that out!! 🙂

packages/react-form/src/hooks/form.ts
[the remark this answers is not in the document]
oluwatimio: Actually, my IDE added the param there automatically on save. Seems like it was missing from this haha

packages/react-form/src/hooks/form.ts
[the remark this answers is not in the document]
oluwatimio: I think the card explanation is used throughout the README for dynamic list so wanted to sort of keep it consistent

packages/react-form/src/types.ts
oluwatimio: I am wondering if this is something we should do in this PR? 🤔 we might have to add some additional logic like resetting all the dynamic lists inside the bag basically we might need some type of reducer like FieldBag.

packages/react-form/README.md
```diff
@@ -742,12 +745,30 @@ You can then use this argument to do as you wish :).
...
+  },
+  onSubmit: async fieldValues => {
+    console.log(fieldValues);
+    return {status: 'success'};
+  },
+});
+```
+
+you can also initialize the form only with the dynamic list `FieldDictionary` but note you will not be able to use the form's reset function or dirty state for the `DynamicList` you will have to use
+the `DynamicList` reset and dirty state instead
```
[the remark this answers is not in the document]
oluwatimio: 😅

packages/react-form/src/hooks/listdirty.ts
oluwatimio: In that case I will do `lists && lists.dynamicList.dirty` and keep as optional param

packages/react-form/src/types.ts
sylvhama: We can discuss about it but I don't think it's okay to add an incomplete feature to a public package :/

oluwatimio: Just added this [here](https://github.com/Shopify/quilt/pull/1828/commits/b4ab1302741510cb30e08764745d54f2be1a92e9)

packages/react-form/src/hooks/form.ts
```diff
@@ -50,36 +60,56 @@ import {useSubmit} from './submit';
...
     setErrors([]);
     basicReset();
-  }, [basicReset, setErrors]);
+    dynamicListReset();
+  }, [basicReset, dynamicListReset, setErrors]);
 
-  const fieldsRef = useRef(fields);
-  fieldsRef.current = fields;
+  const fieldsRef = useRef(fieldsWithLists);
+  fieldsRef.current = fieldsWithLists;
```
TheMallen, line 112: I know this was already here, but a note on the pattern: 

This technically works but may stop working in future react updates. We recently removed similar patterns in [other hooks in quilt](https://github.com/Shopify/quilt/pull/1813) and are considering adding a [lint rule](https://github.com/Shopify/web-configs/issues/229). It might be best to find an alternative way of doing this as a result.

Actually, looking at it closer, since we update it literally every render, why can't we just use the value directly?

packages/react-form/src/hooks/list/baselist.ts
```diff
@@ -86,5 +94,20 @@ export function useBaseList<Item extends object>(
...
+  }, [state.list]);
+
+  const isBaseListDirty = useMemo(
+    () => !isEqual(listWithoutFieldStates, state.initial),
+    [listWithoutFieldStates, state.initial],
+  );
+
+  const fieldsDirty = useDirty({fields});
+
+  const dirty = fieldsDirty || isBaseListDirty;
```
TheMallen, line 110: nit: I wonder if some of these simpler values might actually be nicer just as part of the object literal we return, so there's less to hold in your head / scroll up and reference when you're trying to follow the code

packages/react-form/src/hooks/list/dynamiclist.ts
```diff
@@ -7,11 +7,13 @@ import {
 } from './hooks';
 import {useBaseList, FieldListConfig} from './baselist';
 
-interface DynamicList<Item extends object> {
+export interface DynamicList<Item extends object> {
```
TheMallen, line 10: I like exporting useful types 👍

packages/react-form/src/hooks/list/dynamiclistreset.ts
```diff
@@ -0,0 +1,18 @@
+import {useCallback, useRef} from 'react';
+
+import {DynamicListBag} from '../../types';
+
+export function useDynamicListReset(lists?: DynamicListBag) {
```
TheMallen, line 5: Happy to see this stuff broken up into composable hooks, very in-keeping with the spirit of the existing APIs :)

TheMallen (approved):
One thing I noticed that was already there, an a comment about some readability concerns, but otherwise this is looking pretty nice to me! I didn't tophat, but I'm happy to have this API in the library.

TheMallen:
One thing I didn't mention, probably we should update the docs as well.

packages/react-form/src/hooks/form.ts
oluwatimio: Ahh yes I think we can get rid of the ref since we update it, Thanks! Will try that

packages/react-form/src/hooks/form.ts
```diff
@@ -21,6 +29,9 @@ import {useSubmit} from './submit';
  *     fields: {
  *       title: useField('some default title'),
  *     },
+ *     lists: {
```
attila-berki, line 32: You forgot to update the property name here to `dynamicLists`

packages/react-form/src/hooks/form.ts
```diff
@@ -50,36 +61,56 @@ import {useSubmit} from './submit';
  * @param fields - A dictionary of `Field` objects, dictionaries of `Field` objects, and lists of dictionaries of `Field` objects. Generally, you'll want these to be generated by the other hooks in this package, either `useField` or `useList`. This will be returned back out as the `fields` property of the return value.
  *
  * @param onSubmit - An async function to handle submission of the form. If this function returns an object of `{status: 'fail', error: FormError[]}` then the submission is considered a failure. Otherwise, it should return an object with `{status: 'success'}` and the submission will be considered a success. `useForm` will also call all client-side validation methods for the fields passed to it. The `onSubmit` handler will not be called if client validations fails.
+ * @param lists - Pass in list objects like dynamic list into the form
```
attila-berki, line 64: Same as above ⬆️

packages/react-form/src/hooks/form.ts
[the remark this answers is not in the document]
oluwatimio: Ahh yes 😄 , thanks for spotting it!

attila-berki (approved):
LGTM, nice improvement of the form and the dynamic list 💯

packages/react-form/src/hooks/form.ts
```diff
@@ -41,45 +53,71 @@ import {useSubmit} from './submit';
...
-  const dirty = useDirty(fields);
-  const basicReset = useReset(fields);
+  const fieldsWithLists = useMemo(() => {
+    const fieldsWithList = {...fields};
+    if (dynamicLists) {
+      Object.entries(dynamicLists).forEach(([key, value]) => {
+        (fieldsWithList as any)[key] = value.fields;
+      });
+    }
+    return fieldsWithList;
```
sylvhama, line 98: This might be a micro optimization but to avoid the spreading when there is no dynamic list we could do
```suggestion
    if (dynamicLists) {
      const fieldsWithList = {...fields};
      Object.entries(dynamicLists).forEach(([key, value]) => {
        (fieldsWithList as any)[key] = value.fields;
      });
      return fieldsWithList;
    }
    return fields;
```

sylvhama (commented):
I have tested it in web and it works! However I have noticed a typing issue that should be addressed before merging I think:
Screen Shot 2021-04-16 at 4 31 35 PM

In your codesandbox you use `!` but I think consumers should not have to do this. We need to explore if we can tell TS to detect we have passed dynamic list(s) 🤔

sylvhama:
I think there must be a way e.g. https://www.typescriptlang.org/docs/handbook/2/conditional-types.html (I am not TS expert, this might not be the right thing to look at).

Also another problem comes to my mind, what if I only one to manipulate dynamic lists? then `fields` should become optional 😬

oluwatimio:
> I have tested it in web and it works! However I have noticed a typing issue that should be addressed before merging I think:
> Screen Shot 2021-04-16 at 4 31 35 PM
> 
> In your codesandbox you use `!` but I think consumers should not have to do this. We need to explore if we can tell TS to detect we have passed dynamic list(s) 🤔

In my opinion, the point of the non-null assertion operator is to tell the TypeScript compiler that you already know the variable is not null. I dont see any disadvantages to using it if you know that the dynamic list was passed in

oluwatimio:
We discussed this and we will be returning an empty DynamicList bag when dynamicLists param is undefined. This will have no effect on computation in the form and will remove having to use the assertion operator or checking if dynamic list is defined.

sylvhama:
Proposition to avoid default parameter: https://github.com/Shopify/quilt/pull/1848

PR #1829, opened 2021-04-09
Add build runtime and remove outdated config

BPScott:
## Description

While investigating rollup builds I found some polish that could be applied to our current process.

- Sewing-kit has a concept of "runtimes" that allows you to specify what a project can be compiled for. We weren't using this. While the `buildFlexibleOutputs` plugin doesn't leverage this too heavily, the rollup plugin can be configured to skip builds for irrelevant runtimes. e.g. there's no point doing an esnext build for a package like koa-liveness-ping that is node-only.
- Packages still had a build script referencing `tsc` - We don't use that anymore so it can be removed.
- Packages had a dependency on `tslib` - This is no longer the case as we don't compile with tsc anymore.

This PR:

- Adds runtimes to each package's sewing-kit.config.ts
- Removes all dependencies on `tslib` in project package.json files
- Removes all builds scripts using `tsc` in project package.json files

## Type of change

- [x] EVERY PACKAGE Patch: Bug (non-breaking change which fixes an issue)

## Checklist


kaelig:
👋  Please run:

```
git pull origin main --rebase && git push origin sknext-build-tidyups --force
```

ismail-syed (approved):
Awesome clean up work! 🎉

PR #1833, opened 2021-04-09
Forward ref of component rendered by Async

clauderic:
## Description

This PR adds ref forwarding for the component rendered by `Async` that is returned by the `createAsyncComponent` function. 

## Type of change

It's somewhat debatable whether this is a new feature or a bug fix. I'd tend to categorize the fact that refs are not being properly forwarded currently as a bug (unexpected behaviour for most consumers).

## Checklist

### DIFF
--- a/packages/react-async/src/component.tsx
+++ b/packages/react-async/src/component.tsx
@@ -1,4 +1,5 @@
 import React, {
+  forwardRef,
   useEffect,
   useRef,
   useCallback,
@@ -51,7 +52,8 @@ export function createAsyncComponent<
   Props extends object,
   PreloadOptions extends object = {},
   PrefetchOptions extends object = {},
-  KeepFreshOptions extends object = {}
+  KeepFreshOptions extends object = {},
+  Ref extends object = {}
 >({
   id,
   load,
@@ -87,7 +89,7 @@ export function createAsyncComponent<
     ? AssetTiming.CurrentPage
     : AssetTiming.Immediate;
 
-  function Async(props: Props) {
+  const Async = forwardRef<Ref | undefined, Props>((props, ref) => {
     const {resolved: Component, load, loading, error} = useAsync(resolver, {
       scripts: scriptTiming,
       styles: stylesTiming,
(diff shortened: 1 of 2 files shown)


packages/react-async/src/component.tsx
```diff
@@ -51,7 +52,8 @@ export function createAsyncComponent<
   Props extends object,
   PreloadOptions extends object = {},
   PrefetchOptions extends object = {},
-  KeepFreshOptions extends object = {}
+  KeepFreshOptions extends object = {},
+  Ref extends object = {}
```
clauderic, line 56: Not sure if there's a way to infer this type, I don't believe so but could have overlooked something

clauderic:
> I am biased against refs, since I have never used them for anything outside of UI component code (where you are generally not in a place to use the async library), so take this feedback with a grain of salt.
That's definitely fair, I share the sentiment, though I don't think that debate necessarily needs to be re-hashed here. There are legitimate use-cases and I don't think `react-async` should take an opinionated stance here.

> I find the use of ref here to be particularly confusing in this context, because a) there is no obvious signal we can send to the parent component for when this ref is going to be set, and b) the ref does not apply to loading or error state, which is maybe a sensible default, but seems like it leaves a pretty obvious edge case in solving this (IMO) existing edge case.

Agreed, I don't think there's an elegant solution around this possible given how the loading and error markup is rendered.

I also tend to think it would be acceptable to optimize around the success-scenario here for `refs`. If the component is loading or there was an error, the `ref` will be undefined. If the component has mounted successfully, refs should behave as they normally would.

clauderic:
In `online-store-web`, we recently went through a re-write of our `RichTextEditor` component from scratch. We're planning to run a beta to conditionally render the legacy / next versions.

We had normalized some public API methods for this component such as `blur` and `focus` behind an [imperative handle `ref`](https://reactjs.org/docs/hooks-reference.html#useimperativehandle). This was done to minimize duplication in how we render rich text settings.

Something like:

```tsx
function RichTextSetting() {
  const editorRef = useRef();
  const Component = betaEnabled ? NextRichTextEditor : LegacyRichTextEditor;

  return (
     <Label onClick={() => editorRef.current.focus()}>
         <Component ref={editorRef} />
     </Label>
   )
}
```

Both of these components weigh upwards of 30kb so they are async imported. Our code broke when we changed to async imports for these components.

This limitation forced us to change our approach and have lots of duplicated logic in both `LegacyRichTextEditor` and `RichTextEditorNext` instead.

BPScott:
Its gone stale. lets close it unless @clauderic wants to try to bring this back.

PR #1835, opened 2021-04-12
🐛 Use import type to fix sewing-kit-koa mjs compilation

GoodForOneFare:
## Description
esbuild can't compile the mjs build of sewing-kit-koa because it references a TypeScript type. This switches `Assets` to be imported via `import type`, which gets it stripped out of the mjs build.

## Type of change

- [x] sewing-kit-koa Patch: Bug (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/packages/sewing-kit-koa/CHANGELOG.md
+++ b/packages/sewing-kit-koa/CHANGELOG.md
@@ -10,6 +10,7 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 ### Changed
 
 - Removed dependency on tslib, as we no-longer compile with `tsc`. [#1829](https://github.com/Shopify/quilt/pull/1829)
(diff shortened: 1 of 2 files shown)


BPScott:
This looks like a bug in esbuild.

I would expect that if an item is only used as a type its import should be removed.

I believe that's the case where it's used as a type e.g. 

```
import {aValue, AType} from './blah';

export const x: AType = aValue;
```

should compile to:

```
import {aValue} from './blah';

export const x = aValue;
```

I would exepect the same should occur when the type is used as an `export type`:

```
import {aValue, AType} from './blah';

export const x = aValue;
export type {AType};
```

should compile to:

```
import {aValue} from './blah';

export const x = aValue;
```

GoodForOneFare:
@BPScott why would esbuild process an mjs file using its TypeScript parser?

PR #1837, opened 2021-04-13
[react-form] reset dynamic list

### LINKED ISSUE #1830: useDynamicList does not make the form dirty and has reset issues
## Overview

This issue regroups two problems (we could split them in two tickets if needed). 

- adding / removing an element does not make the form dirty;
- You can reset the stuff you edited in one item, but you cannot reset the action of adding or removing.

[CodeSandbox](https://codesandbox.io/s/vigorous-ramanujan-e9reo?file=/src/App.tsx) to try those bugs.

sylvhama:
## Description

Fixes #1830

Inspired by https://github.com/Shopify/quilt/pull/1828

## Type of change

It does a full reset of a dynamic list instead of resetting each existing field.
So items that have been removed will reappear and items that have been added will disappear. 

- [x] `react-form` Patch: Bug (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/packages/react-form/src/hooks/list/hooks/handlers.ts
+++ b/packages/react-form/src/hooks/list/hooks/handlers.ts
@@ -13,6 +13,7 @@ import {
   updateErrorAction,
   newDefaultAction,
   resetAction,
+  resetListAction,
   ListState,
   ListAction,
 } from './index';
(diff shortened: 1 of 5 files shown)


packages/react-form/src/hooks/list/hooks/handlers.ts
```diff
@@ -69,7 +70,11 @@ export function useHandlers<Item extends object>(
               }
             },
             reset() {
-              dispatch(resetAction({target}));
+              if (state.isDynamic) {
+                dispatch(resetListAction());
+              } else {
+                dispatch(resetAction({target}));
```
sylvhama, line 76: Problem with this: it will reset the whole list for each item :/ Maybe we could find a better way to avoid performances caveats?

oluwatimio: I dont think we should do this because it means you will not be able to reset individual items if you are using a dynamic list. Also I feel like that would be taking away some functionality that might break flows on people already using DynamicList

PR #1838, opened 2021-04-14
Fix react form state mjs

GoodForOneFare:
## Description
Fix parsing of rect-form-state's index.mjs. esbuild is throwing:
```
Error: Client generation failed with errors: [{"location":{"column":24,"file":"node_modules/@shopify/react-form-state/build/esm/index.mjs","length":21,"line":2,"lineText":"import { asChoiceField, ChoiceFieldDescriptor, push, replace, remove } from \"./utilities.mjs\";","namespace":"","suggestion":""},"notes":[],"text":"No matching export in \"node_modules/@shopify/react-form-state/build/esm/utilities.mjs\" for import \"ChoiceFieldDescriptor\""}]
```

To confirm that a stale type is lingering around in the distributed mjs file, [here's unpkg's view of this](https://unpkg.com/browse/@shopify/react-form-state@0.12.4/build/esm/index.mjs):
```ts
import { asChoiceField, ChoiceFieldDescriptor, push, replace, remove } from "./utilities.mjs";
```

## Type of change
- [x] react-form-state Patch: Bug (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/packages/react-form-state/CHANGELOG.md
+++ b/packages/react-form-state/CHANGELOG.md
@@ -5,7 +5,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+- [Patch] Remove TypeScript type from distributed mjs [#1838](https://github.com/Shopify/quilt/pull/1838)
 
 ## 0.12.4 - 2021-04-13
 
--- a/packages/react-form-state/src/index.ts
+++ b/packages/react-form-state/src/index.ts
@@ -1,11 +1,6 @@
 import FormState from './FormState';
-import {
-  asChoiceField,
-  ChoiceFieldDescriptor,
-  push,
-  replace,
-  remove,
-} from './utilities';
+import type {ChoiceFieldDescriptor} from './utilities';
+import {asChoiceField, push, replace, remove} from './utilities';
 
 export {asChoiceField};
 export type {ChoiceFieldDescriptor};


BPScott:
This seems like the same sort of problem as https://github.com/Shopify/quilt/pull/1835 - there are cases where babel isn't removing type imports. I worry that this is be systemic, and I'd like to see if we can be confident of fixing this all in one go instead of having to go around fixing these issues one at a time.

[Babel REPL](https://babeljs.io/repl#?browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=PTAEEsFsAcHsCcAupEE9oFNQG8CCBnfDRAX1ADN5ZJQByAOmAENDj9aBuAKBAhgWQEiifBSo0GzViM5cocJKCFsANDmWkx1OoxbD23LhgAeC5Hmn4S3E2ZTosF4da5A&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=typescript&prettier=false&targets=&version=7.13.15&externalPlugins=) seems to do the right thing so I wonder if this is an artifact of our babel version, and a version update will fix this.

I see two potential solutions:

1) Updating babel fixes the issue and it starts correctly removing unused type imports
2) We set [`"importsNotUsedAsValues": "error"`](https://www.typescriptlang.org/tsconfig#importsNotUsedAsValues), in [tsconfiig.base.json](https://github.com/Shopify/quilt/blob/main/config/typescript/tsconfig.base.json) to enforce that we always use `import type` when importing types across the whole codebase

I'm not too fussed which one we go for. I enabled for `"importsNotUsedAsValues": "error"` in polaris-react but since I did that I feel less strongly about that being a general recommendation - forcing devs to do something that the compiler should already handle without that extra hint feels a bit heavy handed.

PR #1839, opened 2021-04-14
[react-i18n] adjusted types to better reflect runtime

melnikov-s:
## Description

Fixes (issue): N/A

Found this issue during development, I had a type of `string | undefined` and used it as one of the replacements to `i18n.translate`. During runtime the returned type was of `string` but TS reported it as a react element. 

This PR aligns the types to what is actually returned at run time.

Also when a react node is used in one of the replacements the return type is now correctly reported as `(string | React.ReactElement<any>)[]`. (it was just `React.ReactElement<any>` before) 

## Type of change

## Checklist


melnikov-s:
> @melnikov-s Can you explain why this happens? I can't connect the dots...

The `translate` method has several overloads

```javascript
  translate(id: string, replacements?: PrimitiveReplacementDictionary): string;
  translate(
    id: string,
    replacements?: ComplexReplacementDictionary,
  ): React.ReactElement<any>;
```

and based on whether or not you pass in a `PrimitiveReplacementDictionary` or a `ComplexReplacementDictionary` it will determine if the return type is a react element or a string. 

Where those two types were defined as:

```javascript
export interface PrimitiveReplacementDictionary {
  [key: string]: string | number;
}

export interface ComplexReplacementDictionary {
  [key: string]: string | number | React.ReactNode; // <-- I think this can be simplified to just `[key: string]: React.ReactNode`
}
```

since a `ReactNode` can be just about anything (boolean, undefined, null, string, number, etc etc) if we have a  type of `string | undefined` in our replacement dictionary it will fall into the `ComplexReplacementDictionary` since  it satisfies those constraints but not the `PrimitiveReplacementDictionary` as it's missing `undefined` . So TS will infer the return type to be a `React.ReactElement<any>` based on the overload rules. 

During runtime though we use this to determine what to return

```javascript
 const allReplacementsArePrimitives = Object.keys(replacements).every(
    key => typeof replacements[key] !== 'object',
  );
```

so if all replacement are primitives we return a `string`, otherwise an array of react nodes.  Thus the types vs runtime mismatch. 

Looking at it again I think there's another type issue here, the translate method actually returns an array of react elements and not a single react element. It should really be:

` translate(
    id: string,
    replacements?: ComplexReplacementDictionary,
  ): React.ReactElement<any>[];`

probably went unnoticed since the vast majority of the time the results are used as a `ReactNode` which can be either

ismail-syed (commented):
These changes sound sane to me but I'd like to verify that this doesn't break type-check CI in web. You can run a [beta](https://github.com/Shopify/quilt/blob/main/documentation/guides/release-and-deploy.md#releases-for-ing-changes) to verify.

packages/react-i18n/src/i18n.ts
```diff
@@ -134,13 +134,13 @@ export class I18n {
     id: string,
     options: TranslateOptions,
     replacements?: ComplexReplacementDictionary,
-  ): React.ReactElement<any>;
+  ): (string | React.ReactElement<any>)[];
```
dahukish, line 137: ❤️

melnikov-s:
I did not run a beta but I ran it locally and the first commit (https://github.com/Shopify/quilt/pull/1839/commits/6805e9882460f8f89a420bf028a4fe30dd031f12) was OK. 3370f6f7ee1a906fb1bb742c05e2d7ac18569f80 Is the one that caused issues. About ~50 type check failures. 

They're pretty much all because of 

a) We specified a return type of `JSX.Element` where `ReactNode` was meant
b) `() => []` to my suprise can't be used as a JSX element even though react will gladly accept it. (see https://github.com/microsoft/TypeScript/issues/33487). This is an issue because we have several components that are just

```javascript;
function MyComponent() {
   const i18n = usei18n();

   return i18n.translate(\* ...*\);
}
``` 

The fix here would be:
```javascript;
function MyComponent() {
   const i18n = usei18n();

   return <>i18n.translate(\* ...*\)</>;
}
``` 

So some effort will be required to fix all the type errors along with the PR that would bump `react-i18n` with these changes.

mateus:

PR #1843, opened 2021-04-16
Factors out the formatting methods from Address

tapi:
## Description
Factors the formatting methods out of AddressFormatter so that they can be called synchronously for use in POS RN

Discussion [here](https://shopify.slack.com/archives/CNH0GMZ41/p1618595344366400)
TL;DR
We want to use `@shopify/address` in POS RN but have some additional constraints, namely
- We need to be able to pre-warm the cache from a file on disk in case the merchant is offline.
- We may also need to clear the mach in the event the device sends a memory warning to the app.

Since we will have our own cache we are not using AddressFormatter for loading but would still like to take advantage of the formatting functions provided. This PR factors them out so they can be used by POS in this fashion.

## Type of change

- [ ] Address Minor: Existing API functions as it did before. Actual formatting logic is factored out so it is no longer necessary for the `Country` to be loaded from the network.

## Checklist


packages/address/CHANGELOG.md
```diff
@@ -5,7 +5,8 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+- Factored out formatting functions so they can be used by POS. [#1843](https://github.com/Shopify/quilt/pull/1843)
```
atesgoral, line 10:
```suggestion
## Unreleased

- Factored out formatting functions so they can be used synchronously [#1843](https://github.com/Shopify/quilt/pull/1843)

```

packages/address/src/AddressFormatter.ts
```diff
@@ -101,3 +88,25 @@ export default class AddressFormatter {
...
+  return format.split(LINE_DELIMITER).map(fields => {
+    const result = fields.match(FIELD_REGEXP);
+    if (!result) {
+      return [];
+    }
+    return result.map(field => {
+      return FIELDS_MAPPING[field];
+    });
+  });
+}
```
atesgoral, line 112:
```suggestion
}

```

tapi:
Not sure what is causing the test failures they pass when running locally

atesgoral:
@tapi A known issue with a missing `flatMap` polyfill on Node.js 10 -- we're currently wrestling with it.

packages/address/src/AddressFormatter.ts
```diff
@@ -55,10 +55,7 @@ export default class AddressFormatter {
    */
   async format(address: Address): Promise<string[]> {
     const country = await this.getCountry(address.country);
-    const layout = country.formatting.show || DEFAULT_SHOW_LAYOUT;
-    return layout
-      .split(LINE_DELIMITER)
-      .map(fields => renderLineTemplate(country, fields, address).trim());
+    return format(address, country);
```
dahukish, line 58: Can we rename this so there is less confusion with the async `format` above? `buildFormat` maybe?

packages/address/src/AddressFormatter.ts
```diff
@@ -101,3 +88,23 @@ export default class AddressFormatter {
     return null;
   }
 }
+
+export function format(address: Address, country: Country): string[] {
```
dahukish, line 92: Love to see bigger function become smaller ❤️ . Can we add a test for this?

tapi:
Just a question as it bit me during testing. For country code are we using ISO-3166-1? The reason I asked is I had blindly assumed that the code in the existing tests for `format` are correct. 
The existing tests specify 'JA' as the country code for Japan in the `AddressFormatter` which is correct only if the code accepted in the constructor is [FIPS, LOC MARC, Aircraft or NATO (obsolete)](https://en.wikipedia.org/wiki/Country_codes:_J–K#_Japan)

Functionally it makes no difference as JA is mapped to JA, but it does impact readability if everything is supposed to be ISO. I can fix it, just want to make sure there isn't some corner case context I'm missing.

tapi:
iOS ships with a copy of countries on disk and (I think) persists any updates back to disk. There's no documentation as to why, I assume this is to not prevent the use of country data in the event of offline (its used in more than just StC) or if the service is down for any reason.

tapi:
> Just a question as it bit me during testing. For country code are we using ISO-3166-1? The reason I asked is I had blindly assumed that the code in the existing tests for `format` are correct.
> The existing tests specify 'JA' as the country code for Japan in the `AddressFormatter` which is correct only if the code accepted in the constructor is [FIPS, LOC MARC, Aircraft or NATO (obsolete)](https://en.wikipedia.org/wiki/Country_codes:_J%E2%80%93K#_Japan)
> 
> Functionally it makes no difference as JA is mapped to JA, but it does impact readability if everything is supposed to be ISO. I can fix it, just want to make sure there isn't some corner case context I'm missing.

Pass me the stupid hat, Im mixing up country and locale codes

jarredszabadi:
@tapi Is this still something you're interested in pushing and is there any mobile work to do as a result of this? Asking to see if you want to keep this open in the foundations project or if you want to close it.

tapi:
Nah, let ax it. We didn't use it in StC and the only reason for it was to improve code reuse

emileber:
See #2178

PR #1845, opened 2021-04-19
Fix type exports appearing in distributed .mjs files

GoodForOneFare:
[As requested in a previous issue](https://github.com/Shopify/quilt/pull/1838#issuecomment-819211745), this PR bumps Babel dependencies to fix exported types appearing as concrete variables in .mjs files.


BPScott:
Looks like the root issue was https://github.com/babel/babel/pull/12122 which was released in  7.12.0

PR #1848, opened 2021-04-20
improve dynamicLists typing

sylvhama:
You can test with [this condesandbox](https://codesandbox.io/s/fancy-resonance-eh569?file=/src/App.tsx).
I am using function overloading to define two scenarios:

1. we have a classic `useForm` (like all existing ones) that does not use `dynamicLists`;
2. we can a `useForm` relying `dynamicLists`.

Result is:
- No more default value in our hook;
- In `App.tsx`: No need to use `!` or to check if `dynamicLists` is defined.

In the future we could do the same to also turn `fields` optional 🤔

### DIFF
--- a/packages/react-form/src/hooks/form.ts
+++ b/packages/react-form/src/hooks/form.ts
@@ -2,10 +2,12 @@ import {useCallback, useMemo} from 'react';
 import {useLazyRef} from '@shopify/react-hooks';
 
 import {
-  SubmitHandler,
-  FormMapping,
   FieldBag,
+  FormInput,
+  FormWithDynamicListsInput,
+  FormWithoutDynamicListsInput,
   Form,
+  FormWithDynamicLists,
   DynamicListBag,
 } from '../types';
 import {validateAll, makeCleanFields} from '../utilities';
@@ -13,7 +15,7 @@ import {validateAll, makeCleanFields} from '../utilities';
 import {useDirty} from './dirty';
 import {useReset} from './reset';
 import {useSubmit} from './submit';
-import {useDynamicList, useDynamicListDirty, useDynamicListReset} from './list';
+import {useDynamicListDirty, useDynamicListReset} from './list';
 
 /**
  * A custom hook for managing the state of an entire form. `useForm` wraps up many of the other hooks in this package in one API, and when combined with `useField` and `useList`, allows you to easily build complex forms with smart defaults for common cases.
@@ -80,14 +82,22 @@ import {useDynamicList, useDynamicListDirty, useDynamicListReset} from './list';
 export function useForm<T extends FieldBag>({
   fields,
   onSubmit,
+  makeCleanAfterSubmit,
+}: FormWithoutDynamicListsInput<T>): Form<T>;
+
+export function useForm<T extends FieldBag, D extends DynamicListBag>({
+  fields,
   dynamicLists,
+  onSubmit,
+  makeCleanAfterSubmit,
+}: FormWithDynamicListsInput<T, D>): FormWithDynamicLists<T, D>;
+
+export function useForm<T extends FieldBag, D extends DynamicListBag>({
+  fields,
+  dynamicLists,
+  onSubmit,
   makeCleanAfterSubmit = false,
-}: {
-  fields: T;
-  onSubmit?: SubmitHandler<FormMapping<T, 'value'>>;
-  makeCleanAfterSubmit?: boolean;
-  dynamicLists?: DynamicListBag;
-}): Form<T> {
+}: FormInput<T, D>) {
   const fieldsWithLists = useMemo(() => {
     if (dynamicLists) {
       const fieldsWithList = {...fields};
@@ -128,13 +138,7 @@ export function useForm<T extends FieldBag>({
     fieldsRef,
   ]);
 
-  const defaultDynamicListBag = {
-    defaultDynamicList: useDynamicList<{}>([], () => {
-      return {};
-    }),
-  };
-
-  return {
+  const form: Form<T> = {
     fields,
     dirty: dirty || dynamicListDirty,
     submitting,
@@ -143,6 +147,11 @@ export function useForm<T extends FieldBag>({
     validate,
     makeClean,
     submitErrors: errors,
-    dynamicLists: dynamicLists || defaultDynamicListBag,
   };
+
+  if (dynamicLists) {
+    return {...form, dynamicLists};
+  }
+
+  return form;
 }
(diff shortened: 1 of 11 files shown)


packages/react-form/src/hooks/form.ts
```diff
@@ -80,14 +82,22 @@ import {useDynamicList, useDynamicListDirty, useDynamicListReset} from './list';
...
+}: FormWithoutDynamicListsInput<T>): Form<T>;
+
+export function useForm<T extends FieldBag, D extends DynamicListBag>({
+  fields,
+  dynamicLists,
+  onSubmit,
+  makeCleanAfterSubmit,
+}: FormWithDynamicListsInput<T, D>): FormWithDynamicLists<T, D>;
+
+export function useForm<T extends FieldBag, D extends DynamicListBag>({
```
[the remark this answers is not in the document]
sylvhama: nope, as you can see in the codesandbox (App.tsx), I made two form and I don't need to pass the generic types.

oluwatimio (approved):
I think this is good! Nice 💯 . Thanks for digging into it 🙂 . I am a bit concerned about modifying the `useForm` this way but it's looking good in the sandbox! 

I also noticed all the tests in `form.test.tsx` are breaking. The full CI does not seem to be running in this PR for some reason but you can try it locally. I can fix the tests in my PR so dont worry about it.

sylvhama:
Aaaaah I thought it was all green LOL nice catch! I will run tests locally and see what's going on.

oluwatimio:
I am 

> LOL nice catch! I will run tests locally a

Its all good! We just have to change the useForm call to something like this

```
    const {
      submit,
      submitting,
      dirty,
      reset,
      makeClean,
      submitErrors,
      dynamicLists,
    } = useForm({
      fields: {title, description, defaultVariant, variants},
      dynamicLists: {
        dynamicVariants: useDynamicList(data.variants, variantsEmptyFactory),
      },
      onSubmit: onSubmit as any,
      makeCleanAfterSubmit,
    });

    const {
      dynamicVariants: {addItem, fields: dynamicListFields},
    } = dynamicLists;
```

and maybe remove the enable dynamic list prop in productForm

packages/react-form/src/hooks/list/test/dynamiclist.test.tsx
```diff
@@ -73,6 +79,9 @@ describe('useDynamicList', () => {
     });
 
     it('will throw an error if new position is out of index range', () => {
+      const consoleErrorMock = jest.spyOn(console, 'error');
+      consoleErrorMock.mockImplementation();
```
sylvhama, line 83: To avoid 
Screen Shot 2021-04-21 at 10 37 03 AM

packages/react-form/src/hooks/list/test/dynamiclist.test.tsx
```diff
@@ -203,6 +212,66 @@ describe('useDynamicList', () => {
         expect(wrapper).toContainReactComponent(TextField);
       });
     });
+
+    describe('dirty dynamic list', () => {
```
sylvhama, line 216: So far there was no dirty related test

packages/react-form/src/hooks/test/form-with-dynamic-list.test.tsx
```diff
@@ -0,0 +1,145 @@
...
+  isDirty,
+  fakeProduct,
+  hitSubmit,
+  hitReset,
+  waitForSubmit,
+} from './utilities';
+
+import {submitSuccess, submitFail} from '..';
+
+describe('useForm with dynamic list', () => {
```
sylvhama, line 16: those tests are inspired by origin `useForm` test. 

`makeClean` and `makeCleanAfterSubmit` seem broken with dynamic lists 🤔

packages/react-form/src/hooks/test/form.test.tsx
```diff
@@ -1,200 +1,21 @@
...
+  fakeProduct,
+  hitSubmit,
+  hitReset,
+  hitClean,
+  waitForSubmit,
+} from './utilities';
+
+import {submitSuccess, submitFail} from '..';
 
 describe('useForm', () => {
```
sylvhama, line 18: All the code removed here has been moved to `utilities`.

packages/react-form/src/hooks/test/utilities/components/FormWithDynamicVariantList.tsx
```diff
@@ -0,0 +1,95 @@
+import React from 'react';
+
+import {SubmitHandler} from '../../../../types';
+import {notEmpty} from '../../../../validation';
+import {useDynamicList, useField, useForm} from '../../..';
+
+import {SimpleProduct, Variant} from './types';
+import {TextField} from './TextField';
+
+export function FormWithDynamicVariantList({
```
sylvhama, line 10: This is a new component to test dynamic lists.

packages/react-form/src/hooks/test/utilities/components/ProductForm.tsx
```diff
@@ -0,0 +1,82 @@
+import React from 'react';
+
+import {SubmitHandler} from '../../../../types';
+import {positiveNumericString, notEmpty} from '../../../../validation';
+import {useList, useField, useForm} from '../../..';
+
+import {SimpleProduct} from './types';
+import {TextField} from './TextField';
+
+export function ProductForm({
```
sylvhama, line 10: Nothing changed here, I just moved the code.

packages/react-form/src/hooks/list/test/dynamiclist.test.tsx
oluwatimio: hmm 🤔 @sylvhama . There are dirty tests in the baselist since that is where it is originally implemented

oluwatimio: So I dont think we will need to have it in the dynamiclist test

oluwatimio: same case for reset

sylvhama: But `useList` test don't cover cases where we add or remove an item. Plus it does not hurt to make sure nothing break the default behavior.

oluwatimio: Oh I dont mean`useList`, I meant `useBaseList`. Because I noticed in this PR some tests like value changes are being re-created. In that case, could we just write the tests that cover the adding and removing?

PR #1850, opened 2021-04-20
Add core-js polyfills for node 10

dahukish:
## Description

Fixes (issue #)

Fixes main branch (breaking on node 10) https://github.com/Shopify/quilt/runs/2385009314

## Type of change

## Checklist


vsumner (approved):
LGTM. Is it worth adding a note that this should be removed we node 10 is no longer supported?

PR #1858, opened 2021-04-22
fix bad build cache key

dahukish:
## Description

Fixes (issue #)

Failing CI due to bad build cache

## Type of change

## Checklist


BPScott:
Do we need too add this in node-e2e-tests and node-build too/

atesgoral (approved):

PR #1860, opened 2021-04-22
Kill node 10 polyfills

dahukish:
## Description

Fixes (issue #)

Reverts: https://github.com/Shopify/quilt/pull/1850

## Type of change

## Checklist


atesgoral:
Could we do a revert PR instead?

dahukish:
> Could we do a revert PR instead?

We can I just didn't want to revert for no reason (if it didn't fix it)

PR #1865, opened 2021-04-23
[react-form] include `dynamicLists` values in `onSubmit` type

sylvhama:
## Description

Follow up of https://github.com/Shopify/quilt/pull/1828

We noticed that `onSubmit` does not include the values from the dynamic lists.

Thanks to that PR, when using `dynamicLists` you can do something like this:
```ts
useForm<typeof fields, typeof dynamicLists>({...})
```
to make sure `onSubmit` will infer the values types correctly.

I have this [codesandbox](https://codesandbox.io/s/affectionate-snow-69xsx?file=/src/App.tsx) that contains a hardcoded version of `react-form`. Nothing is rendered, the goal is to hover your mouse on variables and see their types.

Screen Shot 2021-04-23 at 5 26 10 PM

Before the fix TS would not include dynamic lists values in `fieldValues`.

## Type of change

- [x] react-form Patch: Bug (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/packages/react-form/README.md
+++ b/packages/react-form/README.md
@@ -644,7 +644,7 @@ A custom hook that adds on `useList` the ability to dynamically add, move and re
 
 Let's simulate a user interface that allows you add as many credit cards as you wish.
 
-You can initialize `useDynamicList` the following way
+You can initialize `useDynamicList` the following way:
 
 ```tsx
 interface Card {
@@ -654,7 +654,7 @@ interface Card {
 }
 
 const emptyCardFactory = (): Card => ({
-  id: new Date.now().toString(),
+  id: Date.now().toString(),
   cardNumber: '',
   cvv: '',
 });
@@ -665,11 +665,11 @@ const {fields, addItem, removeItem, moveItem, reset, dirty} = useDynamicList(
 );
 ```
 
-You can also have a factory that produces multiple cards such as
+You can also have a factory that produces multiple cards such as:
 
 ```tsx
 function generateId() {
-  return new Date.now().toString();
+  return Date.now().toString();
 }
 
 const emptyCardFactory = (): Card[] => {
@@ -685,11 +685,11 @@ const {fields, addItem, removeItem, moveItem, reset, dirty} = useDynamicList(
 );
 ```
 
-AddItem can accept an argument which will be passed to the factory. In this case, the factory could look like
+AddItem can accept an argument which will be passed to the factory. In this case, the factory could look like:
 
 ```tsx
 const emptyCardFactory = ({id, cardNumber, cvv}: Partial<Card>): Card => ({
-  id: id || new Date.now().toString(),
+  id: id || Date.now().toString(),
   cardNumber: cardNumber || '',
   cvv: cvv || '',
 });
@@ -698,7 +698,7 @@ const emptyCardFactory = ({id, cardNumber, cvv}: Partial<Card>): Card => ({
 addItem({cardNumber: '3131 3131 3131 3131'});
 ```
 
-You can choose to initialize the list with no card (same as above) or with existing cards coming from your database for example.
+You can choose to initialize the list with no card (same as above) or with existing cards e.g. coming from your database:
 
 ```tsx
 const loadedCards = [
@@ -713,7 +713,7 @@ const {fields, addItem, removeItem, moveItem, reset, dirty} = useDynamicList(
 
 You can also pass a more complex configuration object specifying a validation dictionary like in `useList`.
 
-Rendering your dynamic list would look like this
+Rendering your dynamic list would look like this:
 
 ```tsx
 <FormLayout>
@@ -755,7 +755,7 @@ Rendering your dynamic list would look like this
 
 You can use `useDynamicList` with `useForm` in two ways:
 
-You can directly pass one or several dynamic lists within `fields`. But note that the `reset` function returned by `useForm` won't reset the dynamic list to its initial value. It will only reset the value of each field for each item. So for example if you remove an item, clicking on reset won't re-add it. Plus the adding or removing a field won't make the form dirty. The form will only be dirty if you edit the value of one item's field.
+You can directly pass one or several dynamic lists within `fields`. But note that the `reset` function returned by `useForm` won't reset the dynamic list to its initial value. It will only reset the value of each field for each item. So, if you remove an item, clicking on reset won't re-add it. Plus, adding or removing a field won't make the form dirty. The form will only be dirty if you edit the value of an item's field.
 
 ```tsx
 const {fields} = useDynamicList<Card>([], emptyCardFactory);
@@ -771,7 +771,7 @@ const form = useForm({
 });
 ```
 
-Or you can pass your dynamic lists within `dynamicLists`. This way the form `reset` and `dirty` will work as expected. Note that you still need to have one "classic field", `fields` cannot be optional at the moment. If you use TypeScript, you might want to pass generic types to `useForm` in order to make sure TypeScript will infer the `fieldValues` types correctly within `onSubmit`.
+Or you can pass your dynamic lists within `dynamicLists`. This way the form `reset` and `dirty` will work as expected. Note that you still need to have one "classic field", `fields` cannot be optional at the moment. If you use TypeScript, you might want to pass generic types to `useForm` to make sure TypeScript infers the `fieldValues` types correctly within `onSubmit`.
 
 ```tsx
 const fields = {
@@ -785,7 +785,7 @@ const form = useForm<typeof fields, typeof customerCards>({
   dynamicLists: {
     customerCards,
   },
-  onSubmit: async fieldValues => {
+  onSubmit: async (fieldValues) => {
     console.log(fieldValues);
     return submitSuccess();
   },


packages/react-form/src/types.ts
```diff
@@ -124,10 +133,8 @@ export interface FieldBag {
   [key: string]: FieldOutput<any>;
 }
 
-export type DynamicListOutput<V extends object> = DynamicList<V>;
```
sylvhama, line 127: I think this is useless 🤔

packages/react-form/src/types.ts
```diff
@@ -74,13 +74,22 @@ export interface FormWithoutDynamicListsInput<T extends FieldBag> {
...
+  fields: T;
   dynamicLists: D;
+  onSubmit?: SubmitHandler<
+    FormMapping<T, 'value'> & FormMapping<DynamicListMapping<D>, 'value'>
+  >;
+  makeCleanAfterSubmit?: boolean;
 }
 
-export interface FormInput<T extends FieldBag, D extends DynamicListBag>
-  extends FormWithoutDynamicListsInput<T> {
```
sylvhama, line 82: I removed `extends` cause it makes it hard to follow.

oluwatimio (approved):
This looks really good!

packages/react-form/README.md
```diff
@@ -698,88 +731,65 @@ You can choose to initialize the list with an existing number of cards or no car
         value={field.cvv.value}
         onChange={field.cvv.onChange}
       />
-      <div>
-        <Button onClick={() => removeItem(index)}>Remove</Button>
-      </div>
```
sylvhama, line 703: I removed the `div` to reduce the density of the component tree.

packages/react-form/README.md
```diff
@@ -698,88 +731,65 @@ You can choose to initialize the list with an existing number of cards or no car
...
   },
   onSubmit: async fieldValues => {
     console.log(fieldValues);
-    return {status: 'success'};
+    return submitSuccess();
   },
 });
 ```
 
-A code sandbox of `useDynamicList` in action can be seen [here](https://codesandbox.io/s/hungry-rubin-exrkz?fontsize=14&hidenavigation=1&theme=dark)
```
oluwatimio, line 779: I am not sure we should remove the sandbox as it helps people visualize how the dynamic list works

sylvhama: If you think something i missing in the README then we can add it tho

packages/react-form/src/hooks/form.ts
```diff
@@ -27,18 +27,14 @@ import {useDynamicListDirty, useDynamicListReset} from './list';
...
  *
  *  function MyComponent() {
- *   const { fields, submit, submitting, dirty, reset, submitErrors, dynamicLists} = useForm({
+ *   const { fields, submit, submitting, dirty, reset, submitErrors } = useForm({
  *     fields: {
  *       title: useField('some default title'),
  *     },
- *     dynamicLists: {
- *       customerPaymentMethods: useDynamicList<Card>([{cardNumber: '4242 4242 4242 4242', cvv: '422'}], emptyCardFactory)
- *     }
```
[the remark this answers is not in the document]
oluwatimio: 🤔 , shouldnt we somehow show the dynamicList param in action here? `useList` might not be there because it goes into the fields param

packages/react-form/README.md
```diff
@@ -638,54 +638,87 @@ function MyComponent() {
...
+
 const emptyCardFactory = (): Card => ({
+  id: new Date.now().toString(),
   cardNumber: '',
   cvv: '',
 });
 
 const {fields, addItem, removeItem, moveItem, reset, dirty} = useDynamicList(
-  [{cardNumber: '4242 4242 4242 4242', cvv: '000'}],
+  [],
```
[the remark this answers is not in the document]
sylvhama: True, that's why on L701 we have a dedicated example for that
> You can choose to initialize the list with no card (same as above) or with existing cards coming from your database for example.

packages/react-form/README.md
```diff
@@ -638,54 +638,87 @@ function MyComponent() {
 
 #### `useDynamicList()`
 
-This offers the same functionality as useList. `useDynamicList` is a hook that adds on `useList` by adding the ability to dynamically add and remove list items. The same way you would utilize `useList` is the same way you would utilize `useDynamicList` except some differences such as the `addItem`, and the `removeItem` function. We also have to pass in a factory into this hook (telling the hook the exact type of object to add and how it should be initialized).
+A custom hook that adds on `useList` the ability to dynamically add, move and remove list tems.
```
attila-berki, line 641:
```suggestion
A custom hook that adds on `useList` the ability to dynamically add, move and remove list items.
```

What is the meaning of `moving a list` item?

packages/react-form/README.md
```diff
@@ -698,88 +731,65 @@ You can choose to initialize the list with an existing number of cards or no car
...
 
 ```tsx
-const {submit, dirty, submitting, reset} = useForm({
+const form = useForm({
   fields: {
-    title: useField('')
-  },
-  dynamicLists: {
-    customerCards: useDynamicList<Card>([{cardNumber: '4242 4242 4242 4242', cvv: '422'}], emptyCardFactory)
+    customerCards: fields, // the fields destructured from useDynamicList
```
attila-berki, line 763: I would add the `useDynamicList` detructuring above, it would be easier to understand what we mean here.

packages/react-form/README.md
```diff
@@ -698,88 +731,65 @@ You can choose to initialize the list with an existing number of cards or no car
...
   onSubmit: async fieldValues => {
     console.log(fieldValues);
-    return {status: 'success'};
+    return submitSuccess();
   },
 });
 ```
 
-you can directly pass a dynamic list within `fields` but note that the `reset` function returned by `useForm` will only reset each field, plus the `dirty` state will only reflect dirty fields within each item.
+Or you can pass your dynamic lists within `dynamicLists`. This way the form `reset` and `dirty` will work as expected. Note that you still need to have one "classic field", `fields` cannot be optionnal at the moment. If you use TypeScript, you might want to pass generic types to `useForm` in order to make sure TypeScript will infer the `fieldValues` types correctly within `onSubmit`.
```
attila-berki, line 772:
```suggestion
Or you can pass your dynamic lists within `dynamicLists`. This way the form `reset` and `dirty` will work as expected. Note that you still need to have one "classic field", `fields` cannot be optional at the moment. If you use TypeScript, you might want to pass generic types to `useForm` in order to make sure TypeScript will infer the `fieldValues` types correctly within `onSubmit`.
```

packages/react-form/README.md
```diff
@@ -795,6 +805,7 @@ const form = useForm(config);
 - `config`, An object has the following fields:
 
   - `fields`, A dictionary of `Field` objects, dictionaries of `Field` objects, and lists of dictionaries of `Field` objects. Generally, you'll want these to be generated by the other hooks in this package, either `useField` or `useList`. This will be returned back out as the `fields` property of the return value.
+  - `useDynamicList`, optionnal dictionaries of `DynamicList`.
```
attila-berki, line 808:
```suggestion
  - `useDynamicList`, optional dictionaries of `DynamicList`.
```

packages/react-form/src/hooks/form.ts
```diff
@@ -18,7 +18,7 @@ import {useSubmit} from './submit';
 import {useDynamicListDirty, useDynamicListReset} from './list';
 
 /**
- * A custom hook for managing the state of an entire form. `useForm` wraps up many of the other hooks in this package in one API, and when combined with `useField` and `useList`, allows you to easily build complex forms with smart defaults for common cases.
+ * A custom hook for managing the state of an entire form. `useForm` wraps up many of the other hooks in this package in one API, and when combined with `useField` and `useList` and `useDynamicList`, allows you to easily build complex forms with smart defaults for common cases.
```
attila-berki, line 21:
```suggestion
 * A custom hook for managing the state of an entire form. `useForm` wraps up many of the other hooks in this package in one API, and when combined with `useField`, `useList` or `useDynamicList`, allows you to easily build complex forms with smart defaults for common cases.
```

packages/react-form/src/hooks/form.ts
```diff
@@ -55,22 +51,16 @@ import {useDynamicListDirty, useDynamicListReset} from './list';
...
- *       <button type="button" onClick={addItem}>Add Payment Method</button>
  *     </form>
  *  );
  *```
  *
  * @param fields - A dictionary of `Field` objects, dictionaries of `Field` objects, and lists of dictionaries of `Field` objects. Generally, you'll want these to be generated by the other hooks in this package, either `useField` or `useList`. This will be returned back out as the `fields` property of the return value.
  *
  * @param onSubmit - An async function to handle submission of the form. If this function returns an object of `{status: 'fail', error: FormError[]}` then the submission is considered a failure. Otherwise, it should return an object with `{status: 'success'}` and the submission will be considered a success. `useForm` will also call all client-side validation methods for the fields passed to it. The `onSubmit` handler will not be called if client validations fails.
- * @param dynamicLists - Pass in dynamic lists objects into the form
+ * @param dynamicLists - optionnal dictionaries of `DynamicList`.
```
attila-berki, line 63:
```suggestion
 * @param dynamicLists - optional dictionaries of `DynamicList`.
```

attila-berki (approved):

packages/react-form/README.md
```diff
@@ -698,88 +731,65 @@ You can choose to initialize the list with an existing number of cards or no car
         value={field.cvv.value}
```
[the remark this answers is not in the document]
sylvhama: There is an example about that for `useField`, I think it's good to show what a field is made of.

packages/react-form/README.md
```diff
@@ -638,54 +638,87 @@ function MyComponent() {
...
-This offers the same functionality as useList. `useDynamicList` is a hook that adds on `useList` by adding the ability to dynamically add and remove list items. The same way you would utilize `useList` is the same way you would utilize `useDynamicList` except some differences such as the `addItem`, and the `removeItem` function. We also have to pass in a factory into this hook (telling the hook the exact type of object to add and how it should be initialized).
+A custom hook that adds on `useList` the ability to dynamically add, move and remove list items.
 
 ##### Using `useDynamicList`
 
-Lets simulate having a user interface that allows you add as many payments methods as you wish.
+Let's simulate a user interface that allows you add as many credit cards as you wish.
 
-1. We have to initialize `useDynamicList` the following way
+You can initialize `useDynamicList` the following way
```
atesgoral, line 647:
```suggestion
You can initialize `useDynamicList` the following way:
```

packages/react-form/README.md
```diff
@@ -638,54 +638,87 @@ function MyComponent() {
...
 
 ```tsx
+interface Card {
+  id: string;
+  cardNumber: string;
+  cvv: string;
+}
+
 const emptyCardFactory = (): Card => ({
+  id: new Date.now().toString(),
```
atesgoral, line 657:
```suggestion
  id: Date.now().toString(),
```

packages/react-form/README.md
```diff
@@ -638,54 +638,87 @@ function MyComponent() {
...
 
 const {fields, addItem, removeItem, moveItem, reset, dirty} = useDynamicList(
-  [{cardNumber: '4242 4242 4242 4242', cvv: '000'}],
+  [],
   emptyCardFactory,
 );
 ```
 
-We can also have a factory that produces multiple cards such as
+You can also have a factory that produces multiple cards such as
```
atesgoral, line 668:
```suggestion
You can also have a factory that produces multiple cards such as:
```

packages/react-form/README.md
```diff
@@ -638,54 +638,87 @@ function MyComponent() {
...
-  {
-    cardNumber: '',
-    cvv: '',
-  },
-  {
-    cardNumber: '',
-    cvv: '',
-  },
+function generateId() {
+  return new Date.now().toString();
```
atesgoral, line 672:
```suggestion
  return Date.now().toString();
```

packages/react-form/README.md
```diff
@@ -638,54 +638,87 @@ function MyComponent() {
...
+  ];
+};
+
+const {fields, addItem, removeItem, moveItem, reset, dirty} = useDynamicList(
+  [],
+  emptyCardFactory,
+);
+```
+
+AddItem can accept an argument which will be passed to the factory. In this case, the factory could look like
```
atesgoral, line 688:
```suggestion
AddItem can accept an argument which will be passed to the factory. In this case, the factory could look like:
```

packages/react-form/README.md
```diff
@@ -698,88 +731,67 @@ You can choose to initialize the list with an existing number of cards or no car
...
 ```
 
-We render our UI representation of the fields by utilizing the `fields.map` function. For each field, we can use the handlers such as onChange, value, onBlur. These are the same handlers that useList provides.
+##### How to use it with `useForm`
 
-We can also utilize the `removeItem`, and `addItem` functions. In this example, these functions are attached to a button press. When we remove an item we pass in the index (to indicate what field item to remove). When we add a item, we utilize the factory we passed in when initializing `useDynamicList`.
+You can use `useDynamicList` with `useForm` in two ways:
 
-Note that addItem can also take in an argument which can then be passed onto the factory. In this case, the factory can be
+You can directly pass one or several dynamic lists within `fields`. But note that the `reset` function returned by `useForm` won't reset the dynamic list to its initial value. It will only reset the value of each field for each item. So for example if you remove an item, clicking on reset won't re-add it. Plus the adding or removing a field won't make the form dirty. The form will only be dirty if you edit the value of one item's field.
```
atesgoral, line 758:
```suggestion
You can directly pass one or several dynamic lists within `fields`. But note that the `reset` function returned by `useForm` won't reset the dynamic list to its initial value. It will only reset the value of each field for each item. So, if you remove an item, clicking on reset won't re-add it. Plus, adding or removing a field won't make the form dirty. The form will only be dirty if you edit the value of an item's field.
```

packages/react-form/README.md
```diff
@@ -698,88 +731,67 @@ You can choose to initialize the list with an existing number of cards or no car
...
   onSubmit: async fieldValues => {
     console.log(fieldValues);
-    return {status: 'success'};
+    return submitSuccess();
   },
 });
 ```
 
-you can directly pass a dynamic list within `fields` but note that the `reset` function returned by `useForm` will only reset each field, plus the `dirty` state will only reflect dirty fields within each item.
+Or you can pass your dynamic lists within `dynamicLists`. This way the form `reset` and `dirty` will work as expected. Note that you still need to have one "classic field", `fields` cannot be optional at the moment. If you use TypeScript, you might want to pass generic types to `useForm` in order to make sure TypeScript will infer the `fieldValues` types correctly within `onSubmit`.
```
atesgoral, line 774:
```suggestion
Or you can pass your dynamic lists within `dynamicLists`. This way the form `reset` and `dirty` will work as expected. Note that you still need to have one "classic field", `fields` cannot be optional at the moment. If you use TypeScript, you might want to pass generic types to `useForm` to make sure TypeScript infers the `fieldValues` types correctly within `onSubmit`.
```

packages/react-form/README.md
```diff
@@ -698,88 +731,67 @@ You can choose to initialize the list with an existing number of cards or no car
...
+};
+
+const customerCards = useDynamicList<Card>([], emptyCardFactory);
+
+const form = useForm<typeof fields, typeof customerCards>({
+  fields,
+  dynamicLists: {
+    customerCards,
   },
   onSubmit: async fieldValues => {
```
atesgoral, line 788:
```suggestion
  onSubmit: async (fieldValues) => {
```

atesgoral (approved):

PR #1867, opened 2021-04-26
Update Jest / Webpack versions

BPScott:
## Description

Update Jest to v26.6.

Update webpack dependency in web-worker, and dev-dep in react-server to be ^4.25.1 from `>4.25.1, <=4.5.0` which i'm not convinced is even a valid version string, as yarn seems to treat it as <=4.5.0).

## Type of change

- [x] web-worker Patch: Bug (non-breaking change which fixes an issue)


michenly (approved):
LGTM

Issue #1855, opened 2021-04-21
[DynamicList] makeCleanAfterSubmit set to true does not work for dynamic lists passed into the DynamicListBag param


oluwatimio:
## Overview

1. When we trigger a submit with makeCleanAfterSubmit to true, the dynamic list does not get cleaned. In other for this to work with dynamic list, we will have to implement a clean function in base list. This function will set the initial list to the current list so the dirty state would return false. This function can then be called in the makeClean utility.

## Consuming repo

> What repo were you working in when this issue occurred?

While working on https://github.com/Shopify/quilt/pull/1828 we found the issue just before merging so we will do a follow up PR to fix it

### Area

- [x] Add any relevant `Area: <area>` labels to this issue

### Scope

`Package: react-form`
---

## Checklist

- [x] Please delete the labels section before submitting your issue
- [x] I have described this issue in a way that is actionable (if possible)

PR #1871, opened 2021-04-27
[react-form | DynamicList] newDefaultValue and value props

oluwatimio:
## Description

Fixes https://github.com/Shopify/quilt/issues/1855

## Summary
Adds a `makeCleanDynamicList` function, `newDefaultValue` function, `defaultValue`, and `value` properties to `useDynamicList`. This also helps fix the issue where the dynamic list is not cleaned when `makeCleanAfterSubmit` is specified. 

## Tophat

Codesandbox [here](https://codesandbox.io/s/nice-surf-f4kml)

## Type of change

## Checklist

### DIFF
--- a/packages/react-form/CHANGELOG.md
+++ b/packages/react-form/CHANGELOG.md
@@ -5,7 +5,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+- Adds a makeCleanDynamicLists function, which adds the ability to set [newDefaultValue, defaultValue] and get the value of a dynamic list. [#1871](https://github.com/Shopify/quilt/pull/1871)
 
 ## 0.13.1 - 2021-04-29
 
--- a/packages/react-form/README.md
+++ b/packages/react-form/README.md
@@ -659,10 +659,17 @@ const emptyCardFactory = (): Card => ({
   cvv: '',
 });
 
-const {fields, addItem, removeItem, moveItem, reset, dirty} = useDynamicList(
-  [],
-  emptyCardFactory,
-);
+const {
+  fields,
+  addItem,
+  removeItem,
+  moveItem,
+  reset,
+  dirty,
+  value,
+  newDefaultValue,
+  defaultValue,
+} = useDynamicList([], emptyCardFactory);
 ```
 
 You can also have a factory that produces multiple cards such as:
@@ -679,10 +686,17 @@ const emptyCardFactory = (): Card[] => {
   ];
 };
 
-const {fields, addItem, removeItem, moveItem, reset, dirty} = useDynamicList(
-  [],
-  emptyCardFactory,
-);
+const {
+  fields,
+  addItem,
+  removeItem,
+  moveItem,
+  reset,
+  dirty,
+  value,
+  newDefaultValue,
+  defaultValue,
+} = useDynamicList([], emptyCardFactory);
 ```
 
 AddItem can accept an argument which will be passed to the factory. In this case, the factory could look like:
@@ -751,6 +765,33 @@ Rendering your dynamic list would look like this:
 </FormLayout>
 ```
 
+You can use the newDefaultValue function to change the defaultValue of the dynamicList. This will also set the value and defaultValue property of the dynamicList to the newValue passed in.
+
+```tsx
+const newValue = [
+  {id: '123456', cardNumber: '4242 4242 4242 4242', cvv: '000'},
+];
+newDefaultValue(newValue);
+```
+
+You can iterate through the values and defaultValues of the dynamicList
+
+```tsx
+value.map(card => (
+  <div>
+    <p>Card number: {card.cardNumber}</p>
+    <p>CVV : {card.cvv}</p>
+  </div>
+));
+
+defaultValue.map(card => (
+  <div>
+    <p>Card number: {card.cardNumber}</p>
+    <p>CVV : {card.cvv}</p>
+  </div>
+));
+```
+
 ##### How to use it with `useForm`
 
 You can use `useDynamicList` with `useForm` in two ways:
@@ -785,7 +826,7 @@ const form = useForm<typeof fields, typeof customerCards>({
   dynamicLists: {
     customerCards,
   },
-  onSubmit: async (fieldValues) => {
+  onSubmit: async fieldValues => {
     console.log(fieldValues);
     return submitSuccess();
   },
--- a/packages/react-form/src/hooks/form.ts
+++ b/packages/react-form/src/hooks/form.ts
@@ -10,7 +10,11 @@ import {
   FormWithDynamicLists,
   DynamicListBag,
 } from '../types';
-import {validateAll, makeCleanFields} from '../utilities';
+import {
+  validateAll,
+  makeCleanFields,
+  makeCleanDynamicLists,
+} from '../utilities';
 
 import {useDirty} from './dirty';
 import {useReset} from './reset';
@@ -109,6 +113,7 @@ export function useForm<T extends FieldBag, D extends DynamicListBag>({
     onSubmit,
     fieldsWithLists,
     makeCleanAfterSubmit,
+    dynamicLists,
   );
 
   const reset = useCallback(() => {
@@ -119,14 +124,16 @@ export function useForm<T extends FieldBag, D extends DynamicListBag>({
 
   const fieldsRef = useLazyRef(() => fieldsWithLists);
   fieldsRef.current = fieldsWithLists;
+  const dynamicListsRef = useLazyRef(() => dynamicLists);
 
   const validate = useCallback(() => {
     return validateAll(fieldsRef.current);
   }, [fieldsRef]);
 
-  const makeClean = useCallback(() => makeCleanFields(fieldsRef.current), [
-    fieldsRef,
-  ]);
+  const makeClean = useCallback(() => {
+    makeCleanFields(fieldsRef.current);
+    makeCleanDynamicLists(dynamicListsRef.current);
+  }, [dynamicListsRef, fieldsRef]);
 
   const form: Form<T> = {
     fields,
--- a/packages/react-form/src/hooks/list/baselist.ts
+++ b/packages/react-form/src/hooks/list/baselist.ts
@@ -46,6 +46,9 @@ interface BaseList<Item extends object> {
   dispatch: React.Dispatch<ListAction<Item>>;
   reset(): void;
   dirty: boolean;
+  defaultValue: Item[];
+  value: Item[];
+  newDefaultValue(newDefaultItems: Item[]): void;
 }
 
 export function useBaseList<Item extends object>(
@@ -63,7 +66,8 @@ export function useBaseList<Item extends object>(
     if (!isEqual(list, state.initial)) {
       dispatch(reinitializeAction(list));
     }
-  }, [list, state.initial, dispatch]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [list, dispatch]);
 
   const validationConfigs = useMemo(
     () =>
@@ -77,6 +81,10 @@ export function useBaseList<Item extends object>(
 
   function reset() {
     dispatch(resetListAction());
+  }
+
+  function newDefaultValue(newDefaultItems: Item[]) {
+    dispatch(reinitializeAction(newDefaultItems));
   }
 
   const handlers = useHandlers(state, dispatch, validationConfigs);
@@ -105,5 +113,13 @@ export function useBaseList<Item extends object>(
 
   const fieldsDirty = useDirty({fields});
 
-  return {fields, dispatch, reset, dirty: fieldsDirty || isBaseListDirty};
+  return {
+    fields,
+    dispatch,
+    reset,
+    dirty: fieldsDirty || isBaseListDirty,
+    defaultValue: state.initial,
+    value: listWithoutFieldStates,
+    newDefaultValue,
+  };
 }
--- a/packages/react-form/src/hooks/list/test/baselist.test.tsx
+++ b/packages/react-form/src/hooks/list/test/baselist.test.tsx
@@ -768,4 +768,130 @@ describe('useBaseList', () => {
       });
     });
   });
+
+  describe('value, newDefaultValue and defaultValue', () => {
+    function TestListWithValue(config: FieldListConfig<Variant>) {
+      const {value, newDefaultValue, defaultValue} = useBaseList<Variant>(
+        config,
+      );
+
+      const onNewDefault = (value: Variant[]) => {
+        newDefaultValue(value);
+      };
+
+      return (
+        <>
+          {value.map(variant => (
+            <>
+              <p>Value: {variant.price}</p>
+              <p>Value: {variant.optionName}</p>
+              <p>Value: {variant.optionValue}</p>
+            </>
+          ))}
+
+          {defaultValue.map(variant => (
+            <>
+              <p>Default: {variant.price}</p>
+              <p>Default: {variant.optionName}</p>
+              <p>Default: {variant.optionValue}</p>
+            </>
+          ))}
+
+          <button type="button" onClick={onNewDefault} />
+        </>
+      );
+    }
+
+    it('returns the value of the baselist', () => {
+      const price = '1.00';
+      const optionName = 'material';
+      const optionValue = 'cotton';
+      const variants: Variant[] = [
+        {
+          price,
+          optionName,
+          optionValue,
+        },
+      ];
+
+      const wrapper = mount(<TestListWithValue list={variants} />);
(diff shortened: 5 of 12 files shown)


sylvhama (commented):
I have made a quick review without looking at the tests yet. I am a bit concerned by the `hasCleaned` state, I need to look at existing tests to understand more that part.

packages/react-form/src/hooks/list/baselist.ts
```diff
@@ -62,10 +66,10 @@ export function useBaseList<Item extends object>(
   const [state, dispatch] = useListReducer(list);
 
   useEffect(() => {
-    if (!isEqual(list, state.initial)) {
+    if (!isEqual(list, state.initial) && !hasCleaned) {
```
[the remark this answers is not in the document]
oluwatimio: but perhaps we could remove state.initial from the dependency list there and avoid this hasCleaned state?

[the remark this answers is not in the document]
oluwatimio: I tried looking for other possibilities but the truth is, we will need a flag since if state.initial is changed, it will try and re-initialize the array. 

Perhaps we could do something better there by checking if state.initial is empty instead of using isEqual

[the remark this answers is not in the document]
oluwatimio: Figured out the solution. The solution they used on `Field.ts` was to remove the initial value from the dependency list as they explained there,

```
We actually do not want this to rerun when our `defaultValue` is updated. 
It can only happen independently of this callback when `newDefaultValue` is called by a user,
and we don't want to undue their hard work by resetting to `value`.
```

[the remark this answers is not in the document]
oluwatimio: Made the changes [here](https://github.com/Shopify/quilt/pull/1871/commits/e6e02a60dd00381bedafd67bd3b5e6df7b76c883) and added a test to make sure reinitialize is called when the list has changed.

packages/react-form/src/hooks/submit.ts
```diff
@@ -7,18 +7,21 @@ import {
...
   getValues,
   makeCleanFields,
+  makeCleanDynamicList,
 } from '../utilities';
 
 export function useSubmit<T extends FieldBag>(
   onSubmit: SubmitHandler<FormMapping<T, 'value'>> = noopSubmission,
   fieldBag: T,
   makeCleanAfterSubmit = false,
+  dynamicLists?: DynamicListBag,
```
attila-berki, line 24: Could we call it `dinamicListBag` for the sake of consistency?

packages/react-form/src/hooks/test/form-with-dynamic-list.test.tsx
```diff
@@ -142,4 +143,111 @@ describe('useForm with dynamic list', () => {
...
+      fillRequiredFields(wrapper);
+
+      expect(isDirty(wrapper)).toBe(true);
+
+      await waitForSubmit(wrapper, promise);
+
+      expect(isDirty(wrapper)).toBe(true);
+    });
+
+    it('does undirty fields after successful submit if makeCleanAfterSubmit is true', async () => {
```
attila-berki, line 164: I think you meant here that it set dirty to false after submit, right?

packages/react-form/src/hooks/test/form-with-dynamic-list.test.tsx
```diff
@@ -142,4 +143,111 @@ describe('useForm with dynamic list', () => {
...
+      fillRequiredFields(wrapper);
+
+      expect(isDirty(wrapper)).toBe(true);
+
+      await waitForSubmit(wrapper, promise);
+
+      expect(isDirty(wrapper)).toBe(true);
+    });
+
+    it('does not undirty fields after if makeCleanAfterSubmit is true but submit is unsuccessful', async () => {
```
attila-berki, line 203:
```suggestion
    it('does not undirty fields if makeCleanAfterSubmit is true but submit is unsuccessful', async () => {
```

packages/react-form/src/hooks/test/form-with-dynamic-list.test.tsx
oluwatimio: Yeah! Will change the wording 🙂

packages/react-form/src/hooks/submit.ts
oluwatimio: I was thinking of following the same naming as useForm since it was called `dynamicLists` there but I think since it uses fieldBag naming here, it's fair to change it too to be consistent. Thanks!

packages/react-form/src/hooks/list/test/baselist.test.tsx
```diff
@@ -768,4 +768,126 @@ describe('useBaseList', () => {
...
+            </>
+          ))}
+
+          {defaultValue.map(variants => (
+            <>
+              <p>Default: {variants.price}</p>
+              <p>Default: {variants.optionName}</p>
+              <p>Default: {variants.optionValue}</p>
+            </>
+          ))}
```
sylvhama, line 798: Nit pick: 
```suggestion
          {value.map(variant => (
            <>
              <p>Value: {variant.price}</p>
              <p>Value: {variant.optionName}</p>
              <p>Value: {variant.optionValue}</p>
            </>
          ))}

          {defaultValue.map(variant => (
            <>
              <p>Default: {variant.price}</p>
              <p>Default: {variant.optionName}</p>
              <p>Default: {variant.optionValue}</p>
            </>
          ))}
```

packages/react-form/src/hooks/list/test/baselist.test.tsx
```diff
@@ -768,4 +768,126 @@ describe('useBaseList', () => {
...
+        {
+          price: newDefaultPrice,
+          optionName: newDefaultOption,
+          optionValue: newDefaultOptionValue,
+        },
+      ]);
+
+      expect(wrapper).toContainReactText(`Default: ${newDefaultPrice}`);
+      expect(wrapper).toContainReactText(`Default: ${newDefaultOption}`);
+      expect(wrapper).toContainReactText(`Default: ${newDefaultOptionValue}`);
```
sylvhama, line 882: I would remove the three expect and inline the "new" values. This part is already tested above. Here we want to test the `useEffect` that reinitializes the list when a new config is passed.

packages/react-form/src/hooks/list/test/baselist.test.tsx
```diff
@@ -768,4 +768,126 @@ describe('useBaseList', () => {
...
+
+      wrapper.setProps({list: [{...variants[0], price: '100.00'}]});
+
+      expect(wrapper).not.toContainReactText(`Default: ${newDefaultPrice}`);
+      expect(wrapper).not.toContainReactText(`Default: ${newDefaultOption}`);
+      expect(wrapper).not.toContainReactText(
+        `Default: ${newDefaultOptionValue}`,
+      );
```
sylvhama, line 890: I would create a brand new array containing a brand new variant:
```tsx
      const nextVariant = {};

      wrapper.setProps({list: [newVariant]});

      expect(wrapper).toContainReactText(`Default: ${nextVariant.price}`);
      // etc..
```

packages/react-form/src/hooks/submit.ts
```diff
@@ -1,40 +1,49 @@
...
+  dynamicListBagRef.current = dynamicListBag;
 
-  const setErrors = useCallback((errors: FormError[]) => {
-    setSubmitErrors(errors);
-    propagateErrors(fieldBagRef.current, errors);
-  }, []);
+  const setErrors = useCallback(
+    (errors: FormError[]) => {
+      setSubmitErrors(errors);
+      propagateErrors(fieldBagRef.current, errors);
```
sylvhama, line 38: Seeing this makes me feel the error handling might be incomplete when passing `dynamicLists` 👀

packages/react-form/src/hooks/submit.ts
```diff
@@ -1,40 +1,49 @@
...
   getValues,
   makeCleanFields,
+  makeCleanDynamicList,
 } from '../utilities';
 
 export function useSubmit<T extends FieldBag>(
   onSubmit: SubmitHandler<FormMapping<T, 'value'>> = noopSubmission,
   fieldBag: T,
   makeCleanAfterSubmit = false,
+  dynamicListBag?: DynamicListBag,
```
sylvhama, line 24: The code is very confusing now, `fieldBag` might or might not include the dynamic lists (see `fieldsWithLists` in form.ts). And in parallel we still pass `dynamicListBag`, I think we should refactor that. 

Maybe we should simply pass `makeCleanDynamicList(dynamicLists)` has a parameter?

packages/react-form/src/hooks/list/test/dynamiclist.test.tsx
```diff
@@ -336,4 +336,105 @@ describe('useDynamicList', () => {
       });
     });
   });
+
+  describe('value', () => {
```
sylvhama, line 340: Please apply feedback from `baseList` here too.

packages/react-form/src/hooks/test/form-with-dynamic-list.test.tsx
```diff
@@ -142,4 +143,111 @@ describe('useForm with dynamic list', () => {
...
+  const optionTextFields = wrapper.findAll(TextField, {label: 'option'});
+  optionTextFields.forEach(textField =>
+    textField.trigger('onChange', 'a value'),
+  );
+
+  const titleTextFields = wrapper.findAll(TextField, {label: 'title'});
+  titleTextFields.forEach(textField =>
+    textField.trigger('onChange', 'a value'),
+  );
+}
```
sylvhama, line 253: Please move this to utilities maybe? I would rename to `fillAllOptionAndTitleFields`

packages/react-form/src/hooks/test/form-with-dynamic-list.test.tsx
```diff
@@ -142,4 +143,111 @@ describe('useForm with dynamic list', () => {
       expect(isDirty(wrapper)).toBe(false);
     });
   });
+
+  describe('makeCleanAfterSubmit', () => {
+    it('does not set dirty to false after successful submit by default', async () => {
```
sylvhama, line 148: This tests seems to be the same as "does not set dirty to false after successful submit if makeCleanAfterSubmit is false", I think we can remove it. We don't care about testing the default values of `FormWithDynamicVariantList`, this is just a test component.

packages/react-form/src/hooks/list/baselist.ts
```diff
@@ -65,7 +68,8 @@ export function useBaseList<Item extends object>(
     if (!isEqual(list, state.initial)) {
       dispatch(reinitializeAction(list));
     }
-  }, [list, state.initial, dispatch]);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [list, dispatch]);
```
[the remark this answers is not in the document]
oluwatimio: https://github.com/Shopify/quilt/pull/1871#discussion_r622489280

packages/react-form/src/hooks/submit.ts
oluwatimio: 🤔 It should not be incomplete because the dynamic lists fields are in the fieldBag. Any errors in the fields should show up here too :

packages/react-form/src/hooks/submit.ts
oluwatimio: I am not sure I understand 😕 . I dont think the code is confusing. Note that `useSubmit` can be used on its own. In that case, we would like to give people the opportunity to pass in a dynamic list to clean their field

packages/react-form/src/hooks/submit.ts
sylvhama: OOps this comment should had been removed, all good!

packages/react-form/src/hooks/submit.ts
sylvhama: My problem is: `fieldBag` already contains `dynamicListBag` (see fieldsWithLists in form.ts) and then we re-pass it a second time. That seems not very clean.

oluwatimio: Ideally it will have been nice if we didnt have to pass in the dynamic list and if cleanFields was able to clean the dynamic list without us having to pass it in but I think this is perhaps one of the drawbacks to not having the whole dynamicList in the FieldBag. It has to be passed into every function that will need to perform an operation on the list itself and not just the fields. Wdyt?

attila-berki (approved):
LGTM 🎩 I tested the `makeCleanAfterSubmit true / false` and the `newDefaultValue` both works as expected.

packages/react-form/src/hooks/test/form-with-dynamic-list.test.tsx
oluwatimio: Moved it but I am thinking we should keep the name `fillRequiredFields` just in case more required fields are added.

packages/react-form/src/hooks/submit.ts
oluwatimio: fieldBag does not contain the DynamicListBag. It only contains fields that are in the dynamic list. Hence, we have to pass the dynamicList.

vsumner (commented):
Could you please update the documentation to include `newDefaultValue` `value` and `defaultValue` and their use?
https://github.com/Shopify/quilt/tree/main/packages/react-form#using-usedynamiclist

oluwatimio:
> Could you please update the documentation to include `newDefaultValue` `value` and `defaultValue` and their use?
> https://github.com/Shopify/quilt/tree/main/packages/react-form#using-usedynamiclist

Thanks for the review @vsumner. Just added documentation [here](https://github.com/Shopify/quilt/pull/1871/commits/e859440c878fd97b8442dd9ac8415093e32949e9) 🙂

packages/react-form/CHANGELOG.md
```diff
@@ -5,7 +5,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+- Adds a makeCleanDynamicLists function, the ability to set a newDefaultValue, defaultValue, and get the value of a dynamic list [#1871](https://github.com/Shopify/quilt/pull/1871)
```
dahukish, line 10: Adds a `makeCleanDynamicLists function`, which adds the ability to set [`newDefaultValue`, `defaultValue`] and get the value of a dynamic list.

packages/react-form/src/hooks/form.ts
```diff
@@ -119,14 +124,17 @@ export function useForm<T extends FieldBag, D extends DynamicListBag>({
 
   const fieldsRef = useLazyRef(() => fieldsWithLists);
   fieldsRef.current = fieldsWithLists;
+  const dynamicListsRef = useLazyRef(() => dynamicLists);
+  dynamicListsRef.current = dynamicLists;
```
dahukish, line 128: This should set `current` on line 127 no?

dahukish: Yeah sorry this: `const dynamicListsRef = useLazyRef(() => dynamicLists);`
and this `const dynamicListsRef = useRef(); dynamicListsRef.current = dynamicLists;`  are equivalent from what I see. You should just be able to do `const dynamicListsRef = useLazyRef(() => dynamicLists);` without assign it to `.current`.

oluwatimio: Ahh I see, yes it should typically set the current value when passed in. Will change it for both! thanks :)

packages/react-form/src/hooks/submit.ts
dahukish: Can we make a new issue for this and as I agree it would be awesome not to have to pass around certain values to everything (in this case `dynamicListBag`).

packages/react-form/README.md
```diff
@@ -659,10 +659,17 @@ const emptyCardFactory = (): Card => ({
...
-);
+const {
+  fields,
+  addItem,
+  removeItem,
+  moveItem,
+  reset,
+  dirty,
+  value,
+  newDefaultValue,
```
dahukish, line 670: This updates what the new default value is. Having `newDefaultValue` and `DefaultValue` side by side seems odd. Could we rename to `updateDefaultValue` as that is what it is doing?

dahukish (commented):
Looking good. Just left some comments around naming and the use if `useLazyRef`.

packages/react-form/README.md
oluwatimio: I sort of wanted to follow the convention of `Field` where the function is called `newDefaultValue` and it also has `defaultValue`. This way its consistent and easily recognizable by those that use `Field`. Wdyt?

dahukish (approved):
One small comment on reassigning to current but looks gtg. Didn't 🎩

jakobbouchard:
I am [having an issue](https://github.com/Shopify/quilt/issues/2418) with this feature. In fact, it's kinda only half working. It does set the value, but the reinitialisation runs twice. Once with the new data, then once with the initial data, which basically nullifies the new data that got added.

PR #1874, opened 2021-04-29
update to lerna 3

michenly:
Part of https://github.com/Shopify/web-foundations/issues/191. Update quilt to use lerna 3 now that shipit supports it.


BPScott (approved):
No idea how to test / prove this is all good but 🎉

PR #1875, opened 2021-04-29
[Do Not Merge] drop node 10 from quilt

michenly:
https://github.com/Shopify/shopify-events/issues/1163

> Node.js 10 will be reaching its End of Life on April 30th. We have already seen dependency bumps breaking our Node.js 10 checks in our CI. Repos should be moving to at least Node.js 12 and should have a plan to move to 14 over the next year.

### DIFF
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [10.x, 12.x, 14.x]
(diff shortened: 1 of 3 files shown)


BPScott:
Hi! Please don't merge this, I'll take care of it :)

My rollup in skn work is close to being done and integrating that into quilt is going to be a breaking change. Rather than have the drop-node-10-support breaking change and then the breaking skn build  change happen in quick succession, I'd like to batch them up to reduce some release churn :)

PR #1877, opened 2021-05-03
Update tryRequire to include logging

### LINKED ISSUE #1737: Easier troubleshooting for SSR failures
## Overview
It would be nice if `@shopify/async` reported errors caught during SSR component lookups.  `@shopify/async/resolver` currently discards error output: https://github.com/Shopify/quilt/blob/997e69e6ff16224a0dc69da88576dbbaefc24cdd/packages/async/src/resolver.ts#L86-L98

## Motivation
We recently adopted `@babel/transform-runtime` in a large app, which (due to the way the transform exposes `default` exports) failed on some server-side circular imports.  The errors silently broke SSR, but were obvious once I added output to async's `// Just ignore failures` block.

cc @vsumner because this is your current headache 😸  cc @lemonmade since this likely affects the lemony equivalents of `async`.

vsumner:
## Description
This PR introduces logging for `@shopify/async` `tryRequire`. 
Fixes #1737

## Type of change

- [X] `@shopify/async`  Minor: New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/packages/async/src/resolver.ts
+++ b/packages/async/src/resolver.ts
@@ -93,14 +93,18 @@ function tryRequire(id: string) {
   ) {
     try {
       return normalize(__webpack_require__(id));
-    } catch {
-      // Just ignore failures
+    } catch (error) {
+      // Log errors
+      // eslint-disable-next-line no-console
+      console.log(`webpack require error: ${error}`);
     }
   } else if (typeof nodeRequire === 'function') {
     try {
       return normalize(nodeRequire(id));
-    } catch {
-      // Just ignore failures
+    } catch (error) {
+      // Log errors
+      // eslint-disable-next-line no-console
+      console.log(`node require error: ${error}`);
     }
   }
 
--- a/packages/async/CHANGELOG.md
+++ b/packages/async/CHANGELOG.md
@@ -9,7 +9,7 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ### Changed
 
-- Updated `tryRequire` to log errors to help with debugging [#1829](https://github.com/Shopify/quilt/pull/1829)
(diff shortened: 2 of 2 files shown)


packages/async/src/resolver.ts
```diff
@@ -93,14 +93,18 @@ function tryRequire(id: string) {
   ) {
     try {
       return normalize(__webpack_require__(id));
-    } catch {
-      // Just ignore failures
+    } catch (error) {
+      // Log errors
+      // eslint-disable-next-line no-console
+      console.log(`webpack require error: ${error}`);
```
vsumner, line 99: I'm open to other suggestions for logging this. I thought about introducing something like `pinojs` but it just seems like overkill.

GoodForOneFare: This definitely improves local debugging experience for known SSR problems.  I would like to explore what automated discovery of niche SSR errors looks like, though.  In an ideal world, these errors would be exposed to consumers to let them decide on how/where to log. Probably via `react-server#extract` .

That's likely way more work than it sounds 😄  I'm opening this can o' worms because:
* One SSR error appears to break _all_ future SSR for a pod
* This would allow a niche admin section to quietly break SSR on individual pods without the obvious, cross-pod metrics dip that we see with major SSR bugs

GoodForOneFare: > One SSR error appears to break all future SSR for a pod

This is based on me glancing at metrics during bad SSR deploys. I could definitely be wrong.

TheMallen: > In an ideal world, these errors would be exposed to consumers to let them decide on how/where to log. Probably via react-server#extract

I agree with this, if we could surface stuff like this through actual runtime errors somewhere like at the extract call we could get a lot more observability at minimum, and maybe even have some automated testing at maximum

GoodForOneFare (approved):
👍  "Perfect is the enemy of good", and this happens often enough that I've changed my view of "good".

PR #1872, opened 2021-04-28
Update graphql-config to ^3.2.0; fix breakages

dahukish:
## Description

Fixes security updates regarding `node-fetch` by updating `graphql-config` to vs 3 as this version strips `graphql-response` which ends up using `node-fetch`.

Once this PR is in: https://github.com/Shopify/quilt/pull/1880 I will rebase.

## Type of change

- [x] `graphql-config-utilities`  Minor: New feature (non-breaking change which adds functionality)
expected)
- [x] `graphql-typescript-definitions`  Patch
expected)
- [x] `graphql-validate-fixtures`  Patch
expected)

## 🎩 

CI passes in sewing-kit (with beta releases): https://github.com/Shopify/sewing-kit/pull/2649

## Checklist

- [ ] I have added a changelog entry, prefixed by the type of change noted above (Documentation fix and Test update does not need a changelog as we do not publish new version)

PR #1878, opened 2021-05-03
Remove node 10 from CI tests

vsumner:
## Description
This PR removes the Node 10x tests from CI.
Fixes (issue #)

PR #1880, opened 2021-05-04
Update eslint-plugin => v39

dahukish:
## Description

- fix linting issues after update
- add `.eslintrc.js` to `graphql-typescript-definitions` for testing fixtures

We need to bump our `eslint-plugin` as it relies on `eslint-plugin-graphql` which end up calling an old version of `graphql-config` which still uses `node-fetch` in its transitive dependencies.

Note: 
- A lot of this is automatic fixes via: `yarn lint --fix` hence the large amount of file changes.
- With the EoL for Nodejs 10 behind us it looks like many transitive dependencies are updating their node support as well:
`error meros@1.1.4: The engine "node" is incompatible with this module. Expected version ">=12". Got "10.24.1"`
With this I have also bumped quilt to node 12 and killed the node 10 check.

required for: https://github.com/Shopify/quilt/pull/1872

## Type of change

N/A


packages/graphql-typescript-definitions/.eslintrc.js
```diff
@@ -0,0 +1,15 @@
+module.exports = {
+    overrides: [
```
atesgoral, line 2: `:yells-at-indentation:`

packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -186,7 +186,9 @@ export class Builder extends EventEmitter {
   }
 
   stop() {
-    this.filesystem?.dispose();
+    if (this.filesystem) {
```
atesgoral, line 189: `?.` no longer allowed here?

packages/graphql-typescript-definitions/test/fixtures/all-clear/types/Generation.ts
```diff
@@ -1,4 +1,4 @@
 export enum Generation {
-  Millenial = 'MILLENIAL',
-  GenX = 'GEN_X',
-}
+  MILLENIAL = "MILLENIAL",
```
atesgoral, line 2: Hmm why double quotes?

packages/graphql-typescript-definitions/test/fixtures/all-clear/types/index.ts
```diff
@@ -1 +1,2 @@
-export {Generation} from './Generation';
+import { Generation } from "./Generation";
+export { Generation };
```
atesgoral, line 2:
```suggestion
import {Generation} from './Generation';
export {Generation};

```

packages/react-form/src/hooks/list/baselist.ts
```diff
@@ -53,11 +53,9 @@ export function useBaseList<Item extends object>(
   validationDependencies: unknown[] = [],
 ): BaseList<Item> {
   const list = Array.isArray(listOrConfig) ? listOrConfig : listOrConfig.list;
-  const validates: FieldListConfig<Item>['validates'] = Array.isArray(
-    listOrConfig,
-  )
-    ? {}
-    : listOrConfig.validates || {};
+  const validates: FieldListConfig<Item>['validates'] = useMemo(() => {
```
atesgoral, line 56: Why memoize this? This doesn't look like an expensive operation at all.

packages/graphql-typescript-definitions/src/index.ts
dahukish: linter yell at this for some reason

packages/graphql-typescript-definitions/.eslintrc.js
dahukish: Ironic no?

dev.yml
```diff
@@ -2,8 +2,8 @@ name: quilt
 type: node
 up:
   - node:
-      version: v10.18.1
-      yarn: 1.22.5
+      version: v12.18.0
+      yarn: 1.22.4
```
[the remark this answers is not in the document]
dahukish: nope I wil fix this.

packages/graphql-typescript-definitions/test/fixtures/all-clear/types/Generation.ts
dahukish: these are generated files

packages/react-form/src/hooks/list/baselist.ts
dahukish: Linter wants it to be I can ignore but then we should look at the rule

packages/graphql-typescript-definitions/test/fixtures/all-clear/types/Generation.ts
```diff
@@ -1,4 +1,4 @@
...
+  GEN_X = "GEN_X",
+}
```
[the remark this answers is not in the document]
dahukish: auto gen files

packages/statsd/src/test/.eslintrc.js
```diff
@@ -0,0 +1,13 @@
+module.exports = {
+  overrides: [
+    {
+      files: [
+        'client.test.ts',
+
```
atesgoral, line 6:
```suggestion
```

packages/graphql-typescript-definitions/.eslintrc.js
atesgoral: Very :) WHO LINTS THE LINTER CONFIG?

packages/graphql-typescript-definitions/src/index.ts
atesgoral: :yells-at-linter:

packages/graphql-typescript-definitions/test/fixtures/all-clear/types/Generation.ts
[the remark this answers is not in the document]
atesgoral: Could we still match them with our style since they're being checked in? Or will they get regenerated and to lose those tweaks?

BPScott:
per https://github.com/Shopify/quilt/pull/1875 and https://github.com/Shopify/quilt/pull/1878 can you hold fire on the removal on node 10 support. That's going to be a breaking change and my rollup changes will be breaking too, so in order to avoid churn I'd like to do the both of them at the same time.

packages/graphql-typescript-definitions/test/fixtures/all-clear/types/Generation.ts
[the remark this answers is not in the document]
dahukish: They will get regen on a yarn build an blown away again.

packages/react-form/src/hooks/list/baselist.ts
dahukish:
```
The 'validates' conditional could make the dependencies of useMemo Hook (at line 80) change on every render. Move it inside the useMemo callback. Alternatively, wrap the initialization of 'validates' in its own useMemo() Hook.eslintreact-hooks/exhaustive-deps
```

I think the linter is correct.

dahukish:
> per #1875 and #1878 can you hold fire on the removal on node 10 support. That's going to be a breaking change and my rollup changes will be breaking too, so in order to avoid churn I'd like to do the both of them at the same time.

Sure thing.

packages/graphql-typescript-definitions/.eslintrc.js
```diff
@@ -0,0 +1,15 @@
+module.exports = {
```
BPScott, line 1: Personal nitpick: I'm not a big fan of .eslintrc.js files dotted around a repo, I'd prefer it if all overrides were in a single .eslintrc.js file at the root

packages/graphql-typescript-definitions/test/document.test.ts
```diff
@@ -1,5 +1,6 @@
 import * as path from 'path';
 
+// eslint-disable-next-line @shopify/typescript/prefer-build-client-schema
```
BPScott, line 3: The [migration to fix this seems pretty straightfoward](https://github.com/Shopify/web-configs/blob/dab62e4f72e14c1f33ae6a6a626f890e2a872f3e/packages/eslint-plugin/docs/rules/typescript/prefer-build-client-schema.md). are we able to do it?

packages/graphql-typescript-definitions/src/print/document/utilities.ts
```diff
@@ -1,5 +1,5 @@
 import {upperCaseFirst} from 'upper-case-first';
-/* eslint-disable @typescript-eslint/ban-ts-ignore */
+/* eslint-disable @typescript-eslint/ban-ts-comment */
```
BPScott, line 2: I think we can change the `@ts-ignore` linnes to be `@ts-expect-error TS needs this implict reference though we never use it explicitly` then we can remove these eslint-disable lines

packages/graphql-validate-fixtures/test/validate.test.ts
```diff
@@ -2,6 +2,7 @@
 
 import {join} from 'path';
 
+// eslint-disable-next-line @shopify/typescript/prefer-build-client-schema
```
BPScott, line 5: Mentioned this migration above. same question.

packages/jest-mock-apollo/src/test/MockApolloLink.test.ts
```diff
@@ -1,6 +1,7 @@
 import {readFileSync} from 'fs';
 import path from 'path';
 
+// eslint-disable-next-line @shopify/typescript/prefer-build-client-schema
```
BPScott, line 4: as above

packages/react-i18n/src/test/manager.test.ts
```diff
@@ -524,7 +524,7 @@ function createTranslationPromise(
     };
   });
 
-  // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
   // @ts-ignore (It thinks that the resolve/ reject are not assigned yet,
```
BPScott, line 528: use `@ts-expect-error It thinks that the resolve/ reject are not assigned yet` and you shouldn't need the eslint-disable

packages/graphql-typescript-definitions/package.json
```diff
@@ -30,7 +30,6 @@
     "chalk": "^4.0.0",
     "change-case": "^4.1.1",
     "chokidar": "^3.3.1",
-    "core-js": "^3.0.0",
```
vsumner, line 33: good catch

BPScott (approved):
All looking pretty good.

A few little nitpicky things inline but we can come back to them later if desired as it's mostly personal preference stuff. Nothing that needs to block this being merged.

packages/graphql-typescript-definitions/package.json
dahukish: Yeah it slipped by many times. Thanks!

.eslintignore
```diff
@@ -7,3 +7,6 @@ packages/*/build
 **/*.js
 **/*.esnext
 **/*.mjs
+**/sewing-kit.config.ts
```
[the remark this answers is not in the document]
dahukish: like this: https://github.com/Shopify/quilt/blob/068b50ce8a31164364016a3de744c1418b07aaf2/config/sewing-kit/index.ts#L24

vsumner (approved):
LGTM!

dahukish:
> All looking pretty good.
> 
> A few little nitpicky things inline but we can come back to them later if desired as it's mostly personal preference stuff. Nothing that needs to block this being merged.

Updating to `buildClientSchema` is pretty non-trivial as we would need to rebuild the type definitions into introspection queries and from graphql-js you just use buildSchema: https://graphql.org/graphql-js/constructing-types/ 🤷

PR #1887, opened 2021-05-07
Update  release docs

dahukish:
## Description

The current release process for production is no longer valid. Updating to eliminate friction for contributors.


dahukish:
These changes are no valid.

PR #1888, opened 2021-05-07
Update lerna.json and yarn release command

dahukish:
## Description

Currently deploys in quilt are broken: https://shipit.shopify.io/shopify/quilt/production/deploys/1256411
As of the lerna 3 bump.

This should fix them.

## Type of change

This is a fix for the main quilt package itself.

## Checklist

 N/A


vsumner (changes requested):
LGTM, but drop the `release-beta` command too. That will also break.

PR #1894, opened 2021-05-14
Fix to get broken graphql libs out the door

dahukish:
## Description

Fixes the issue with the missing bins in `graphql-typescript-definitions` and `graphql-validate-fixtures`.

## Type of change

## Checklist

### DIFF
--- a/packages/graphql-config-utilities/CHANGELOG.md
+++ b/packages/graphql-config-utilities/CHANGELOG.md
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Changed
+
+- Refactor `getGraphQLProjects` to take default projects into consideration. [#1894](https://github.com/Shopify/quilt/pull/1894)
 
 ## 2.0.1 - 2021-05-07
 
--- a/packages/graphql-typescript-definitions/CHANGELOG.md
+++ b/packages/graphql-typescript-definitions/CHANGELOG.md
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Fixed
+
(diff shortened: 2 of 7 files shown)


shipit.yml
```diff
@@ -15,3 +15,7 @@ ci:
 dependencies:
   post:
     - yarn run build: {timeout: 1800}
+
+deploy:
+  override:
+    - node_modules/.bin/lerna publish from-git --yes
```
dahukish, line 21: This is temporary and will need to be deployed via a throw-away release branch and shipit stack until we get shipit fixed.

tests/consistent-package-json.test.ts
```diff
@@ -75,6 +75,11 @@ packages.forEach(
             // We want to make sure the first set of items are always in a fixed
             // order. Most importantly, the `!*.tsbuildinfo` exclusion must be
             // after `build/*`.
+            /* eslint-disable jest/no-if */
+            if (packageJSON?.bin) {
```
dahukish, line 79: hacky but only some of the packages need to export `bin/*` with the files entry in `package.json` we could also break this into to sets of looping tests.

shipit.yml
BPScott: I'm pretty ok with this being merged to master, so quilt uses `from-git` for releases for a bit. That seems better than "deploys from master are blocked'

BPScott (approved):

vsumner (approved):
I am using this in a beta in my branch for core-js and it worked as expected in web.

packages/graphql-config-utilities/CHANGELOG.md
```diff
@@ -5,7 +5,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+- Refactor `getGraphQLProjects` to take default projects into consideration. [#1894](https://github.com/Shopify/quilt/pull/1894)
```
atesgoral, line 10: Add a `### Changed` section above?

packages/graphql-typescript-definitions/CHANGELOG.md
```diff
@@ -5,7 +5,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+- Fix broken file exports. [#1894](https://github.com/Shopify/quilt/pull/1894)
```
atesgoral, line 10: `### Fixed` section?

PR #1897, opened 2021-05-17
force the registry to be public

vsumner:
## Description
Force the registry to be public. This prevents matching on old private packages.

Note there is a version of `@shopify/react-server` that has not been published yet.

### DIFF
--- a/shipit.yml
+++ b/shipit.yml
@@ -14,4 +14,6 @@ ci:
     - node-build (14.x)
 dependencies:
   post:
+    - npm config set @shopify:registry https://registry.yarnpkg.com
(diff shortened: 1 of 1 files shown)


GoodForOneFare (approved):

packages/react-server/src/index.ts
```diff
@@ -1,3 +1,4 @@
+// release comment
```
atesgoral, line 1: Assuming Lerna looks at commit hashes and not file diffs (?), you could do a `git commit --alow-empty -m "Release commit"` to tickle git with no code changes.

vsumner: It looks for file changes.

PR #1902, opened 2021-05-18
[react-i18n] Fix unformatCurrency utility stripping significant digits on RSD currency

lhoffbeck:
## Description
Part of a fix to resolves Shopify/shopify issue [#284738](https://github.com/Shopify/shopify/issues/284738), along with [shopify/web PR #43731](https://github.com/Shopify/web/pull/43731).

RSD supports [2 decimal places](https://github.com/Shopify/quilt/blob/0ca29bf2108fda54ae20b0552cd5d6658f9f9a24/packages/react-i18n/src/constants/currency-decimal-places.ts#L120), but (per [Christian Jaekl](https://shopify.slack.com/archives/C7TJQLVC7/p1621350321010700?thread_ts=1621347490.003900&cid=C7TJQLVC7)) 
> ...they've become worthless and fallen out of use.  (1 para ~= $0.0001 USD currently, so 1RSD is approximately 1 US cent). ... Stripe still considers RSD to have subunits.  In Chrome, `Intl.NumberFormat` treats RSD as if it has no decimals, which is what you're getting if you call formatCurrency without a `form:` parameter.

For example:

```
new Intl.NumberFormat('en-CA', {style: 'currency', currency: 'RSD'}).format('1.55')
	-> "RSD 2"
new Intl.NumberFormat('en-CA', {style: 'currency', currency: 'CAD', form: 'short}).format('1')
	-> "$1.55"
```

Because of this, when our [unformatCurrency](https://github.com/Shopify/quilt/blob/ae01acaa8e02f494a793c044d019fc03c202a6cf/packages/react-i18n/src/i18n.ts#L266) utility calls [currencyDecimalSymbol](https://github.com/Shopify/quilt/blob/ae01acaa8e02f494a793c044d019fc03c202a6cf/packages/react-i18n/src/i18n.ts#L578), the `formatCurrency` rounds all RSD values to the nearest Int--unless you pass a `form: short`.

This change adds the `form: short` option to the `currencyDecimalSymbol` utility and updates the tests to check for this use case.

To 🎩 you can remove the change and the RSD test should fail.

## Type of change

## Checklist

### DIFF
--- a/packages/react-i18n/CHANGELOG.md
+++ b/packages/react-i18n/CHANGELOG.md
@@ -5,9 +5,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-## Unreleased
+<!-- Unreleased -->
 
-- Fixed `currencyDecimalSymbol` function returning `DECIMAL_NOT_SUPPORTED` for RSD
+- Fixed `currencyDecimalSymbol` function returning `DECIMAL_NOT_SUPPORTED` for RSD [#1902](https://github.com/Shopify/quilt/pull/1902)
 
 ## 5.3.4 - 2021-04-13
 


packages/react-i18n/CHANGELOG.md
```diff
@@ -5,7 +5,8 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+- Fixed `currencyDecimalSymbol` function returning `DECIMAL_NOT_SUPPORTED` for RSD
```
atesgoral, line 9:
```suggestion
<!-- Unreleased -->

- Fixed `currencyDecimalSymbol` function returning `DECIMAL_NOT_SUPPORTED` for RSD [#1902](https://github.com/Shopify/quilt/pull/1902)
```

lhoffbeck: saw this just after I pushed the change up 🙈

lhoffbeck: ahh will fix :)

atesgoral (approved):

atesgoral:
@lhoffbeck `@shopify/react-i18n@5.3.8` is released

PR #1906, opened 2021-05-20
Bump required node version to >=12.14.0  and build using rollup

BPScott:
## Description

A pair of breaking changes.

- Update required node version to be >=12.14.0, and add an engines.node key to each package.json to enforce that.
- Update sewing-kit-next build config to use the rollup build.

This PR also contains some CI simplifications copied from #1908. Hopefully that can get merged then I'll rebase this PR.

## Type of change

Every package: MAJOR


packages/graphql-tool-utilities/package.json
```diff
@@ -20,6 +20,9 @@
     "url": "https://github.com/Shopify/quilt/issues"
   },
   "homepage": "https://github.com/Shopify/quilt/blob/main/packages/graphql-tool-utilities/README.md",
+  "engines": {
+    "node": ">=12.14.0"
+  },
   "dependencies": {
     "apollo-codegen-core": "0.36.5",
     "core-js": "^3.0.0",
```
atesgoral, line 28: Remove?

gems/quilt_rails/dev.yml
```diff
@@ -4,7 +4,7 @@ up:
   - bundler
   - node:
       yarn: v1.19.1
```
atesgoral, line 6: Should we bump yarn to 1.22.5 to match the parent dev.yml?

gems/quilt_rails/dev.yml
```diff
@@ -4,7 +4,7 @@ up:
...
-      version: v10.18.1
+      version: v12.14.0
```
atesgoral, line 7: `version` coming after `yarn` is giving me a cognitive dissonance :/ Could we reorder them while we're at it?

dahukish (approved):
:shipit:

packages/graphql-tool-utilities/package.json
dahukish: This keeps getting added always

BPScott: Yep. I removed the usage, but somehow missed the dependency.

BPScott:
Alrighty then. Major release time. Let's do this!

PR #1907, opened 2021-05-20
[react-i18n] Add `none` option for formatCurrency

lhoffbeck:
## Description

On shopify/web, we occasionally have a need to do currency-style formatting _without_ a currency symbol--most prominent use-case is a text field that shows a money value but the currency symbol shouldn't be editable (and shown as a prefix instead).

This PR adds a `none` type to the `CurrencyFormatOptions`.

## Type of change

- [x] **react-i18n** Minor: New feature (non-breaking change which adds functionality)

## Checklist


packages/react-i18n/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- Unreleased -->
 
+- Added a `none` option to the `CurrencyFormatOptions` that returns a currency-formatted string without a currency symbol [#1907](https://github.com/Shopify/quilt/pull/1907)
```
michenly, line 10: This would be a minor (new feature added) instead of a path update.

michenly (changes requested):
The code looks great 😄  

Please add this new interface to the documentation and than we are good to go!

michenly (approved):
🚢

PR #1908, opened 2021-05-20
Simplify CI config

BPScott:
## Description

This PR refactors our CI suite, moving from multiple simultaneous jobs to running building, linting, unit tests and e2e tests in sequence.

The net result is one job that has a slightly slower wall clock time for all jobs to complete (4.5ish mins => 5.5ish mins), but takes less CPU time overall - as the multiple jobs duplicate effort in doing yarn installs.

This means one job per node version (i.e. 3 checks total) instead of 4 jobs per node version (i.e. 12 checks total) which is easier to read in the checks list, and easier to configure. This is worth the extra minute or so taken IMO.

TODO once I get approval: Update required checks in GH settings


atesgoral (approved):

.github/workflows/node-ci.yml
```diff
@@ -32,167 +32,37 @@ jobs:
...
-          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
-          restore-keys: |
-            ${{ runner.os }}-yarn-
-
-      - uses: actions/cache@v2
-        id: build-cache
-        name: Restore build cache
-        with:
-          # index packages files need to be cached for e2e tests
           path: |
```
BPScott, line 43: I don't really like this cache config. Restoring build output feels like it is an avenue to false confidence where we think content is being generated but it's actually just restored from the cache.

shipit.yml
```diff
@@ -1,17 +1,8 @@
 ci:
   require:
-    - node-unit-tests (10.x)
-    - node-unit-tests (12.x)
-    - node-unit-tests (14.x)
```
dahukish, line 5: 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥

dahukish (approved):
:shipit:

atesgoral (approved):

PR #1909, opened 2021-05-21
fix: resolve url to only be path and search

rdadoune:
## Description

Fix Koa mock-context so that `ctx.url` (and `ctx.originalUrl`) works the same way as it does in Koa and as it's [defined by Node.js](https://nodejs.org/api/http.html#http_message_url).

> Request URL string. This contains only the URL that is present in the actual HTTP request. Take the following request:
>
> ``` 
> GET /status?name=ryan HTTP/1.1
> Accept: text/plain 
> ```

## Type of change

## Checklist

### DIFF
--- a/packages/jest-koa-mocks/src/create-mock-context/test/create-mock-context.test.ts
+++ b/packages/jest-koa-mocks/src/create-mock-context/test/create-mock-context.test.ts
@@ -11,7 +11,7 @@ describe('create-mock-context', () => {
     const context = createContext({method, url});
 
     expect(context.method).toBe(method);
-    expect(context.url).toBe(url);
+    expect(context.url).toBe('/admin');
   });
 
   it('defaults status to 404', () => {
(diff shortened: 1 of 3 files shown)


atesgoral (approved):

atesgoral (approved):

packages/jest-koa-mocks/src/create-mock-context/test/create-mock-context.test.ts
```diff
@@ -62,22 +62,23 @@ describe('create-mock-context', () => {
     const context = createContext({url: STORE_URL});
 
     const {url, originalUrl, host, origin, path, protocol} = context;
-    expect(url).toBe(STORE_URL);
-    expect(originalUrl).toBe(STORE_URL);
+    expect(url).toBe('/admin');
```
ttemrahcelik, line 65: these parts are gonna be fun when unified admin removes this prefix. 😞

ttemrahcelik (approved):

mstar-kk (commented):
I have the same problem, please merge as soon as possible.

PR #1910, opened 2021-05-21
[storybook-a11y] Adds option to disable animation

dfmcphee:
## Description

This adds the ability to disable animations when running the a11y tests to allow us to fix issues like https://github.com/Shopify/web/issues/42730.

## Type of change

- [x] storybook-a11y-test - Minor: New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/packages/storybook-a11y-test/src/index.ts
+++ b/packages/storybook-a11y-test/src/index.ts
@@ -96,13 +96,19 @@ const isAutcompleteNope = violation => {
   return isAutocompleteAttribute && hasNope;
 };
 
-const testPage = (iframePath, browser, timeout) => {
+const testPage = (iframePath, browser, timeout, disableAnimation) => {
   return async id => {
     console.log(` - ${id}`);
 
     try {
       const page = await browser.newPage();
       await page.goto(`${iframePath}?id=${id}`, {waitUntil: 'load', timeout});
+      if (disableAnimation) {
+        await page.addStyleTag({
+          content:
+            '* {transition: none !important; animation: none !important; }',
+        });
+      }
       const result = await page.evaluate(() =>
         window.axe.run(document.getElementById('root'), {}),
       );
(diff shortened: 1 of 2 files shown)


alex-page (approved):
Nice feature. Thanks @dfmcphee

packages/storybook-a11y-test/src/index.ts
```diff
@@ -96,13 +96,19 @@ const isAutcompleteNope = violation => {
...
     console.log(` - ${id}`);
 
     try {
       const page = await browser.newPage();
       await page.goto(`${iframePath}?id=${id}`, {waitUntil: 'load', timeout});
+      if (disableAnimation) {
+        await page.addStyleTag({
+          content:
+            '* {transition: none !important; animation: none !important; }',
+        });
```
kaelig, line 110: Is it possible that it could break scripts that rely on `animationend`?

If that's fine, then let's make it more aggressive, and also target pseudo elements ([source](https://github.com/zaqqaz/visual-unit-tests/blob/e4ebade1b8e046dbdbf659983b987ae21566221f/src/utils/testUtils.ts#L61-L70)):

```css
*,
*::after,
*::before {
  transition-delay: 0s !important;
  transition-duration: 0s !important;
  animation-delay: -0.0001s !important;
  animation-duration: 0s !important;
  animation-play-state: paused !important;
  caret-color: transparent !important;
}
```

However, if that's going to be an issue, I suggest using this hack to accelerate animations and might work as well:

```js
await page._client.send('Animation.setPlaybackRate', {playbackRate: 10});
```

kaelig (changes requested):

packages/storybook-a11y-test/src/index.ts
dfmcphee: I tried disable with the client animation options but couldn't get them to work as I expected them too. I updated the styles to include pseudo elements and instead of pausing and changing the delay, I changed all of the durations to a millisecond to make sure the `animationend` events still fire. I also was unsure about why the caret color was needed there, so I didn't include that part.

kaelig (approved):

PR #1913, opened 2021-05-22
Update badges to use shields.io and add downloads

atesgoral:
## Description

Update badges to use shields.io and add downloads column


kaelig:
It gets a bit messy, any chance you can specify widths so all buttons look like they have the same font size and align nicely in the column?

Screen_Shot_2021-05-24_at_5_17_00_PM

atesgoral:
Closing for now. Will have to deal with the layout regression + update all the individual READMEs.

atesgoral:
The wrong references are a different problem: #1916

PR #1918, opened 2021-05-25
Add URL validation to predicates and validators

mzlock:
## Description

I recently had to write my own URL validation in a form I was building. This seems like some handy validation to add to the quilt stack as a shared resource, so I'm proposing adding `isURL` and `isSecureURL` validation to the predicates and validators.

## Type of change

- `predicates` Minor: adds new `isURL` and `isSecureURL` predicates
- `react-form` Minor: adds new `isURL` and `isSecureURL` validators

## Checklist


GoodForOneFare:
Megan is no longer at Shopify, and this is 2 years out of date.  Feel free to open in a new PR if these could be useful!

PR #1920, opened 2021-05-27
Update countries_en fixture

lhoffbeck:
## Description

While [updating shopify/web to the latest version of address-mocks]( https://github.com/Shopify/web/pull/43910#issuecomment-847906777), we found that we were missing the `provinceKey` field from the `countries_en.ts`. ProvinceKey was removed in [this PR](https://github.com/Shopify/quilt/pull/1610), where `The mocks for the countries call were created by doing a call to the live service.`

Live service is returning the `provinceKey` again so I just called the service and regenerated the mocks.

## Type of change

- [x] **address-mocks** Patch: Bug (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/packages/address-mocks/src/fixtures/countries_en.ts
+++ b/packages/address-mocks/src/fixtures/countries_en.ts
@@ -4,9 +4,10 @@ const data = {
       {
         name: 'Afghanistan',
         code: 'AF',
+        continent: 'Asia',
         phoneNumberPrefix: 93,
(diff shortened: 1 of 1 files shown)


lhoffbeck:
@hassaanali1 FYI

atesgoral (approved):

atesgoral:
@lhoffbeck `@shopify/address-mocks@2.0.1` has been released

PR #1922, opened 2021-05-27
Use scoped names for badges

### LINKED ISSUE #1916: The recently migrated graphql-* packages are not scoped to @shopify
We should:

- [ ] Publish updates to those packages to add "moved to @shopify/..." notice at the top of their READMEs.
- [ ] Change the package names to add the `@shopify` scope
- [ ] Publish
- [ ] Track down consuming libs/repos and do the updates

atesgoral:
## Description

Closes #1916 

Consider unscoped packages when generating badges. Also use the full name with scope in package links.

## Type of change

Just a root README fix.

## Checklist


plopfile.js
```diff
@@ -169,7 +169,12 @@ function getPackages(type = 'js') {
     if (existsSync(packageJSONPath)) {
       const {name, description} = require(packageJSONPath);
 
-      acc.push({name: name.replace('@shopify/', ''), description});
+      acc.push({
+        name: name.replace('@shopify/', ''),
```
BPScott, line 173: I think i'd find this less confusing if `name` was the full name (`@shopifiy/async` etc) and we had a 'shortName' prop that stripped off the namespaces (`async`)

PR #1923, opened 2021-05-28
Prioritise using allowlist from ShopifySecurityBase

dougedey-shopify:
## Description

Fixes warnings in applications that use quilt rails in combination with Shopify Security Base

```
DEPRECATION WARNING: whitelist is deprecated and will be removed from Rails 6.2 (use allowlist instead) (called from call_proxy at /Users/dougedey/.gem/ruby/2.7.3/gems/quilt_rails-3.4.2/lib/quilt_rails/react_renderable.rb:23)
```

The whitelist method is deprecated and we should be using allowlist when it is available, this was we're compatible with apps using an old version of ShopifySecurityBase where the allowlist method was added.

This is not a breaking change

## Type of change

- [x] Quilt_rails Patch: Bug (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/gems/quilt_rails/CHANGELOG.md
+++ b/gems/quilt_rails/CHANGELOG.md
@@ -7,11 +7,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 <!-- ## Unreleased -->
 
+### Changed
+
+- Prioritize using `allowlist` over `whitelist` from ShopifySecurityBase ([#1923](https://github.com/Shopify/quilt/pull/1923))
+
 ## 3.4.2 - 2021-02-23
 
 ### Changed
 
-- Claerify quilt_rails error ([#1752](https://github.com/Shopify/quilt/pull/1752))
+- Clarify quilt_rails error ([#1752](https://github.com/Shopify/quilt/pull/1752))
 
 ## 3.4.1 - 2021-02-18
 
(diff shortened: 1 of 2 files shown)


gems/quilt_rails/CHANGELOG.md
```diff
@@ -7,11 +7,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 <!-- ## Unreleased -->
 
+### Changed
+
+- Prioritize using `allowlist` over `whitelist` from ShopifySecurityBase (#1923)[https://github.com/Shopify/quilt/pull/1923]
```
atesgoral, line 12:
```suggestion
- Prioritize using `allowlist` over `whitelist` from ShopifySecurityBase ([#1923](https://github.com/Shopify/quilt/pull/1923))
```

michenly (approved):
I dont know ruby/rails all that well. But a basic 🎩 does run without error.

atesgoral (approved):

dougedey-shopify:
After this is merged, I'll do a new PR to bump the dependency on Shopify Security base and remove this check, I think that will be a minor release (but open to it being a major)

dougedey-shopify:
Actually, I can't force an optional dependency version in the Gemspec, I'd have to enforce the Dependency in the gemspec, not a `if this dependency exists it must be above this version`, unless I do that in the code, but that isn't something I like doing.

PR #1926, opened 2021-05-31
Downgrade formdata-polyfill

vsumner:
## Description
This PR downgrades and locks `formdata-polyfill` to `3.0.20` to remove the dependency on `fetch-blob` which doesn't support versions of node less than 14.

Fixes (issue #)

## Type of change

- [X] `@shopify/polyfills` Patch: Bug (non-breaking change which fixes an issue)

## Checklist


atesgoral (approved):

BPScott (approved):
Preemptive approval pending CI

PR #1927, opened 2021-05-31
Update eslint-plugin 40.2.3

vsumner:
## Description

Fixes (issue #)

## Type of change

## Checklist


BPScott:
Cheeky request, but while you're touching basically everything, would it be ok to change prettier's [`avoidParens`](https://github.com/Shopify/quilt/blob/main/.prettierrc#L2) value to `avoid` so it matches [our suggested config](https://github.com/Shopify/web-configs/blob/main/packages/prettier-config/index.json)

vsumner:
> Cheeky request, but while you're touching basically everything, would it be ok to change prettier's [`avoidParens`](https://github.com/Shopify/quilt/blob/main/.prettierrc#L2) value to `avoid` so it matches [our suggested config](https://github.com/Shopify/web-configs/blob/main/packages/prettier-config/index.json)

Done!

BPScott (approved):
I'm suprised by the `Array<THING>` to `THING[]`  changes - I thought we had `THING[]` originally and eslint recently changed us to `Array<THING>`.

I do prefer `THING[]` but I'm suprised by the churn of changing and then changing back. Was there an update to the shared config that caused this reversion? i can't see anything locally that might do this?

vsumner:
> I'm suprised by the `Array<THING>` to `THING[]` changes - I thought we had `THING[]` originally and eslint recently changed us to `Array<THING>`.
> 
> I do prefer `THING[]` but I'm suprised by the churn of changing and then changing back. Was there an update to the shared config that caused this reversion? i can't see anything locally that might do this?

I think I fixed it in: https://github.com/Shopify/web-configs/pull/212

BPScott:
Oh delightful, I totally missed that PR :D

PR #1928, opened 2021-05-31
Change all require statement to import (exclude any require statement…

michenly:
… wrap in if statement)

As part of the Vite experiment, it was discovered that our `@shopify/polyfills` packages are actually shipping `mjs` files that uses `require` statement. Which make this package NOT a true ESM module. 

## Description

This PR changes
- most require statement to import statement
- remove empty `export {}` (because we no longer need the signal that those file are ESM)

What was NOT done but maybe should be consider:
these 3 require statements are only use in jest mode, and also wrap in a if statement. Questionable on how it should be updated to use import statement.
https://github.com/Shopify/quilt/blob/polyfill-proper-esm/packages/polyfills/src/formdata.jest.ts#L3-L5
https://github.com/Shopify/quilt/blob/polyfill-proper-esm/packages/polyfills/src/intersection-observer.jest.ts#L3-L5
https://github.com/Shopify/quilt/blob/polyfill-proper-esm/packages/polyfills/src/mutation-observer.jest.ts#L3-L6

## Type of change

## Checklist

### DIFF
--- a/packages/polyfills/CHANGELOG.md
+++ b/packages/polyfills/CHANGELOG.md
@@ -7,6 +7,10 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## Unreleased -->
 
+### Fixed
+
+- Replace `require` statements with `import` statements (except for the conditional Jest polyfills) so we get a proper strict ESM module.
+
 ## 3.0.1 - 2021-05-31
 
 ### Fixed
--- a/packages/polyfills/src/fetch.browser.ts
+++ b/packages/polyfills/src/fetch.browser.ts
@@ -1,1 +1,1 @@
-require('whatwg-fetch');
+import 'whatwg-fetch';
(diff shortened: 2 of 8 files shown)


vsumner (approved):
Makes sense to me. Transpile should handle changing these back to requires if necessary.

packages/polyfills/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## Unreleased -->
 
+### Fixed
+
+- Replace `require` statement with `import` statement (except for the jest polyfills wrap in if statement) so we get a proper stric ESM module.
```
atesgoral, line 12:
```suggestion
- Replace `require` statements with `import` statements (except for the conditional Jest polyfills) so we get a proper strict ESM module.
```

PR #1931, opened 2021-06-02
[DO NOT MERGE] Update quilt to support webpack 5

vsumner:
## Description
This PR updates `graphql-mini-transforms` to support webpack 5.

Closes https://github.com/Shopify/sewing-kit/issues/2689
Closes https://github.com/Shopify/sewing-kit/issues/2693

## Type of change
- [X] `graphql-mini-transforms` Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] `web-worker` Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X] `sewing-kit-koa` Major: Breaking change to support webpack 5 + remove `vendor.js`
- [X] `react-server` Major: Breaking change to support webpack 5
- [X] `magic-entries-webpack-plugin` Major: Breaking change to support webpack 5

## Checklist

### DIFF
--- a/packages/graphql-mini-transforms/package.json
+++ b/packages/graphql-mini-transforms/package.json
@@ -38,17 +38,18 @@
   },
   "dependencies": {
     "@jest/transform": "^27.0.2",
-    "@types/fs-extra": "^9.0.11",
-    "@types/webpack": "^4.41.11",
     "fs-extra": "^9.1.0",
     "graphql": ">=14.5.0 <16.0.0",
-    "graphql-typed": "^1.1.0",
-    "loader-utils": "^2.0.0"
+    "graphql-typed": "^1.1.0"
   },
   "devDependencies": {
     "@types/common-tags": "^1.8.0",
     "@types/loader-utils": "^1.1.3",
-    "common-tags": "^1.8.0"
+    "common-tags": "^1.8.0",
+    "webpack": "^5.40.0"
+  },
+  "peerDependencies": {
+    "webpack": ">=5.38.0"
   },
   "files": [
     "build/",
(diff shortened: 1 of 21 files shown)


packages/graphql-mini-transforms/package.json
```diff
@@ -25,18 +25,19 @@
   },
   "dependencies": {
     "@jest/transform": "^26.6.2",
-    "@types/fs-extra": "^8.1.0",
-    "@types/webpack": "^4.41.11",
+    "@types/fs-extra": "^9.0.11",
```
vsumner, line 28: bumped to match `fs-extra` version.

packages/graphql-mini-transforms/package.json
```diff
@@ -25,18 +25,19 @@
...
-    "loader-utils": "^2.0.0"
+    "graphql-typed": "^1.0.0"
   },
   "devDependencies": {
     "@types/common-tags": "^1.8.0",
-    "@types/loader-utils": "^1.1.3",
     "common-tags": "^1.8.0"
   },
+  "peerDependencies": {
+    "webpack": ">=5.38.0"
```
vsumner, line 39: I think this can be looser but for now it makes sense.

atesgoral:
Still trying to understand the motivation of https://github.com/Shopify/quilt/blob/main/packages/web-worker/src/webpack-parts/loader.ts#L100-L110. The `Compilation` object no longer has a `cache` property and I can't figure out the webpack 5 analogue without understanding what this code aims to do.

Some archeology led me to believe it might have been inspired by https://github.com/webpack-contrib/worker-loader/blob/v0.8.0/index.js#L45-L53

Which today has evolved into: https://github.com/webpack-contrib/worker-loader/blob/master/src/index.js#L106-L113

PR #1932, opened 2021-06-02
Add `appendToEntries` to allow outside sources to add to the final entries object before compilation

dahukish:
## Description

Essentially we need a way to expose the final entries generated by this plugin without exposing to much or have to know about anything besides entries. The new callback allows the entries to still be passed to webpack while allowing for optional final edits before compilation. This is currently needed to support HMR properly using sewing-kit.

Related: https://github.com/Shopify/sewing-kit/pull/2686

## Type of change

## Checklist

### DIFF
--- a/packages/magic-entries-webpack-plugin/src/magic-entries-webpack-plugin.ts
+++ b/packages/magic-entries-webpack-plugin/src/magic-entries-webpack-plugin.ts
@@ -3,37 +3,76 @@ import {basename, resolve} from 'path';
 import globToRegExp from 'glob-to-regexp';
 import {Compiler, Entry} from 'webpack';
 
+type EntryOption = Compiler['options']['entry'];
+
 export interface Options {
   pattern: string;
   folder: string | string[];
   nameFromFile: (file: string) => string;
+  appendToEntries: {[entryName: string]: string};
 }
-
-type EntryOption = Compiler['options']['entry'];
 
 /**
  * A webpack plugin that automatically configures webpack entries by checking the filesystem
- * @param config
+ * @param folder - string or array of strings to look for when auto detection of entries happens
+ * @param pattern - regex to be used to compare files that will be included as entries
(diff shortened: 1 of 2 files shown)


packages/magic-entries-webpack-plugin/src/magic-entries-webpack-plugin.ts
```diff
@@ -3,37 +3,46 @@ import {basename, resolve} from 'path';
...
   folder: string | string[];
   nameFromFile: (file: string) => string;
+  appendToEntries: {[entryName: string]: string};
 }
 
-type EntryOption = Compiler['options']['entry'];
-
 /**
  * A webpack plugin that automatically configures webpack entries by checking the filesystem
  * @param config
```
TheMallen, line 17: could we add docs here for what `appendToEntries` does? I wasn't really sure from just the name until I read through the code a few times

packages/magic-entries-webpack-plugin/src/test/magic-entries-webpack-plugin.test.ts
```diff
@@ -243,46 +243,63 @@ describe('magic-entries-webpack-plugin', () => {
...
+              'basic-ts.entry.client.ts': ENTRY_B,
+              'basic-tsx.entry.client.tsx': ENTRY_B,
+            }),
+            MagicEntriesPlugin.client(),
+          ],
+        };
+        const name = 'node-virtual-module-entrypoints';
+        await withWorkspace(name, async ({workspace}) => {
+          await workspace.write('index.js', ENTRY_A);
+          await workspace.write('cats.js', CATS_MODULE);
```
TheMallen, line 283: 🐱 🐈

TheMallen (approved):
Haven't tophatted but code and approach checks out to me. Would like a bit of documentation around the new callback but otherwise 👍

PR #1940, opened 2021-06-09
Fix references in the polyfills migration guide

MitMaro:
## Description

The migration guide for the `@shopify/polyfills` package contain references that are missing the "s" from "polyfills". This change fixes those references.

## Type of change

This is a simple documentation fix that does not require a new release.

## Checklist


vsumner (approved):
Thank you!

PR #1945, opened 2021-06-16
Update @jest/transforms to 27

vsumner:
## Description
This PR updates `@jest/transform` in `graphql-mini-transforms` to `27`.

Fixes (issue #)

## Type of change

- [X] `graphql-mini-transforms`  Minor: New feature (non-breaking change which adds functionality)

## Checklist


packages/graphql-mini-transforms/package.json
```diff
@@ -24,8 +24,8 @@
     "node": ">=12.14.0"
   },
   "dependencies": {
-    "@jest/transform": "^26.6.2",
     "@types/fs-extra": "^9.0.11",
+    "@jest/transform": "^27.0.2",
```
BPScott, line 28: We should keep this alphabetical

BPScott (commented):
Do we need to modify the file jest / jest-simple transforms to be default exports of a single object rather than multiple named exports as part of jest 27 support?

packages/graphql-mini-transforms/package.json
vsumner: Fixed.

BPScott (approved):

PR #1946, opened 2021-06-16
Update `fs-extra` to latest version used in `lerna@4.0.0`

dahukish:
## Description

We currently have many versions of `fs-extra` lingering about. The current version used in v4 of lerna drops support for older versions of node < 10.x. This change also needed so we can updated sewing-kit to the latest version of lerna as well.

## Type of change

## Checklist


vsumner (approved):
LGTM. Since these are dependencies instead of devDependencies could you please add a changelog entry?

dahukish:
> LGTM. Since these are dependencies instead of devDependencies could you please add a changelog entry?

100 percent. Was going to clean up after review.

TheMallen (approved):
🙏

PR #1952, opened 2021-06-22
add integration_test_error config

yask123:
## Description
In Partners App we want to run end to end testing using cypress which boots the rails app in `test` environment for end to end testing using headless browser. 

Added `config.integration_test_error`(default `true`) option to disable raising error in rails `test` environment

## Checklist


michenly:
Can you add something in the documentation that expose this feature and explanation on when this is needed?

gems/quilt_rails/lib/quilt_rails/configuration.rb
```diff
@@ -13,6 +13,7 @@ def initialize
       self.react_server_protocol = ENV['REACT_SERVER_PROTOCOL'] || "http"
       self.logger                = ::Logger.new($stdout)
       self.mount                 = true
+      self.integration_test_error = true
```
TheMallen, line 16: I think I'd rather see this called like `allow_integration_test` and default to false, since I think that is a bit clearer than `integration_test_error`

TheMallen (changes requested):
I'd like to see the option have a bit nicer naming, other than that looks good

TheMallen (approved):
🎉

PR #1958, opened 2021-06-23
⏫  Upgrade React, React testing supports react 17

### LINKED ISSUE #1955: Add support for React 17.x to react-testing
## Overview

React 17 does not work with the current version of `@shopify/react-testing` out of the box for a few reasons:
- `react-reconciler` is out of date, and needs to be bumped to `0.26.1`
- `react@17.x` changes the name of an internal property [we use (abuse)](https://github.com/Shopify/quilt/blob/1fb9cea87b419de5721276d5ada31a23e9dd427e/packages/react-testing/src/root.tsx#L259-L259) from `_reactInternalFiber` to `_reactInternals` which causes our tree wrapping and traversal to fail
- our peer-dependency yells at the user because it is set to `>=16.8.0 <17.0.0`

I believe we can safely support both 17 and 16, simply by checking for either property and using whichever is there in our [fiber traversal logic](https://github.com/Shopify/quilt/blob/1fb9cea87b419de5721276d5ada31a23e9dd427e/packages/react-testing/src/root.tsx#L259-L259).

## Motivation

React 17 has been out for a while, and we should support it so that folks can take advantage of the newest versions as they are released

### LINKED ISSUE #1957: Upgrade Quilt's React to 17
## Overview

Quilt should build with 17, once the testing library supports it.

## Motivation

React 17 good

TheMallen:
closes #1957 
closes #1955 

## Description
This PR does the minimum work to get quilt building with React 17, mostly making `@shopify/react-testing` play nice with it.

**Full breakdown**
- upgrades quilt's local copy of React to `17.0.2`
- upgrades `react-reconciler` to latest
- fixes a breakage with `@shopify/react-testing` using a renamed internal property
- adds tests to ensure we don't break 16.x consumers
- widens `react-testing`'s stated version support
- updates docs

There will still be WARNs on package install because the peer dependencies in our other packages are unchanged. (I will perform a follow-up PR to change them).

## Tophatting
The tests should be sufficient, but if you want you can yalc this version of react-testing into another app (eg. Shrink-Ray) and run the test suite there. Next upgrade that repo to React 17 and run the tests again. Things should work swimingly.

## Checklist


TheMallen:
**update** I pulled the react 17 upgrade out of this PR as there were some non react-testing related issues I'd like to sort out separately. I'll be updating the description to be more reflective shortly.

packages/react-testing/src/compat.ts
```diff
@@ -0,0 +1,11 @@
+import {ReactInstance, Fiber} from './types';
+
+export function getInternalFiber(instance: ReactInstance): Fiber {
```
BPScott, line 3: Silly nitpick - perhaps run with the most recent name `getInternals` so our terminology matches the most recent react naming

packages/react-testing/README.md
```diff
@@ -665,6 +665,10 @@ Enzyme is a very popular testing library that heavily inspired the approach this
 
 While the premise of writing tests that mirror user actions is compelling, basing all tests off the raw DOM being produced becomes unmanageable for larger apps.
 
+### What versions of React does this support?
+
+The React versions this library supports are spelled out via a [peer dependency in the package.json](https://github.com/Shopify/quilt/blob/main/packages/react-testing/package.json#L47-L47)
```
BPScott, line 670: Looks like we need to update the react/react-dom peer deps to `"^16.8.0 || ^17.0.0"`

BPScott (commented):
Remember to update the react/react-dom peer deps to `"^16.8.0 || ^17.0.0"`.

What's our plan for updating the version of react installed in development? Keep it at 16.whatever for now?

vsumner (approved):
🎩 ✅

TheMallen:
@BPScott 
> What's our plan for updating the version of react installed in development? Keep it at 16.whatever for now?
I originally had that rolled into this PR but it broke a couple other things, so my current plan is to fix those as individual PRs and then do one to upgrade.

BPScott (approved):
Love how straightforward this was :D

PR #1964, opened 2021-06-30
Switch hook on useMountedRef from useEffect to useLayoutEffect to support React 17

### LINKED ISSUE #1959: React-hooks should work with React 17
## Overview

[`useMountedRef`](https://github.com/Shopify/quilt/blob/react-testing-support-react-17/packages/react-hooks/src/hooks/mounted-ref.ts#L3) seems to break in tests when upgrading to React 17. Since this is a relatively foundational hook it could be an outsize issue. Let's fix it.

image

lucabezerra:
## Description

`useEffect`'s cleanup function is no longer synchronous, so we switched to `useLayoutEffect` to preserve the behaviour of `useMountedRef` in React 17+.

Source: https://reactjs.org/blog/2020/08/10/react-v17-rc.html#effect-cleanup-timing

Fixes #1959

## Checklist

## Tophat Instructions

On package.json, upgrade React and ReactDOM to the following:
```
    "react": "17.0.2",
    "react-dom": "17.0.2",
```
And the resolutions to:
```    
    "@types/react": "17.0.11",
    "@types/react-dom": "17.0.8"
```
Then run `yarn && yarn test react-hooks`. The tests should pass!

### DIFF
--- a/packages/react-hooks/src/hooks/mounted-ref.ts
+++ b/packages/react-hooks/src/hooks/mounted-ref.ts
@@ -2,7 +2,6 @@ import {useRef} from 'react';
 
 import {useIsomorphicLayoutEffect} from './isomorphic-layout-effect';
 
-
 export function useMountedRef() {
   const mounted = useRef(true);
 
--- a/packages/react-hooks/src/hooks/mounted-ref.ts
+++ b/packages/react-hooks/src/hooks/mounted-ref.ts
@@ -1,9 +1,12 @@
-import {useRef, useLayoutEffect} from 'react';
+import {useRef} from 'react';
+
+import {useIsomorphicLayoutEffect} from './isomorphic-layout-effect';
+
 
 export function useMountedRef() {
   const mounted = useRef(true);
 
-  useLayoutEffect(() => {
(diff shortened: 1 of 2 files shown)


atesgoral (approved):

BPScott:
IIRC recall useLayoutEffect doesn't work on the server-side. If we make this change then I think we're going get a lot of logspam or errors about that.

https://github.com/Shopify/quilt/pull/1813 seems to be a related issue where useEffect being async is a pain (though that was an issue in the mount rather than unmount phase. Perhaps that might offer some inspiration

TheMallen:
@BPScott  I'm not sure we need `useMountedRef` to work on the server, since "mounted" doesn't have a super clear meaning there. We could use the isomorphic version anyway I guess though.

BPScott (approved):

BPScott:
Gave this a tick. I'd love to see a test for this but I'm not sure that would work

TheMallen:
@BPScott the existing tests fail without this changeset, so we're covered imo

BPScott:
Oh perfect! in that case ignore me :)

PR #1972, opened 2021-07-15
[react-i18n] Fixes i18n date format timezone humanize bug

qt314:
currently when you pass options like `timeZoneName` with a `humanize` style to `formatDate`, they come out with time repeated and the second timezone as lower case.

Screen Shot 2021-07-15 at 1 36 24 PM

## Description

Fixes (issue #)
fix by making sure that when breaking apart the date object for sending to translation, the irrelevant display properties on the date objects are undefined. I also removed the `.toLocaleLowerCase();` since it wouldn't do anything other than make time zones or eras lowercase which isn't inline with polaris anyways https://polaris.shopify.com/content/grammar-and-mechanics#section-dates-numbers-and-measurements

## Type of change

## Checklist


packages/react-i18n/src/i18n.ts
```diff
@@ -499,9 +500,11 @@ export class I18n {
     const timeZone = options?.timeZone;
     const time = this.formatDate(date, {
       ...options,
+      ...undefinedDateDisplayProps,
+      timeZoneName: options?.timeZoneName,
       hour: 'numeric',
       minute: '2-digit',
-    }).toLocaleLowerCase();
```
qt314, line 504: this couldn't do anything other than making the time zone name or era lowercase, which is not in line with [polaris](https://polaris.shopify.com/content/grammar-and-mechanics#section-dates-numbers-and-measurements).

Probably was introduced to make am/pm lower case but they are already lowercase

packages/react-i18n/src/constants/index.ts
```diff
@@ -38,6 +38,18 @@ export enum Weekday {
   Saturday = 'saturday',
 }
 
+export const undefinedDateDisplayProps = {
+  era: undefined,
```
atesgoral, line 42: I don't think that it's a good idea to maintain a list of properties to wipe. We'd have to keep updating this as we add support for additional options.

A better fix could be to pluck the options that we allow in that context, instead of spreading the entire options object.

But, I think the root cause of this issue could be something else: I suspect something to do with an unexpected DateStyle being passed in.

atesgoral:
Can you point to the app/route this issue was observed in? The problem might be originating from the code that uses the `humanizePastDate` method.

packages/react-i18n/src/constants/index.ts
qt314: passing in options still works (see my tests). I only use this when the functions isolate certain things like `monthDay` and `time`. right now, these aren't working properly. For example if I pass in `timeZoneName` as an option, `monthDay` will display the time zone.

qt314: `humanize` splits up the date and the time so it can send it to a translation i.e. `{date} at {time}`. Right now the functions aren't splitting them up properly when options are passed in. I need to clear the options and only specify the properties that are associated with `date` and `time` separately. So I pass the options through, but only the `time` props to the time consts and only the `date` props to the date consts

qt314:
> Can you point to the app/route this issue was observed in? The problem might be originating from the code that uses the `humanizePastDate` method.

I haven't used it in web yet but if you run my tests without my changes you can see the problem. when you pass in additional options to `formatDate` with `style: DateStyle.Humanize` like `timeZoneName`, format Date breaks

packages/react-i18n/src/i18n.ts
```diff
@@ -499,9 +500,11 @@ export class I18n {
     const timeZone = options?.timeZone;
     const time = this.formatDate(date, {
       ...options,
+      ...undefinedDateDisplayProps,
```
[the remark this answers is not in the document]
atesgoral: It's almost always a better pattern to explicitly pick the properties that are relevant instead of removing properties you don't want. That way, the contract is clear. We should probably do a destructing assignment from the options object to pick out the properties that are allowed to passed down and recompose the options object with those.

joswong13-shopify:
An example is we are trying to use `i18n.formateDate(new Date(`some date from graphql`), { timeZoneName: 'short', style: DateStyle.Humanize})` and the return value would be this `Jul. 1, MST at 11:44 a.m. mst` where the timezone is lower cased and repeated.

packages/react-i18n/src/i18n.ts
```diff
@@ -564,10 +563,56 @@ export class I18n {
...
+      timeStyle,
+      calendar,
+      dayPeriod,
+      numberingSystem,
+      hourCycle,
+      fractionalSecondDigits,
+      localeMatcher,
+      formatMatcher,
+      hour12,
+      timeZone,
```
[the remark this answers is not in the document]
atesgoral: O boy. Yeah, if it's a different story if we have sooo many properties we want to pass down. Let me sleep over this :)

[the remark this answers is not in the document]
qt314: I actually think I added way to many properties, I was looking at another `DateTimeFormatOptions` type definition. It should be way smaller now :)

dahukish (approved):
LGTM. I do wonder if this might break some test output if there are snapshots etc expecting lower case strings.

qt314:
@dahukish ooo thats a good point. Just thinking about it now, the only things that would be affected by `toLowerCase` would be adding in an option to display `timezone` or `epoch`. but since both of those options were broken in the first place I'm assuming no one used them so it _should_ be safe from that

joswong13-shopify (approved):

PR #1973, opened 2021-07-19
[react-i18n] Enable async translation loading for Metro (React Native bundler) with static paths

samrmur:
## Description
**TLDR**
Adds an option `useStaticImports` to use static imports instead of dynamic. Enables asynchronous loading for Metro bundler which is used in React Native development.

**Full Description**
Updates translation template to allow for generating static paths. This is required by the Metro bundler is it needs to know what modules to package at build time.

From [Metro documentation](https://facebook.github.io/metro/docs/bundling): `When bundling, each of the modules gets assigned a numeric id, meaning no dynamic requires are supported. Requires are changed by its numeric version, and modules are stored in different possible formats.`

Generated template with static paths would look similar to:
```javascript
import _en from './translations/en.json';
import React from 'react';
import {useI18n} from '@shopify/react-i18n';

export default function MyComponent() {
  const [i18n] = useI18n({
    id: 'MyComponent_3259328325',
    fallback: _en,
    translations(locale) {
      if (["de", "fr", "zh-TW"].indexOf(locale) < 0) {
        return;
      }
  
      return (async () => {
        let dictionary;

        if (locale === "de") {
          dictionary = await import(/* webpackChunkName: "MyComponent_${defaultHash}-i18n", webpackMode: "lazy-once" */ "./translations/de.json")
        } else if (locale === "fr") {
          dictionary = await import(/* webpackChunkName: "MyComponent_${defaultHash}-i18n", webpackMode: "lazy-once" */ "./translations/fr.json")
        } else {
          dictionary = await import(/* webpackChunkName: "MyComponent_${defaultHash}-i18n", webpackMode: "lazy-once" */ "./translations/zh-TW.json")
        }

        return dictionary && dictionary.default;
      })();
    }
  });
  return i18n.translate('key');
}
```

This is far less performant than using template strings but enables bundlers without dynamic import support to use asynchronous loading. An option `useStaticImports` has been added to babel options to allow developers to enable this feature. By default it will be off so that other applications using this library with dynamic imports will be unaffected by the change.

Option Usage:
```js
// babel.config.js
{
  plugins: [
    ['@shopify/react-i18n/babel', {useStaticImports: true}],
  ],
}
```

## Type of change

## Checklist


BPScott:
react-native also does not support dynamic imports, and our readme details how to set `mode: 'from-dictionary-index'` to make imports synchronous: https://github.com/Shopify/quilt/blob/main/packages/react-i18n/README.md#how-do-i-get-this-i18n-library-to-work-with-react-native

How does this PR's new option differ from the existing behaviour of setting `mode: 'from-dictionary-index'`? Is there a way that we can consolidate these two approaches? What happens if you mix and match those two options, e.g. `{mode: 'from-dictionary-index', importMode: 'dynamic'}` or `{mode: 'from-dictionary-index', importMode: 'static'}?`

michenly:
Is the source of the problem just that metro is not using webpack as a bundler and cannot determine which files should included in the bundle?

michenly:
IMO if that is the case, we can support 3 modes of `mode: 'bundle-with-webpack' | 'bundle-with-metro' | 'from-dictionary-index' = 'bundle-with-webpack'` just to make it really clear there will never be cross over b/t the different modes.

I also like that we make it really clear that webpack is a dependency to use the default transform.

samrmur:
> Is the source of the problem that react native is not using webpack as a bundler and cannot determine which files should be included in the bundle?

Exactly, Metro needs to know the exact files to include in the resulting bundle which it can't determine if the import is dynamic.

BPScott:
Heya sorry, stuff got busy there. 

It sounds like this new mode is a 3rd option to use instead of the existing  'from-generated-index' and 'from-dictionary-index' rather than additional behaviour that modifies both of those modes. Rather than creating a new `importMode` or `bundler` option, it feels like should just be a new value on the `mode` option. That way you make it clear you're selecting from a single list of four transform outcomes (default, from-dictionary-index, from-generated-index, this new mode), rather than suggesting that `mode` and `importMode` have some funky combinatorial relationship where 3 `mode`s, and two `importMode`s give you six possible codepaths.

I'm pretty sure this new behaviour will be useful for handling translations inside npm libraries, as rollup also wants to know the exact list of importable files rather than being given an ambiguous `import(someVariable + '.json')` too so I wouldn't want to tie the new name to `metro` or name things based on the bundler that shall consume it.

I don't really like the naming involving `dynamic/static` either - when I think dynamic or static imports I think using the `import('./somefile.js')` function is a dynamic import and using the `import BLAH from './somefile.js'` keyword is a static import. But in this cas `importMode: static` is still using an `import('./somefile.js')` function, but the argument is an explicit filename rather than containing a variable. when I read this PR initially i expected 'importMode: static' to use be using `import BLAH from './somefile.js'` keywords but that's not the case.

---

Rephrased as concrete suggestions rather than just moaning:

What if we remove the concept of `importMode`, instead opt into this behaviour by setting a new `mode` value.  i've not got a really good suggestion for this though. Perhaps `from-explicit-paths`?

samrmur:
> Heya sorry, stuff got busy there.
> 
> It sounds like this new mode is a 3rd option to use instead of the existing 'from-generated-index' and 'from-dictionary-index' rather than additional behaviour that modifies both of those modes. Rather than creating a new `importMode` or `bundler` option, it feels like should just be a new value on the `mode` option. That way you make it clear you're selecting from a single list of four transform outcomes (default, from-dictionary-index, from-generated-index, this new mode), rather than suggesting that `mode` and `importMode` have some funky combinatorial relationship where 3 `mode`s, and two `importMode`s give you six possible codepaths.
> 
> I'm pretty sure this new behaviour will be useful for handling translations inside npm libraries, as rollup also wants to know the exact list of importable files rather than being given an ambiguous `import(someVariable + '.json')` too so I wouldn't want to tie the new name to `metro` or name things based on the bundler that shall consume it.
> 
> I don't really like the naming involving `dynamic/static` either - when I think dynamic or static imports I think using the `import('./somefile.js')` function is a dynamic import and using the `import BLAH from './somefile.js'` keyword is a static import. But in this cas `importMode: static` is still using an `import('./somefile.js')` function, but the argument is an explicit filename rather than containing a variable. when I read this PR initially i expected 'importMode: static' to use be using `import BLAH from './somefile.js'` keywords but that's not the case.
> 
> Rephrased as concrete suggestions rather than just moaning:
> 
> What if we remove the concept of `importMode`, instead opt into this behaviour by setting a new `mode` value. i've not got a really good suggestion for this though. Perhaps `from-explicit-paths`?

Yeah I agree there, definitely removes the ambiguity of how the relationships between `mode` and `importMode`. The new behaviour though is actually meant to enable both `default` and `from-generated-index` for bundlers with explicit file names. What do you think of two additional options: `with-explicit-paths` & `from-generated-index-with-explicit-paths`?

michenly:
I would rather not add new additional options that we need to maintain if its not going to be use. Is mobile planning on using both `from-generated-index-with-explicit-paths` and `with-explicit-paths`?

BPScott:
I agree with Michelle, we should avoid adding new modes unless there is a proven need for them.

samrmur:
> I would rather not add new additional options that we need to maintain if its not going to be use. Is mobile planning on using both `from-generated-index-with-explicit-paths` and `with-explicit-paths`?

Shopify Mobile currently only has a need for `from-explicit-paths`. I can add that option instead and if the need does come up for requiring `from-generated-index` in mobile open another PR to enable it.

packages/react-i18n/babel-plugin.md
```diff
@@ -290,3 +290,20 @@ module.exports = {
...
+##### Usage
+
+```js
+// webpack.config.js
+{
+  plugins: [
+    ['@shopify/react-i18n/babel', {mode: 'with-explicit-paths'}],
+  ],
+}
+```
```
BPScott, line 309: Would be great to see a demo of how the output is transformed here

packages/react-i18n/src/babel-plugin/test/babel-plugin.test.ts
```diff
@@ -380,6 +380,132 @@ describe('babel-plugin-react-i18n', () => {
...
+
+                  switch (locale) {
+                    case "de":
+                      dictionary = await import(/* webpackChunkName: "MyComponent_${defaultHash}-i18n", webpackMode: "lazy-once" */ "./${TRANSLATION_DIRECTORY_NAME}/de.json");
+                      break;
+                    case "fr":
+                      dictionary = await import(/* webpackChunkName: "MyComponent_${defaultHash}-i18n", webpackMode: "lazy-once" */ "./${TRANSLATION_DIRECTORY_NAME}/fr.json");
+                      break;
+                    default:
+                      dictionary = await import(/* webpackChunkName: "MyComponent_${defaultHash}-i18n", webpackMode: "lazy-once" */ "./${TRANSLATION_DIRECTORY_NAME}/zh-TW.json");
```
BPScott, line 494: slightly outside the bounds of this pr but a nice change might be to avoid the intermediate variable, and thus possibly slightly simplify the output could be to chain the promise so you can immediatly return

```
const returnDefault = (mod) => mod && mod.default;

// then in each case
case: "fr":
  return await import(/*magic comment*/"./${TRANSLATION_DIRECTORY_NAME}/zh-TW.json").then(returnDefault);

```

packages/react-i18n/src/babel-plugin/test/babel-plugin.test.ts
```diff
@@ -405,6 +531,43 @@ function optionsForFile(filename, withTranslations = false) {
   };
 }
 
+function staticImportOptionsForFile(filename, translationOption) {
+  let dummyPath;
+
+  switch (translationOption) {
+    case 'fallbackOnly':
+      dummyPath = path.join(
```
BPScott, line 539: it looks like the import code is longer than the size of the fixtures. inline the fixtures here rather than storing in separate files perhaps?

packages/react-i18n/babel-plugin.md
```diff
@@ -102,7 +102,7 @@ useI18n({
 
 ### `mode`
 
-Type: `undefined`, `from-generated-index`, `from-dictionary-index`
+Type: `undefined`, `from-generated-index`, `from-dictionary-index`, `with-explicit-paths`
```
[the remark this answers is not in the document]
BPScott: perhaps `with-dynamic-paths` as the default? It'd be nice to keep a bit of symetry between the name of the default and the name of the mode being added in this PR. If folks could come up with a better sounding pairing I'd be happy with that too ;)


I'd like to keep the magic comment when using `with-explicit-paths`. I suspect that in my rollup library building stuff I'll find this mode useful - When I compile a library with rollup I want to keep the webpack magic comment in the code present in the compiled output of the library, as that code shall then be ran through webpack when a consuming application bundles the library into its build output, and we'll want to provide those compilation hints to webpack.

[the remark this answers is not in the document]
BPScott: nice! it looks like `webpackMode=lazy` is the default so we could omit that part of the comment at least

packages/react-i18n/src/babel-plugin/test/babel-plugin.test.ts
samrmur: I think they need to separate folders/files so that the plugin detect what files are available in the adjacent `translations` folder. I only need the fixtures to test how the code is generated based on the amount of available locales. Added some symlinks to existing test files.

packages/react-i18n/src/babel-plugin/test/babel-plugin.test.ts
```diff
@@ -380,6 +380,122 @@ describe('babel-plugin-react-i18n', () => {
...
+
+                return (async () => {
+                  const returnDefault = (dict) => dict && dict.default;
+
+                  switch (locale) {
+                    case "de":
+                      return await import(/* webpackChunkName: "MyComponent_${defaultHash}-i18n" */ "./${TRANSLATION_DIRECTORY_NAME}/de.json").then(returnDefault);
+                    case "fr":
+                      return await import(/* webpackChunkName: "MyComponent_${defaultHash}-i18n" */ "./${TRANSLATION_DIRECTORY_NAME}/fr.json").then(returnDefault);
+                    default:
```
BPScott, line 486: Using default as the final option feels weird. Should this be `case: 'zh-TW'`?

Now that we're using `then` rather than await to deal with what to do with the result of a promise, I don't think we need the `await` keyword here anymore, `return import(...` might be enough. At that point, we're not using await, so I don't think we need this to be an immediately invoked async function anymore.

```js
 translations(locale) {
  if (["de", "fr", "zh-TW"].indexOf(locale) < 0) {
    return;
  }

  const returnDefault = (dict) => dict && dict.default;

   switch (locale) {
    case "de":
      return import(/* webpackChunkName: "MyComponent_${defaultHash}-i18n" */ "./${TRANSLATION_DIRECTORY_NAME}/de.json").then(returnDefault);
  // other cases here
  }
```

packages/react-i18n/src/babel-plugin/test/fixtures/fallbackOnlyTranslation/translations/en.json
```diff
@@ -0,0 +1 @@
+../../adjacentTranslations/translations/en.json
```
BPScott, line 1: Not a fan of the symlinks tbh. I think it'd be better to keep these as regular files, even if there is a bit of duplication.

packages/react-i18n/src/babel-plugin/test/babel-plugin.test.ts
```diff
@@ -380,6 +380,122 @@ describe('babel-plugin-react-i18n', () => {
...
+        await normalize(
+          `import _en from './${TRANSLATION_DIRECTORY_NAME}/en.json';
+          import React from 'react';
+          import {useI18n} from '@shopify/react-i18n';
+
+          export default function MyComponent() {
+            const [i18n] = useI18n({
+              id: 'MyComponent_${defaultHash}',
+              fallback: _en,
+              translations(locale) {
```
BPScott, line 402: In this case can we immediately return? There doesn't seem to be much value in having these empty checks.

Would `translations(local) {return} be enough?

BPScott (commented):
Left some ideas inline about how this could be simplified. I think we can handle this whole thing with using `.then()` rather than needing any async/await syntax, and thus we don't need the immediatly invoked function

packages/react-i18n/src/babel-plugin/babel-templates.ts
```diff
@@ -89,3 +96,62 @@ export function i18nGeneratedDictionaryCallExpression(
...
+  const fileName = isExplicit ? file.id : `\${${file.id}}`;
+
+  const fileLocation = `./${TRANSLATION_DIRECTORY_NAME}/${fileName}.json`;
+  const wrappedLocation = isExplicit
+    ? `"${fileLocation}"`
+    : `\`${fileLocation}\``;
+  const webpackMode = isExplicit ? '' : ', webpackMode: "lazy-once"';
+
+  return `import(/* webpackChunkName: "${bindingId}-i18n"${webpackMode} */ ${wrappedLocation})`;
+}
```
michenly, line 149: I find this block of logic difficult to understand, how about simply:

```suggestion
function generateImportDefinition(
  bindingId: string,
  explicitFileName?: string,
) {
  if (explicitFileName) {
    return `import(/* webpackChunkName: "${bindingId}-i18n" */ "./${TRANSLATION_DIRECTORY_NAME}/${explicitFileName}.json")`;
  }

  return `import(/* webpackChunkName: "${bindingId}-i18n", webpackMode: "lazy-once" */ \`./${TRANSLATION_DIRECTORY_NAME}/\${locale}.json\`)`;
}
```

packages/react-i18n/src/babel-plugin/index.ts
```diff
@@ -107,7 +111,7 @@ export default function injectWithI18nArguments({
             return;
           }
 
-          const {mode, defaultLocale} = state.opts;
+          const {mode = 'with-dynamic-paths', defaultLocale} = state.opts;
```
michenly, line 114: 😍

packages/react-i18n/src/babel-plugin/test/babel-plugin.test.ts
michenly: translations is actually optional, so we can 🔥 the entire translate method when there is no additional lang other than default

michenly (changes requested):

michenly (approved):

packages/react-i18n/src/babel-plugin/test/babel-plugin.test.ts
samrmur: Michelle mentioned this as an overall improvement we can make since the `translations` function is not required: https://github.com/Shopify/quilt/pull/1973#discussion_r691377071. I agree with Michelle that it should be a separate PR since it's an overall general improvement. I can add it to my workload as something to do right after this PR.

packages/react-i18n/babel-plugin.md
```diff
@@ -102,9 +101,101 @@ useI18n({
...
+  id: 'MyComponent_<hash>',
+  fallback: _en,
+  async translations(locale) {
+    if (['jp'].indexOf(locale) < 0) {
+      return;
+    }
+
+    return (async () => {
+      return import(
+        /* webpackChunkName: "MyComponent_<hash>-i18n" */ `./translations/jp.json`
```
michenly, line 160:
```suggestion
        /* webpackChunkName: "MyComponent_<hash>-i18n" */ "./translations/jp.json"
```

packages/react-i18n/babel-plugin.md
```diff
@@ -102,9 +101,101 @@ useI18n({
...
+
+      switch (locale) {
+        case 'fr':
+          return import(
+            /* webpackChunkName: "MyComponent_<hash>-i18n" */ `./translations/fr.json`
+          ).then(returnDefault);
+        default:
+          return import(
+            /* webpackChunkName: "MyComponent_<hash>-i18n" */ `./translations/jp.json`
+          ).then(returnDefault);
```
michenly, line 193:
```suggestion
          return import(
            /* webpackChunkName: "MyComponent_<hash>-i18n" */ "./translations/fr.json"
          ).then(returnDefault);
        default:
          return import(
            /* webpackChunkName: "MyComponent_<hash>-i18n" */ "./translations/jp.json"
          ).then(returnDefault);
```

packages/react-i18n/src/babel-plugin/test/babel-plugin.test.ts
BPScott: Cool, happy to do that in a follow up :)

BPScott:
re-review comments - you don't seem to have acted upon the suggestions in https://github.com/Shopify/quilt/pull/1973#discussion_r694247657 that should allow for simpler output. Did you give this a try and it didn't work?

samrmur:
> re-review comments - you don't seem to have acted upon the suggestions in [#1973 (comment)](https://github.com/Shopify/quilt/pull/1973#discussion_r694247657) that should allow for simpler output. Did you give this a try and it didn't work?

The suggestions were implemented but were not reflected in the documentation. Just updated the documentation to account for removal of `default`

PR #1977, opened 2021-07-28
changed short timezone to uppercase in humanized dates

sophschneider:
## Description

Fixes (issue #)

we need `toLocaleLowercase` on humanized times to ensure that `am` and `pm` are lowercase, but this also affects short timezones that should be uppercase. This separates out the timezone so that it remains uppercase

## Type of change

## Checklist


dahukish (approved):
LGTM

PR #1978, opened 2021-07-28
⬆️ support graphql 15.x

TheMallen:
## Description

Our version ranges meant that projects with dependencies that want `graphql@15.x` would have (possibly worrisome) duplicate graphql dependencies.

## Checklist

### DIFF
--- a/packages/graphql-validate-fixtures/src/validate.ts
+++ b/packages/graphql-validate-fixtures/src/validate.ts
@@ -480,19 +480,23 @@ function validateValueAgainstType(
   }
 
   if (isEnumType(type)) {
-    return type.parseValue(value) == null
-      ? [
-          error(
-            keyPath,
-            `value does not match enum ${nameForType(
-              type,
-            )} (available values: ${type
-              .getValues()
-              .map((enumValue) => enumValue.value)
-              .join(', ')})`,
-          ),
-        ]
-      : [];
+    try {
+      return type.parseValue(value) == null
+        ? [
+            error(
+              keyPath,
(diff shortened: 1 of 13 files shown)


packages/graphql-validate-fixtures/test/__snapshots__/validate.test.ts.snap
```diff
@@ -1179,7 +1179,7 @@ Object {
   "validationErrors": Array [
     Object {
       "keyPath": "generation",
-      "message": "value does not match enum Generation (available values: FIRST, SECOND, THIRD)",
+      "message": "Value \\"SEVENTH\\" does not exist in \\"Generation\\" enum.",
```
TheMallen, line 1182: these snapshots changed more because they come from the try / catch we added (they used to be a custom error)

packages/graphql-validate-fixtures/src/validate.ts
```diff
@@ -480,19 +480,23 @@ function validateValueAgainstType(
...
-              type,
-            )} (available values: ${type
-              .getValues()
-              .map((enumValue) => enumValue.value)
-              .join(', ')})`,
-          ),
-        ]
-      : [];
+    try {
+      return type.parseValue(value) == null
```
TheMallen, line 484: we still try to do this so that we support 14.x

TheMallen:
hmm looks like we got some TS errors, weird since I didn't see those running `build` locally. I'll look into them

dahukish (approved):
LGTM

PR #1980, opened 2021-07-30
Update to the latest version of skn

BPScott:
## Description

sewing-kit-next has undergone some juicy changes. Simplifying and reworking its architecture and plugins. Here's a PR to pull our skn dependencies up to the latest versions. The build output of all packages remains identical to main, with the exception of some extra ts-expect-error comments added because TS is confusing itself as I haven't updated the contents of `sewing-kit-plugin-quilt`- to use the latest versions of the sk packages.

Stuff of note:

- the `@sewing-kit/{config,hooks,plugins}` packages have all be consolidated into the core package.
- `plugin-javascript` and its built-in babel config is gone, replaced by `plugin-babel`, which requires you explicitly state the babel preset that you wan to use. Unsuprisingly we've chosen `@shopify/babel-preset` as the baseline babel config.
-  We no longer generate d.ts file entrypoints. This means that the `types` and `typesVersions` keys in package.json now point into the build folder directly

### Testing

To compare the old and new build output i did the following:

- Checked out  the `main` branch into `~/projects/quilt`, Ran `rm -rf .sewing-kit; git clean -f -x packages/*/build; yarn build` to generate a fresh build of main.
- Checkout out this branch into another folder an ran `rm -rf .sewing-kit; git clean -f -x packages/*/build; yarn build` to generate a fresh build of this branch
- Ran `for PACKAGENAME in $(ls -1 packages) do; diff -ru packages/$PACKAGENAME/build ~/projects/quilt/packages/$PACKAGENAME/build -x '*.tsbuildinfo'` to run diffs of every packages's build output.
- Noted that the only differences were as a result of code changes in `sewing-kit-plugin-quilt`, and d.ts files no longer being generated at the root (as we now point types/typesVersions directly into the build folder)

## Type of change

- all: patch


config/sewing-kit/index.ts
```diff
@@ -2,42 +2,37 @@ import {
...
+} from '@sewing-kit/core';
+import {babel} from '@sewing-kit/plugin-babel';
 import {packageBuild} from '@sewing-kit/plugin-package-build';
 import {} from '@sewing-kit/plugin-jest';
 
-import {addLegacyDecoratorSupport} from './plugin';
-
 export function quiltPackage({
   jestEnv = 'jsdom',
-  useReact = false,
```
BPScott, line 17: I've removed `useReact` as there is no harm in enabling it for all packages

config/sewing-kit/index.ts
```diff
@@ -2,42 +2,37 @@ import {
...
   jestTestRunner = 'jest-circus',
+  polyfill = true,
```
BPScott, line 13: The only package that overrides this and sets it to false is `polyfills` because we want that package to uh, provide the polyfills.

config/sewing-kit/index.ts
```diff
@@ -2,42 +2,37 @@ import {
...
+            },
+          ],
+        ],
+      },
+    }),
     packageBuild({
       nodeTargets: 'node 12.14.0',
       browserTargets: 'extends @shopify/browserslist-config',
     }),
-    createProjectBuildPlugin('Quilt.PackageBuild', ({hooks}) => {
```
BPScott, line 28: babelIgnorePatterns no longer exists, and we don't need to add legacy decorator support because `@shopify/babel-preset` already enables it correctly

dahukish (approved):
Did a pair review with @BPScott LGTM

PR #1991, opened 2021-08-10
Deprecate and remove sewing-kit-plugin-quilt

BPScott:
## Description

It uses webpack 4 and legacy sewing-kit-next plugins. It represents
an old way of thinking about sewing-kit plugins.  It's easier to
remove it and start afresh.

## Type of change

- @shopify/sewing-kit-plugin-quilt: BREAKING


vsumner (approved):
LGTM

PR #1996, opened 2021-08-11
Update saddle-up

BPScott:
## Description

Removes an unneeded dependency on old sewing-kit-next packages

## Type of change

Technically react-server patch.  But it only touches a dev dependency so not distributed code has changed


michenly (approved):
byeee

PR #1997, opened 2021-08-12
Update typescript to v4.3.5

BPScott:
## Description

Keeping up to date. We get some perf improvements with incremental builds, and we can stop emitting and publishing typescript generated js files while keeping incremental build support, which leads to smaller package sizes as we're sending less cruft to npm.

You probably want to review this commit by commit, as most of the files changed are the changelogs which were added in the 3rd commit.

## Type of change

- Everything: patch.

### DIFF
--- a/packages/dates/CHANGELOG.md
+++ b/packages/dates/CHANGELOG.md
@@ -9,8 +9,8 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ### Changed
 
-- `formatDate`'s options object now expects its `hourCycle` value to be `'h11', 'h12', 'h23' or 'h24' rather than an arbitary string. Any other value would have caused an error at runtime, this helps us catch that error at compile time. [[#1997](https://github.com/Shopify/quilt/pull/1997)]
-- Updated bulid tooling, types are now compiled with TypeScript 4.3. [[#1997](https://github.com/Shopify/quilt/pull/1997)]
+- `formatDate`'s options object now expects its `hourCycle` value to be 'h11', 'h12', 'h23' or 'h24' rather than an arbitary string. Any other value would have caused an error at runtime, this helps us catch that error at compile time. [[#1997](https://github.com/Shopify/quilt/pull/1997)]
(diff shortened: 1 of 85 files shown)


packages/react-form-state/src/FormState.ts
```diff
@@ -140,14 +140,14 @@ export default class FormState<
 
   // eslint-disable-next-line @shopify/react-prefer-private-members
   public validateForm() {
-    return new Promise((resolve) => {
+    return new Promise<void>((resolve) => {
```
BPScott, line 143: TS now warns if you don't return anything from a promise as it suspects that might be a bug. If that's what you absolutely wanted to do then this is how you say "yes Typescript I really wanted to return nothing from this promise'

packages/dates/CHANGELOG.md
```diff
@@ -5,7 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Changed
+
+- `formatDate`'s options object now expects its `hourCycle` value to be `'h11', 'h12', 'h23' or 'h24' rather than an arbitary string. Any other value would have caused an error at runtime, this helps us catch that error at compile time. [[#1997](https://github.com/Shopify/quilt/pull/1997)]
```
GoodForOneFare, line 12:
```suggestion
- `formatDate`'s options object now expects its `hourCycle` value to be `h11`, `h12`, `h23` or `h24` rather than an arbitary string. Any other value would have caused an error at runtime, this helps us catch that error at compile time. [[#1997](https://github.com/Shopify/quilt/pull/1997)]
```

packages/dates/CHANGELOG.md
```diff
@@ -5,7 +5,12 @@ All notable changes to this project will be documented in this file.
...
+- Updated bulid tooling, types are now compiled with TypeScript 4.3. [[#1997](https://github.com/Shopify/quilt/pull/1997)]
```
GoodForOneFare, line 13: `bulid` => `build` throughout

GoodForOneFare (approved):
👍 Banal nits aside, looks good.

Issue #1895, opened 2021-05-17
Add April21 ApiVersion enum 


karthik-ameex:
latest version does not contsains latest apiversion enum.

atesgoral:
Fixed by #2001

PR #2000, opened 2021-08-14
Updated the enum Api version

Resousse:
## Description

Fixes (issue #1895)

Removal of old unsupported version from both GraphQL proxy and Webhook packages.
Add of supported version that were missing in packages.

## Type of change

- [X]  Patch: Bug (non-breaking change which fixes an issue)

## Checklist


peter-hamilton:
Thanks for bringing this to our attention.

We decided to replace the enum type with a looser string union type in #2001 . This way new versions can be used without waiting for the package to be updated. The change also includes up to date versions.

Resousse:
> Thanks for bringing this to our attention.
> 
> We decided to replace the enum type with a looser string union type in #2001 . This way new versions can be used without waiting for the package to be updated. The change also includes up to date versions.

Ok, great, closing this one!

PR #2002, opened 2021-08-17
add more info to the react-i18n babel plugin README

michenly:
Add some more information on react-i18n babel plugin.

- Separate it from the main package README to babel-plugin/README
- Make it clear that there are 3 different mode in this plugin
- Modify the examples to make it clear what the transform will generate

### DIFF
--- a/packages/react-i18n/README.md
+++ b/packages/react-i18n/README.md
@@ -162,9 +162,9 @@ Translations are provided using two keys in the `withI18n` decorator:
 - `fallback`: a translation file to use when translation keys are not found in the locale-specific translation files. These will usually be your English translations, as they are typically the most complete.
 - `translations`: a function which takes the locale and returns one of: nothing (no translations for the locale), a dictionary of key-value translation pairs, or a promise of one of the above. The `translations` function can also throw and `react-i18n` will handle the situation gracefully. Alternatively, you can pass an object where the keys are locales, and the values are either translation dictionaries, or promises for translation dictionaries.
 
-We recommend that you co-locate your translation files in a `./translations` directory and that you include an `en.json` file ([schema specification](docs/react_i18n_schema.md)) in that directory as your fallback. We give preferential treatment to this structure via a [babel plugin](#Babel) that will automatically fill in the arguments to `useI18n`/ `withI18n` for you.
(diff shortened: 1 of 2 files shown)


packages/react-i18n/src/babel-plugin/README.md
```diff
@@ -0,0 +1,266 @@
...
+  module: {
+    rules: [
+      {
+        test: /\.jsx?$/,
+        use: [
+          {
+            loader: 'babel-loader',
+            options: {
+              plugins: [
+                '@babel/plugin-syntax-dynamic-import',
```
BPScott, line 121: In the first config demo, we show babel.config.js style config here we switch to webpack, and in the final example we claim "webpack.config.js" but the code looks the same as the first.

Can we keep a consistent style. I like the one used in the first config. So this block would be 👍 

```js
plugins: [
  ['@shopify/react-i18n/babel', {mode: 'from-generated-index'}],
],
```

packages/react-i18n/src/babel-plugin/README.md
```diff
@@ -0,0 +1,266 @@
...
+      );
+      return dictionary && dictionary.default;
+    })();
+  },
+});
+```
+
+#### `from-dictionary-index`
+
+_*Use this mode to statically embedding locale-specific translations.*_
```
BPScott, line 180:
```suggestion
_*Use this mode to statically embed locale-specific translations.*_
```

packages/react-i18n/src/babel-plugin/README.md
```diff
@@ -0,0 +1,266 @@
...
+Type: `string`
+
+Default: `en`
+
+#### Setting the default locale to something other than `en`
+
+If you want your default locale to be something else than English because it's not your primary locale, you can pass the `defaultLocale` option to the babel plugin:
+
+```js
+// webpack.config.js
```
BPScott, line 260:
```suggestion
// babel.config.js
```

BPScott (approved):
Looking great!

Some little grammar nits, and a thought on example code inline.

I'd be tempted to have the babel readme be in packages/react-i18n/README-BABEL.md as a sibling of the regular readme so we're not pushing folks into src directories to read documentation. What do you reckon?

packages/react-i18n/src/babel-plugin/README.md
michenly: I am actually going to change it all to `webpack.config.js` instead since webpack is actually a dependency on the default transform

michenly:
also moved the `src/babel-plugin/README.md` to `babel-plugin.md`

PR #2003, opened 2021-08-17
Append virtual modules to ignored paths list

atesgoral:
## Description

Fixes (issue #)

## Type of change

## Checklist

### DIFF
--- a/packages/react-server/src/webpack-plugin/webpack-plugin.ts
+++ b/packages/react-server/src/webpack-plugin/webpack-plugin.ts
@@ -36,6 +36,8 @@ export class ReactServerPlugin {
     (virtualModules as any).apply(compiler);
 
     compiler.hooks.afterEnvironment.tap('ReactServerPlugin', () => {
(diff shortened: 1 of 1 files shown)


packages/react-server/src/webpack-plugin/webpack-plugin.ts
```diff
@@ -34,6 +34,10 @@ export class ReactServerPlugin {
     const modules = this.modules(compiler);
     const virtualModules = new VirtualModulesPlugin(modules);
     (virtualModules as any).apply(compiler);
+
+    compiler.hooks.afterEnvironment.tap('ReactServerPlugin', () => {
+      compiler.watchFileSystem.paths.push(...Object.keys(modules));
+    });
```
michenly, line 40: what if we make use of [`WatchIgnorePlugin`](https://github.com/webpack/webpack/blob/c181294865dca01b28e6e316636fef5f2aad4eb6/lib/WatchIgnorePlugin.js) and use it within this plugin on what to ignore.

atesgoral: that could work

PR #2004, opened 2021-08-19
Fix the formatCurrencyShort that results in double negative nordic regions

egjiri:
## Description

This PR fixes an issue that results in a double negative symbol in the 3 nordic regions (Norway, Sweden, and Finland) that when run through the Intl formatter result in a different negative symbol than the standard one which does not get stripped in our replace. I have updated the replace matcher with a regex that can capture both the standard minus symbol and the nordic one. See screenshot below outlining the problem where I have highlighted the nordic negative symbol in yellow.

### Screenshot showing the issue in the Payouts List page (currently being webified)
Screen Shot 2021-08-19 at 4 21 33 PM

### Screenshot showing the results of the Intl.NumberFormat being run through different locales
Screen Shot 2021-08-19 at 4 18 42 PM

## Type of change

## Checklist

### DIFF
--- a/packages/react-i18n/src/i18n.ts
+++ b/packages/react-i18n/src/i18n.ts
@@ -431,7 +431,7 @@ export class I18n {
       : `${formattedAmount}${shortSymbol.symbol}`;
 
     return amount < 0
-      ? `-${formattedWithSymbol.replace('-', '')}`
+      ? `-${formattedWithSymbol.replace(/[-−]/, '')}`
       : formattedWithSymbol;
   }
 
--- a/packages/react-i18n/CHANGELOG.md
+++ b/packages/react-i18n/CHANGELOG.md
@@ -7,6 +7,10 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Fixed
+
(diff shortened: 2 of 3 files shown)


packages/react-i18n/src/i18n.ts
```diff
@@ -431,7 +431,7 @@ export class I18n {
       : `${formattedAmount}${shortSymbol.symbol}`;
 
     return amount < 0
-      ? `-${formattedWithSymbol.replace('-', '')}`
+      ? `-${formattedWithSymbol.replace(/[-−]/, '')}`
```
movermeyer, line 434: Shouldn't this use whatever negative symbol is used by the language?

(Why does this logic exist in the first place?)

egjiri: I guess for now I won't go again Polaris especially since the current Rails page format all currencies with the minus sign at the front and reverting this logic would result in React & Rails being different and could cause us problems with things being consistent while we're still webifying sections. Maybe once we're at 100% React we can revisit this decision.

For now I'll try to push through with this bug fix

qt314 (approved):

dahukish:
@movermeyer Are you ok with change from an i18n perspective?

movermeyer (approved):

PR #2018, opened 2021-08-27
Use a separate gemspec parser for gems

atesgoral:
## Description

~~Not sure why they were there in the first place. Might be vestiges of some old code.~~

~~*Update*: It's used by the README template to populate gems~~

~~*Update*: Wrote a separate parser for gemspec files.~~

For the top-level README generator where we list Node.js packages and Ruby gems, clone the package info loader and create one specific for gems. We were previously using the hack of including a dummy package.json file in gems to let them get picked up by the generic package crawler.

1. Adding a gem-specific crawler to do rudimentary parsing of gemspec files to pull out key/value pairs (and do some rudimentary parsing)
2. Removing all Node.js stuff from the one and only gem.


TheMallen (approved):
🤔 seems fine, though I don't know if we'll ever have more gems in here :p

atesgoral:
@TheMallen That crossed my mind. But this is more ENGINEERING.

PR #2019, opened 2021-08-30
Update graphql-tool-utilities to allow graphql 15

vsumner:
## Description
This PR updates `apollo-codegen-core` which allows graphql 15 to be installed and removes the last reference to version 14.
Fixes (issue #)

## Type of change

- [X] `graphql-tool-utilities` Minor: New feature (non-breaking change which adds functionality)

## Checklist


michenly (approved):
All CI passes so we should we ok

TheMallen (approved):

Issue #2020, opened 2021-08-30
add ts-jest-mock to allow jest inferred mocks


lbenie:
## Overview

Improves the DX when writing unit tests.

Removes the need to cast everything with `jest.Mock` or other `jest` type constructs.

It works with:
- [x] classes
- [x] functions
- [x] objects

## Type

- [x] New feature
- [ ] Changes to existing features

## Motivation

> What inspired this feature request? What problems were you facing,
> or what else makes you think this should be in `quilt`?

When unit testing with `jest` and `TypeScript` there's a compiler problem that occurs because `TypeScript` doesn't know that a module has been mocked.

We are forced to use casts such as `jest.Mock` and many other variants. It leads to wrongly inferred mocks.

Let's say we have a function called `myFn`

```ts
export const myFn = (a: number, b: number) => a + b
```

For example:

```ts
import {myFn} from 'myModule'

jest.mock('myModule')

it('some test', () => {
  myFn.mockReturnValue() <-- TS doesn't know it's mocked by jest, so the compiler fails
})

```

with `ts-jest-mock`:

```ts
import {myFn} from 'myModule'
import {createMock} from 'ts-jest-mock'

jest.mock('myModule')

const myFnMock = createMock(myFn) <-- this is where the magic happens

it('some test', () => {
  myFn.mockReturnValue('') <-- would fail the TypeScript compiler 
})

it('some test 2', () => {
  myFn.mockReturnValue(2) <-- would pass the TypeScript compiler 
})

```

### Area

- [ ] Add any relevant `Area: <area>` labels to this issue

### Scope

---

## Checklist

- [x] Please delete the labels section before submitting your issue
- [x] I have described this issue in a way that is actionable (if possible)

PR #2021, opened 2021-08-30
feat: adds ts-jest-mock to created type inferred mocks

lbenie:
## Description

Fixes (issue #2020)

I have added [ts-jest-mock](https://github.com/lbenie/ts-jest-mock), a library util helper to `react-testing` which provides a function `createMock` and a type `JestMock` to help with the DX and jest.

I'm simply re-exporting the function and type so it can be used everywhere at in Shopify.

## Type of change

- [x] react-testing Minor: New feature (non-breaking change which adds functionality)

## Checklist


packages/react-testing/src/index.ts
```diff
@@ -4,3 +4,5 @@ export type {Node} from './types';
 export * from './mount';
 export * from './destroy';
 export * from './toReactString';
+export {createMock} from 'ts-jest-mock';
```
TheMallen, line 7: I don't like us adding jest-specific stuff to this entrypoint, currently everything jest related is cordoned off in the matchers one.

packages/react-testing/package.json
```diff
@@ -33,7 +33,8 @@
   "dependencies": {
     "@shopify/useful-types": "^3.0.4",
     "jest-matcher-utils": "^26.6.2",
-    "react-reconciler": "^0.26.2"
+    "react-reconciler": "^0.26.2",
+    "ts-jest-mock": "^1.0.1"
```
TheMallen, line 37: if this is just some external lib why not just use it directly when needed?

TheMallen (changes requested):
I'm not sure about this, I don't like adding jest-dependant stuff to the main entrypoint, and I don't see why folks can't just import this as needed

lbenie:
> I'm not sure about this, I don't like adding jest-dependant stuff to the main entrypoint, and I don't see why folks can't just import this as needed

Thanks for the comment, it is a valid concern and I hear ya. 

I agree that lib could be imported directly when needed, I created this PR after a talk I had with @caropinzonsilva where we could possibly add this here instead so it would be the only entry point for that.

The only benefit I can think of is that the official React testing guides are using jest in their examples and that's the default test runner as well with CRA. With that being said, unless we plan on switching to a different test runner, it helps in the sense that this becomes the entry point to empower devs with more DX and type inferred jest mocks.

I'm still figuring out how we operate here (I'm new!).

thanks again for the feedback let me know your thoughts.

Issue #2029, opened 2021-09-03
Humanize in react-i18n does not distinguish between "1 minute ago" and "1 minutes ago"


aktill:
## Overview

@doctorhyde mentioned over on slack that the web timeline was showing events that happened between 1 and 2 minutes ago with the linguistically awkward description `1 minutes ago` as opposed to `1 minute ago`. Asked if someone could investigate

![shows singular as plural as above](https://screenshot.click/21-09-bjpgl-mg49s.png)
## Motivation

Investigation seemed to indicate that this could be tweaked in i18n.ts in quilt by inserting a case handling `isOneMinuteAgo` between `isLessThanOneMinuteAgo` and `isLessThanOneHourAgo`
<img width="592" alt="Screen Shot 2021-09-03 at 8 13 17 AM" src="https://user-images.githubusercontent.com/24965135/132045940-93082e3a-cfe7-4793-9875-e9a60b1e079f.png">

Have a possible solution underway

## Checklist

- [ X ] Please delete the labels section before submitting your issue
- [ X ] I have described this issue in a way that is actionable (if possible)

movermeyer:
Users need to provide `one` and `other` keys for `lessThanOneHourAgo`:

```json
"lessThanOneHourAgo": {
  "one": "{count} minute ago",
  "other": "{count} minutes ago", 
},
```

Which means that the incorrect `quilt` docs need to get corrected: https://github.com/Shopify/quilt/blob/51fabc952487933f14ed62e9bdaea07606e40713/packages/react-i18n/migration-guide.md#L87-L123

aktill:
Addressed by above linked PR

PR #2030, opened 2021-09-03
[React-i18n] Add case to humanize to handle less than two minutes and more than one minute (replace "1 minutes ago") output

aktill:
## Description

Fixes: https://github.com/Shopify/quilt/issues/2029

Implements logic to handle humanizing of a time that is between 1 and 2 minutes ago, so that `A minute ago` can be output instead of `1 minutes ago`. Reported by PSN as being awkward looking.

This addition adds a `isOneMinuteAgo` check to the `humanizePastDate` function in React-i18n

## Type of change

## Checklist

## Tophat Instructions (mainly for me)

1) publish in quilt using `PKG_NAME=react-i18n yarn tophat`
2) add translation key to web
3) link in web using `yalc link --no-pure @shopify/react-i18n`, remove with `yalc remove @shopify/react-i18n`

### DIFF
--- a/packages/address-consts/index.d.ts
+++ b/packages/address-consts/index.d.ts
@@ -1,0 +1,1 @@
+export * from "./build/ts/index";
--- a/packages/address-mocks/index.d.ts
+++ b/packages/address-mocks/index.d.ts
@@ -1,0 +1,1 @@
+export * from "./build/ts/index";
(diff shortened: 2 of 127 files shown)


aktill:
Not required, alternate fix implemented here: https://github.com/Shopify/web/pull/49140

PR #2032, opened 2021-09-09
Fix doc to give more proper example for pluralization 

iCynara:
## Description
Fix examples


tayre (approved):
Lgtm

trishrempel (approved):

sjahu (approved):
Looks good. I was initially concerned that the calling code wouldn't know what to do with the new keys, but [under the hood](https://github.com/Shopify/online-store-ui/blob/73135aa3df446da9b98c0d0cb4e643841b571aef/src/components/DateTime/DateTime.tsx#L34) it's just making a regular i18n.translate call so it should work just fine. Nice fix

sjahu:
Hmmm, actually is what I just wrote true? @vsumner . Don't you need the `count` variable to be named for pluralization to work?

https://github.com/Shopify/online-store-ui/blob/73135aa3df446da9b98c0d0cb4e643841b571aef/src/components/DateTime/DateTime.tsx#L37

iCynara:
oh right, it just need a merge, but the button is grey out for me, do i need permission or something? @vsumner

PR #2034, opened 2021-09-10
Enable/fix test typing

peter-hamilton:
## Description

Enable type-checking and fix type errors for a batch of packages.

### Changes

- react-async
  - add missing `cancelled` property.
  - changed `noop()` to match expected type.
- react-bugsnag
  - removed unused `ReactPlugin` plugin
- react-form
  - Change return type for `mockField` to allow for `Field<undefined>`
  - cast `onClick` callbacks to `any` in a test where the event is programmatically triggered.
  - add non-null assertions to controlled `Root.find()` returns.
  - add missing argument to `randomVariants()` call
- react-import-remote
  - Used unsafe type casts to mock parts of the DOM.
- react-graphql-universal-provider
  - Removed unused variables
    - In one case, eslint `disable-next-line no-new` was suppressed for a test that with `new ApolloLink()`
- react-hooks
  - define argument types for mock functions
  - add non-null assertions to controlled `Root.find()` returns
- react-html
  - fix typo `nomodule` -> `noModule`
- react-network
  - Correct function return type. 
- react-router
  - Aligned the types returned by `createDefaultHistory` and `createDefaultLocation`. **NOTE:** These functions aren't used within Quilt so it's not clear how to test the changes.
- react-server
  - Import saddle-up matcher in `src/tests/setup.ts`
  - cast Context.body as `NodeJS.ReadableStream` after `createRender` is called.
  - Changed `MockApp` to return `null` instead of `void` in order to make it a valid JSX component.
- react-testing
  - Added missing `options: MountOptions` argument for `mountWithContext` being used in `e2e.test.tsx`. **NOTE:** It's possible `options` SHOULD be an optional parameter. If that's the case, `.createMount` typing should be changed instead.
  - Used hard cast for a case where `Node<unknown>` was being compared to `Element<{}> `

## Type of change

- [X] react-testing  Patch: Bug (non-breaking change which fixes an issue)
- [X] react-server  Patch: Bug (non-breaking change which fixes an issue)
- [X] react-router  Patch: Bug (non-breaking change which fixes an issue)
- [X] react-network  Patch: Bug (non-breaking change which fixes an issue)
- [X] react-import-remote  Patch: Bug (non-breaking change which fixes an issue)
- [X] react-html  Patch: Bug (non-breaking change which fixes an issue)
- [X] react-hooks  Patch: Bug (non-breaking change which fixes an issue)
- [X] react-graphql-universal-provider  Patch: Bug (non-breaking change which fixes an issue)
- [X] react-form  Patch: Bug (non-breaking change which fixes an issue)
- [X] react-bugsnag  Patch: Bug (non-breaking change which fixes an issue)
- [X] react-async  Patch: Bug (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/packages/react-form/src/hooks/list/tests/baselist.test.tsx
+++ b/packages/react-form/src/hooks/list/tests/baselist.test.tsx
@@ -797,7 +797,7 @@ describe('useBaseList', () => {
             </>
           ))}
 
-          <button type="button" onClick={onNewDefault} />
+          <button type="button" onClick={onNewDefault as any} />
         </>
       );
     }
@@ -847,7 +847,7 @@ describe('useBaseList', () => {
           optionName: newDefaultOption,
           optionValue: newDefaultOptionValue,
         },
-      ]);
+      ] as any);
 
       expect(wrapper).toContainReactText(`Default: ${newDefaultPrice}`);
       expect(wrapper).toContainReactText(`Default: ${newDefaultOption}`);
@@ -881,7 +881,7 @@ describe('useBaseList', () => {
           optionName: newDefaultOption,
           optionValue: newDefaultOptionValue,
         },
-      ]);
+      ] as any);
 
       expect(wrapper).toContainReactText(
         `Default: ${newDefaultPrice}Default: ${newDefaultOption}Default: ${newDefaultOptionValue}`,
--- a/packages/react-testing/src/tests/element.test.tsx
+++ b/packages/react-testing/src/tests/element.test.tsx
@@ -458,9 +458,7 @@ describe('Element', () => {
     it('finds all matching nodes', () => {
       const element = new Element(defaultTree, [divOne], defaultRoot);
       expect(
-        element.findAllWhere(
-          (element) => ((element as unknown) as Element<{}>) === componentTwo,
-        ),
+        element.findAllWhere((element) => element.type === componentTwo.type),
       ).toHaveLength(0);
 
       expect(
--- a/packages/react-testing/src/tests/element.test.tsx
+++ b/packages/react-testing/src/tests/element.test.tsx
@@ -198,7 +198,6 @@ describe('Element', () => {
     it('concatenates the text contents of all child elements and child text', () => {
       const childTextOne = 'foo ';
       const childTextTwo = 'bar';
-      const descendantText = ' baz?';
 
       const elementChild = new Element(defaultTree, [], defaultRoot);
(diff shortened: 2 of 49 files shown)


packages/react-testing/src/tests/element.test.tsx
```diff
@@ -458,9 +457,10 @@ describe('Element', () => {
   describe('#findAllWhere()', () => {
     it('finds all matching nodes', () => {
       const element = new Element(defaultTree, [divOne], defaultRoot);
-
       expect(
-        element.findAllWhere((element) => element === componentTwo),
+        element.findAllWhere(
+          (element) => ((element as unknown) as Element<{}>) === componentTwo,
+        ),
```
BPScott, line 463: Thanks to https://github.com/Shopify/quilt/pull/1999, I think you can do

```suggestion
        element.findAllWhere<Element<{}>>((element) => element === componentTwo),
```

BPScott (commented):
Looking very nice! I've dropped a few comments inline asking about `any` usage

packages/react-form/src/hooks/list/tests/baselist.test.tsx
```diff
@@ -797,7 +797,7 @@ describe('useBaseList', () => {
             </>
           ))}
 
-          <button type="button" onClick={onNewDefault} />
+          <button type="button" onClick={onNewDefault as any} />
```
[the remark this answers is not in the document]
peter-hamilton: I chose the brute force cast because the test doesn't deal with MouseEvents and I wanted to minimize changes to the original test. I followed the example of [clickEvent() and changeEvent()](https://github.com/Shopify/quilt/blob/f96703c5c6d9338586c4de13888f21516b94f833/packages/react-form/src/hooks/list/tests/utils.tsx#L53-L56) used throughout react-forms tests. I agree it seems wonky, but I'm not sure how to avoid it without replacing `Variant[]` values with fake `React.MouseEvent` objects.

packages/react-testing/src/tests/element.test.tsx
peter-hamilton: The generic type isn't applied to the predicate argument, only the return type. Should the implementation of `findWhere` and `findAllWhere` be changed?

BPScott (approved):
Dropped a comment regarding the element test inline.

I'm happy to leave the react-form tests as is for now.

PR #2036, opened 2021-09-13
Fixing an issue with the SSR Apollo link

alexandcote:
## Description
This PR fixed an issue we saw when the online-store-web application migrate to the universal graphql provider.

## Type of change
- [x]  Patch: Bug

## Checklist


packages/react-graphql/src/links.ts
```diff
@@ -46,7 +46,7 @@ export class SsrExtractableLink extends ApolloLink {
           operationDone();
         },
         next: observer.next.bind(observer),
-        error: observer.next.bind(observer),
+        error: observer.error.bind(observer),
```
dahukish, line 49: 🎖️

PR #2037, opened 2021-09-13
Only add the SSR link on the server

alexandcote:
## Description
We don't need the SSR link on the client. We don't need to keep track of those promises on the client. Let's save some memory.

## Type of change

## Checklist


michenly (approved):
LGTM

PR #2039, opened 2021-09-15
Migrate from sewing-kit-next to loom

BPScott:
## Description

`sewing-kit-next` has been renamed to `loom`, and has been moved into the `shopify` npm namespace. Code behaviour in the latest packages from the `@sewing-kit` namespace is identical to the corresponding package in the `@shopify` namespace aside from the binary name, the config file name and the metadata directory.

This PR:

- Replaces `@sewing-kit/core` with `@shopify/loom`, `@sewing-kit/cli` with `@shopify/loom-cli` and `@sewing-kit/plugin-*` with `@shopify/loom-plugin-*`
- Uses the `loom` executable instead of `skn`
- Renames `sewing-kit.config.ts` to `loom.config.ts`
- Ignores `.loom` instead of `.sewing-kit` in ignore files

## Type of change

-All:  patch

Build output should be identical

## Checklist


dahukish (approved):
:shipit:

PR #2040, opened 2021-09-16
Migrate to Apollo 3

### LINKED ISSUE #1677: Apollo 3 support
## Overview

With the advent of [Apollo 3](https://www.apollographql.com/docs/react/migrating/apollo-client-3-migration/), the `ApolloClient`, some breaking changes were introduced, including centralization of imports and typedef changes.
These changes cause apps that have migrated to Apollo 3 and are currently using `graphql-testing` to load additional packages that now are part of `@apollo/client`.

That being said, I wanted to confirm that supporting Apollo 3 is something that's on the roadmap and also offer to create a PR to perform the initial migration if that's the case (and no one's currently working on it).

## Type

- [ ] New feature

## Motivation

Client apps that are consuming the library and have migrated to Apollo 3 will load additional packages on their dependency graph, such as:

- apollo-cache-inmemory
- apollo-client
- apollo-link

One additional example of a breaking change is that the `GraphQL` type (from `graphql-controller.ts`) is using the `ApolloClient` type from `apollo-client`, which doesn't contain the `setLink` method, now present in the Apollo client exported by `@apollo/client`, causing errors when matching types from the library and this library.

ViktorMihov:
## Description

Closes #1677

Add 
- `@apollo/client`

Remove 
- `@apollo/react-common`
- `@apollo/react-hooks`
- `apollo-client`
- `apollo-link`
- `apollo-link-error`
- `apollo-link-http-common`
- `apollo-cache`
- `apollo-cache-inmemory`
- `graphql-tag`
- `react-apollo`

[Apollo Client Changelog](https://github.com/apollographql/apollo-client/blob/d3c7218a0bcd1722f4e0c4636632744564456fb6/CHANGELOG.md)
[Migrating to Apollo Client 3.0
](https://www.apollographql.com/docs/react/migrating/apollo-client-3-migration/)

## Type of change

## Checklist


packages/graphql-testing/src/graphql-controller.ts
```diff
@@ -33,14 +28,18 @@ export class GraphQL {
     mock: GraphQLMock | undefined,
     {unionOrIntersectionTypes = [], cacheOptions = {}}: Options = {},
   ) {
+    const possibleTypes = {};
+
+    unionOrIntersectionTypes.forEach((supertype) => {
```
dahukish, line 33: We could make this a reducer and save the `forEach` and `map` calls.

dahukish, line 33: I wonder if we can add some more type safety. The tools we use to generate the the unions file almost always looks like this: https://github.com/Shopify/flow-app/blob/master/src/graphql/schema/fragmentTypes.json but would be nicer than `any[]` to have `UnionOrIntersection[]` and `PossibleType`.

BPScott:
Going to close this in favor of https://github.com/Shopify/quilt/pull/2302

PR #2045, opened 2021-09-22
[storybook-a11y-test] Set best practices for dealing with failing rules

kaelig:
## Description

At the moment, folks have no other choice but to entirely disable accessibility tests in a story whenever it has a single violation. This is detrimental to accessibility testing as it could hide many issues.

The best practice is to either:

- disable a rule at an element-level,
- set a violation as `reviewOnFail` so that teams can get back to it later

This PR allows for very granular axe configuration settings that work both in Storybook and in this automated testing suite.

### Before

```js
MyStory.parameters = {
  a11y: {
    // Don't do this!
    disable: true,
  }
};
```

### After

```js
MyStory.parameters = {
  a11y: {
    // Instead, use one of these 👇
    // @see axe-core optionsParameter (https://github.com/dequelabs/axe-core/blob/develop/doc/API.md#options-parameter)
    options: {
      rules: [
        {
          // Exclude some elements
          // <explain why here>
          id: 'autocomplete-valid',
          selector: '*:not([autocomplete="nope"])',
        },
        {
          // When there's a false positive, it's okay to disable a specific rule, for example:
          // Color contrast ratio doesn't need to meet 4.5:1, as elements are disabled
          // @link https://dequeuniversity.com/rules/axe/4.1/color-contrast?application=axeAPI
          id: 'color-contrast',
          enabled: false,
        },
        {
          // You may also want to
          // This needs to be fixed in the future
          // <explain why it was set to reviewOnFail>
          id: 'landmark-complementary-is-top-level',
          reviewOnFail: true,
        },
      ],
    },
  },
};
```

## 🎩  instructions

```
dev clone quilt
git fetch
git checkout allow-axe-rule-overrides
dev up
PKG_NAME=storybook-a11y-test yarn tophat
```

In your project (your mileage may vary regarding `yarn` scripts:

```
npx yalc link --no-pure @shopify/storybook-a11y-test
yarn run build-storybook
yarn run storybook:a11y
```

## Type of change

## Checklist

---

Tested in both Shopify/web and Shopify/polaris-react.

Polaris didn't have any errors.

In Shopify/web, these errors popped up, probably due to the version of axe-core being used:

Screen Shot 2021-09-22 at 12 27 47 AM

### DIFF
--- a/packages/storybook-a11y-test/src/index.ts
+++ b/packages/storybook-a11y-test/src/index.ts
@@ -22,13 +22,13 @@ type A11yParametersElement = A11yParameters['element'] | NodeList;
 
 const FORMATTING_SPACER = '    ';
 
-const getBrowser = () => {
+function getBrowser() {
   return puppeteer.launch({
     args: ['--no-sandbox', '--disable-setuid-sandbox'],
   });
-};
+}
 
-export const getStoryIds = async (iframePath: string) => {
+export async function getStoryIds(iframePath: string) {
   const browser = await getBrowser();
   const page = await browser.newPage();
   await page.goto(iframePath);
@@ -48,42 +48,39 @@ export const getStoryIds = async (iframePath: string) => {
   await browser.close();
 
   return storyIds.filter((storyId) => !disabledStoryIds.includes(storyId));
-};
+}
 
-const removeSkippedStories = (skippedStoryIds: StoryId[]) => {
+function removeSkippedStories(skippedStoryIds: StoryId[]) {
   return (selectedStories: StoryId[] = [], storyId = '') => {
     if (skippedStoryIds.every((skipId) => !storyId.includes(skipId))) {
       return [...selectedStories, storyId];
     }
     return selectedStories;
   };
-};
+}
 
-export const getCurrentStoryIds = async ({
+export async function getCurrentStoryIds({
   iframePath,
   skippedStoryIds = [],
 }: {
   iframePath: string;
   skippedStoryIds: StoryId[];
-}) => {
+}) {
   const stories =
     process.argv[2] == null
       ? await getStoryIds(iframePath)
       : process.argv[2].split('|');
 
   return stories.reduce(removeSkippedStories(skippedStoryIds), []);
-};
+}
 
-const formatFailureNodes = (nodes: axeCore.NodeResult[]) => {
+function formatFailureNodes(nodes: axeCore.NodeResult[]) {
   return `${FORMATTING_SPACER}${nodes
     .map((node) => node.html)
     .join(`\n${FORMATTING_SPACER}`)}`;
-};
+}
 
-const formatMessage = (
-  id: axeCore.Result['id'],
-  violations: axeCore.Result[],
-) => {
+function formatMessage(id: axeCore.Result['id'], violations: axeCore.Result[]) {
   return violations
     .map((fail) => {
       return `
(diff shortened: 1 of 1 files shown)


packages/storybook-a11y-test/src/index.ts
```diff
@@ -139,7 +168,7 @@ export const testPages = async ({
   iframePath,
   storyIds = [],
   concurrentCount = os.cpus().length,
-  timeout = 3000,
+  timeout = 30000,
```
BPScott, line 171: Having a 30 second timeout per page means this is going to run forever when there's some weird bug that stops it from resolving. Perhaps a bump to 5000 if 3000 isn't enough?

packages/storybook-a11y-test/src/index.ts
```diff
@@ -115,22 +123,43 @@ const testPage = (iframePath, browser, timeout, disableAnimation) => {
...
+        const axeOptions: axeCore.RunOptions =
+          (story.parameters.a11y && story.parameters.a11y.options) || {};
+
+        window.axe.configure(axeConfig);
+
+        return window.axe
+          .run(axeElement || document.getElementById('root')!, axeOptions)
+          .then((results) => {
+            return results;
+          })
```
BPScott, line 147: This then block seems to return exactly what was passed to it so it doesn't seem to do anything and thus can be removed


```suggestion
```

packages/storybook-a11y-test/src/index.ts
kaelig: This was a fix! Initially this was set to match Puppeteer's default, but a zero was missing. I agree 30s is a lot, tho.

packages/storybook-a11y-test/src/index.ts
```diff
@@ -115,22 +123,43 @@ const testPage = (iframePath, browser, timeout, disableAnimation) => {
...
       }
-      const result = await page.evaluate(() =>
-        window.axe.run(document.getElementById('root'), {}),
-      );
+      const result: axeCore.AxeResults = await page.evaluate((id) => {
+        const story = window.__STORYBOOK_STORY_STORE__.fromId(id)!;
+
+        // optional selector which element to inspect
+        const axeElement: axeCore.ElementContext | undefined =
+          story.parameters.a11y && story.parameters.a11y.element;
```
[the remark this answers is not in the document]
kaelig: This part of the code is definitely a good one to refactor, thanks for the pointers.

What I ended up doing is aligning with what addon-a11y's runner does:

https://github.com/storybookjs/storybook/blob/78c580eac4c91231b5966116492e34de0a0df66f/addons/a11y/src/a11yRunner.ts

Let me know what you think!

---

You'll see this:

```ts
import type {A11yParameters} from '@storybook/addon-a11y/dist/ts3.9/params';
```

Ideally this type would be easier to import, so I opened a PR in Storybook, seeking a more elegant solution: https://github.com/storybookjs/storybook/pull/16128

packages/storybook-a11y-test/README.md
```diff
@@ -2,65 +2,153 @@
...
     console.log(results.join('\n'));
     process.exit(1);
   } else {
-    console.log('🧚‍♀️ Accessibility is all g');
+    console.log('🧚  Accessibility tests passed');
   }
 })();
 ```
 
+### Ignoring specific violations in a Story
```
dfmcphee, line 88: Could we add something here about only doing this if there are false positives, not to get around an existing accessibility problem being flagged?

packages/storybook-a11y-test/src/index.ts
dfmcphee: Same thought here. 30s seems like a lot, especially because we run these in parallel.

dfmcphee (approved):
This looks good to me now other than Ben's comment about converting those arrow functions / variables into regular functions.

Do we have a plan for getting teams to re-enable these in web after this change goes in?

BPScott (approved):
No ran but logic seems good to me

kaelig:
>Do we have a plan for getting teams to re-enable these in web after this change goes in?

The first step is to replace `disable: true` with a `reviewOnFail: true` on each failing rule (see example below):

```ts
PrototypeComponent.parameters = {
  a11y: {
    options: {
      rules: [
        {
          // Page-level semantics cause a violation and need to be reworked.
          // Currently discussing solutions with the accessibility team.
          //
          // @link https://github.com/Shopify/shopify/issues/123
          // @link https://dequeuniversity.com/rules/axe/4.3/landmark-complementary-is-top-level
          id: 'landmark-complementary-is-top-level',
          reviewOnFail: true,
        },
      ],
    },
  },
};
```

Once I'm done, we'll have a better idea of the scale of the problem and we can look into next steps (as in: how to nudge team toward fixing these issues).

PR #2047, opened 2021-09-27
[storybook-a11y-test] Update docs with with config parameters

m-ocana:
## Description
It updates the docs after the last update. As discussed in https://github.com/Shopify/web/pull/50146, the current docs use the [config.rules format](https://github.com/dequelabs/axe-core/blob/develop/doc/API.md#parameters-1) instead of the [options.rules format](https://github.com/dequelabs/axe-core/blob/develop/doc/API.md#options-parameter). However the rules as referenced as **options.rules**

This PR just replaces the a11y **options** paremeters with **config**.

## Type of change
It updates the usage examples in the README file

## Checklist

### DIFF
--- a/packages/storybook-a11y-test/README.md
+++ b/packages/storybook-a11y-test/README.md
@@ -104,8 +104,8 @@ MyStory.parameters = {
     // 🙌 Instead, override single rules on specific elements.
     // 👇 see guidelines below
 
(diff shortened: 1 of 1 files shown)


nickpresta:
Thanks @m-ocana 

For the @Shopify/web-foundations folks, I don't think this requires a version bump since it's just a documentation change. Can you confirm? Thank you

nickpresta (approved):

PR #2052, opened 2021-10-04
Import named ApolloClient

alexandcote:
## Description
During my exploration with SWC, I discovered that we should use the named export instead of the default import for ApolloClient.

## Type of change
- [x]  Patch: Bug (non-breaking change which fixes an issue)expected)

## Checklist


TheMallen (approved):
Seems legit (assuming CI greens)

PR #2054, opened 2021-10-04
[react-i18n] add feature to provide clock format of locale

### LINKED ISSUE #2053: [react-i18n] add function to detect clock format (12/24 hour) of locale
## Overview

It can be useful to know whether or not the given locale uses a 24 hour or 12 hour clock format.

Currently, time pickers in web/online-store don't have a need for this as they simply display time options in the correct format by formatting the input, e.g. `date.toLocaleTimeString(i18n.locale, TIME_FORMATTER)`.

If a design calls for a separate meridiem (AM/PM) picker, then this kind of feature will make it easy to know whether or not it's necessary to show a meridiem picker.

## Type

- [x] New feature
- [ ] Changes to existing features

## Motivation

> What inspired this feature request? What problems were you facing,
> or what else makes you think this should be in `quilt`?

This feature could be used in this timepicker component:

image

https://github.com/Shopify/online-store-web/blob/master/app/sections/Themes/ThemeIndex/components/SchedulingModal/hooks/useClockFormat.ts

braheem:
## Description

Fixes #2053 

Adds method that returns the clock format of the given locale. FEDs can leverage this for any use cases where a meridiem (AM/PM) timepicker is needed.

## Type of change

- [x] react-i18n Minor: New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/packages/react-i18n/src/i18n.ts
+++ b/packages/react-i18n/src/i18n.ts
@@ -377,6 +377,19 @@ export class I18n {
     return Boolean(easternNameOrderFormatter);
   }
 
+  hasTwelveHourClockFormat(locale?: string) {
+    const afternoonDateTime = new Date(2020, 1, 1, 13);
+    try {
+      const dateParts = new Intl.DateTimeFormat(locale || this.locale, {
+        hour: 'numeric',
+      }).formatToParts(afternoonDateTime);
+      const hourValue = dateParts.find((part) => part.type === 'hour')?.value;
+      return Number(hourValue) < 13;
+    } catch {
+      return false;
+    }
+  }
+
   @memoize()
   numberSymbols() {
     const formattedNumber = this.formatNumber(123456.7, {
--- a/packages/react-i18n/README.md
+++ b/packages/react-i18n/README.md
@@ -148,6 +148,9 @@ The provided `i18n` object exposes many useful methods for internationalizing yo
 - `ordinal()`: formats a number as an ordinal according to the locale, e.g. `1st`, `2nd`, `3rd`, `4th`
 - `hasEasternNameOrderFormatter()`: returns true when an eastern name order formatter corresponding to the locale/language exists.
 - `numberSymbols()`: returns an object specifying the current locale's decimal and thousand symbols. Example: For the `es-ES` locale the output would be `{ decimalSymbol: ',', thousandSymbol: '.' }`
+- `hasTwelveHourClockFormat()`: Returns true if the locale uses 12-hour clock format. Uses the current locale if no locale is passed.
+  - `hasTwelveHourClockFormat('en-US')` will return true
+  - `hasTwelveHourClockFormat('fr-CA')` will return false
 
 Most notably, you will frequently use `i18n`’s `translate()` method. This method looks up a key in translation files that you supply based on the provided locale. This method is discussed in detail in the next section.
 
--- a/packages/react-i18n/src/tests/i18n.test.ts
+++ b/packages/react-i18n/src/tests/i18n.test.ts
@@ -2115,4 +2115,46 @@ describe('I18n', () => {
       expect(i18n.hasEasternNameOrderFormatter()).toStrictEqual(false);
     });
   });
+
+  describe('#hasTwelveHourClockFormat', () => {
+    it('returns true for en-CA locale', () => {
+      const i18n = new I18n(defaultTranslations, {locale: 'en-CA'});
+
+      expect(i18n.hasTwelveHourClockFormat()).toStrictEqual(true);
+    });
+
+    it('returns true for en-US locale', () => {
+      const i18n = new I18n(defaultTranslations, {locale: 'en-US'});
+
+      expect(i18n.hasTwelveHourClockFormat()).toStrictEqual(true);
+    });
+
+    it('returns true for en-AU locale', () => {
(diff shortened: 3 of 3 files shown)


braheem:
I understand the use case for this might be very limited so i'm not sure if it makes sense to add it. Let me know if you're okay with adding this so I can update the changelog.

atesgoral:
Looks like a useful feature to include!

packages/react-i18n/src/i18n.ts
```diff
@@ -377,6 +377,19 @@ export class I18n {
     return Boolean(easternNameOrderFormatter);
   }
 
+  hasTwelveHourClockFormat(locale?: string) {
+    const afternoonDateTime = new Date(2020, 1, 1, 13);
+    try {
```
BPScott, line 382: Under what situation will this throw? Is there a way to detect that with if/else conditionals rather than a try/catch?

braheem: It will throw when an invalid locale is passed to `Intl.DateTimeFormat()`, like empty string or gibberish. 

On a related note, I noticed in the tests that there are test cases where locale is passed as 'unknown'. I tried using 'unknown' with the expectation that it would throw and default to 24 hour format but Intl parses 'unknown' as 'en-US'. 

I could possibly switch to if/else by validating the locale before I pass it to `Intl.DateTimeFormat`, but I figured a try/catch was the easier and more robust solution. Happy to change it if you have a preference and know a good way to validate a locale string.

BPScott: Thanks for the context.

So this function:
- returns true if the locale uses a 12h clock
- returns false if the locale uses a 24h clock or the locale is gibberish.

I think the throwing is valid so that we can determine the difference a garbage locale and one with a 24h clock. It looks like that is what occurs if you feed a garbage locale into `formatDate()`

BPScott (commented):
Reading https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/hourCycle I note that there's actually 4 types of hour cycle  and that some locales change how they want to count present hours - is midnight 00:00 or 24:00 in a 24h system, or 00:00am or 12:00 am in a 12 hour system? Is this a concern for how you lay out your UIs? It seems that if you want to fully adhere to the locale information then creating a an additional UI component for am/pm adds a lot of development burden around handling that box in edge-cases. Having a hours list that goes `12, 1, 2, 3` with a separate box for am/pm for a "12h - midnight is 12:00am" seems very confusing.

---

I note that react-i18n doesn't contain any other date formatting / date information stuff - it all lives in `@shopify/dates` and [is reexported in react-i18n](https://github.com/Shopify/quilt/blob/f461292bf45e22c89ab348f4a62134286bd38ce7/packages/react-i18n/src/i18n.ts#L1-L14)

I wonder if the meat of this code should live in [`getDateTimeParts` in `@shopify/dates`](https://github.com/Shopify/quilt/blob/main/packages/dates/src/get-date-time-parts.ts#L8) and then this function becomes `return getDateTimeParts(new Date(2020, 1, 1, 13)).getHourParts() < 13` which has the bonus of caching the information about the locale as in the past folks have found that creating new Intl instances is pretty expensive.

braheem:
> Reading https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/hourCycle I note that there's actually 4 types of hour cycle and that some locales change how they want to count present hours - is midnight 00:00 or 24:00 in a 24h system, or 00:00am or 12:00 am in a 12 hour system? Is this a concern for how you lay out your UIs? It seems that if you want to fully adhere to the locale information then creating a an additional UI component for am/pm adds a lot of development burden around handling that box in edge-cases. Having a hours list that goes `12, 1, 2, 3` with a separate box for am/pm for a "12h - midnight is 12:00am" seems very confusing.
> 
> I note that react-i18n doesn't contain any other date formatting / date information stuff - it all lives in `@shopify/dates` and [is reexported in react-i18n](https://github.com/Shopify/quilt/blob/f461292bf45e22c89ab348f4a62134286bd38ce7/packages/react-i18n/src/i18n.ts#L1-L14)
> 

Good point about hour cycle, I wasn't aware of those differences. I think if the user understands the hourCycle for their locale, then there shouldn't be any confusion around the time and meridiem picker if the time options are based on the locale. For something like 'h11' for example, 0:00 pm would be noon, 0:00 am would be midnight. 

It does add some development burden in that the options for time wouldn't be hardcoded. We could use the same code currently used for admin timepicker options: make a list of dates and use `date.toLocaleTimeString(i18n.locale, TIME_FORMATTER)`. But at that point maybe it makes more sense show AM/PM in the timepicker option as well (just like admin currently does)? It does seem like the need for this feature is based on a UX problem more than anything else. I can't think of another use case for this 🤔 

> I wonder if the meat of this code should live in [`getDateTimeParts` in `@shopify/dates`](https://github.com/Shopify/quilt/blob/main/packages/dates/src/get-date-time-parts.ts#L8) and then this function becomes `return getDateTimeParts(new Date(2020, 1, 1, 13)).getHourParts() < 13` which has the bonus of caching the information about the locale as in the past folks have found that creating new Intl instances is pretty expensive.

That would be great but i'm not very familiar with that code, would this work? Doesn't look like I can pass the locale to that function.

mateus:

PR #2013, opened 2021-08-26
Update quilt packages to support webpack 5

vsumner:
## Description
This PR updates packages with `webpack` support to version 5.

Fixes (issue #)

## Type of change

- [ ] `@shopify/jest-dom-mocks` Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] `graphql-mini-transforms` Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] `@shopify/react-server` Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] `@shopify/sewing-kit-koa` Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] `@shopify/web-worker` Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

- [x] I have added a changelog entry, prefixed by the type of change noted above (Documentation fix and Test update does not need a changelog as we do not publish new version)

PR #2056, opened 2021-10-18
Shuffle web-worker and RPC packages between Quilt and remote-ui

lemonmade:
## Description

Fixes part of https://github.com/Shopify/remote-ui/issues/96

This PR shuffles around some of our web-worker packages between remote-ui and quilt. Per that issue, the plan is that these monorepos have the following packages after these changes are complete:

- `quilt` has `@shopify/web-worker`. This package is an opinionated way of implementing web workers that relies on a specific set of technologies that we use at Shopify (webpack and babel). This library depends on the next library for remote procedure calls (RPC) between the main page and a web worker.
- `remote-ui` has `@remote-ui/rpc`, the RPC library used by `@shopify/web-worker` and the other `@remote-ui` libraries. The RPC library is part of remote-ui’s implementation — all the other libraries assume it is being used for communication, and use that assumption to perform smart memory management.

## Type of change

I will do a major bump for both `@shopify/web-worker` and `@shopify/rpc`. `@shopify/rpc` is being deprecated in this PR, so I've removed all the source code for it, and the next major version will be the last for that package. `@shopify/web-worker` is moving to a mostly-compatible API and probably doesn't strictly need a major bump, but for safety and simplicity in consumers I will use a major version to more clearly draw the line between the old and new implementations.

## Checklist


packages/web-worker/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## Unreleased -->
```
BPScott, line 8:
```suggestion
## Unreleased
```

packages/web-worker/README.md
```diff
@@ -14,25 +14,25 @@ $ yarn add @shopify/web-worker
...
 ## Usage
 
-This library contains three parts that must be used in tandem:
+This library contains three parts that must be used together:
 
 1. The public API of the package provided by `@shopify/web-worker`
-1. A webpack plugin provided by `@shopify/web-worker/webpack`
-1. A babel plugin provided by `@shopify/web-worker/babel`
+1. A babel plugin provided by `@shopify/remote-ui/babel` that identifies uses of `createWorkerFactory` that need to be processed
+1. A webpack plugin provided by `@shopify/remote-ui/webpack` that outputs the worker as a dedicated chunk
```
BPScott, line 21: On first glance of this diff it sounded like `@shopify/web-worker/{babel,webpack}` were removed but they're still present and referenced several times in this readme. Should this have been kept the same?

packages/web-worker/README.md
```diff
@@ -83,12 +83,63 @@ const worker = createWorker();
...
+By default, this library will create a worker by calling `new Worker` with a blob URL for the worker script. This is generally all you need, but some use cases may want to construct the worker differently. For example, you might want to construct a worker in a sandboxed iframe to ensure the worker is not treated as same-origin, or create a worker farm instead of a worker per script. To do so, you can supply the `createMessenger` option to the function provided by `createWorkerFactory`. This option should be a function that accepts a `URL` object for the location of the worker script, and return a `MessageEndpoint` compatible with being passed to [`@remote-ui/rpc`’s `createEndpoint`](https://github.com/Shopify/quilt/tree/master/packages/rpc#usage) API.
```
BPScott, line 139:
```suggestion
By default, this library will create a worker by calling `new Worker` with a blob URL for the worker script. This is generally all you need, but some use cases may want to construct the worker differently. For example, you might want to construct a worker in a sandboxed iframe to ensure the worker is not treated as same-origin, or create a worker farm instead of a worker per script. To do so, you can supply the `createMessenger` option to the function provided by `createWorkerFactory`. This option should be a function that accepts a `URL` object for the location of the worker script, and return a `MessageEndpoint` compatible with being passed to [`@remote-ui/rpc`’s `createEndpoint`](https://github.com/Shopify/quilt/tree/main/packages/rpc#usage) API.
```

BPScott (commented):
Is this for webpack 4 / 5 / both?

There was a bunch of webpack 5 support stuff in https://github.com/Shopify/quilt/pull/2013 and I'm not sure how this would interact with those changes.

I understand none of this but I trust you.
Fixup the test fails - looks like there's some type diferences and I'm happy to give this a 👍 .

lemonmade:
@dahukish I finally got the tests passing here (Jest and webpack conspired together to make it incredibly hard to debug some leftover `@remote-ui/web-workers` that I had :P). Ben's on vacation now, so was wondering if you would be able to take another look at this?

atesgoral:
CI was stuck for some reason. Rebased against main and updated yarn.lock. 🤞

packages/web-worker/loom.config.ts
```diff
@@ -7,7 +7,41 @@ export default createPackage((pkg) => {
   pkg.entry({root: './src/index'});
   pkg.entry({name: 'babel', root: './src/babel-plugin'});
   pkg.entry({name: 'webpack', root: './src/webpack-parts'});
-  pkg.entry({name: 'webpack-loader', root: './src/webpack-parts/loader'});
   pkg.entry({name: 'worker', root: './src/worker'});
   pkg.use(quiltPackage());
+
+  // The library references the worker "wrapper" files as relative paths from
+  // the source code. Because they aren’t in the dependency graph, Rollup
+  // does not bundle them. This sewing-kit plugin adds a copy rollup plugin
```
atesgoral, line 15:
```suggestion
  // does not bundle them. This Loom plugin adds a copy rollup plugin
```

atesgoral (commented):
LGTM! :)

PR #2064, opened 2021-11-01
Always require statsd-instrument

bradsokol:
## Description

quilt_rails makes use of the statsd-instrument gem but does not explicitly require it. If statsd-instrument has not been loaded by the app or another of the app's gems, a `NameError` will be raised:

```
NameError (uninitialized constant Quilt::Performance::Client::StatsD):
  
quilt_rails (3.5.1) lib/quilt_rails/performance/client.rb:30:in `distribution'
quilt_rails (3.5.1) lib/quilt_rails/performance/client.rb:57:in `block in report_events'
quilt_rails (3.5.1) lib/quilt_rails/performance/client.rb:53:in `each'
quilt_rails (3.5.1) lib/quilt_rails/performance/client.rb:53:in `report_events'
quilt_rails (3.5.1) lib/quilt_rails/performance/client.rb:48:in `process_report!'
...
```

I suspect that this happens infrequently since this gem is often used with others that do load stats-instrument, such as cusco or shopify_metrics. Tests pass because statsd-instrument is explicitly loaded in `Quilt::PerformanceReportableTest`.

The fix was to move `require "statsd-instrument"` from `Quilt::PerformanceReportableTest` to `quilt_rails.rb` where the other dependencies are `require`'d.

## Type of change

- [x] quilt_rails Patch: Bug (non-breaking change which fixes an issue)

## Checklist


gems/quilt_rails/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### Changed
+
+- Always `require "statsd-instrument"` so that apps don't need to do so.
```
atesgoral, line 12:
```suggestion
- Always `require "statsd-instrument"` so that apps don't need to do so. ([#2064](https://github.com/Shopify/quilt/pull/2064))
```

atesgoral (approved):

PR #2065, opened 2021-11-02
Run on latest 12.x

atesgoral:
## Description

Tweak CI matrix to run on 12.x instead of 12.14 (Don't know the historical context of being on 12.14)

## Type of change

Just a CI environment tweak. No change to any code.

### DIFF
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: ['12.14', '14.x', '16.x']
+        node-version: ['12.x', '14.x', '16.x']
     timeout-minutes: 15
 
     steps:


michenly:
Wondering if this should sync to the v12 version or not https://github.com/Shopify/quilt/blob/548cb40c25e340a64cbc900532a9831d62cc1780/config/loom/index.ts#L31 (maybe this is why we test for 12.14)

BPScott:
> Don't know the historical context of being on 12.14

We currently say 12.14.0 is our minimum supported version. With '12.x'  CI will run on the latest v12 version (12.22.7 according to [this](https://github.com/Shopify/quilt/runs/4081255629?check_suite_focus=true#step:3:8))

Consider a world where node added some APIs between the minimum stated supported version (12.14) and the minimum version we run CI against (12.22). It we were to use those APIs then things would blow up for consumers using 12.14 but we wouldn't know because we never run our tests against the minimum version that we claim support.

Strictly speaking, If we were to require a specific minor version in v14/v16  then the same would go for v14/v16 releases too (e.g. in the loom repo, packages have an engines field of `"^12.22.0 || ^14.17.0 || >=16.0.0"` so we should test on v`14.17.0`, rather than `14.x`). However in quilt's case this doesn't apply for as the quilt packages have engines fields of `>=12.14.0`. So for quilt I think it's fine for the test matrix to be `['12.14', '14.x', '16.x']`.


I think it is useful for CI to run against the minimum versions that we claim to support rather the latest release in a major version so that we can ensure that we don't accidentally use a feature that is not present in old-but-supported node versions and thus break consumers that are still using that version. This is the same rational for configuring dev.yml to use the oldest supported node version (in quilt's case 12.14) rather than the latest version - except now can no longer do that thanks to local dev requiring node 16 thanks to M1 chips.

PR #2066, opened 2021-11-02
Updated .nvmrc to use v16

ViktorMihov:
## Description

Updated `.nvmrc` to align node version with `dev.yml`


michenly:
Thank you!

PR #2077, opened 2021-11-11
Include offsets in humanized timezones

qt314:
## Description
Closes https://github.com/Shopify/web/issues/52384

Old regex only included word characters (excluded `+`/`-`) which caused offset timezones (i.e. `GMT+1`) to not be matched. These timezones are used in from `Intl.DateTimeFormat` when there is no timezone abbreviation in that language.

## Tophat instructions
note: there are tests for this in the PR 

1. https://regex101.com/
2. use the pattern from below `\s([\w()+\-:.]+$)`
3. make sure it selects the timezone for these cases:
- `6:00pm EST`
- `10:00am GMT+1`
- `4:30pm UTC-2:30`
- `4:30pm UTC+1.5`

## Type of change

## Checklist

### DIFF
--- a/packages/react-i18n/src/i18n.ts
+++ b/packages/react-i18n/src/i18n.ts
@@ -582,7 +582,7 @@ export class I18n {
       hour: 'numeric',
     });
 
-    const zoneMatchGroup = /\s(\w*$)/i.exec(hourZone);
+    const zoneMatchGroup = /\s([\w ()+-:]*$)/i.exec(hourZone);
 
     return zoneMatchGroup ? zoneMatchGroup[1] : '';
   }
--- a/packages/react-i18n/src/i18n.ts
+++ b/packages/react-i18n/src/i18n.ts
@@ -582,7 +582,7 @@ export class I18n {
       hour: 'numeric',
     });
 
-    const zoneMatchGroup = /\s([\w(+\-:.)]+$)/.exec(hourZone);
+    const zoneMatchGroup = /\s([\w()+\-:.]+$)/.exec(hourZone);
 
     return zoneMatchGroup ? zoneMatchGroup[1] : '';
   }
--- a/packages/react-i18n/src/i18n.ts
+++ b/packages/react-i18n/src/i18n.ts
@@ -582,7 +582,7 @@ export class I18n {
       hour: 'numeric',
     });
 
-    const zoneMatchGroup = /\s([\w ()+-:]*$)/i.exec(hourZone);
+    const zoneMatchGroup = /\s([\w()+-:]*$)/i.exec(hourZone);
 
     return zoneMatchGroup ? zoneMatchGroup[1] : '';
   }
--- a/packages/react-i18n/src/i18n.ts
+++ b/packages/react-i18n/src/i18n.ts
@@ -582,7 +582,7 @@ export class I18n {
       hour: 'numeric',
     });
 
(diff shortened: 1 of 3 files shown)


packages/react-i18n/src/tests/i18n.test.ts
```diff
@@ -1428,6 +1428,33 @@ describe('I18n', () => {
...
+        i18n.formatDate(date, {
+          timeZoneName: 'short',
+          style: DateStyle.Humanize,
+        });
+
+        expect(translate).toHaveBeenCalledWith(
+          'date.humanize.lessThanOneWeekAgo',
+          {
+            pseudotranslate: false,
+            replacements: {weekday: 'Tuesday', time: '6:00 a.m. GMT+1'},
```
qt314, line 1451: before this would just say `6:00 am` since the `GMT+1` wasn't matched because of the `+`

packages/react-i18n/src/i18n.ts
```diff
@@ -582,7 +582,7 @@ export class I18n {
       hour: 'numeric',
     });
 
-    const zoneMatchGroup = /\s(\w*$)/i.exec(hourZone);
+    const zoneMatchGroup = /\s([\w ()+-:]*$)/i.exec(hourZone);
```
dahukish, line 585: will there always be a space?

qt314: nope, im gonna remove it

packages/react-i18n/src/i18n.ts
```diff
@@ -582,7 +582,7 @@ export class I18n {
       hour: 'numeric',
     });
 
-    const zoneMatchGroup = /\s(\w*$)/i.exec(hourZone);
+    const zoneMatchGroup = /\s([\w()+-:]*$)/i.exec(hourZone);
```
atesgoral, line 585: [GMT+9.5](http://www.timebie.com/timelocal/gmt+95.php) uses a decimal point

atesgoral: (We may need to include `.` in the set)

atesgoral: Also, we should probably use `+` instead of `*`. The result will be the same: there will be a match but the matched group will yield a `''` if there's nothing following the `\s`. But using a `+` would make this more intuitive.

qt314: It still matched `1.5` i guess since its still a number but I added into the set just in case :)

qt314: Im gonna have to figure out how to not make these a range of chars

packages/react-i18n/src/i18n.ts
```diff
@@ -582,7 +582,7 @@ export class I18n {
       hour: 'numeric',
     });
 
-    const zoneMatchGroup = /\s(\w*$)/i.exec(hourZone);
+    const zoneMatchGroup = /\s([\w(+\-:.)]*$)/.exec(hourZone);
```
atesgoral, line 585: Also please change `*` to `+`

packages/react-i18n/src/i18n.ts
qt314: i made it `/\s([\w(+\-:.)]+$)/

dahukish (approved):
:shipit:

royleejr:
@qt314 These changed have now been implemented in the new quilt packages release! @shopify/react-i18n@6.2.4

PR #2078, opened 2021-11-12
[useful-types]: adding DeepRequired and DeepNonNullable

weslleyaraujo:
## Description

Fixes (issue #)

## Type of change

## Checklist


packages/useful-types/README.md
```diff
@@ -109,18 +109,18 @@ The following type aliases are provided by this library:
 
 - `DeepOmit<T, K>` Recursively maps over all properties in a type and omits those matching `K`.
 
-```ts
-interface Obj {
-  __typename: string;
-  foo: string;
-  bar: {
+  ```ts
```
weslleyaraujo, line 112: I've also indented this block as it seems broken: https://github.com/Shopify/quilt/tree/main/packages/useful-types

PR #2079, opened 2021-11-15
Escape Unicode characters in JSON

alexandcote:
Related to: https://github.com/Shopify/shoppingmall/issues/313
May be related to: https://github.com/Shopify/banking/issues/6049

## Description
Fixing an issue when data contains non-ASCII characters. HTTP Header doesn't seem to support UTF-8 and seems to only support `ISO-8859-1`. For that reason, we are forcing the JSON string to fully escape all Unicode characters (the JSON spec support this, but doesn't mandate it) so if the JSON string contains any non-ASCII character, the node sidecar will receive the right value.

## Type of change

## Checklist

### DIFF
--- a/gems/quilt_rails/test/quilt/react_renderable_test.rb
+++ b/gems/quilt_rails/test/quilt/react_renderable_test.rb
@@ -11,7 +11,7 @@ module Quilt
         render_react,
         reverse_proxy(
           url,
-          headers: { 'X-Request-ID': request.request_id, 'X-Quilt-Data': {}.to_json }
+          headers: { 'X-Request-ID': request.request_id, 'X-Quilt-Data': '{}' }
         )
       )
     end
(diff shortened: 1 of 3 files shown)


gems/quilt_rails/test/quilt/react_renderable_test.rb
```diff
@@ -24,7 +24,7 @@ def test_render_react_calls_with_custom_headers
       headers = {
         'x-custom-header': 'test',
         'X-Request-ID': request.request_id,
-        'X-Quilt-Data': {}.to_json,
+        'X-Quilt-Data': {}.to_json.encode("ISO-8859-1"),
```
atesgoral, line 27: Don't need it for `{}`

atesgoral:
By directly using the JSON library, we can force the proper escaping of all Unicode characters:

```
o = {name: "Ateş Göral"}
JSON.generate(o, ascii_only: true)
=> "{\"name\":\"Ate\\u015f G\\u00f6ral\"}"
```

atesgoral (approved):

PR #2081, opened 2021-11-22
Add import specifier prevent duplicates

cartogram:
## Description

This PR prevents duplicate specifiers from being added when using the `addImportSpecifier` transform. If duplicates end up in the source, they cause a syntax error for transforms down the line.

## Type of change

## Checklist

### DIFF
--- a/packages/ast-utilities/src/javascript/addImportStatement/addImportStatement.ts
+++ b/packages/ast-utilities/src/javascript/addImportStatement/addImportStatement.ts
@@ -19,5 +19,19 @@ export default function addImportStatement(statement: string | string[]) {
         path.node.body.unshift(ast);
       });
     },
+    File(path: traverse.NodePath<t.Program>) {
+      if (path.node.body.length === 0) {
+        path.node.body = [...path.node.body, ...addImports(statements)];
+      }
+    },
   };
+}
+
+function addImports(statements: string[]) {
+  return statements.map((statement) => {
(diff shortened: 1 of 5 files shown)


cartogram:
@Shopify/web-foundations do you know what the ruby-tests failures in CI are about?

atesgoral:
@cartogram The build issue on main is fixed.

cartogram:
@atesgoral Thanks for following up!

Would I also be able to get a review 🍏 from you or someone on the web-foundations team?

packages/ast-utilities/src/javascript/addImportStatement/addImportStatement.ts
```diff
@@ -19,5 +19,19 @@ export default function addImportStatement(statement: string | string[]) {
         path.node.body.unshift(ast);
       });
     },
+    File(path: traverse.NodePath<t.Program>) {
```
atesgoral, line 22: Could we add a test for this case?

cartogram: Good catch! I can't with my current test utilities because the `File` node is not part of the fixture (they start at `Program` because they are coming from strings not files). I tried to get around this without any luck.

I suggest we merge without a test for this and I'll come back to it. These utilities are already well tested given how experimental they are, and this particular logic exists to catch the rare edge case that they run this transform on an empty file.

What do you think?

PR #2083, opened 2021-11-22
Upgrade Ruby, update instructions

atesgoral:
## Description

Updating to latest Ruby.

## Type of change

Environment change.

## Checklist


gems/quilt_rails/dev.yml
```diff
@@ -1,7 +1,6 @@
-name: react-server
+name: quilt-rails
 up:
-  - ruby: 2.6.3
-  - bundler
```
BPScott, line 4: By taking this away, we no longer install any bundler dependencies when `dev up` is ran, and we no longer ensure that the version of bundler used locally is the same as the version that it was previously. That doesn't see useful to me

atesgoral: I saw web's dev.yml didn't have an explicit `bundler` item, so I removed it. I assumed this would mean to default to whatever bundler version comes with the specified Ruby version, but maybe I'm wrong.

atesgoral: Hmmm. Good question.

Looking at shopify instead, there's the `bundler` entry still. So I'll actually model this after that and add the line back.

atesgoral: Something to do with k8s fixtures: https://github.com/Shopify/web/blob/master/dev.yml#L116-L129

BPScott (approved):

PR #2086, opened 2021-11-23
New transform `wrapJsxChildren()`

cartogram:
## Description

Adds a new transform for wrapping a the children of a React JSX Element with a new React JSX Element. For example, if you wanted to wrap everything under a `<Suspense>` tag with a new `<Provider>` component.

## Type of change

## Checklist

### DIFF
--- a/packages/ast-utilities/CHANGELOG.md
+++ b/packages/ast-utilities/CHANGELOG.md
@@ -7,6 +7,10 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Added
+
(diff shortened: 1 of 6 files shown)


packages/ast-utilities/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
 
+### Added
+
+- `replaceJsxBody()` support root nodes
```
cartogram, line 12: See https://github.com/Shopify/quilt/pull/2085, will do the minor changelog clean up before merging.

PR #2087, opened 2021-11-24
Adjust DeepThunk type for changes in typescript 4.5

peter-hamilton:
## Description

Changes `DeepThunk` type to support new [Restrictions on Assignability to Conditional Types](https://devblogs.microsoft.com/typescript/announcing-typescript-4-5/#restrictions-on-assignability-to-conditional-types) in Typescript 4.5.

These changes address errors being faced updating [shopify/web to 4.5](https://github.com/Shopify/web/pull/53227).

`DeepThunk` was changed to separate `| null | undefined` from the conditional types `a?b:c`. It should be equivalent to the previous definition.

## Type of change

- [x] graphql-fixtures Patch: Bug (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/packages/graphql-fixtures/CHANGELOG.md
+++ b/packages/graphql-fixtures/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


lucabezerra (approved):
LGTM (but I might be biased haha)

dahukish:
Change seems reasonable. Have we tried this in web or shrink-ray either as a yalc package or a alpha?

lucabezerra:
> Change seems reasonable. Have we tried this in web or shrink-ray either as a yalc package or a alpha?

Tried it locally on web via `yarn link` and it seemed to work for the type-checking. Haven't run the any tests, though. Should I?

Issue #1090, opened 2019-10-03
Deprecation warnings for Performance API usage


lemonmade:
When running our performance libraries in web, I see the following warnings:

![Screen Shot 2019-10-03 at 12 22 21 PM](https://user-images.githubusercontent.com/3012583/66145331-9959b300-e5d8-11e9-9637-d0c8c935b6d0.png)

cc/ @TheMallen @michenly maybe something that could be addressed in your sprint if you have time?

michenly:
`The PerformanceObserver does not support buffered flag with the entryTypes argument.`
from https://github.com/Shopify/quilt/blob/c5721a6104343c2b46a13bfd0d6dd50ab5797c6a/packages/performance/src/utilities.ts#L42-L45


`Deprecated API for given entry type.`
from https://github.com/Shopify/quilt/blob/c5721a6104343c2b46a13bfd0d6dd50ab5797c6a/packages/performance/src/utilities.ts#L31

lemonmade:
I logged out the entry type before each of the warnings:

![Screen Shot 2019-10-17 at 10 24 48 AM](https://user-images.githubusercontent.com/3012583/67162902-ea86d800-f336-11e9-9098-6f677f4b1b7a.png)

donnielrt:
It would be awesome if we could get rid of these, they really do clutter up the console!

atesgoral:
The deprecated API warning is easy to repro in the dev console: 

```
performance.getEntriesByType('longtask')
VM179:1 Deprecated API for given entry type.
```

But I still haven't been able to track down any spec around what/why is deprecated and what the course of action is.

PR #2089, opened 2021-11-24
Use type when using buffered: true

atesgoral:
## Description

Part of #1090

In Chrome's DevTools console, we're getting a bunch of these errors:

```
The PerformanceObserver does not support buffered flag with the entryTypes argument.
```

We're using a single type, and wrapping it in an array to pass to `entryTypes` while we have `buffered: true`. We should just use the `type` property.

https://developer.mozilla.org/en-US/docs/Web/API/PerformanceObserver/observe

> - entryTypes: An array of DOMString objects, each specifying one performance entry type to observe. May not be used together with the "type" or "buffered" options.
> - type: A single DOMString specifying exactly one performance entry type to observe. May not be used together with the entryTypes option.
> - buffered: A boolean flag to indicate whether buffered entries should be queued into the observer's buffer. Must be used only with the "type" option.

## Type of change

- [x] `performance` Patch: Bug (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/packages/performance/CHANGELOG.md
+++ b/packages/performance/CHANGELOG.md
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Fixed
+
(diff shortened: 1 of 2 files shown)


BPScott (approved):
Changelog entry then :shipit:

BPScott:
I'm looking forward to the follow up PR that fixes the `Deprecated API for given entry type.` warnings too :D

PR #2090, opened 2021-11-25
Clean up test output in AppBridgeUniversalProvider

vsumner:
## Description
This PR fixes `Error: Not implemented: navigation (except hash changes)` console errors in `AppBridgeUniversalProvider` tests.

Fixes (issue #)

## Type of change

- [X] `@shopify/app-bridge-universal-provider` Patch: Bug (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/packages/react-app-bridge-universal-provider/src/tests/AppBridgeUniversalProvider.test.tsx
+++ b/packages/react-app-bridge-universal-provider/src/tests/AppBridgeUniversalProvider.test.tsx
@@ -4,23 +4,17 @@ import {Provider as AppBridgeProvider} from '@shopify/app-bridge-react';
 import {extract} from '@shopify/react-effect/server';
 import {mount} from '@shopify/react-testing';
 import {HtmlManager, HtmlContext} from '@shopify/react-html';
+import {location} from '@shopify/jest-dom-mocks';
 
 import {AppBridgeUniversalProvider} from '../AppBridgeUniversalProvider';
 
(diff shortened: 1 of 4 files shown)


packages/react-app-bridge-universal-provider/src/tests/AppBridgeUniversalProvider.test.tsx
```diff
@@ -8,6 +8,21 @@ import {HtmlManager, HtmlContext} from '@shopify/react-html';
 import {AppBridgeUniversalProvider} from '../AppBridgeUniversalProvider';
 
 describe('<AppBridgeUniversalProvider />', () => {
+  const assign = window.location.assign;
+
+  beforeAll(() => {
```
BPScott, line 13: Could we use [`@shopify/jest-dom-mocks`'](https://github.com/Shopify/quilt/tree/main/packages/jest-dom-mocks#standard-mocks) location mock? it looks like that  handles stubbing/restoring [assign](https://github.com/Shopify/quilt/blob/main/packages/jest-dom-mocks/src/location.ts#L16)

vsumner: Good catch. Fixed.

BPScott:
Remember to add jest-dom-mocks as a reference in react-app-bridge-provider's tsconfig 
https://github.com/Shopify/quilt/runs/4324225069?check_suite_focus=true#step:8:84

PR #2091, opened 2021-11-25
Clean graphql config utilities test output

vsumner:
## Description
Tests for `graphql-config-utilities` outputs warnings when testing `resolveProjectName`. This PR disables them because they are expected.

Fixes (issue #)

## Type of change

- [X] `graphql-config-utilities` Patch: Bug (non-breaking change which fixes an issue)

## Checklist


BPScott:
Do we need an afterEach to restore the original implementation?

PR #2093, opened 2021-11-25
Get all entries and filter them ourselves

### LINKED ISSUE #1090: Deprecation warnings for Performance API usage

atesgoral:
## Description

Closes #1090

Get rid of a deprecation warning that gets logged by Chrome when certain types like `longtask` are used with `window.performance.getEntriesByType()` by using `window.performance.getEntries()` and doing the filtering ourselves.

## Type of change

- [x] `performance` Patch: Bug (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/packages/performance/CHANGELOG.md
+++ b/packages/performance/CHANGELOG.md
@@ -11,6 +11,7 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


BPScott (approved):
chagnelog entry and :shipit:

PR #2096, opened 2021-11-30
Deprecate @shopify/magic-entries-webpack-plugin

vsumner:
## Description
This PR deprecates `@shopify/magic-entries-webpack-plugin`. This plugin is not used by node and is not compatible with webpack 5. We have made the choice to deprecate instead of continuing to support this library.
Fixes (issue #)

## Type of change

- [X] `@shopify/magic-entries-webpack-plugin` Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist


dahukish (approved):
🔥

PR #2100, opened 2021-12-03
Adding adding a way to fill a fragment

alexandcote:
Resolve: https://github.com/Shopify/online-store-web/issues/5949

## Description
The `createFiller` function now return an object containing `fillOperation` and `fillFragment`. It's now possible to simply pass a fragment to `fillFragment` without a query. `fillOperation` has the same behaviour and didn't change. We only change the way we expose it.

**Worth mentioning I have 0 contexts on that and I simply copy and adapt the code [already in our repo.](https://github.com/Shopify/online-store-web/blob/master/packages/%40shopify/graphql-fixtures/src/fill.ts#L153-L163)**

## Type of change

## Checklist

### DIFF
--- a/packages/graphql-fixtures/src/fill.ts
+++ b/packages/graphql-fixtures/src/fill.ts
@@ -37,56 +37,32 @@ export interface ResolveDetails {
   parentFields: FieldDetails[];
 }
 
-export interface Resolver<
-  T = any,
-  Data = {},
-  Variables = {},
-  DeepPartial = {},
-> {
-  (
-    request: GraphQLRequest<Data, Variables, DeepPartial>,
-    details: ResolveDetails,
-  ): T;
+export interface Resolver<Request, T = any> {
+  (request: Request, details: ResolveDetails): T;
 }
 
-export type Thunk<T, Data, Variables, DeepPartial> =
-  | T
-  | Resolver<T, Data, Variables, DeepPartial>;
+export type Thunk<Request, T> = T | Resolver<Request, T>;
 
-export type DeepThunk<T, Data, Variables, DeepPartial> =
+export type DeepThunk<Request, T> =
   | (T extends object
       ? {
           [P in keyof T]: Thunk<
+            Request,
             | (T[P] extends (infer U)[]
-                ? Thunk<
-                    DeepThunk<U, Data, Variables, DeepPartial>,
-                    Data,
-                    Variables,
-                    DeepPartial
-                  >[]
+                ? Thunk<Request, DeepThunk<Request, U>>[]
                 :
                     | (T[P] extends ReadonlyArray<infer U>
-                        ? ReadonlyArray<
-                            Thunk<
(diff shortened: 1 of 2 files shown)


packages/graphql-fixtures/src/fill.ts
```diff
@@ -148,33 +149,56 @@ export function createFiller(
...
 
       return fillObject(
-        operation.rootType,
-        operation.rootType,
-        // the root type is kind of weird, since there is no "field" that
-        // would be used in a resolver. For simplicity in the common case
-        // we just hack this type to make it conform.
-        [operation as any],
+        fragment.possibleTypes[0],
+        fragment.possibleTypes[0],
```
lemonmade, line 163: To match the filling the way the rest of the autofilling stuff works, I think you'd need to randomly pick the concrete type you are filling for, rather than just taking the first one?

lemonmade (commented):
I'd probably split this into its own, dedicated API, rather than changing `createFiller()`. That's mostly to avoid the breaking change, but also because I think this API should be very secondary to filling at the query level.

alexandcote:
>I'd probably split this into its own, dedicated API, rather than changing createFiller(). That's mostly to avoid the breaking change, but also because I think this API should be very secondary to filling at the query level.

Maybe I can keep `createFiller` as is and expose `createFillers`? I would like to only pass the options and schema once and expose every useful filler from there. What do you think?

atesgoral:
Is this still on?

alexandcote:
Yes, I simply didn't have the time to update it. I will try to do it next week

mateus:

packages/graphql-fixtures/src/fill.ts
```diff
@@ -142,6 +147,9 @@ const defaultResolvers = {
   ID: () => faker.datatype.uuid(),
 };
 
+/**
+ * @deprecated You should use `createFillers` instead.
+ **/
 export function createFiller(
```
alexandcote, line 153: Deprecated 😄

mateus:
https://github.com/Shopify/web/pull/89748 bumped the package in web to see what would happen and some tests are failing https://buildkite.com/shopify/web-ci-builder/builds/655408. I did not dig into it, but I'd suggest investigating before moving forward. I will be able to revisit this next week if that's ok 🙏🏻 .

mateus:
Just a single type-check error left now

```
Property 'variables' does not exist on type 'GraphQLRequest<B2BCustomerSearchQueryData, Variables, B2BCustomerSearchQueryPartialData> | null'.
```

https://buildkite.com/shopify/web-ci-builder/builds/655504#0187819c-2458-4d83-a3bf-ff194e766814

It should be occurring on any file that contains text that matches`fillGraphQL\(.+?, \(`.

alexandcote:
@mateus I used to have this as a minor change but the type changes are breaking changes so because of that I decided to make it a major bump.

PR #2102, opened 2021-12-06
✨ Add cleanup option to createMount in react-testing

marutypes:
## Description

This PR adds a `cleanup` function to `createMount` in `@shopify/react-testing` that can be used to handle any additional cleanup a wrapper or its context might need. This was originally proposed by @GoodForOneFare for stopping any ongoing gql requests and other such async handlers that may be leaking out of our context objects.

## Type of change
- [x] `@shopify/react-testing` Minor: New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/packages/react-testing/CHANGELOG.md
+++ b/packages/react-testing/CHANGELOG.md
@@ -9,7 +9,7 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ### Added
 
-- Added new `cleanup` option to `createMount`
+- Added new `cleanup` option to `createMount` [[#2102](https://github.com/Shopify/quilt/pull/2102)]
 
 ## 3.2.9 - 2021-11-23
 


dahukish (approved):
:shipit:

BPScott (approved):
Looking great to me! 

Remember the PR link in the changelog and :shipit:

packages/react-testing/src/mount.ts
```diff
@@ -127,6 +133,14 @@ export function createMount<
       resolveRoot: (root) => root.find(element.type),
     });
 
+    if (cleanup) {
+      const originalDestroy = wrapper.destroy.bind(wrapper);
+      wrapper.destroy = () => {
+        cleanup(wrapper, options);
```
GoodForOneFare, line 139: Would it be a lot of effort to support async `cleanup`? A common pattern in web is "loading state tests need end with a `graphQL.resolveAll()` to complete pending `React.act` calls".

If we could automate that, adding an `afterEach` check for dangling acts becomes less noisy.

packages/react-testing/CHANGELOG.md
```diff
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Added
+
+- Added new `cleanup` option to `createMount`
```
atesgoral, line 12: `:yells-at-pr-self-reference:`

packages/react-testing/README.md
```diff
@@ -148,6 +148,10 @@ This function allows you to perform additional logic after a component has been
 
 If this option returns a `Promise`, the result of calling `mount()` will become a promise that resolves to the custom `Root` instance. Otherwise, it will synchronously return the `Root` instance. If you specify the `Async` generic argument as `true`, you **must** pass this option.
 
+##### `cleanup(root: CustomRoot, options: MountOptions): void`
+
+This function allows you to perform additional cleanup when a wrapper is destroyed It gets called with a special [`Root`](#root) instance that has one additional property: `context`, the object with the context you created in `context()` (or an empty object). You can use this hook to clean up any dangling promises or other operations when your test suite is being discarded.
```
GoodForOneFare, line 153:
```suggestion
This function allows you to perform additional cleanup when a wrapper is destroyed. It gets called with a special [`Root`](#root) instance that has one additional property: `context`, the object with the context you created in `context()` (or an empty object). You can use this hook to clean up any dangling promises or other operations when your test suite is being discarded.
```

GoodForOneFare (approved):
👍 I have a question about async support, but this can ship as is.

packages/react-testing/src/mount.ts
TheMallen: My worry is that `destroy` is currently sync, which means that existing uses could break if its unexpectedly changed, and simply overwriting it with an async function in some cases doesn't fix types. I can take a crack at it though.

TheMallen: also @GoodForOneFare  if you have any ideas about how to deal with the points above (or think they're not a big deal) that's also good info :P

GoodForOneFare: Those are good points! I figured it wouldn't be trivial.

Overnight, I talked myself into wanting an alternative solution. Something more explicit, and part of general test flow.

The bad idea version of this looks like:

```ts
await component.trigger('onFocus', {andTotesResolveGraphQLForMe: true});
```

PR #2103, opened 2021-12-07
[Quilt Rails]: Ignore any problem with performance report and process in background

miry:
## Description

Performance report is a proxy controller that broke shopstatus' SLO
when users tried to send bad request.

Update the code to ignore any exceptions from reporting and run process in background.

## Checklist


miry:
@Shopify/admin-web-foundations pls review. It seems team is not added to the Collobarators of the repo, that I can request review from team.

heathernfran (approved):

PR #2106, opened 2021-12-09
[@shopify/storybook-a11y-test] Use @axe-core/puppeteer

dfmcphee:
## Description

This updates the `@shopify/storybook-a11y-test` package to be compatible with the changes storybook introduced in `6.4.0`. They changed the way global variables are handled in the iframe and we can no longer access `axe` the way we were in these tests. I instead updated it to use `@axe-core/puppeteer` which handle the injection into the iframe for us.

You can see the a11y tests erroring out in this PR here: https://github.com/Shopify/polaris-react/pull/4796

And I tested the output of these changes in the build here: https://github.com/Shopify/polaris-react/runs/4471933720?check_suite_focus=true

## Type of change
- [x] `@shopify/storybook-a11y-test` Minor: New feature (non-breaking change which adds functionality)

## Checklist


caropinzonsilva (approved):
Changes look good! I did not 🎩

PR #2107, opened 2021-12-09
Add .length getter to jest-dom-mock Storage

mawburn:
## Description

When trying to write tests that were trying to utilize the Storage mock, I was unable to use the length feature from the Storage object. 

This change should allow us to use the length property of Storage.

https://developer.mozilla.org/en-US/docs/Web/API/Storage/length

## Type of change

I'm not sure if this is worthy of a minor bump or just a patch, since it's functionality that exists on Storage and not in this library, I feel like a Patch is appropriate. 

- [x] `jest-dom-mock` Patch: Bug (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/packages/jest-dom-mocks/CHANGELOG.md
+++ b/packages/jest-dom-mocks/CHANGELOG.md
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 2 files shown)


BPScott (changes requested):
You don't want that publish commit or any of those tags you've possibly created in here.

Just the commit with your change is fine. Please rebase to so this contains just your change and no version wrangling.

Then we'll do the version release stuff against the master branch

mawburn:
Thanks @BPScott, this should be fixed. That felt a little heavy handed, but I was trying to follow the docs. Apologies.

BPScott (approved):

PR #2111, opened 2021-12-16
Handle tryRequire errors

vsumner:
## Description

Fixes (issue #)

## Type of change

## Checklist

### DIFF
--- a/packages/react-async/src/hooks.ts
+++ b/packages/react-async/src/hooks.ts
@@ -80,6 +80,10 @@ export function useAsync<T>(
     try {
       const resolved = await resolver.resolve();
 
+      if (resolver.requireError) {
+        throw resolver.requireError.error;
+      }
+
       if (mounted.current) {
         // It's important to use the function form of setValue here.
         // Resolved is going to be a function in most cases, since it's
--- a/packages/react-async/src/tests/component.test.tsx
+++ b/packages/react-async/src/tests/component.test.tsx
@@ -42,16 +42,27 @@ function Loading() {
   return <>Loading…</>;
 }
 
+jest.mock('@shopify/async', () => {
+  return {
+    ...jest.requireActual('@shopify/async'),
+    createResolver: jest.fn(),
+  };
+});
+
+const mockCreateResolver = jest.requireMock('@shopify/async').createResolver;
+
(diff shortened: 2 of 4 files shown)


packages/react-async/src/tests/component.test.tsx
```diff
@@ -42,16 +42,27 @@ function Loading() {
...
 }
 
+jest.mock('@shopify/async', () => {
+  return {
+    ...jest.requireActual('@shopify/async'),
+    createResolver: jest.fn(),
+  };
+});
+
+const mockCreateResolver = jest.requireMock('@shopify/async').createResolver;
```
vsumner, line 52: I really don't like how I'm handling this mock and would be VERY open to suggestions here.

I couldn't figure out another way of handling this.

dahukish: I think the cleanest way would be what we chatted about before and break these out into another test suite 🤷

vsumner: yeah. Done

packages/react-async/src/hooks.ts
```diff
@@ -80,6 +80,10 @@ export function useAsync<T>(
     try {
       const resolved = await resolver.resolve();
 
+      if (resolver.requireError) {
+        throw resolver.requireError.error;
```
GoodForOneFare, line 84: I've never fully understood webpack's behaviour during SSR breakages, so I'm fuzzy on if this is a behavioural change or not 😅 

My understanding of old behaviour is:
* The first failing webpack `require` is silently swallowed 
* `null` is returned to the caller
* The caller renders nothing for this component (and all children), and HTML contains a shallow component tree

So does this:
1. Behave exactly as before
2. Display an error for the first failing request, then shallow render subsequent requests
3. Throw on every request until a rollback

I think it's _2_, but may be completely wrong about everything 😸

GoodForOneFare:
I had pictured the end result looking more like this:

```ts
  const hydrationManager = new HydrationManager({assetErrorCallback: (error) => {
    metricsService.increment('require_error');
    bugsnag.logException(error);
  }});
```

I'm very rusty, but it looks like all of the helpers in `createAsyncComponent` can `useHydrationManager`, so maybe my suggestion is possible.

PR #2124, opened 2022-01-10
Fix disableAnimation

kyledurand:
## Description

Fixes an issue I was seeing in https://github.com/Shopify/polaris-react/pull/4796 where animated component tests were flakey. 

I'm not sure why values like 0.0001s were being used and also why the one for animation delay was a negative so if this change breaks something let me know.

## Type of change

More explicitly setting transition and animation properties to cancel animation

## Checklist


kyledurand:
I guess this could break `animationend` 🤔 I'll try reverting those values and see if the modal tests still pass

dfmcphee (approved):

PR #2125, opened 2022-01-11
[react-shortcuts]Fix node and allow to override ignored input tags

moimael:
## Description

_Taking over for @moimael_ (@loic-d)

This PR fixes:
- A false positive with following test: `calls shortcuts that are scoped to a specific node only when that node is focused`
- Fixes issue with early return in `useShortcut()` preventing shortcut to be registered when `node` is provided
- Exposes a new `ignoreTags` flag allowing consumers to override the default ignored tags (`textarea`, `select`, `input`)

🎩  ping me for links to the Web Admin PR consuming these changes + Spin URL

## Type of change

## Checklist

### DIFF
--- a/packages/react-shortcuts/src/Shortcut/hooks.ts
+++ b/packages/react-shortcuts/src/Shortcut/hooks.ts
@@ -3,6 +3,10 @@ import React from 'react';
 import {ShortcutContext} from '../ShortcutProvider';
 import Key, {HeldKey} from '../keys';
 
+const DEFAULT_IGNORED_TAGS = ['INPUT', 'SELECT', 'TEXTAREA'] as const;
+
+export type DefaultIgnoredTag = typeof DEFAULT_IGNORED_TAGS[number];
+
 export interface Subscription {
   unsubscribe(): void;
 }
@@ -11,6 +15,7 @@ export interface Options {
   held?: HeldKey;
   node?: HTMLElement | null;
   ignoreInput?: boolean;
+  acceptedDefaultIgnoredTags?: DefaultIgnoredTag[];
   allowDefault?: boolean;
 }
 
@@ -21,13 +26,15 @@ export default function useShortcut(
 ) {
   const shortcutManager = React.useContext(ShortcutContext);
   const subscription = React.useRef<Subscription | null>(null);
-  const {node, held, ignoreInput, allowDefault} = options;
+  const {
+    node,
+    held,
+    ignoreInput,
+    acceptedDefaultIgnoredTags,
+    allowDefault,
+  } = options;
 
   React.useEffect(() => {
-    if (node != null) {
-      return;
-    }
-
     if (shortcutManager == null) {
       return;
     }
@@ -38,6 +45,7 @@ export default function useShortcut(
       node,
       held,
       ignoreInput: ignoreInput || false,
+      ignoredTags: getIgnoredTags(acceptedDefaultIgnoredTags),
       allowDefault: allowDefault || false,
     });
 
@@ -51,10 +59,21 @@ export default function useShortcut(
   }, [
     node,
     ordered,
-    onMatch,
     held,
     ignoreInput,
+    acceptedDefaultIgnoredTags,
     allowDefault,
     shortcutManager,
+    onMatch,
   ]);
+}
+
+function getIgnoredTags(acceptedDefaultIgnoredTags?: DefaultIgnoredTag[]) {
+  if (!acceptedDefaultIgnoredTags?.length) {
+    return (DEFAULT_IGNORED_TAGS as unknown) as DefaultIgnoredTag[];
+  }
+
+  return DEFAULT_IGNORED_TAGS.filter(
+    (tag) => !acceptedDefaultIgnoredTags.includes(tag),
+  );
 }
--- a/packages/react-shortcuts/src/tests/ShortcutManager.test.tsx
+++ b/packages/react-shortcuts/src/tests/ShortcutManager.test.tsx
@@ -3,7 +3,7 @@ import {mount} from '@shopify/react-testing';
 import {timer} from '@shopify/jest-dom-mocks';
 
 import Key, {HeldKey, ModifierKey} from '../keys';
-import Shortcut from '../Shortcut';
+import Shortcut, {DefaultIgnoredTag} from '../Shortcut';
 import ShortcutProvider from '../ShortcutProvider';
 
 import ShortcutWithFocus from './ShortcutWithRef';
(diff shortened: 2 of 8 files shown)


packages/react-shortcuts/src/Shortcut/hooks.ts
```diff
@@ -11,23 +11,22 @@ export interface Options {
   held?: HeldKey;
   node?: HTMLElement | null;
   ignoreInput?: boolean;
+  ignoreTags?: HTMLElement['tagName'][];
```
loic-d, line 14: For context, this is needed to support undo/redo using keyboard shortcuts on Web's `Editor` component which uses a `textarea` under the hood.

packages/react-shortcuts/src/Shortcut/hooks.ts
```diff
@@ -51,10 +51,11 @@ export default function useShortcut(
   }, [
     node,
     ordered,
-    onMatch,
     held,
     ignoreInput,
+    ignoreTags,
     allowDefault,
     shortcutManager,
+    onMatch,
```
loic-d, line 59: Moving callback to the end

packages/react-shortcuts/src/tests/ShortcutManager.test.tsx
```diff
@@ -120,13 +120,24 @@ describe('ShortcutManager', () => {
   it('calls shortcuts that are scoped to a specific node only when that node is focused', () => {
     const spy = jest.fn();
 
-    const app = mount(
+    const ShortcutWithFocusAndProvider = ({spy, focusNode}) => (
       <ShortcutProvider>
-        <ShortcutWithFocus spy={spy} />
-      </ShortcutProvider>,
+        <ShortcutWithFocus spy={spy} focusNode={focusNode} />
```
loic-d, line 125: We had a false positive here: 
- ref in `ShortcutWithFocus` was initialized with `null`
- `useShortcut` hook was called with `node: null` (not what we expected)
- `onMatch` spy was called because of the `partiallyMatching` code path (not `partiallyMatching && onFocusedNode` because `node` is null)

packages/react-shortcuts/src/tests/ShortcutManager.test.tsx
```diff
@@ -120,13 +120,24 @@ describe('ShortcutManager', () => {
...
+    );
+
+    const app = mount(
+      <ShortcutWithFocusAndProvider spy={spy} focusNode={false} />,
     );
 
-    app.forceUpdate();
+    keydown('z');
+
+    expect(spy).not.toHaveBeenCalled();
```
loic-d, line 135: Not hasn't been focused yet, `onMatch` should not be called

packages/react-shortcuts/src/tests/ShortcutManager.test.tsx
```diff
@@ -120,13 +120,24 @@ describe('ShortcutManager', () => {
...
+
+    app.setProps({
+      focusNode: true,
```
loic-d, line 138: Force a focus on the node

packages/react-shortcuts/src/tests/ShortcutManager.test.tsx
```diff
@@ -225,6 +236,67 @@ describe('ShortcutManager', () => {
     expect(event.preventDefault).toHaveBeenCalled();
   });
 
+  it.each(['textarea', 'input', 'select'])(
```
loic-d, line 239: Missing test: by default shortcuts are not called for these elements

packages/react-shortcuts/src/tests/ShortcutWithRef.tsx
```diff
@@ -4,23 +4,41 @@ import Shortcut from '../Shortcut';
...
-  const {spy} = props;
-  const node = React.useRef<HTMLButtonElement | null>(null);
+export default function ShortcutWithFocus({
+  spy,
+  focusNode = true,
+  tagName = 'button',
+  ignoreTags,
+}: Props) {
+  const [node, setNode] = React.useState<HTMLElement | null>(null);
+  const elementWithFocus = React.createElement(tagName, {
```
loic-d, line 19: `button` by default, but test can provide custom elements

packages/react-shortcuts/src/Shortcut/hooks.ts
```diff
@@ -11,23 +11,22 @@ export interface Options {
...
   onMatch: (matched: {ordered: Key[]; held?: HeldKey}) => void,
   options: Options = {},
 ) {
   const shortcutManager = React.useContext(ShortcutContext);
   const subscription = React.useRef<Subscription | null>(null);
-  const {node, held, ignoreInput, allowDefault} = options;
+  const {node, held, ignoreInput, ignoreTags, allowDefault} = options;
 
   React.useEffect(() => {
-    if (node != null) {
```
loic-d, line 27: I believe the `node` feature was not working all together as this early return was preventing the shortcut to be registered when a `node` was provided 🤔

packages/react-shortcuts/src/Shortcut/hooks.ts
```diff
@@ -11,23 +11,22 @@ export interface Options {
   held?: HeldKey;
   node?: HTMLElement | null;
   ignoreInput?: boolean;
+  ignoreTags?: HTMLElement['tagName'][];
   allowDefault?: boolean;
 }
 
+const DEFAULT_IGNORE_TAGS = ['INPUT', 'SELECT', 'TEXTAREA'];
```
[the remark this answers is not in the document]
loic-d: Update: going with `acceptDefaultIgnoredTags`

packages/react-shortcuts/src/Shortcut/hooks.ts
loic-d: That's the idea. We are providing our editor's `textarea` as `node` and making sure `TEXTAREA` is removed from the list of ignored tags so we can hit this code path in `ShortcutManage.tsx`:

```

        if (node) {
          const onFocusedNode = document.activeElement === node;
          return partiallyMatching && onFocusedNode;
        }
```

packages/react-shortcuts/src/tests/ShortcutWithRef.tsx
```diff
@@ -4,23 +4,41 @@ import Shortcut from '../Shortcut';
...
-export default function ShortcutWithFocus(props: Props) {
-  const {spy} = props;
-  const node = React.useRef<HTMLButtonElement | null>(null);
+export default function ShortcutWithFocus({
+  spy,
+  focusNode = true,
+  tagName = 'button',
+  ignoreTags,
+}: Props) {
+  const [node, setNode] = React.useState<HTMLElement | null>(null);
```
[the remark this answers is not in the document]
loic-d: This is not about the focus, it's about the `node` being updated when the ref is available.
We don't have access to the ref until the element is rendered (we can't provide a React element as the `node` to `Shortcut`). This forces us to use a ref callback... Surprisingly, [there's no infinite loop](https://codesandbox.io/s/laughing-sun-4mkc1?file=/src/App.js) so I would expect `React.createElement` to do some kind of memoization?

packages/react-shortcuts/src/Shortcut/hooks.ts
```diff
@@ -11,23 +17,32 @@ export interface Options {
   held?: HeldKey;
   node?: HTMLElement | null;
   ignoreInput?: boolean;
+  acceptedDefaultIgnoredTags?: DefaultIgnoredTag[];
   allowDefault?: boolean;
 }
 
+const DEFAULT_IGNORED_TAGS = [
```
ryanwilsonperkin, line 24: Honestly I'd probably just go with a `as const` declaration on this instead of the enum, and then export this value. That way we don't need to worry about adding something to the enum but missing it here, and we'll still get the typing benefits by doing `typeof DEFAULT_IGNORED_TAGS[number]` for the option type definition

loic-d: Good call 👍

ryanwilsonperkin (approved):

PR #2132, opened 2022-01-18
Updates faker to reflect change in library maintainers ("faker" --> "@faker-js/faker")

ecbrodie:
## Description

This PR upgrades the `faker` library across all of Quilt's NPM packages, mainly to point the library towards a new package name to reflect a change in the library's maintainers.

Earlier this month, the maintainer of FakerJS went rogue and deleted all code from the repo ([news article](https://www.bleepingcomputer.com/news/security/dev-corrupts-npm-libs-colors-and-faker-breaking-thousands-of-apps/) [old repo link](https://github.com/marak/Faker.js/)). In response to this act, [the open source community banded together](https://fakerjs.dev/update.html#an-update-from-the-faker-team) to fork off the original library and create a new one called [`@faker-js/faker`](https://github.com/faker-js/faker).

It is in Quilt's best interest to point its usage of Faker away from the abandoned package towards the new forked package with a long-term maintenance plan in place.

## Type of change

All packages have had their dev dependencies effectively upgraded via the switch to `@faker-js/faker`. I upgraded these libraries to the latest version before all the Faker changes occurred, `5.5.3`. There is a `6.0.x` version in the works that Quilt should eventually adopt (for example, it rewrites the package into TypeScript), but it's currently in Alpha and I didn't feel comfortable adopting an alpha version at this time.

There were several `devDependency` entries for Faker in nested packages that I did not feel like we're necessary because we were already specifying the `devDependency` for Faker in the root `package.json`, therefore I removed them. If reviewers wish for these entries to be kept, I'll happily add them back in.

_The one library that is publicly affected by this change is `graphql-fixtures`, since it lists Faker as a dependency._ I've added a CHANGELOG entry for this and assumed this will be a Minor release. I'm also open to feedback on this.

- [x] `graphql-fixtures` Minor: New feature (non-breaking change which adds functionality)

## Checklist


vsumner (approved):

packages/react-form/package.json
```diff
@@ -27,8 +27,8 @@
     "react": ">=16.8.0 <18.0.0"
   },
   "devDependencies": {
-    "@types/faker": "^4.1.5",
-    "faker": "^4.1.0"
+    "@faker-js/faker": "^5.5.3",
```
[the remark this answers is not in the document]
ecbrodie: This was simply a case of my updating this package first, before realizing that faker was everywhere. Forgot to loop back to here, so I'll remove it. Thanks.

ecbrodie:
@vsumner @dahukish thanks for the reviews! What are the next steps for this PR? Does a member of your team merge it and release a new version of `graphql-fixtures`, or am I expected to act further from here?

melodyhabbouche (approved):
LGTM code wise! (from what I am familiar with)

caropinzonsilva:
@ecbrodie can you fix the lint errors? https://github.com/Shopify/quilt/runs/4856332216?check_suite_focus=true

ecbrodie:
@caropinzonsilva I'm happy to fix the linting issues. Most of them seem related to import ordering with Faker, but not all of them seem related to this PR. So, just be warned that there may be changes in this PR that are out of scope to a Faker lib upgrade.

caropinzonsilva:
> I'm happy to fix the linting issues. Most of them seem related to import ordering with Faker, but not all of them seem related to this PR. So, just be warned that there may be changes in this PR that are out of scope to a Faker lib upgrade.

Sounds good, thank you so much! As long as they are import order changes we should be ok, but I'll take a look at the fixes after you push them. Otherwise, we can't really merge your changes :/

ecbrodie:
Looks like I was wrong about the linting issues being unrelated. Most of them were caused by an accidental double semicolon. That is fixed now.

Unfortunately I'm currently stuck on this linting error (reproducible locally with `yarn lint`):

```
ESLint found lint errors.

error: Parsing error: "parserOptions.project" has been set for @typescript-eslint/parser.
The file does not match your project config: faker.d.ts.
The file must be included in at least one of the projects provided at faker.d.ts:
```

I'm unsure exactly how to fix this. I tried linking tsconfig to this custom type definition file via the `"include"` option in `tsconfig.base.json`, but that didn't help. I'm out of time for now to continue looking at this. Does anyone have an idea how to work around this one?

ecbrodie:
Okay, looks like I figured it out. TLDR, `yarn clean` is your friend :)

I'll merge once all jobs finish green.

packages/graphql-fixtures/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## Unreleased -->
 
+## Unreleased - 2022-01-18
```
BPScott, line 10: When formatting changelogs, uncomment the unreleased heading and don't add a date (that will be added when we trigger the publish, as "date you did a PR" and "date a publish was done" may note be the same)

```suggestion
## Unreleased
```

tsconfig.eslint.json
```diff
@@ -6,7 +6,8 @@
     "./config/**/*.ts",
     "./tests/**/*.ts",
     "./packages/**/*.ts",
-    "./packages/**/*.tsx"
+    "./packages/**/*.tsx",
+    "faker.d.ts"
```
BPScott, line 10: This line can be removed if you move the d.ts file into the config directory per my earlier comment

BPScott (changes requested):
Can you move the faker.d.ts file into `config/typescript` with our other d.ts files.

I've added a suggestion that should keep the "how to format the changelog" linting happy

packages/graphql-fixtures/package.json
```diff
@@ -24,9 +24,9 @@
     "node": ">=12.14.0"
   },
   "dependencies": {
+    "@faker-js/faker": "^5.5.3",
     "@shopify/useful-types": "^3.0.5",
-    "@types/faker": "^4.1.2",
-    "faker": "^4.1.0",
+    "@types/faker": "^5.5.9",
```
BPScott, line 29: This feels like it shall be kinda janky - we're depending on `@faker-js/faker` but we're not shipping any types for that package. We currently rely on your faker.d.ts file to say "hey they types for `@faker-js/faker` are the same as `faker`" but that would only apply to the quilt repo, not in any other project.

ecbrodie: I hear you about this feedback and somewhat agree with the concern. However, do note that I'm not changing any of the existing behaviour regarding how `graphql-fixtures` is exporting the `@types/faker` package. I'm simply just preserving existing behaviour, except with a necessary upgrade.

May I suggest that we hold off on any further changes to this until Quilt upgrades to Faker v6? This way, Quilt can take advantage of having Faker being rewritten into TypeScript and thus not requiring a DefinitelyTyped dependency anymore.

packages/graphql-fixtures/CHANGELOG.md
ecbrodie: Awesome. I suggest that a note about this be added to `CONTRIBUTING.md`, but I'll leave that to this repo's maintainers to decide :)

packages/graphql-fixtures/package.json
BPScott: My fear was that we have publicly exposed function - `createFiller` whose type signature depends types that come from faker - and now that we're using `@faker-js/faker` instead of `faker` consumers won't know where to find the types for `@faker-js/faker`. We provide types for `@faker-js/faker` in this repo in the modules.d.ts file, but there's nothing teaching the published package that.

[Looking into the the types](https://unpkg.com/browse/graphql-fixtures@1.1.5/build/ts/fill.d.ts) it doesn't seem like our our exports's parameters or return types don't actually depend upon faker, so I think we get away with this.

BPScott (approved):
It turns out that we don't use the types from faker as part of the public API of graphql-fixtures so my original concern isn't applicable.

Good to go once you make the changelog linter happy

packages/graphql-fixtures/CHANGELOG.md
```diff
@@ -5,7 +5,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
```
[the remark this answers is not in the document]
ecbrodie: Hah, of course it does! Will update.

PR #2135, opened 2022-01-20
Add format function to @shopify/dates

### LINKED ISSUE #2101: [@shopify/dates] - A format function that works like moment().format
## Overview

Would it be desired to add a function which operates similarly to `moment().format()` ([docs](https://momentjs.com/docs/#/displaying/format/))?

This function is used a lot in various repos, and it wouldn't be too hard to implement the most used format strings across Shopify.

Some usage examples:

https://github.com/Shopify/iBookify-react/blob/aa59b9a610fc834ca17842878f27b2ce45d6e852/ibookify-react/src/components/BookingDatePicker.js#L34

https://github.com/Shopify/estuary/blob/9918d0b1742a13cecf222802b2886ad0de73db90/app/ui/NotebooksShow/components/MarkdownEditor/utilities.ts#L163

https://github.com/Shopify/fbs/blob/f82f7ffd11399d6ce5a7b78cf4d23c4394a5dc49/app/ui/MerchantApp/sections/OrderForecast/utilities/index.ts#L33-L35

## Motivation

We [wanted to replace `moment` in the `fbs` repo](https://github.com/Shopify/fbs/pull/33041), and noticed that we had to write custom formatting functions to generate date strings, since the formatting function in `@shopify/dates` is only a wrapper around [`Intl.DateTimeFormat().format()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/format).

I would be happy to open a PR with a proposed function to do this, but wanted to check first if it made sense before proceeding.

### Area

- [x] Add any relevant `Area: <area>` labels to this issue

### Scope

  - [x] Tag it with the `Package: <package_name>` label.

---

## Checklist

kathawala:
## Description

Fixes #2101

Adds a `format` function to `@shopify/dates` which serves as a drop-in replacement for `moment().format` usage.

Please see the linked issue for more rationale as to why this is wanted: https://github.com/Shopify/quilt/issues/2101

It works as follows:

```ts
import {format} from '@shopify/dates';

const date = new Date(2010, 1, 14, 15, 25, 50, 125);
const formattedDate = format(date, 'dddd, MMMM D YYYY, h:mm:ss a');
formattedDate === 'Sunday, February 14 2010, 3:25:50 pm'; // true
```

This handles the following format string options:

|              |      |                                        |
| -----------: | ---- | -------------------------------------- |
|        Month | M    | 1 2 ... 11 12                          |
|              | MM   | 01 02 ... 11 12                        |
|              | MMM  | Jan Feb ... Nov Dec                    |
|              | MMMM | January February ... November December |
| Day of Month | D    | 1 2 ... 30 31                          |
|              | DD   | 01 02 ... 30 31                        |
|  Day of Week | ddd  | Sun Mon ... Fri Sat                    |
|              | dddd | Sunday Monday ... Friday Saturday      |
|         Year | YY   | 70 71 ... 29 30                        |
|              | YYYY | 1970 1971 ... 2029 2030                |
|         Hour | H    | 0 1 ... 22 23                          |
|              | HH   | 00 01 ... 22 23                        |
|              | h    | 1 2 ... 11 12                          |
|              | hh   | 01 02 ... 11 12                        |
|       Minute | m    | 0 1 ... 58 59                          |
|              | mm   | 00 01 ... 58 59                        |
|       Second | s    | 0 1 ... 58 59                          |
|              | ss   | 00 01 ... 58 59                        |
|        AM/PM | A    | AM PM                                  |
|              | a    | am pm                                  |

## Type of change

## Checklist

### DIFF
--- a/packages/dates/CHANGELOG.md
+++ b/packages/dates/CHANGELOG.md
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Added
+
+- Added `format()` function, matching `moment().format`. [[#2135](https://github.com/Shopify/quilt/pull/2135)]
 
 ## 1.0.13 - 2022-01-06
 
--- a/packages/dates/README.md
+++ b/packages/dates/README.md
@@ -31,6 +31,45 @@ const timeZone2 = 'America/Toronto';
 const newDate = applyTimeZoneOffset(date, timeZone1, timeZone2); //'2018-06-01T02:00:00.000Z'
 ```
 
+### `format`
+
+Lighter replacement for [`format()` from the `moment` library](https://momentjs.com/docs/#/displaying/format/).
+Can output to any given locale / timezone (defaults to the system's locale & timezone).
+Handles a subset of the tokens `moment().format()` handles, namely the following:
+
+|              |      |                                        |
+| -----------: | ---- | -------------------------------------- |
+|        Month | M    | 1 2 ... 11 12                          |
+|              | MM   | 01 02 ... 11 12                        |
+|              | MMM  | Jan Feb ... Nov Dec                    |
+|              | MMMM | January February ... November December |
+| Day of Month | D    | 1 2 ... 30 31                          |
+|              | DD   | 01 02 ... 30 31                        |
+|  Day of Week | ddd  | Sun Mon ... Fri Sat                    |
+|              | dddd | Sunday Monday ... Friday Saturday      |
+|         Year | YY   | 70 71 ... 29 30                        |
+|              | YYYY | 1970 1971 ... 2029 2030                |
+|         Hour | H    | 0 1 ... 22 23                          |
+|              | HH   | 00 01 ... 22 23                        |
+|              | h    | 1 2 ... 11 12                          |
+|              | hh   | 01 02 ... 11 12                        |
+|       Minute | m    | 0 1 ... 58 59                          |
+|              | mm   | 00 01 ... 58 59                        |
+|       Second | s    | 0 1 ... 58 59                          |
+|              | ss   | 00 01 ... 58 59                        |
+|        AM/PM | A    | AM PM                                  |
+|              | a    | am pm                                  |
+
+```ts
+import {format} from '@shopify/dates';
+
+const date = new Date(2021, 0, 14, 13, 2, 3);
+const dateStr = format(date, 'YYYY-MM-DD h:mm:ss A'); // 2021-01-14 1:02:03 PM
+
+const date2 = new Date(Date.UTC(2021, 1, 1));
+const dateStr2 = format(date, 'M/D/YY', 'UTC', 'en-US'); // 1/1/21
+```
+
 ### `formatDate`
 
 Takes in a date object and two additional parameters, the locale and an optional options object. Returns a new date string with the applied locale and options.
--- a/packages/dates/src/format.ts
+++ b/packages/dates/src/format.ts
@@ -1,0 +1,151 @@
+import {memoizedGetDateTimeFormat} from './utilities';
+
+interface DateTimeFormatOptions {
+  localeMatcher?: 'lookup' | 'best fit';
+  weekday?: 'long' | 'short' | 'narrow';
+  era?: 'long' | 'short' | 'narrow';
+  year?: 'numeric' | '2-digit';
+  month?: 'numeric' | '2-digit' | 'long' | 'short' | 'narrow';
+  day?: 'numeric' | '2-digit';
+  hour?: 'numeric' | '2-digit';
+  minute?: 'numeric' | '2-digit';
+  second?: 'numeric' | '2-digit';
+  timeZoneName?: 'long' | 'short';
+  formatMatcher?: 'basic' | 'best fit';
+  hour12?: boolean;
+  timeZone?: string;
+}
+
+type Token =
+  | 'YYYY'
+  | 'YY'
+  | 'MMMM'
+  | 'MMM'
+  | 'MM'
+  | 'M'
+  | 'DD'
+  | 'D'
+  | 'HH'
+  | 'H'
+  | 'hh'
+  | 'h'
+  | 'mm'
+  | 'm'
+  | 'ss'
+  | 's'
+  | 'A'
+  | 'a'
+  | 'dddd'
+  | 'ddd';
+
+type TokenDateTimeFormatOptionsMap = {
+  [key in Token]: {
+    type: Intl.DateTimeFormatPartTypes;
+    options: DateTimeFormatOptions;
+    extraTransform?: (str: string) => string;
+  };
+};
+
+const TRANSFORM_MAP = {
+  numericFromTwoDigit: (time: string) =>
+    time.startsWith('0') && time.length === 2 ? time.charAt(1) : time,
+  twoDigitFromNumeric: (time: string) =>
+    time.length === 1 ? `0${time}` : time,
+  lowercase: (time: string) => time.toLowerCase(),
+};
+
+const TOKEN_MAP: TokenDateTimeFormatOptionsMap = {
+  YYYY: {type: 'year', options: {year: 'numeric'}},
+  YY: {type: 'year', options: {year: '2-digit'}},
+  MMMM: {type: 'month', options: {month: 'long'}},
+  MMM: {type: 'month', options: {month: 'short'}},
+  MM: {type: 'month', options: {month: '2-digit'}},
+  // eslint-disable-next-line id-length
+  M: {type: 'month', options: {month: 'numeric'}},
+  DD: {type: 'day', options: {day: '2-digit'}},
+  // eslint-disable-next-line id-length
+  D: {type: 'day', options: {day: 'numeric'}},
+  HH: {type: 'hour', options: {hour: '2-digit', hour12: false}},
+  // eslint-disable-next-line id-length
+  H: {
+    type: 'hour',
+    options: {hour: 'numeric', hour12: false},
+    // extraTransform is needed when Node and Browsers don't respect the 2-digit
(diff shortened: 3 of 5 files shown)


BPScott:
It feels like you could do something with [formatToParts](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/formatToParts) to get many segments of a datetime in an array, and then you wouldn't need to call `.format()` a once for each part of the datetime

packages/dates/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@shopify/dates",
-  "version": "1.0.13",
+  "version": "1.1.0",
```
BPScott, line 3: version numbers get update at publish time. revert this.

packages/dates/src/format.ts
```diff
@@ -0,0 +1,201 @@
...
+  date: Date,
+  fmtStr: string,
+  tz?: string,
+  localeStr?: string,
+): string {
+  const options = memoizedGetDateTimeFormat(localeStr, {
+    timeZone: tz,
+  }).resolvedOptions();
+  let result = fmtStr;
+  result = replaceToken(result, 'YYYY', getYear(date, 'numeric', options));
```
BPScott, line 13: This looks like you're calling the get* function even if the token was not found in the fmtStr that was requested. This means a `format(SOMEDATE, '')` shall make 20 calls to IntlDateTime. toLocaleTimeString / toLocaleDateString (as each get* call makes one request). That seems absurdly inefficient. You'll need to make this smart enough request the minimal amount of toLocaleTimeString / toLocaleDateString requests.

kathawala: Thanks for the suggestion. I modified the function to use `Intl.DateTimeFormat().formatToParts` like you suggested and now it also only calls the aforementioned function for each given token in the format string (as opposed to calling it 20 times for every string).

I had to call it for each token as opposed to calling it once with all the parts in the format string and pulling the parts out because some of the parts get modified by the presence of the others. Thus to get accurate reads for each part, I have to get each part in isolation from the others.

kathawala:
@BPScott Thanks a lot for the suggestions! I think this version is much more improved. Now, I'm using `formatToParts` and only making a call to it for each token in a given string.

packages/dates/README.md
```diff
@@ -31,6 +31,45 @@ const timeZone2 = 'America/Toronto';
...
+|        AM/PM | A    | AM PM                                  |
+|              | a    | am pm                                  |
+
+```ts
+import {format} from '@shopify/dates';
+
+const date = new Date(2021 0 14 13 2 3);
+const dateStr = format(date, 'YYYY-MM-DD h:mm:ss A'); // 2021-01-14 1:02:03 PM
+
+const date2 = new Date(Date.UTC(2021 1 1));
```
atesgoral, line 69:
```suggestion
const date2 = new Date(Date.UTC(2021, 1, 1));
```

packages/dates/src/format.ts
```diff
@@ -0,0 +1,223 @@
...
+  fmtStr: string,
+  tz?: string,
+  localeStr?: string,
+): string {
+  const {locale, timeZone} = memoizedGetDateTimeFormat(localeStr, {
+    timeZone: tz,
+  }).resolvedOptions();
+  let result = fmtStr;
+
+  // the strategy is as follows:
```
atesgoral, line 135: Couldn't this have been done with one pass of `fmtStr.replace(GIANT_REGEXP_OF_ALL_TOKENS, () => ...)`?

kathawala: I think, yes it could be done this way. In this circumstance, the callback would go through each match and make the replacement as specified by `TOKEN_MAP`, and then we wouldn't have to do multiple regex passes or surrounding the replacements with braces.

I'll try it out and see! Thanks for suggesting, I hadn't thought of it that way. 👍

packages/dates/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## Unreleased -->
 
+## 1.1.0 - 2022-01-20
+
+- Added format function, matching `moment().format`
```
BPScott, line 12:
```suggestion
## Unreleased

### Added

- Added `format()` function, matching `moment().format`. [[#2135](https://github.com/Shopify/quilt/pull/2135)]
```

Don't add the version number stuff, that gets added as part of the release.
Add the PR reference in the change.

BPScott (commented):
Added suggestion to properly format changelog.

I've been trying to make time for this but my todo list is has been growing faster than I've been knocking stuff off it recently. It looks like @atesgoral is gonna take this on, thanks Ates!

packages/dates/CHANGELOG.md
kathawala: Ah, thanks for letting me know how it should be done 👍

packages/dates/src/format.ts
atesgoral: Also, I'd be happy to pair on it!

packages/dates/src/format.ts
```diff
@@ -0,0 +1,166 @@
...
+  fmtStr: string,
+  tz?: string,
+  localeStr?: string,
+): string {
+  const {locale, timeZone} = memoizedGetDateTimeFormat(localeStr, {
+    timeZone: tz,
+  }).resolvedOptions();
+  const allTokens = new RegExp(
+    Object.keys(TOKEN_MAP)
+      .map((ch) => `(${ch})`)
```
atesgoral, line 135: You probably want `^(FOO|BAR|BAZ)$` instead of `(FOO)|(BAR)|(BAZ)`

1. Move the capture parens outside
2. Anchor the start and end -- you don't want to accidentally capture other crap

packages/dates/src/format.ts
```diff
@@ -0,0 +1,167 @@
...
+    timeZone: tz,
+  }).resolvedOptions();
+
+  return fmtStr.replace(ALL_TOKENS_REGEX, (token: Token) => {
+    const timeUnit = getTimeUnitFromToken(token, date, locale, timeZone);
+    return timeUnit ?? '';
+  });
+}
+
+function getTimeUnitFromToken(
```
atesgoral, line 147: This should be inlined with the above function to make things more efficient -- esp. since it's only used by the function above, and also then the above function is not doing a lot on its own :)

packages/dates/src/format.ts
```diff
@@ -0,0 +1,167 @@
...
+function lowercase(time: string) {
+  return time.toLowerCase();
+}
+
+const ALL_TOKENS_REGEX = new RegExp(
+  Object.keys(TOKEN_MAP)
+    .map((ch) => `(${ch})`)
+    .join('|'),
+  'g',
+);
```
atesgoral, line 129:
```suggestion
const ALL_TOKENS_REGEX = new RegExp(
  `^(${Object.keys(TOKEN_MAP).join('|')})$`,
  'g',
);
```

packages/dates/src/format.ts
kathawala: Yeah I can take you up on that!

packages/dates/src/format.ts
atesgoral: Nope. As we saw with the failing tests, we shouldn't anchor :)

atesgoral: But should still `/(FOO|BAR|BAZ)/` to reduce the number of match tokens

packages/dates/src/format.ts
```diff
@@ -0,0 +1,155 @@
...
+      second: '2-digit',
+      hour12: true,
+    },
+    extraParser: lowercase,
+  },
+  dddd: {type: 'weekday', options: {weekday: 'long'}},
+  ddd: {type: 'weekday', options: {weekday: 'short'}},
+};
+
+function numericFromTwoDigit(time: string) {
```
atesgoral, line 112: I would bundle these parsers into a `const PARSER_MAP = {` object above the TOKEN_MAP definition to signify that these are not just floating functions that are used elsewhere, but they are just parsers needed by the TOKEN_MAP. (Also despite hoisting, I personally find it easier to reason about code when functions that are needed are defined above, not below)

kathawala: Yeah, that's a good point. I myself had that thought when coming back to this PR after a day or two haha.

The bundling is also a good suggestion. Thanks!

packages/dates/src/format.ts
```diff
@@ -0,0 +1,155 @@
...
+  }).resolvedOptions();
+
+  return fmtStr.replace(ALL_TOKENS_REGEX, (token: Token) => {
+    const {type, options, extraParser} = TOKEN_MAP[token];
+
+    const timeUnit = Intl.DateTimeFormat(locale, {
+      timeZone,
+      ...options,
+    })
+      .formatToParts(date)
```
atesgoral, line 146: I could be sweet to either memoize this, or bake into to the TOKEN_MAP, but it could be a future optimization consideration.

atesgoral: (assuming time zone and locale are persistent, or we want to incorporate them into the memoization)

packages/dates/src/format.ts
```diff
@@ -0,0 +1,155 @@
...
+type TokenDateTimeFormatOptionsMap = {
+  [key in Token]: {
+    type: Intl.DateTimeFormatPartTypes;
+    options: DateTimeFormatOptions;
+    extraParser?: (str: string) => string;
+  };
+};
+
+const PARSER_MAP = {
+  numericFromTwoDigit(time: string) {
```
atesgoral, line 50: Nitpick / stylistic concern: I almost would want to see these defined as arrow functions here

kathawala: I've made both changes now 👍

atesgoral (approved):
LGTM!

packages/dates/src/format.ts
kathawala: I see. I could do this using Map and having an object key which contains all the arguments to the Intl.DateTimeFormat function.

Let me make a follow up issue for it, then I can get around to it when I have some more time 👍

PR #2139, opened 2022-01-25
🐛 react-i18n: Make unformatCurrency callable by its own result

### LINKED ISSUE #2137: 🐛 react-i18n: unformatCurrency needs to be callable on its own result
## Overview

Ref issue: https://github.com/Shopify/core-issues/issues/33330

With the [latest change in 6.3.1](https://github.com/Shopify/quilt/pull/2126), if the locale uses a period as a thousands separator, you could not safely call `unformatCurrency` on the original value and its result and get the same result.

Example with Danish locale:
```
unformatCurrency('16,32') // '16.32'
unformatCurrency('16.32') // '1632.00'
```

Because some downstream code has this expectation of being able to use `unformatCurrency` on values with English locale formatting, even for locales that use different thousand and decimal separators, the library needs to maintain backwards-compatibility. 

It also needs to support locales like Vietnamese with the Vietnamese đồng currency, which uses a period as a thousands separator and zero decimal places.

## Consuming repo

`Shopify/web` with [this issue](https://github.com/Shopify/core-issues/issues/33330).

trishrempel:
## Description

Fixes #2137 

With the [latest change in 6.3.1](https://github.com/Shopify/quilt/pull/2126), if the locale uses a period as a thousands separator, you could not safely call `unformatCurrency` on the original value and its result and get the same result.

Example with Danish locale:
```ts
unformatCurrency('16,32', 'EUR') // '16.32'
unformatCurrency('16.32', 'EUR') // '1632.00'
```

Because some downstream code has this expectation of being able to use `unformatCurrency` on values with English locale formatting, even for locales that use different thousand and decimal separators, the library needs to maintain backwards-compatibility. 

It also needs to support locales like Vietnamese with the Vietnamese đồng currency, which uses a period as a thousands separator and zero decimal places.

### What approach did you choose and why?

I updated `normalizedNumber` to override with English locale number formatting rules in the following circumstances:
- The locale's decimal symbol is not `.` AND
  - The input string has exactly one period AND
  - The input string has `maxDecimalPlaces` or fewer numerals after the period

The `maxDecimalPlaces` variable is the currency's number of decimal places or 2, whichever is higher.
This covers the scenario where `unformatCurrency` is being called for a currency with three digits (e.g. `BHD`) and with a locale that uses a period as a thousands separator (e.g. Italian). In this case, if the input doesn't supply decimal places, the period is assumed to be a decimal symbol with English locale formatting.

For example, in the `it` locale:
```ts
unformatCurrency('123.456', 'BHD') // '123.456'
unformatCurrency('123.456,000', 'BHD') // '123456.000'
unformatCurrency('9.123.456', 'BHD') // '9123456.000'
```

### What should reviewers look for?

I think this PR hits closer to the mark of being backwards-compatible and recognizing English locale formatting while also respecting locales that don't use periods as decimal separators.

Is there anything I might be missing in terms of unanticipated negative impact for other locales and currencies? 
Any missing use cases or test cases?

### Examples

| Locale | Thousand | Decimal | Currency (places) | Value to unformat | Result |
|---|---|---|---|---|---|
| `vi` | `.` | `,` |  |  |  |
|  |  |  | VND (0) | `20.000` | `20000.00` |
|  |  |  | VND (0) | `2.000,00` | `2000.00` |
|  |  |  | VND (0) | `2.000.0` | `20000.00` |
|  |  |  | VND (0) | `2.000.00` | `200000.00` |
|  |  |  | VND (0) | `2000,00` | `2000.00` |
|  |  |  | VND (0) | `200.00` | `200.00` |
|  |  |  | VND (0) | `2000.0` | `2000.00` |
|  |  |  | VND (0) | `2,000.00` | `2000.00` |
|  |  |  | VND (0) | `2,345.67` | `2346.00` |
|  |  |  | USD (2) | `2.345,67` | `2345.67` |
|  |  |  | USD (2) | `2,345.67` | `2345.67` |
|  |  |  | BHD (3) | `2.000` | `2.000` |
|  |  |  | BHD (3) | `2.000.0` | `20000.000` |
|  |  |  | BHD (3) | `2.000,0` | `2000.000` |
|  |  |  | BHD (3) | `2,000.000` | `2000.000` |
|  |  |  | BHD (3) | `2.00` | `2.000` |
| `fr` | ` ` | `,` |  |  |  |
|  |  |  | USD (2) | `1 234,50` | `1234.50` |
|  |  |  | USD (2) | `1234,50` | `1234.50` |
|  |  |  | USD (2) | `1234.50` | `1234.50` |
|  |  |  | USD (2) | `1,234.50` | `1234.50` |
|  |  |  | USD (2) | `1,234` | `1.23` |
|  |  |  | BHD (3) | `1 234,567` | `1234.567` |
|  |  |  | BHD (3) | `1234.50` | `1234.500` |
|  |  |  | BHD (3) | `1,234.567` | `1234.567` |
| `en` | `,` | `.` |  |  |  |
|  |  |  | USD (2) | `1234.50` | `1234.50` |
|  |  |  | USD (2) | `1,234.50` | `1234.50` |
|  |  |  | USD (2) | `1234` | `1234.00` |
|  |  |  | USD (2) | `1.234` | `1.23` |
|  |  |  | BHD (3) | `1.234` | `1.234` |
|  |  |  | BHD (3) | `1.2345` | `1.235` |
|  |  |  | BHD (3) | `1.1` | `1.100` |
|  |  |  | BHD (3) | `1234` | `1234.000` |
|  |  |  | BHD (3) | `1,234.5` | `1234.500` |

## Type of change

## Checklist

### DIFF
--- a/packages/react-i18n/src/i18n.ts
+++ b/packages/react-i18n/src/i18n.ts
@@ -62,8 +62,9 @@ export interface TranslateOptions {
 
 // Used for currecies that don't use fractional units (eg. JPY)
 const PERIOD = '.';
-const COMMA = ',';
 const DECIMAL_VALUE_FOR_CURRENCIES_WITHOUT_DECIMALS = '00';
+const REGEX_NON_DIGITS = /\D/g;
+const REGEX_PERIODS = /\./g;
 
 export class I18n {
   readonly locale: string;
@@ -120,7 +121,7 @@ export class I18n {
     this.defaultCurrency = currency;
     this.defaultTimezone = timezone;
     this.pseudolocalize = pseudolocalize;
-    this.onError = onError || defaultOnError;
+    this.onError = onError || this.defaultOnError;
     this.loading = loading || false;
   }
 
@@ -159,7 +160,7 @@ export class I18n {
 
     if (optionsOrReplacements == null) {
       normalizedOptions = {pseudotranslate: pseudolocalize};
-    } else if (isTranslateOptions(optionsOrReplacements)) {
+    } else if (this.isTranslateOptions(optionsOrReplacements)) {
       normalizedOptions = {
         ...optionsOrReplacements,
         replacements,
@@ -236,13 +237,9 @@ export class I18n {
   }
 
   unformatNumber(input: string): string {
-    const {thousandSymbol, decimalSymbol} = this.numberSymbols();
+    const {decimalSymbol} = this.numberSymbols();
 
-    const normalizedValue = normalizedNumber(
-      input,
-      decimalSymbol,
-      thousandSymbol,
-    );
+    const normalizedValue = this.normalizedNumber(input, decimalSymbol);
 
     return normalizedValue === '' ? '' : parseFloat(normalizedValue).toString();
   }
@@ -266,13 +263,9 @@ export class I18n {
   }
 
   unformatCurrency(input: string, currencyCode: string): string {
-    const {thousandSymbol, decimalSymbol} = this.numberSymbols();
+    const {decimalSymbol} = this.numberSymbols();
 
-    const normalizedValue = normalizedNumber(
-      input,
-      decimalSymbol,
-      thousandSymbol,
-    );
+    const normalizedValue = this.normalizedNumber(input, decimalSymbol);
 
     if (normalizedValue === '') {
       return '';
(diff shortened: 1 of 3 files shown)


packages/react-i18n/src/i18n.ts
```diff
@@ -632,64 +625,57 @@ export class I18n {
...
+    let lastIndexOfRealDecimal = lastIndexOfNominalDecimal;
 
-  const isNegative = input.trim().startsWith('-');
-  const negativeSign = isNegative ? '-' : '';
+    // For locales that do not use period as the decimal symbol, users may still input a period
+    // and expect it to be treated as the decimal symbol for their locale.
+    if (
+      decimalSymbol !== PERIOD &&
+      (input.match(REGEX_PERIODS) || []).length === 1 &&
+      this.decimalValue(input, lastIndexOfPeriod).length <= 2
```
movermeyer, line 640: I guess there are no locales that have groupings of less than 3 (for the least significant digits)?

(India uses groups of 2, but only after the first 3 least significant digits)

---

A hypothetical locale that uses period for grouping, digit groups of 2, and a decimal symbol of `,`...

In such a hypothetical, a number of `1234`, represented as `12.34,` would parse to `12.34`

I don't think such a grouping exists though (through my brief internet searching)

movermeyer: I guess it also won't work for currencies that use more than 2 decimal digits (e.g., BHD uses three decimal digits; Bitcoin uses 8 decimal digits), or in scenarios where sub-cent accuracy is needed (e.g., banking typically uses 4 decimal digits)

e.g., A Baharaini-Frenchmen making a transfer of `123.50 BHD` would write `123.500`, but this would parse it as `123000 BHD`.

trishrempel: > I guess there are no locales that have groupings of less than 3 (for the least significant digits)?

Not that I could see, but it's totally possible for me to have missed something.

trishrempel: > e.g., A Baharaini-Frenchmen making a transfer of 123.50 BHD would write 123.500, but this would parse it as 123500 BHD?

Yes; in the French locale, how would we be able to guess if they mean 123500 or 123.500? In cases of ambiguity like this, I think it makes sense to fall back to locale formatting, e.g. `123,500`. 

Unfortunately, I think this would break the idea of being able to call `unformatCurrency` multiple times, because `123,500` would become `123.500`, which would become `123500.00`. I'm not sure how to get around this though, so open to ideas if you have any.

trishrempel: I suppose, I could use the currency's number of decimal places to check for 3 or less instead of 2 or less if the currency's decimal places is 3. 

That would mean though, in the French locale and `BHD` currency scenario, that I would assume `123.500` to be `123.500`, not `123500`. If someone wanted to actually use `.` as a thousands separator when the number of places after the period is 3 or less, they would have to also supply the decimal places (e.g. `123.500,000`).

I don't know if this is any better though or if it just invites more potential problems.

packages/react-i18n/src/i18n.ts
```diff
@@ -632,64 +625,57 @@ export class I18n {
...
-      // where the thousand symbol is a period and a comma appears before a period
-      (thousandSymbol === PERIOD &&
-        lastIndexOfComma > -1 &&
-        lastIndexOfPeriod > -1 &&
-        lastIndexOfPeriod > lastIndexOfComma))
-  ) {
-    lastIndexOfRealDecimal = lastIndexOfPeriod;
-  }
+  private normalizedNumber(input: string, decimalSymbol: string) {
+    const lastIndexOfNominalDecimal = input.lastIndexOf(decimalSymbol);
```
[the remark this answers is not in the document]
trishrempel: Thanks for this - this was left over from the previous iteration of this function.

packages/react-i18n/src/i18n.ts
trishrempel: Note - I just created a second commit where I explore this scenario - please let me know what you think!

> I suppose I could use the currency's number of decimal places to check for 3 or less instead of 2 or less if the currency's decimal places is 3.
> 
> That would mean though, in the French locale and BHD currency scenario, that I would assume 123.500 to be 123.500, not 123500. If someone wanted to actually use . as a thousands separator when the currency's decimal places is 3, they would have to also supply the decimal separator (e.g. 123.500,000).

movermeyer (approved):

Bruno125 (approved):

trishrempel:
Bump please for @Shopify/admin-web-foundations approval!
After this is shipped, could you please set up a beta release for testing?

packages/react-i18n/src/tests/i18n.test.ts
```diff
@@ -718,12 +718,13 @@ describe('I18n', () => {
         expect(i18n.unformatNumber('1.234.567,56')).toBe('1234567.56');
       });
 
-      it('does not treat . as the decimal symbol if , is not used as a decimal symbol', () => {
+      it('treats . as the decimal symbol if . is used as a decimal symbol', () => {
```
caropinzonsilva, line 721: I don't have a lot of context on react-i18n, so I just want to make sure I understand what we are solving.

The tests cases that we are changing here were actually covered before but the assumption was that the result should be different?
So for example. with the previous changes we thought that `unformatNumber('1234.50') === 123450` but in order to make it backwards compatible we need it to be `unformatNumber('1234.50') === 1234.5` instead?

trishrempel: Yes; previously I was making the assumption that `unformatCurrency('1234.50', 'EUR')` should be formatted as `123450.00`, but that's because I wanted to support the Vietnamese locale and I didn't think of a better way to do that while remaining backwards compatible.

So I'm essentially restoring the functionality I [changed in my previous PR](https://github.com/Shopify/quilt/pull/2126/files#diff-a745a62ae45742714f76cc54da399803f886c30142880938f387687063fc7c85L1169), while also being compatible with Vietnamese number formatting (e.g. `unformatCurrency('1.234', 'VND') == '1234.00'`)

packages/react-i18n-universal-provider/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@shopify/react-i18n-universal-provider",
-  "version": "2.1.21",
+  "version": "2.1.22-currency.0",
```
znja, line 3: I'm not sure about the version control here. Why do we need it to be xx-curency.0?

trishrempel: Yes, I made a beta release yesterday to test it, I'm ok with moving it to `2.1.22`

znja (approved):

PR #2140, opened 2022-01-26
[GraphQL-Fixtures] - Replace deprecated faker.random.boolen

attila-berki:
## Type of change

## Checklist


BPScott:
What faker version added `faker.datatype.boolean`? Do we need to update our dependency on faker to match that minimum version or are we already covered?

BPScott (approved):

ecbrodie:
This was an oopsie on my part, that I missed this in #2132 . My bad and thanks for updating @attila-berki . I only detected this minutes ago.

There is one more deprecation I forgot to update: `faker.random.number()` --> `faker.datatype.number()`. Going to fix that now...

PR #2143, opened 2022-01-27
Add export for RouteProps to react-router

robertmeessen:
## Description

Adding type export for `RouteProps` to `react-router` package.

Enabling us to use `import {RouteProps} from '@shopify/react-router'` from projects as the to be able to extend the `Route` component types e.g. for adding a custom `ProtectedRoute` component or similar.

We could also import the types directly from `react-router` as the package is around but the linter would yell at us saying `import/no-extraneous-dependencies` so this PR would be the better way.

## Type of change

## Checklist


ryanwilsonperkin (approved):
LGTM, thanks!

PR #2146, opened 2022-01-30
Fix ctx.set in jest-koa-mocks

### LINKED ISSUE #2067: @shopify/jest-koa-mocks: context.set is not working
## Overview

When trying to upgrade `@shopify/jest-koa-mocks` to `3.0.6` our tests break due to `context.set` not using the built-in function anymore due to this change: https://github.com/Shopify/quilt/commit/b16f54290afbaaea70021e1a457e2fefbe5ac403#diff-f8661886e00c5d85521278962660352c556bf2b223429bbf448c467a3e528ea1R115

I cannot read about the origin of the change since there is no PR nor an issue connected to it. If this is how it's intended to work, I'd be grateful to get some help on how to use the built-in `set` function instead or maybe see this added to the readme.

Could it be an option to make this optional? Adding a `useBuiltinSet` option, or something? 🤔 

Side note: I do not have permissions to add labels to this issue.

## Consuming repo

It's a repo on a Github Enterprise instance.

ide:
## Description
Fixes #2067

Mock context objects from `@shopify/jest-koa-mocks` stopped supporting `ctx.set()` in #2063, which set the implementation of that method to a no-op. This is a useful method that does no I/O and doesn't need to be stubbed.

I suspect the reason this method was stubbed out is because two tests in `react-server` had code like this: `ctx = {...createMockContext(), x: y}`, which breaks the prototype chain of the mocked context and causes `ctx.set` to be undefined. This issue doesn't affect only `set` -- every property higher up on the prototype chain would need to be stubbed or reimplemented which doesn't appear to be the intent of `@shopify/jest-koa-mocks`.

How: This PR stops overwriting `ctx.set` with a no-op and changes the two tests to use `Object.assign` instead of a spread so that the mock contexts preserve their prototype chains.

Test Plan: Ran `yarn test react-server` and verified all tests pass without warning about `ctx.set` not being defined.

## Type of change

- [X] jest-koa-mocks Patch: Bug (non-breaking change which fixes an issue)expected)

## Checklist


packages/jest-koa-mocks/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## Unreleased -->
```
atesgoral, line 8:
```suggestion
## Unreleased
```

PR #2148, opened 2022-01-31
Reorder exports map

heathernfran:
## Description

https://github.com/Shopify/web/issues/56636

Webpack 5 can resolve exports from package.json. Reordering the exports map ensures that when libraries ship with only commonjs, esm compiling will take precedence.

## Type of change

## Checklist

### DIFF
--- a/packages/graphql-config-utilities/CHANGELOG.md
+++ b/packages/graphql-config-utilities/CHANGELOG.md
@@ -9,6 +9,7 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ### Changed
 
+- Reorder exports map to prioritize the `esnext` condition. [[#2148](https://github.com/Shopify/quilt/pull/2148)]
 - Clean test output. [[#2091](https://github.com/Shopify/quilt/pull/2091)]
 
 ## 3.0.6 - 2021-11-22
--- a/packages/koa-shopify-webhooks/CHANGELOG.md
+++ b/packages/koa-shopify-webhooks/CHANGELOG.md
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Changed
+
+- Reorder exports map to prioritize the `esnext` condition. [[#2148](https://github.com/Shopify/quilt/pull/2148)]
 
 ## 4.4.0 - 2022-01-31
 
--- a/packages/storybook-a11y-test/CHANGELOG.md
+++ b/packages/storybook-a11y-test/CHANGELOG.md
@@ -10,6 +10,7 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 ### Changed
 
 - Updated puppeteer to 13.1.3 ([#2149](https://github.com/Shopify/quilt/pull/2149))
+- Reorder exports map to prioritize the `esnext` condition. [[#2148](https://github.com/Shopify/quilt/pull/2148)]
 
 ## 0.4.3 - 2022-01-10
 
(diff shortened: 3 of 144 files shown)


packages/koa-shopify-webhooks/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ### Changed
 
+- Reorder exports map to prioritize the `esnext` condition. [[#2148](https://github.com/Shopify/quilt/pull/2148)]
```
BPScott, line 12: Might need to rebase, as I think @vsumner released a new version of this package this morning.

vsumner (approved):
Love the tests!

tests/consistent-package-json.test.ts
```diff
@@ -139,6 +139,18 @@ packages.forEach(
...
           });
+
+          it('specifies the expected exports', () => {
+            expect(packageJSON.exports).toStrictEqual(
+              expect.objectContaining(expectedPackageJSON.exports),
+            );
+          });
+
+          it('specifies esnext as the first expected export', () => {
+            expect(packageJSON.exports['.'].esnext).toBe(
```
BPScott, line 150: This seems like it asserts the value of a specific key instead of anything to do with the ordering. It also seems to only check the `.` export key rather than all keys.

I would expect that something like the following would cause this test to fail.

```
exports: {
  "./foo" : { 'require': './foo.js', esnext: './foo.esnext' }
}
```

tests/consistent-package-json.test.ts
```diff
@@ -139,6 +139,18 @@ packages.forEach(
           it('specifies the expected types', () => {
             expect(packageJSON.types).toBe(expectedPackageJSON.types);
           });
+
+          it('specifies the expected exports', () => {
```
BPScott, line 143: It is possible for packages to have an exports key that does not contain `"."` - e.g. `graphql-persisted`,  so I don't think this test is correct.

BPScott (commented):
Code changes look good. I think this needs a rebase and an adjustment of the changelogs as @vsumner  has done a release this morning.

The testd doesn't seem like they're testing what we'd hope

dahukish (commented):
Thanks for this @heathernfran. Just need to address @BPScott testing comments.

packages/koa-shopify-webhooks/CHANGELOG.md
heathernfran: Fixed 👍

packages/graphql-config-utilities/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ### Changed
 
+- Reorder exports map to prioritize the `esnext` condition. [[#2148](https://github.com/Shopify/quilt/pull/2148)]
 - Clean test output. [[#2091](https://github.com/Shopify/quilt/pull/2091)]
```
heathernfran, line 13: This hasn't been released, right?

packages/storybook-a11y-test/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 ### Changed
 
 - Updated puppeteer to 13.1.3 ([#2149](https://github.com/Shopify/quilt/pull/2149))
+- Reorder exports map to prioritize the `esnext` condition. [[#2148](https://github.com/Shopify/quilt/pull/2148)]
```
heathernfran, line 13: Fix from earlier merge conflict

tests/consistent-package-json.test.ts
```diff
@@ -139,6 +139,18 @@ packages.forEach(
           it('specifies the expected types', () => {
             expect(packageJSON.types).toBe(expectedPackageJSON.types);
           });
+
+          it('specifies esnext, import, and require as the ordered keys in the exports map', () => {
```
dahukish, line 143: ❤️

packages/graphql-config-utilities/CHANGELOG.md
BPScott: Looks like it hasn't - the last change was just in test files, which are ignored when [determining what to release](https://github.com/Shopify/quilt/blob/main/lerna.json#L7)

BPScott (approved):
Fantastic!

Issue #1970, opened 2021-07-13
Review polyfills based on updated target versions


BPScott:
UPDATE: 2022-05 - We've shipped #2277 which drops support for node 12 and Safari 10, 11 and 12.

Thus the "Update browser targets" and "Update node versions" items have been completed.

We still need to audit our polyfills to remove polyfills that are now supported natively by browsers/node/jest.

---

Recently [we dropped support for Safari/iOS 10, 11 and 12 in `@shopify/browserslist-config`](https://github.com/Shopify/web-configs/issues/271).

To confirm the list of currently supported browsers run `npx browserslist` in an up-to-date copy of the web codebase.

We now target desktop Safari 13.1 and up and iOS Safari v13.4 and up (technically ioOS 13.6 is the minimum supported version but caniuse data says that iOS v13.4 through 13.7 all used the same browser and thus are grouped together). This saves a bunch of transpiling, notably `async`/`await` no longer needs to be transpiled, so we'll have smaller builds.
This begets some tidyiing

### DONE! ~Update browser targets~

- Update `caniuse-lite` to its latest version to ensure we get latest browser version info
- Update `@shopify/browserslist-config` to v3.0.0, so that build target the new change

### Update polyfills

Removing browser support means we can also remove polyfills of features that are now present in all the browsers we target. Identify cases where polyfills can be removed, and remove them.

For instance [intl.pluralRules](https://caniuse.com/intl-pluralrules) and [intersectionObserver](https://caniuse.com/intersectionobserver) is now supported by all browsers we support, and thus we can remove the intl and intersectionobserver polyfills.

While adjusting our polyfills and making a major version bump it is worth claiming we only support newer versions of jest too. Check what features the version of jsdom that jest 26/27 supports, as we may be able to remove some jest-only polyfills. For instance [mutation observer was added in jsdom v13.2](https://github.com/jsdom/jsdom/releases/tag/13.2.0), so the mutation-observer polyfill may be able to be removed.

### DONE! ~Bonus: Update node targets~

We're close to being able to go all in on esmodules for node (this is going to be useful for skn internals). Let's ensure we are ready for that future by bumping the required node versions to the versions where esmodules are considered stable `^12.22.0 || ^14.17.0 || >=16.0.0`. This is a little hairier than the browser version updates as this will force consumers to update their node versions, but I think it is worth forcing that issue and doing the bump at the same time as the browser targets so that we avoid needing another major version somewhere down the line.

[...]

PR #2150, opened 2022-01-31
Update loom to v1

BPScott:
## Description

I keep telling people "look at quilt" for a demo of using loom in a monorepo, but quilt uses an outdated version, so thus it's not a great demo tool.

This PR updates Loom to v1, and update the plugins it uses along the way.

Babel is updated and deduplicated, which explains most of the yarn.lock churn.
Build output is borderline identical - there's a small change to some babel helper functions, but aside from that our code remains functionally identical (and byte-for-byte identical in most cases).

## Type of change

- all: Bug (non-breaking change which fixes an issue)

## Checklist


config/loom/index.ts
```diff
@@ -16,7 +16,19 @@ export function quiltPackage({
   // occur we need to turn some options off
   const polyfillOptions = polyfill ? {} : {useBuiltIns: false, corejs: false};
 
+  const targets = isIsomorphic
+    ? 'extends @shopify/browserslist-config, node 12.14.0'
+    : 'node 12.14.0';
```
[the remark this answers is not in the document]
BPScott: We're retaining 12.14.0 as our minimum supported version, and thus the version we transpile to. (This matches the minimum version for the that we specify we support in the `engines` field in the various `package.json`s).

Increasing this means Would mean we are dropping support for some versions of node, which is a breaking change. We want to do that [at some point](https://github.com/Shopify/quilt/issues/1970) but not right now.

loom.config.ts
```diff
@@ -1,15 +1,18 @@
...
+import {
+  buildLibrary,
+  buildLibraryWorkspace,
+} from '@shopify/loom-plugin-build-library';
 import {eslint} from '@shopify/loom-plugin-eslint';
-import {jest} from '@shopify/loom-plugin-jest';
 import {prettier} from '@shopify/loom-plugin-prettier';
-import {workspaceTypeScript} from '@shopify/loom-plugin-typescript';
+
+import type {} from '@shopify/loom-plugin-jest';
```
heathernfran, line 9: Curious about why it's importing an empty object? Is this intentional?

BPScott: It is intentional, but it is not obvious why.

Loom does some kinda funky things with augmenting objects, that don't quite always work super cleanly.  
This is needed so that typescript realises that `hooks.jestConfig` is a valid key on the hooks object and that it has a particular shape of its own. Without this it thinks that jestConfig is not a valid key.

I'll add a comment explaining this

packages/react-async/loom.config.ts
```diff
@@ -1,10 +1,9 @@
...
+import {createPackage} from '@shopify/loom';
 
 import {quiltPackage} from '../../config/loom';
 
 export default createPackage((pkg) => {
-  pkg.runtimes(Runtime.Browser, Runtime.Node);
-  pkg.entry({root: './src/index'});
-  pkg.entry({name: 'testing', root: './src/testing'});
+  pkg.entry({root: './src/index.ts'});
+  pkg.entry({name: 'testing', root: './src/testing.tsx'});
```
heathernfran, line 7: Also curious about the use case for the entry file being `testing.tsx` instead of `testing.ts`. Wondering if it's related to using the [`jsdom` testing environment](https://github.com/Shopify/quilt/pull/2150/files#diff-e4cab4a4ad90614b98cde109db889b0ac44ebbfdd533fb09c63a33d55c8e4098R13)?

packages/react-effect/loom.config.ts
```diff
@@ -1,10 +1,9 @@
...
+import {createPackage} from '@shopify/loom';
 
 import {quiltPackage} from '../../config/loom';
 
 export default createPackage((pkg) => {
-  pkg.runtimes(Runtime.Browser, Runtime.Node);
-  pkg.entry({root: './src/index'});
-  pkg.entry({name: 'server', root: './src/server'});
+  pkg.entry({root: './src/index.ts'});
+  pkg.entry({name: 'server', root: './src/server.tsx'});
```
heathernfran, line 7: Same question as above about the `.tsx` server file

BPScott: [That's the filename in the repo](https://github.com/Shopify/quilt/blob/main/packages/react-effect/src/server.tsx). It's tsx, not ts because it contains JSX content.

config/loom/index.ts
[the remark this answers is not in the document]
heathernfran: > Increasing this means would mean we are dropping support for some versions of node, which is a breaking change. We want to do that at some point but not right now.

Makes sense, I agree that dropping support is out of scope for this PR.

> The change in the ci config means say we support 12.14, but only run tests against 12.22, which isn't ideal, but I think that is a better state of affairs than being trapped on an old loom version until we make the breaking change to drop support for node 12

Thanks for clarifying/reiterating the node 12.14 and 12.22 use cases 👍

loom.config.ts
heathernfran: Thanks for elaborating and for adding the comment 🎉

PR #2151, opened 2022-02-01
[FED] Add support to namespace on @shopify/react-i18n

pedrodurek:
## Description

Fixes https://github.com/Shopify/shopify-markets/issues/3855

This PR aims to support namespace on `useI18n` hooks, the use of namespace is recommended [here](https://github.com/Shopify/web/blob/main/documentation/decisions/06%20-%20Deprecating%20global%20translation%20files.md#what-are-my-alternatives).

Before:
```tsx
const [i18n] = useI18n();
i18n.translation('Namespace.key1')
i18n.translation('Namespace.key2')
i18n.translation('Namespace.key3')
// or
const [i18n] = useI18n();
const scope = ['Namespace']
i18n.translation('key1', {scope})
i18n.translation('key2', {scope})
i18n.translation('key3', {scope})
```

After:
```tsx
const [i18n] = useI18n({namespace: 'Namespace');
i18n.translation('key1')
i18n.translation('key2')
i18n.translation('key3')
```

## Type of change

## Checklist

### DIFF
--- a/packages/react-i18n/src/hooks.tsx
+++ b/packages/react-i18n/src/hooks.tsx
@@ -1,13 +1,14 @@
 import React from 'react';
 import {useLazyRef} from '@shopify/react-hooks';
 
+import {isObjectEmpty} from './utilities';
 import {I18n} from './i18n';
-import {I18nManager, RegisterOptions} from './manager';
+import {I18nManager, I18nOptions, RegisterOptions} from './manager';
 import {I18nContext, I18nIdsContext, I18nParentContext} from './context';
 
 type Result = [I18n, React.ComponentType<{children: React.ReactNode}>];
 
-export function useI18n(options?: Partial<RegisterOptions>): Result {
+export function useI18n(options: Partial<I18nOptions> = {}): Result {
   const manager = React.useContext(I18nContext);
 
   if (manager == null) {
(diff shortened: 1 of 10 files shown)


GoodForOneFare:
IMO, the way we do i18n needs a rethink, and adding another indirect path for static analysis will make automating our way out of it harder.

That said: I have no power here; we may never get around to a rewrite; "just copy and paste, bro" is easy to say when you're not living with the problem.

BPScott:
@TheMallen is looking at alternative ideas to i18n, and probably has some thoughts regarding how this would impact any evolution.

I think I agree with @GoodForOneFare - adding another way for namespaces to be specified has the potential to make any future wholesale reworking harder, and not accepting this PR and getting you to sit tight while we get a better handle on how i18n shall look in the future is might be best.

pedrodurek:
Hey @GoodForOneFare and @BPScott, thanks for the feedback.

What exactly do you guys have in mind for the future of the i18n lib? If the idea is to follow the same steps as the [official react-i18next](https://github.com/i18next/react-i18next), this PR won't make the migration harder as the official one also relies on namespaces.

If you guys want, I may be able to help you. I've been contributing to i18next organization for almost two years now.

BPScott:
Following a second PR that adds this behaviour and a [small revival of this discussion](https://github.com/Shopify/quilt/pull/2305#issuecomment-1159096835), I've made a call.

Merging `scope/namespace` support creates additional codepaths that we don't want to support and we think will impede future investigations into iterative improvements by making migration to new approaches harder. As such, we are not interested in accepting this PR at this time.

The team currently does not have a roadmap for when those investigations would occur as our focus currently lies elsewhere.


Thank you for thinking about this @pedrodurek and for considering adding behaviour to the shared library and I hope this rejection doesn't put you off working with us in the future.

PR #2152, opened 2022-02-01
graphql-fixtures: Load a single faker locale and improve test suites load time

sbalay:
## Description

By default faker will load a bunch of locales when you require the default export of the library.
Loading a single locale makes noticeable improvements in the load time of test suites

This PR sets `en` as the default locale for the library, and skips loading all the other locales available in `faker`.

### Scope

This only updates `graphql-fixtures` because it's what I managed to test in Shop, but we probably should apply this change to other libs of quilt using faker default export unless we intentionally want to load all locales for some reason.

Happy to include those changes in this PR or open a follow up PR for doing so.

### Shop

In Shop, [this PR](https://github.com/Shopify/arrive-client/pull/14446) **reduced the test suite time from ~360s to ~280s** in local environments.

## Type of change

Not so sure here... if for some reason someone is relying on graphql-fixtures loading faker's locales then this should be considered a breaking change. This would be very rare though, and we might want to consider this change a minor instead?

- [x] `graphql-fixtures` Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

## Credits

@notjosh for the research and finding of faker affecting tests load time


sbalay:
> Great find! If you don't want to take on updating the other uses of faker, could you open an issue please?

Hey, happy to do the update! 

I'll add the changes to this PR and request the review again, unless you think it's better to do that in a separate PR?

Also, regarding type of change, any thought on this being a breaking change vs minor?

sbalay:
@vsumner changed other places as well.

Another question, regarding changelog, I was going to add an entry in the changelog of the affected packages... but then decided not do so, as they don't really expose functionality based on faker. graphql-fixtures seems to be the only package to do so.

vsumner:
> @vsumner changed other places as well.
> 
> Another question, regarding changelog, I was going to add an entry in the changelog of the affected packages... but then decided not do so, as they don't really expose functionality based on faker. graphql-fixtures seems to be the only package to do so.

Makes sense.

packages/graphql-fixtures/CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## Unreleased -->
 
+## Unreleased
+
+### Changed
+
+- Load a single locale of faker (en) instead of loading all locales.
```
BPScott, line 14:
```suggestion
- Load a single locale of faker (en) instead of loading all locales. [[#2152](https://github.com/Shopify/quilt/pull/2152)]
```

BPScott (approved):
Looks good to me.

As you note, `graphql-fixtures` is the only package that depends upon faker as part of its public API, all the other changes are to test files only, which don't require a changelog entry.

I'm tempted to say `minor` version for this - I think it's highly unlikely that anybody has been using faker for gql data with non-english locales, and it was never something we consciously supported in the first place.

vsumner (approved):

sbalay:
@BPScott do you know if there's anything wrong with the `CLA` check? Seems to be hanged in `Waiting for status to be reported`

PR #2155, opened 2022-02-02
Re-export faker from graphql-fixtures

### LINKED ISSUE #2154: graphql-fixtures should re-export faker instance
## Overview

`graphql-fixtures` should re-export the `faker` instance that it imports in the fill module so that it can be used by consumers to ensure they're using the same instance which has been seeded with the same value.

This would also resolve any regressions related to https://github.com/Shopify/quilt/issues/2006

## Motivation

`graphql-fixtures` imports faker (now from `@faker-js/faker`) and uses it to seed a value based on the keypath so that certain seeded values will be consistent throughout the filled graphql object:

https://github.com/Shopify/quilt/blob/490472f04169c1c86612a2ff82936d08fe07b338/packages/graphql-fixtures/src/fill.ts#L289-L303

This allows us to ensure that if a particular object type is filled multiple times without specifying a specific value like "id" then that value will be returned consistently each time. When the value, especially an "id" value, is different for the same object, it confuses some graphql libraries like Apollo and prevents them from returning the correct value.

When users of this library attempt to use faker themselves to provide a filler function, they'll find that unless they use the _exact_ same path for import `@faker-js/faker` as `graphql-fixtures`, and unless they have the same version installed, they'll end up with a difference instance of the `Faker` object which does not have the seed provided. This results in their faked data not being seeded the same way, and resulting in different values. We can work around this by re-exporting the `faker` instance so that it can be re-used directly.

## Example

Here's an example of such a case that is currently broken. `graphql-fixtures` recently updated so that:

```js
import faker from '@faker-js/faker';
```

became

```js
import faker from '@faker-js/faker/locales/en';
```

If a consuming library had a filler like so:

```js
import faker from '@faker-js/faker';
const fill = createFiller(schema, {
  resolvers: {
    ID: (_, {parent}) => `gid://${parent.name}/${faker.datatype.number}`,
  },
});
```

they would now find that their tests that reuse an object with the same ID in multiple queries is now failing. What happens is that the default export from `@faker-js/faker` is not the same object as is default exported from `@faker-js/faker/locales/en`. Since the consumer is importing a different object, it hasn't been seeded, and so it won't result in the consistent value for this ID.

## Checklist

ryanwilsonperkin:
## Description

Fixes #2154

Allows consumers to consume the exact faker instance that we're seeding to ensure that they will be consistent with our package. This makes it easier for consumers to use faker in their filler methods so that they don't need to match our exact version in their dependencies.

## Type of change

- [x] `graphql-fixtures` Minor: New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/packages/graphql-fixtures/CHANGELOG.md
+++ b/packages/graphql-fixtures/CHANGELOG.md
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Changed
+
+- Re-export the instance of faker used in graphql-fixtures so that it can be leveraged easily by consumers [[#2155]](https://github.com/Shopify/quilt/pull/2155)
 
 ## 1.3.0 - 2022-02-01
 
--- a/packages/graphql-fixtures/README.md
+++ b/packages/graphql-fixtures/README.md
@@ -57,6 +57,20 @@ If you need to have the filled object immediately, you can invoke `fill` as foll
 const data = fill(myQuery, data)({query: myQuery});
 ```
 
+#### Using faker
+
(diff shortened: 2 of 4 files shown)


BPScott (approved):
Fantastic!

PR #2156, opened 2022-02-03
 Delete empty magic-entries-webpack-plugin and rpc packages

BPScott:
## Description

These packages have been emptied, and that empty version has been published, so now they can be cleared away.


caropinzonsilva (approved):
Looks good! 🧹 ✨

PR #2161, opened 2022-02-07
[graphql-fixtures] Remove unused call to faker.seed()

notjosh:
## Description

Hi! We're seeing _lots_ of calls to `faker.seed()` when using this library, and it turns out (predictably) that's because we call it a lot in some situations! 😱

The problem here, is that `init_genrand()` is quite an expensive operation ([source](https://github.com/faker-js/faker/blob/f1884becbc108c3ddbad952d879cec0fc5196e08/src/vendor/mersenne.ts#L104-L125)). It was showing to be using 2-4% of time spent in test suites!

As a starting point, this PR cuts the calls to `init_genrand()` to around 50%, without any noticeable negative side effect.

Discussion: For future, we could replace the Mersenne Twister implementation for less cryptographically secure but cheaper + good enough for our needs. For example, [an implementation](https://github.com/chrisakroyd/random-seedable/blob/b23fce2e77cc17c3bf3964688e80a5456045ab95/src/lcg.js#L46-L48) of an LCG with an essentially free seeding function. I have a patch for this that I used in profiling to swap out the implementation in Faker.js, but a) it's obviously fragile to patch a core function like that, and b) it _will_ break existing tests relying on specific seeded output.

(This ventures deep into "should we even still use Faker?" territory though, but it's food for thought)

## Type of change

- [x] `graphql-fixtures` Patch: Bug (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/packages/graphql-fixtures/CHANGELOG.md
+++ b/packages/graphql-fixtures/CHANGELOG.md
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Changed
+
+- Removes call to `faker.seed()` that uses `Math.random()` as a seed value [[#2161]](https://github.com/Shopify/quilt/pull/2161)
 
 ## 1.4.1 - 2022-02-09
 
--- a/packages/graphql-fixtures/src/fill.ts
+++ b/packages/graphql-fixtures/src/fill.ts
@@ -299,9 +299,7 @@ function withRandom<T>(keypath: FieldDetails[], func: () => T, seedOffset = 0) {
       ),
     ) + seedOffset,
   );
-  const value = func();
-  faker.seed(Math.random() * 10000);
-  return value;
+  return func();
 }
 
 function createValue<T>(


atesgoral:
> should we even still use Faker?

My first instinct is, no.

I don't see why we want unpredictable/nondeterministic test fixtures.

atesgoral:
What I implicitly (or in my own head) meant was: Even with seeding, tests might be prone to breaking the moment there's a slight tweak to how the PRNG or data synthesis works. I hard-code the test input and outputs in tests.

notjosh:
> What I implicitly (or in my own head) meant was: Even with seeding, tests might be prone to breaking the moment there's a slight tweak to how the PRNG or data synthesis works. I would hard-code the input and outputs in tests.

Ohhh right, yes. If we change the PRNG, tests will (and do!) definitely break, so I haven't touched that. This change doesn't change any of the output used for fixtures at all, and gets re-seeded with a deterministic seed every time it's called.

I'd imagine the only way this could break tests would be extremely specific, i.e. tests that a) mock `Math.random()` for some reason (maybe to imitate seeding the value), _and_ b) rely specifically on the instance of Faker set up with this module (which would be left with a different seed after execution).

And for completeness: other libs pulling in Faker wouldn't be impacted, as they get (re-)seeded separately in their own instance.

BPScott:
@ryanwilsonperkin you've been knee deep in [faker seeding funkyness](https://github.com/Shopify/quilt/pull/2155). Got any thoughts?

ryanwilsonperkin:
Copying a couple comments over from slack:

This might break the odd thing in web, but probably not too to much because we already came across and fixed a few of the places where we were asserting on specific values that we expected to come back from the seed.

We’d still need some determinism within a given test though, I’m not sure we can remove all the calls to seed. Simplest example to demonstrate would be if you had multiple useQuery calls within the same component and you expected them to each resolve (potentially nested) a Widget type. If you don’t provide an id value in your partial fill argument, then you’ll get a deterministic id value for the Widget in each of the query responses. If we remove the seeding, then this would result in each one getting a different id, which then breaks Apollo (it returns undefined)

notjosh:
^Totally agreed, but this is only removing _one_ of the calls to seed which uses `Math.random()` as the seed (so not deterministic anyway), so the `withRandom` method will still be deterministic in the same way it was before

PR #2163, opened 2022-02-08
use typescript native types

m4thieulavoie:
## Description

This PR is a first of two steps to get rid of `Omit`, `Arguments`, `ConstructorArguments` and `ThenType` in favour of their native Typescript counterpart.

Fixes (issue #)

## Type of change

 - `react-async` Patch: No impact `Omit` usage change (non-breaking change which fixes an issue)
 - `react-form-state` Patch: No impact `Omit` usage change (non-breaking change which fixes an issue)
 - `react-graphql` Patch: No impact `Omit` usage change (non-breaking change which fixes an issue)
 - `react-testing` Patch: No impact `Omit` and `Arguments` usage change (non-breaking change which fixes an issue)
 - `useful-types` Patch: No impact `ArgumentAtIndex` and `FirstArgument` change to accept `Parameters`  (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/packages/react-html/CHANGELOG.md
+++ b/packages/react-html/CHANGELOG.md
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Changed
+
+- Remove dependency on `@shopify/useful-types` by using built-in types. [[#2163](https://github.com/Shopify/quilt/pull/2163)]
 
 ## 11.1.15 - 2022-02-09
 
--- a/packages/react-server/CHANGELOG.md
+++ b/packages/react-server/CHANGELOG.md
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Changed
+
+- Remove dependency on `@shopify/useful-types` by using built-in types. [[#2163](https://github.com/Shopify/quilt/pull/2163)]
 
 ## 2.1.5 - 2022-02-09
 
--- a/packages/graphql-testing/CHANGELOG.md
+++ b/packages/graphql-testing/CHANGELOG.md
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Changed
+
+- Remove devDependency on `@shopify/useful-types` by using built-in types. [[#2163](https://github.com/Shopify/quilt/pull/2163)]
 
 ## 5.1.6 - 2022-02-09
 
--- a/packages/graphql-testing/package.json
+++ b/packages/graphql-testing/package.json
@@ -38,7 +38,6 @@
     "jest-matcher-utils": "^26.6.2"
   },
   "devDependencies": {
-    "@shopify/useful-types": "^3.1.0",
     "graphql-typed": "^1.1.3"
   },
   "files": [
--- a/packages/graphql-testing/tsconfig.json
+++ b/packages/graphql-testing/tsconfig.json
@@ -11,5 +11,5 @@
     "./src/**/*.tsx"
   ],
   "exclude": ["**/test/**/*", "**/tests/**/*"],
-  "references": [{"path": "../graphql-typed"}, {"path": "../useful-types"}]
+  "references": [{"path": "../graphql-typed"}]
 }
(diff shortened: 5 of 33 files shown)


BPScott:
As we're now no longer depending on `@shopify/useful-types` in several places, we might have packages that no longer depend upon `@shopify/useful-types`. For instance it looks like that was the only usage in react-form-state.

Can you identify these cases where a package no longer depends upon `@shopify/useful-types`, and then in those packages remove the dependency from package.json and the tsconfig.json.

BPScott:
MaybeFunctionReturnType looks very similar to ReturnType, I wonder if that's a meaningful difference, and thus if it's worth removing MaybeFunctionReturnType?

```
export type MaybeFunctionReturnType<T> = T extends (...args: any[]) => infer U
  ? U
  : never;


type ReturnType<T extends (...args: any) => any> = T extends (...args: any) => infer R ? R : any;
```

m4thieulavoie:
> As we're now no longer depending on `@shopify/useful-types` in several places, we might have packages that no longer depend upon `@shopify/useful-types`. For instance it looks like that was the only usage in react-form-state.
> 
> Can you identify these cases where a package no longer depends upon `@shopify/useful-types`, and then in those packages remove the dependency from package.json and the tsconfig.json.

That's done, but it affected some unit tests. Let me know if I handled that properly!

m4thieulavoie:
> MaybeFunctionReturnType looks very similar to ReturnType, I wonder if that's a meaningful difference, and thus if it's worth removing MaybeFunctionReturnType?
> 
> ```
> export type MaybeFunctionReturnType<T> = T extends (...args: any[]) => infer U
>   ? U
>   : never;
> 
> 
> type ReturnType<T extends (...args: any) => any> = T extends (...args: any) => infer R ? R : any;
> ```

Tried to do it yesterday, but I guess I was tired, but [browsing reddit](https://www.reddit.com/r/typescript/comments/smx08y/extract_class_method_arguments_utility_type_in_ts/) I found _exactly_ what I was missing 😅  that should be done!

m4thieulavoie:
> I've pushed a commit that reverts the typescript-project-references.test.ts changes and applies the proper fix.
> 
> That test asserts that "if you use a dependency on another package in this monorepo in a package's package.json, then there should be a corresponding entry in that package's tsconfig.json reference field" (as this repo uses project references - one ts project per package)
> 
> You've removed react-server's dependency on useful-types in the package.json, but you did not remove the reference to the useful-types folder in react-server's tsconfig.json. And thus the test fails, as there should be mention of a dependant package (or in this case no mention of a package that is not depended upon) in both package.json and tsconfig.json.
> 
> The fix is not to modify the typescript-project-references.test.ts, but to update the tsconfig.json files, removing the references to useful-types in the packages that no longer depend upon useful-types

Nice! I totally missed that packages had internal `tsconfig.json`, that's my bad!

BPScott (commented):
Just pushed up another commit where I spotted dangling usages of FirstArgument and Arguments.

- Packages that no longer depend upon useful-types should have an entry in their CHANGELOG.md. Remember to reference this PR, like the other changelog entries
- I reckon we can revert the change in useful-types - we're due to delete these types once we update this PR anyway so I don't think there's much value in adjusting them now.

I reckon we should revert the

packages/useful-types/src/types.ts
```diff
@@ -2,10 +2,13 @@ export type ThenType<T> = T extends Promise<infer U> ? U : T;
 
 export type Arguments<T> = T extends (...args: infer U) => any ? U : never;
 export type ArgumentAtIndex<
-  Func,
```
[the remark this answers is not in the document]
m4thieulavoie: Agreed, just reverted it

packages/useful-types/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 ### Added
 
 - Added `PartialSome` and `RequireSome` types to set specified fields of a property to optional or required
+- Use Typescript native types and pave the ground to deprecate legacy types
```
BPScott, line 13: This can be removed, as we're no longer touching useful-types

BPScott (commented):
Code changes look good to me.

Can you add changelog entries to the packages that no longer depend upon useful-types

packages/useful-types/CHANGELOG.md
m4thieulavoie: Done

packages/react-html/CHANGELOG.md
```diff
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Changed
+
+ - Remove dependency on `@shopify/useful-types` by using built-in types. [[#2163](https://github.com/Shopify/quilt/pull/2163)]
```
BPScott, line 12:
```suggestion
- Remove dependency on `@shopify/useful-types` by using built-in types. [[#2163](https://github.com/Shopify/quilt/pull/2163)]
```

packages/react-server/CHANGELOG.md
```diff
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Changed
+
+-Remove dependency on `@shopify/useful-types` by using built-in types. [[#2163](https://github.com/Shopify/quilt/pull/2163)]
```
BPScott, line 12:
```suggestion
- Remove dependency on `@shopify/useful-types` by using built-in types. [[#2163](https://github.com/Shopify/quilt/pull/2163)]
```

PR #2164, opened 2022-02-08
Fix spin snapshots and boostrap the environment

vsumner:
## Description
This PR fixes spin snapshots and bootstraps the environment

## How can I tophat this?
```
spin up quilt@fix-spin-snapshots --no-snapshots;
spin shell
journalctl -fu spin-environment@shopify--quilt.service
```

Wait until you see:
```
Feb 08 18:47:49 systemd[1]: Finished [Spin] Compile environment {shopify--quilt}.
```


GoodForOneFare (approved):
> Snapshots have no value for quilt.

That's absolutely not true, and it's somewhat disappointing that multiple admin FED leads have expressed this opinion.  If you want new developers to quickly understand and contribute to these libraries, letting them skip a lengthy (disk eating) `yarn install`, and warming the build/lint caches should be a priority.

fwiw, the reason these snapshots exist is because **I** hack on these libraries in Spin, and our automation picked up on that fact.

vsumner:
> > Snapshots have no value for quilt.
> 
> That's absolutely not true, and it's somewhat disappointing that multiple admin FED leads have expressed this opinion. If you want new developers to quickly understand and contribute to these libraries, letting them skip a lengthy (disk eating) `yarn install`, and warming the build/lint caches should be a priority.
> 
> fwiw, the reason these snapshots exist is because **I** hack on these libraries in Spin, and our automation picked up on that fact.

Thank you for educating me on the value they can provide for a library. Could you recommend a better setup so that quilt can take advantage of these?

ryanwilsonperkin:
Agreed! I've used spin in the past for editing quilt and have been trying to avoid `dev` as much as possible lately 🙂

ryanwilsonperkin (approved):

PR #2162, opened 2022-02-07
⬆️ Update Puppeteer to version 13.2.0

emileber:
## Description

Fixes #2160 

`storybook-a11y-test` has recently [updated Puppeteer from `10.4.0` to `13.1.2`](https://github.com/Shopify/quilt/pull/2149). Puppeteer `12.0.0` updated Chromium to version 94, which had issues when run inside a container, which affected CIs. Puppeteer [released version `13.2.0` hours ago](https://github.com/puppeteer/puppeteer/releases/tag/v13.2.0) with Chromium 99 which doesn't have the issue of v94.

This PR updates `puppeteer` to version `13.2.0`.

## Type of change

- [x] `storybook-a11y-test` Patch: Bug (non-breaking change which fixes an issue)
- [ ]  Minor: New feature (non-breaking change which adds functionality)
- [ ]  Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

- [x] I have added a changelog entry, prefixed by the type of change noted above (Documentation fix and Test update does not need a changelog as we do not publish new version)

PR #2166, opened 2022-02-08
🔧 Deduplicate dependencies with yarn-deduplicate

emileber:
## Description

Following [a comment on PR 2162](https://github.com/Shopify/quilt/pull/2162#pullrequestreview-876230059) that was made a minute too late, I created this PR which deduplicates all dependencies.

## Type of change

- [x] (All packages possibly?) Patch: Bug (non-breaking change which fixes an issue)

## Checklist


BPScott (approved):
Thanks!

emileber:
I'm not sure which packages it can affect, but then again, the risk is minimal. Should we take additional steps to be safe?

PR #2178, opened 2022-02-16
Expose address formatting functions publicly so they can be used synchronously

emileber:
## Description

These changes were recycled from #1843, which were abandoned at the time since the motivation behind the changes had changed, but I believe that the changes themselves are still relevant regardless.

When formatting address and rendering an address edition form, there's an obligatory async operation from the `AddressFormatter` API. To overcome this issue, we've started to cache the countries and ordered fields, but can't really cache the formatting operation itself, so we still suffer a flash of skeleton.

Exposing a synchronous API would really simplify the implementation.

## Type of change

## Checklist

### DIFF
--- a/packages/address/CHANGELOG.md
+++ b/packages/address/CHANGELOG.md
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Added
+
+- Exposes `formatAddress`, `buildOrderedFields` formatting functions publicly so they can be used synchronously [[#2178](https://github.com/Shopify/quilt/pull/2178)]
 
 ## 3.0.16 - 2022-02-25
 
--- a/packages/address/src/AddressFormatter.ts
+++ b/packages/address/src/AddressFormatter.ts
@@ -1,14 +1,7 @@
 import {Address, FieldName, Country} from '@shopify/address-consts';
 
-import {renderLineTemplate, FIELDS_MAPPING} from './utilities';
+import {formatAddress, buildOrderedFields} from './format';
 import {loadCountry, loadCountries} from './loader';
-
-const FIELD_REGEXP = /({\w+})/g;
-const LINE_DELIMITER = '_';
-const DEFAULT_FORM_LAYOUT =
-  '{firstName}{lastName}_{company}_{address1}_{address2}_{city}_{country}{province}{zip}_{phone}';
-const DEFAULT_SHOW_LAYOUT =
-  '{lastName} {firstName}_{company}_{address1} {address2}_{city} {province} {zip}_{country}_{phone}';
 
 const ORDERED_COUNTRIES_CACHE: {
   [locale: string]: Country[];
@@ -55,10 +48,7 @@ export default class AddressFormatter {
    */
   async format(address: Address): Promise<string[]> {
     const country = await this.getCountry(address.country);
-    const layout = country.formatting.show || DEFAULT_SHOW_LAYOUT;
-    return layout
-      .split(LINE_DELIMITER)
-      .map((fields) => renderLineTemplate(country, fields, address).trim());
+    return formatAddress(address, country);
   }
 
   /* Returns an array that shows how to order fields based on the country code
@@ -76,17 +66,7 @@ export default class AddressFormatter {
   async getOrderedFields(countryCode: string): Promise<FieldName[][]> {
     const country = await this.getCountry(countryCode);
 
-    const format = country ? country.formatting.edit : DEFAULT_FORM_LAYOUT;
-
-    return format.split(LINE_DELIMITER).map((fields) => {
-      const result = fields.match(FIELD_REGEXP);
-      if (!result) {
-        return [];
-      }
-      return result.map((field) => {
-        return FIELDS_MAPPING[field];
-      });
-    });
+    return buildOrderedFields(country);
   }
 
   private loadCountryFromCache(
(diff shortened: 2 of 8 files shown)


packages/address/src/AddressFormatter.ts
```diff
@@ -1,15 +1,8 @@
...
 
-import {renderLineTemplate, FIELDS_MAPPING} from './utilities';
+import {formatAddress, buildOrderedFields} from './utilities';
 import {loadCountry, loadCountries} from './loader';
 
-const FIELD_REGEXP = /({\w+})/g;
-const LINE_DELIMITER = '_';
-const DEFAULT_FORM_LAYOUT =
-  '{firstName}{lastName}_{company}_{address1}_{address2}_{city}_{country}{province}{zip}_{phone}';
-const DEFAULT_SHOW_LAYOUT =
```
emileber, line 10: These were duplicated in utilities. Now that the actual logic lives in utilities, they're no longer needed here.

packages/address/src/tests/utilities.test.ts
```diff
@@ -1,10 +1,18 @@
...
+import {
+  buildOrderedFields,
+  formatAddress,
+  renderLineTemplate,
+} from '../utilities';
 
 const Canada = fixtures.country.EN.data.country;
+const Japan = fixtures.countries.JA.data.countries.find(
+  ({code}) => code === 'JP',
+)!;
```
emileber, line 13: I'm using a non-null assertion here, though I'm not a fan of this, since it's tests based on fixtures, I figured the risk was minimal.

packages/address/src/tests/utilities.test.ts
```diff
@@ -1,10 +1,18 @@
...
 
-const address = {
+const address: Address = {
```
emileber, line 15: Explicitly typing literals ensure that no extra properties are defined, like typos.

BPScott: We could also do `Parameters[typeof formatAddress][0]` to avoid the extra import. This also describes how this it is intending to be used.

packages/address/CHANGELOG.md
```diff
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Added
+
+- Exposes formatting functions publicly so they can be used synchronously [#1843](https://github.com/Shopify/quilt/pull/1843) [#2178](https://github.com/Shopify/quilt/pull/2178)
```
BPScott, line 12: A consumer will read "new functions exposed" and will then immediately think "neat, what are they" - let's just tell them rather than making them hunt:

```suggestion
- Exposes `formatAddress`, `buildOrderedFields` formatting functions publicly so they can be used synchronously [#1843](https://github.com/Shopify/quilt/pull/1843) [#2178](https://github.com/Shopify/quilt/pull/2178)
```

packages/address/src/utilities.ts
```diff
@@ -69,3 +75,24 @@ function getZone(zones: Zone[], zoneCode: string): Zone {
     }
   );
 }
+
+export function formatAddress(address: Address, country: Country): string[] {
```
BPScott, line 79: `utilities` feels like a file that should just be for internal usage rather than exposed publicly. What do you reckon to keeping these functions in AddressFormatter.tsx, and exposing them from there - that way AddressFormatter.tsx is the public api and utilities.ts is just for private stuff.

BPScott (commented):
Thanks for the PR!

Code behaviour looks good to me at the macro level. A few small comments inline, the main one being `utilities` feels like a private only name, once we add publicly exposed things in there it's not clear what should be public and what should be private. Let's sidestep that by keeping the public API all in AddressFormatter.tsx

packages/address/src/tests/utilities.test.ts
emileber: I'm not sure I see the benefit 🤔 

To me, it feels more confusing and tightly coupled than the simple independent fixture object it is supposed to be. There's also a similar fixture object typed using `Address` in the `AddressFormatter.test.ts` file.

If there's any concern I did not anticipate, should we change both test files?

packages/address/src/utilities.ts
BPScott: All fair comments, I've got no strong opinion about this and I'm not going to force it. I'm happy for this to go out as-is. Thanks for the viewpoint.

BPScott (approved):
Thanks for the replies. I didn't have a strong opinion on what I wrote. I'm happy for this to go out without those nitpicky thoughts.

Rebase atop main and merge when you're ready.

packages/address/src/utilities.ts
emileber: Thanks for the thorough review, your comments are valuable as they bring other view points on the changes. 😃 

I think I'll split with a new `format.ts` file, and also split the tests as well, to keep it clear and scoped.

packages/address/CHANGELOG.md
```diff
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Added
+
+- Exposes `formatAddress`, `buildOrderedFields` formatting functions publicly so they can be used synchronously [#1843](https://github.com/Shopify/quilt/pull/1843) [#2178](https://github.com/Shopify/quilt/pull/2178)
```
BPScott, line 12:
```suggestion
- Exposes `formatAddress`, `buildOrderedFields` formatting functions publicly so they can be used synchronously [[#2178](https://github.com/Shopify/quilt/pull/2178)]
```

#1843 wasn't merged so I don't think it's useful to mention in the changelog

BPScott (approved):
Great idea with moving these functions into format.ts.

On small nitpick with the changelog line and this is good to go.

PR #2179, opened 2022-02-16
Add resetCache method to AddressFormatter and improve address mocks

emileber:
## Description

To avoid side-effects that the cache can have in different situations, like testing, I've added a way to reset the cache.

### Tests and fixtures

While making sure this package tests didn't have any side-effects, I discovered that there were in fact side-effects, where the cache was used in place of the mocks some times, meaning that some tests would fail in isolation while they passed when the whole suite was run.

To _address_ this (pun intended), I changed how the address mocks return the countries fixtures and errors so that now, it supports all countries in the FR, JA, and EN locales.

## Type of change

- [x] `@shopify/address-mocks` Minor: New feature (non-breaking change which adds functionality)

## Checklist


emileber:
Though it is not required, I'd like to merge https://github.com/Shopify/quilt/pull/2178 first.

packages/address/src/AddressFormatter.ts
```diff
@@ -1,20 +1,22 @@
...
-const DEFAULT_SHOW_LAYOUT =
-  '{lastName} {firstName}_{company}_{address1} {address2}_{city} {province} {zip}_{country}_{phone}';
-
-const ORDERED_COUNTRIES_CACHE: {
+let countriesCache: {
   [locale: string]: Country[];
 } = {};
 
 export default class AddressFormatter {
+
```
emileber, line 11: This line conflicts with `prettier` rules, but is required by `lines-around-comment` rule...

packages/address/src/AddressFormatter.ts
```diff
@@ -10,11 +10,19 @@ const DEFAULT_FORM_LAYOUT =
...
   [locale: string]: Country[];
 } = {};
 
 export default class AddressFormatter {
+  /**
+   * Useful in tests or any situation where the cache has undesirable
+   * side-effects.
+   */
+  static resetCache() {
+    countriesCache = {};
```
BPScott, line 23: Rather than recreate the whole object you could empty it out:

```
for (var key in ORDERED_COUNTRIES_CACHE) {
  if (ORDERED_COUNTRIES_CACHE.hasOwnProperty(key)) {
    delete ORDERED_COUNTRIES_CACHE[key];
  }
}
```

That would allow it to continue to be a `const` and you wouldn't need to rename the variable

packages/address/src/AddressFormatter.ts
```diff
@@ -89,11 +97,13 @@ export default class AddressFormatter {
...
-  private loadCountryFromCache(
-    countryCode: string,
-  ): Country | undefined | null {
-    if (ORDERED_COUNTRIES_CACHE[this.locale]) {
-      return ORDERED_COUNTRIES_CACHE[this.locale].find((country) => {
+  /**
+   * Synchronous API to consult the cached countries before having to resort to
+   * fetching.
+   */
+  loadCountryFromCache(countryCode: string): Country | undefined | null {
```
BPScott, line 104: I don't understand why we'd need to make this public. What is the the use case for needing to access the value from the cache (and returning null if its not found in the cache is empty) instead of calling `getCountry()`

BPScott (commented):
Can you please expand more on the use-cases for this?

I can kinda understand the need for clearing the cache, but I don't understand why you'd need to access the value from `loadCountryFromCache` (and returning null if its not found in the cache is empty) instead of calling `getCountry`?

packages/address/src/AddressFormatter.ts
emileber: Returning `null` or `undefined` was left as-is, not to change the internal API. I guess `null` could mean the cache isn't set and `undefined` could mean the `countryCode` wasn't found. This could be either documented or changed so that it's clearer.

As for the motivation behind the change: since `getCountry` is async, it forces an async behavior all the time even though a sync usage could be possible after the first load. This forces a flash of skeleton in React components even when all the data is already cached.

emileber:
While using the new cache reset in this package tests, I discovered that some tests now fail, so I'll investigate to make sure they're not passing because of any side-effects of the cache.

packages/address/src/AddressFormatter.ts
BPScott: Converting it to be a Map sounds like a great idea. I like it.

packages/address/src/AddressFormatter.ts
BPScott: > Returning null or undefined was left as-is, not to change the internal API. 

I get that, but I'm not sure why its useful to expose this function. Exposing this publicly we have two mechanisms of getting a country - getCountry and  loadCountryFromCache - one of which shall always return correct data, and the other is a footgun that won't work unless a cache is already primed. That's difficult to understand and teach and leaves room for errors - people will become confused why sometimes their content load and sometimes it doesn't if they use loadCountryFromCache.


Can you please expand on why you believe you need this. I wonder if there's is a solution like "we let you prime the AddressFormatter cache" that is perferable.

emileber: Totally valid concerns you're bringing up! 

I guess the name of the function itself is really misleading as it doesn't (never did) load anything, it only _looks for_ the country inside the cache, so renaming it like `getCountryFromCache` might help clear that up a little.

## The situation: Async side-effects

A promise is a side-effect by definition in React, which leads to an in-between state while waiting, even though the data is cached for most of the app lifetime.

Here's a super simplified flow of what's happening:

1. Component renders a skeleton
2. An effect uses the async `format` function and saves the result into a state.
3. Finally replace the skeleton with the formatted address state.
4. User event triggers a change in the address, go back to step 1.

## The goal: Maximize sync flow

To avoid this forced flash of skeleton, I was in the process of managing our own cache on top of the existing one (which led me to expose a synchronous API for this package #2178), but I figured that having read access to the `AddressFormatter` cache would be best. With the cache, we can then default to a sync behavior whenever available and only fallback to async if necessary.

1. Use the sync `formatAddress` right away if the country is available in the cache.
2. If not,
   1. render skeleton
   2. let the effect load the country so that every sub-sequent renders do not need to go through another async effect.
   3. Replace skeleton with formatted address.

> people will become confused why sometimes their content load and sometimes it doesn't if they use loadCountryFromCache.

I believe the `...FromCache` part is a clear enough distinction. It's meant for an optimization use-case, if a dev only cares about getting a country, `getCountry` would be the obvious choice.

---

Priming wouldn't help since we would still _suffer_ from the async behavior we're trying to solve. And I don't want to expose write access to the cache because there's no reason to prime the cache ourselves, nor would it be safer to do so.

emileber: After a quick meeting, here's what I'll be doing about this particular change:
- I'll revert exposing `loadCountryFromCache` to keep the API internal
- I'll explore a way to make a React binding on our side to optimize rendering synchronously whenever possible.

packages/address-mocks/src/index.ts
```diff
@@ -1,48 +1,59 @@
...
-            return (
-              body.operationName === operationName &&
-              body.variables.locale === locale
-            );
-          }
-          return false;
-        },
-        fixtures[operationName][locale],
-        {overwriteRoutes: false},
+  fetch.post(
```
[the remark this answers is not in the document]
emileber: It's a bug fix.

It doesn't change how consumers are using it, though if they were calling `getCountry('JP')`, they would actually get Canada instead of Japan, unless the whole country list was in the cache.

packages/address/src/AddressFormatter.ts
```diff
@@ -17,19 +23,18 @@ export default class AddressFormatter {
...
-
-    return country;
+    return Promise.resolve(
+      this.loadCountryFromCache(countryCode) ||
+        loadCountry(this.locale, countryCode),
+    );
   }
 
   async getCountries(): Promise<Country[]> {
+    const cachedCountries = ORDERED_COUNTRIES_CACHE.get(this.locale);
```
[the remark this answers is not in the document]
emileber: Good point!

Unless the list of countries we are serving through the atlas service ever change, it could prevent a full refetch in a single session, though a page refresh (or the next browser session) would get the updated list.

I personally think it was an oversight, and now that we have a `resetCache` method, any frontend that wants to force a refetch of the list could do it, but I don't believe it's a real use-case. 

Noting it in the changelog is a good idea, if you're up with keeping this change.

packages/address-mocks/src/index.ts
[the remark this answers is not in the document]
emileber: Though it's definitely a minor change (non-breaking but adds new features like error messages)

packages/address/src/AddressFormatter.ts
```diff
@@ -17,19 +23,18 @@ export default class AddressFormatter {
   }
 
   async getCountry(countryCode: string): Promise<Country> {
-    let country = this.loadCountryFromCache(countryCode);
```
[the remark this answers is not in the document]
emileber: I partially rolled back. Now, it uses a `const` and avoids the useless `await`, but it's otherwise the same exact behavior in the same exact order/structure.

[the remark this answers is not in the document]
BPScott: Cool, this looks good to me

packages/address-mocks/src/index.ts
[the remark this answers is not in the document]
BPScott: Thanks for the clarification, a change that would surface incorrect usage is minor/patch to me :)

PR #2180, opened 2022-02-17
🔧 Update eslint to v8 and eslint-plugin-shopify to v41

emileber:
## Description

```ts
class Example {
  /**
   * This comment was triggering either `lines-around-comment` or `prettier/prettier`
   */
  constructor() { /**/ }
}
```

Updating both `eslint` and the `@shopify/eslint-plugin` packages seem to have solved the issue. 

## Type of change

Since it's related to dev tools configuration, I do not believe it warrants any version bump.

## Checklist


emileber:
Most changes are related to the new [`jest/prefer-to-be` rule](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/prefer-to-be.md) added in the new [`eslint-plugin-jest@25.0.0` version](https://github.com/jest-community/eslint-plugin-jest/blob/main/CHANGELOG.md#2500-2021-10-10) used by `@shopify/eslint-plugin`.

BPScott (approved):
Changes look great to me. Thanks for updating eslint and our preset!

PR #2181, opened 2022-02-18
[react-server] Pass isServer to the react element created on the server

benkovy:
## Description

`[react-server]`

Convenience? 

When using quilt and you do not provide your own server we still get server side rendering for free thanks to this plugin, feels like this might add some convenience to checking whether or not you are on the server. 

We probably have some util already for this so please tell me if we do!

## Checklist


BPScott:
This assumes that the App component accepts a prop called `isSever` which, currently none of our bootstraps contain.

Having to drill this prop down into any components you actually use pr jam it into some context would be pretty clunky, so I doubt most folks would need this.


The easiest way to check if you're rendering on the server is to check for the absence of the `window` global. Is there a reason that you wouldn't be able to do this: 

```js
export const isServer = typeof window === 'undefined';`
```

benkovy:
sweet that makes sense - thanks @BPScott

PR #2184, opened 2022-02-22
update deprecated export pattern

dahukish:
## Description

Fixes (issue https://github.com/Shopify/web/issues/56743)

The `./` pattern for package exports is deprecated and ignored https://webpack.js.org/guides/package-exports/#support

Updated too `./*` which is viable. Also sorted the exported order to reflect: https://webpack.js.org/guides/package-exports/#notes-about-ordering

## Type of change

- [x] (All packages) Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

### DIFF
--- a/packages/ast-utilities/package.json
+++ b/packages/ast-utilities/package.json
@@ -68,12 +68,6 @@
   "module": "index.mjs",
   "esnext": "index.esnext",
   "exports": {
-    "./": "./",
-    ".": {
-      "esnext": "./index.esnext",
-      "import": "./index.mjs",
-      "require": "./index.js"
-    },
     "./javascript": {
       "esnext": "./javascript.esnext",
       "import": "./javascript.mjs",
@@ -83,6 +77,12 @@
       "esnext": "./markdown.esnext",
       "import": "./markdown.mjs",
       "require": "./markdown.js"
-    }
+    },
+    ".": {
+      "esnext": "./index.esnext",
+      "import": "./index.mjs",
+      "require": "./index.js"
+    },
+    "./*": "./"
   }
 }
--- a/packages/graphql-mini-transforms/package.json
+++ b/packages/graphql-mini-transforms/package.json
@@ -75,12 +75,6 @@
   "module": "index.mjs",
   "esnext": "index.esnext",
   "exports": {
-    "./": "./",
-    ".": {
-      "esnext": "./index.esnext",
-      "import": "./index.mjs",
-      "require": "./index.js"
-    },
     "./jest": {
       "esnext": "./jest.esnext",
       "import": "./jest.mjs",
@@ -95,7 +89,13 @@
       "esnext": "./webpack-loader.esnext",
       "import": "./webpack-loader.mjs",
       "require": "./webpack-loader.js"
-    }
+    },
+    ".": {
+      "esnext": "./index.esnext",
+      "import": "./index.mjs",
+      "require": "./index.js"
+    },
+    "./*": "./"
   },
   "tags": [
     "graphql",
--- a/packages/graphql-testing/package.json
+++ b/packages/graphql-testing/package.json
@@ -54,16 +54,16 @@
   "module": "index.mjs",
   "esnext": "index.esnext",
   "exports": {
-    "./": "./",
+    "./matchers": {
+      "esnext": "./matchers.esnext",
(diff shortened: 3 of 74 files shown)


heathernfran:
Updating this test to `./*` should fix the tests

https://github.com/Shopify/quilt/blob/main/tests/consistent-package-json.test.ts#L145

heathernfran (approved):

packages/ast-utilities/package.json
```diff
@@ -68,12 +68,6 @@
   "module": "index.mjs",
   "esnext": "index.esnext",
   "exports": {
-    "./": "./",
-    ".": {
```
BPScott, line 72: What's with thinking behind the reordering of the . import? It's doesn't make any difference but I'd say the what you get when you import  `@shopify/PACKAGENAMEHERE` would be the most important export to see, and thus it deserves to be at the top of the list

packages/graphql-mini-transforms/package.json
```diff
@@ -95,7 +89,13 @@
       "esnext": "./webpack-loader.esnext",
       "import": "./webpack-loader.mjs",
       "require": "./webpack-loader.js"
-    }
+    },
+    ".": {
```
BPScott, line 93: Same Q

packages/graphql-testing/package.json
```diff
@@ -54,16 +54,16 @@
   "module": "index.mjs",
   "esnext": "index.esnext",
   "exports": {
-    "./": "./",
+    "./matchers": {
```
BPScott, line 57: Same Q

packages/polyfills/package.json
```diff
@@ -224,11 +223,6 @@
       "import": "./idle-callback.node.mjs",
       "require": "./idle-callback.node.js"
     },
-    ".": {
```
BPScott, line 227: One more reorder

packages/react-async/package.json
```diff
@@ -56,16 +56,16 @@
   "module": "index.mjs",
   "esnext": "index.esnext",
   "exports": {
-    "./": "./",
+    "./testing": {
```
BPScott, line 59: One more reorder

packages/react-effect/package.json
```diff
@@ -51,16 +51,16 @@
   "module": "index.mjs",
   "esnext": "index.esnext",
   "exports": {
-    "./": "./",
+    "./server": {
```
BPScott, line 54: One more reorder

packages/react-html/package.json
```diff
@@ -59,16 +59,16 @@
   "module": "index.mjs",
   "esnext": "index.esnext",
   "exports": {
-    "./": "./",
+    "./server": {
```
BPScott, line 62: One more reorder

packages/react-network/package.json
```diff
@@ -62,16 +62,16 @@
   "module": "index.mjs",
   "esnext": "index.esnext",
   "exports": {
-    "./": "./",
+    "./server": {
```
BPScott, line 65: One more reorder

packages/react-server/package.json
```diff
@@ -82,16 +82,16 @@
   "module": "index.mjs",
   "esnext": "index.esnext",
   "exports": {
-    "./": "./",
+    "./webpack-plugin": {
```
BPScott, line 85: One more reorder

packages/react-testing/package.json
```diff
@@ -53,16 +53,16 @@
   "module": "index.mjs",
   "esnext": "index.esnext",
   "exports": {
-    "./": "./",
+    "./matchers": {
```
BPScott, line 56: One more reorder

packages/web-worker/package.json
```diff
@@ -21,12 +21,6 @@
     }
   },
   "exports": {
-    ".": {
```
BPScott, line 24: One more reorder

BPScott (commented):
Code changes look good.


What's with thinking behind the reordering of the . import? It's doesn't make any difference but I'd say the what you get when you import `@shopify/PACKAGENAMEHERE` would be the most important export to see, and thus it deserves to be at the top of the list, above what you get when importing `@shopify/PACKAGENAMEHERE/blah`

PR #2185, opened 2022-02-22
chore(): add type assertion tests for useful-types

KevinMind:
## Description

Fixes (issue #)

## Type of change

## Checklist


BPScott (changes requested):
Jest tests are runtime tests. The asserts that you're writing will always pass as at runtime time you're always doing `expect(input).toStrictEqual(input);`, which has nothing to do with type systems.

It seems very strange to write runtime tests to assert on type output. This works because we run tsc over our test files but it's got nothing to do with what gets ran within jest (as the jest output is always true). There's no way to test error states because any error would cause our type-checking to fail as this code is treated as part of the codebase (and codebase code must always pass).

I'd prefer an approach where we explicitly test the output of tsc somehow. This would allow us to test both sucess states and error states of the types, and would make it clear that we're testing the type-system rather than anything at runtime.

packages/useful-types/src/tests/types.test.ts
```diff
@@ -0,0 +1,233 @@
...
+) {
+  if (typeof actual === typeof expected && !isNoMatch(expected)) {
+    throw new Error(
+      `cannot submit values of the same type. actual: ${actual}, expected: ${expected}`,
+    );
+  }
+  expect(actual).toStrictEqual(actual);
+}
+
+function extractCompilerErrors(code: string) {
```
KevinMind, line 105: This approach just runs the snippet against ts-node with a custom tsconfig. The issue here is actually accessing the types defined in `./types` Will keep plucking away to get it to work.

packages/useful-types/test-d/types.test-d.ts
```diff
@@ -0,0 +1,20 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
+import {expectType, expectAssignable, expectNotAssignable} from 'tsd';
+
+import {ArrayElement} from '../build/ts/types';
```
BPScott, line 4: Rather than relying on build output are you able to test against `../src` instead?

BPScott:
I've just took the liberty of adding a commit to this PR. It adds a build step in projects that contain a `test-d` folder that runs tsd against that project. This means that tsd is ran as part of doing `yarn run build`, which feels a little strange at first but makes more sense if we frame it as "tsd must run against `.d.ts` files, thus we are testing build output, not source code".

This ensures that the tsd stuff is ran after a build and thus when the type files are always present and means we don't have to say "in order for tests to pass we run a build first (as tests should test source code, not compiled output)".

See it in action be introducing an error to `packages/useful-types/test-d/ArrayElement.test-d.ts` and running `yarn run build` and you'll get:

```
[18:34:15.646] failed during step build package useful-types variant


TSD found errors.


  packages/useful-types/test-d/ArrayElement.test-d.ts:9:43
  ✖  9:43  Argument of type string is not assignable to parameter of type string[].

  1 error
```

packages/useful-types/test-d/types.test-d.ts
KevinMind: Replying to both ^ and `I don't understand why you're trying to jam this into jest,`

I like using the CLI since it works out of the box and if we have a nice way of running it then I'm all in! Thanks for setting up the orchestration!

Regrading some of the feedback, I didn't set out to use jest, just didn't want to / didn't know how to add another CI step while it was possible to do without. Also I think there is at least _some_ sense in running this check in the "test" phase since it is logically most similar to a unit test, as in setup, render, assert.

.eslintrc.js
```diff
@@ -78,7 +78,7 @@ module.exports = {
       },
     },
     {
-      files: ['**/tests/**/*.ts', '**/tests/**/*.tsx', '**/loom.config.ts'],
+      files: ['**/tests/**/*.ts', '**/tests/**/*.tsx', '**/loom.config.ts', '**/*.test-d.ts', '**/*.test.d.tsx'],
```
BPScott, line 81: Should that  `**/*.test.d.tsx` be  `**/*.test-d.tsx`?

```suggestion
      files: ['**/tests/**/*.ts', '**/tests/**/*.tsx', '**/loom.config.ts', '**/*.test-d.ts', '**/*.test-d.tsx'],
```

Though do we even need '**/*.test-d.tsx'? When would we want these type-checking files to contain JSX content?

packages/react-async/test-d/types.test-d.ts
```diff
@@ -0,0 +1,5 @@
+import {expectAssignable} from 'tsd';
```
BPScott, line 1: We've  proved that the the loom integration works across multiple packages, but this test itself isn't that useful. Let's remove this

documentation/guides/testing-exported-typescript-types.md
```diff
@@ -0,0 +1,77 @@
...
+
+## How does it work?
+
+TSD is run as a part of the build and any errors will be logged and cause your build to fail. Since TSD is testing `built` typescript typings files, we need to first build the package, and then test against the exported types.
+
+All quilt packages reference a `types` property in their package.json. This is the typings file TSD will test against.
+
+## Why not just use `yarn type-check`?
+
+Typechecking quilt source code ensures that there are no type errors in the source code. TSD allows us to assert that types built by quilt packages implement the logical constraints we intend them to. You can test erorr cases, negative cases, and a range of positive cases. This leads to more resilient types and easier refactoring of types.
```
BPScott, line 56:
```suggestion
Typechecking quilt source code ensures that there are no type errors in the source code. TSD allows us to assert that types built by quilt packages implement the logical constraints we intend them to. You can test error cases, negative cases, and a range of positive cases. This leads to more resilient types and easier refactoring of types.
```

config/loom/index.ts
```diff
@@ -56,5 +60,43 @@ export function quiltPackage({
         ]);
       });
     }),
+    runTsd(),
   ]);
 }
+
+/**
+ * TSD allows us to run assertions against the types that are generated as part
+ * of the build
```
BPScott, line 69: Adding a bit of exposition about why this is a build step

```suggestion
 * TSD allows us to run assertions against the types that are generated as part
 * of the build. This is ran as a post-build step because tests the compiled
 * types in `packages/PACKAGENAME/build/ts` rather than source code.
```

PR #2193, opened 2022-03-01
Update export maps to export cjs only for node consumers

vsumner:
## Description
This PR updates export maps to only export cjs for code that is intended for node consumers.

Fixes (issue #)

## Type of change

I'm torn between patch and major. This is closer to a bug fix than a breaking change as the esnext/esm code could never work as intended.

`@shopify/async`
`@shopify/react-i18n`
`@shopify/web-worker`
`graphql-mini-transforms`

## Checklist


packages/graphql-mini-transforms/CHANGELOG.md
```diff
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Changed
+
+- Export cjs by default for `babel`, `webpack`, `jest`, and `jest-simple`. [[#2193](https://github.com/Shopify/quilt/pull/2193)]
```
royleejr, line 12: You mention in the change log that you export cjs for `babel`, `webpack`, `jest`, and `jest-simple` but in `packages/graphql-mini-transforms/package.json` I don't see one for `babel`.

royleejr (approved):
Other than that 1 comment for the changelog. Looks good!

tests/consistent-package-json.test.ts
```diff
@@ -29,6 +29,7 @@ const KNOWN_TEMPLATE_KEYS = [
 ];
 
 const GLOB_PATH = './*';
+const CJS_EXPORT_KEYS = ['webpack', 'jest', 'jest-simple', 'babel'];
```
BPScott, line 32: This works, but claiming "export keys with these names are allowed to use the `default` export only" feels kinda weird. There's nothing inherently magical about these names that apply to all packages so blanket excluding them them feels odd.

I wonder If I'd feel better about this if the whitelist contained the package names too (e.g. `web-worker/babel`, so know we're talking about specific entrypoints in specific packages.

I'd also like to see a test that shows these keys use just a `default` value - currently we're omiting them from any validation

BPScott (commented):
Package.json changes are good. I'm leaning towards patch too - in the cases of graphql-mini-transforms and web-worker the mjs versions never worked thanks to their usage of __filename and __dirname so this is a fix.

I believe it's possible to use just a string instead of needing an object that only contains default: e.g. `"./babel": "./babel.js"` and you get the same behaviour. Dunno how you feel about that - if it's any cleaner or consistent. I could go either way.

One thought inline about the test for validating the exports map shape - I think I'd like to make it clear what gets exceptions rather than "'babel' is a magic name has a special case across all packages".

packages/graphql-mini-transforms/CHANGELOG.md
vsumner: good catch

PR #2194, opened 2022-03-01
Upgrade storybook a11y addon and update deprecated methods

itwasmattgregg:
## Description

This PR introduces changes to storybook-a11-tests to improve performance and to support storyStoreV7 that can provide speed gains to storybook as well.

- Refactors to class base approach to simplify browser instance re-usage
- Runs static server to bypass any issues with CORS that puppeteer could have
- Uses `loadStory` API when available
- Provides additional method to collect story ids from StoryStoreV7
- To be able to properly run tests, I had to add a lot of dependencies and a mini storybook
- I treated the build storybook as test fixtures and committed then to git
-`store.raw()/store.extract()` and friends that list all stories require a prior call to `store.cacheAllCSFFiles()` (which is async). This will load all stories, and isn't generally a good idea in an addon, as it will force the whole store to load.

Superseedes: https://github.com/storybookjs/storybook/pull/16398/files

The code changes here are being used in

web: https://github.com/Shopify/web/pull/58844
email: https://github.com/Shopify/email/pull/11832

## Type of change

- [ ] storybook-a11y-test Major: Ugrade to use class based approach and supports storyStoreV7

## Checklist


itwasmattgregg:
I will need help updating the package version itself. I haven't done that before

packages/storybook-a11y-test/README.md
```diff
@@ -49,36 +49,39 @@ For optimal test performance, break the build and accessibility testing steps in
 Make sure you have built your Storybook instance and there is an `iframe.html` file that you can point the test towards.
```
itwasmattgregg, line 49: This should be changed to not reference the iframe

felipeleusin (approved):
Paired with @itwasmattgregg on this and 🎩  with email and web

itwasmattgregg:
This looks good from my end. Tophatted in web

felipeleusin (approved):

packages/storybook-a11y-test/src/index.ts
```diff
@@ -22,86 +25,201 @@ type ViolationResult = Unpacked<AxeResults['violations']>;
...
-      .filter((story) => story.parameters.a11y && story.parameters.a11y.disable)
-      .map((story) => story.id),
-  );
+    if (!address || typeof address === 'string') {
+      throw new Error('Unable to start local server');
+    }
 
-  await page.close();
-  await browser.close();
+    this.iframePath = `http://127.0.0.1:${address.port}/iframe.html`;
```
BPScott, line 53: Is there a reason that using a `file://` url is now insufficient?

packages/storybook-a11y-test/stories/Button.stories.tsx
```diff
@@ -0,0 +1,53 @@
...
+  // More on argTypes: https://storybook.js.org/docs/react/api/argtypes
+  argTypes: {
+    backgroundColor: {control: 'color'},
+  },
+} as ComponentMeta<typeof Button>;
+
+// More on component templates: https://storybook.js.org/docs/react/writing-stories/introduction#using-args
+const Template: ComponentStory<typeof Button> = (args) => <Button {...args} />;
+
+export const Primary = Template.bind({});
```
BPScott, line 19: If we're all about modern behaviours, then we could try out [CSF v3's new object format](https://storybook.js.org/blog/component-story-format-3-0/#new-in-csf-3-0)

Something like: 
```js
export const Primary = {args: { primary: true, label: 'Button' }}
```

packages/storybook-a11y-test/tests/index.test.ts
```diff
@@ -0,0 +1,59 @@
+/* eslint-env node */
+import path from 'path';
+
+import {A11yTestRunner} from '../src/index';
+
+const buildDir = path.join(__dirname, './fixtures/storybook');
+
+describe('can test a story', () => {
```
BPScott, line 8: Testing against fixtured data feels strange. It means we're testing against a specific point in time rather than using a storybook built with whatever version that is defined in the package.json.

I can see a point in time when the we update storybook in the dev dependencies, but don't touch the test fixtures and we think everything is fine because tests pass, but then when it comes to test the real latest storybook version everything breaks. I think that'd be suprising and it wouldn't be clear that you'd need to update the fixture data whenever you updated the storybook version.

I think running `yarn workspace @shopify/storybook-a11y-test run build-storybook --modern --output-dir=tests/fixtures/storybook` build in a `beforeAll` step here and testing against what that generates would be more bulletproof at the expense of taking a bit longer.


I also note that the creation of this fixture is not deterministic - I ran `yarn workspace @shopify/storybook-a11y-test run build-storybook --modern --output-dir=tests/fixtures/storybook` and it produced a bunch of different files compared to this output. It's not clear if anything meaningful actually changed, so keeping this fixture up to date has the potential to be churny.

BPScott (commented):
I LOVE that we've now got tests for this! That is super cool. The update to reuse a single fewer browser instances is also fantastic and shall lead to a nice speed improvement.

However I'm a little concerned by how clunky the fixture output is and how it could potentially lead to giving us false confidence in the future (an update to storybook wouldn't result in updating the fixture data and I bet people will forget). I think it'd be worth doing a storybook build in a beforeAll test set rather than committing fixtured output. It'll be slower overall but I think the extra confidence is worth it.

I think that we can also avoid installing some of these storybook plugins too - we're not going to be using this storybook UI so having fewer dependencies is prefereable.

packages/storybook-a11y-test/.storybook/main.js
```diff
@@ -0,0 +1,33 @@
+require('core-js/stable');
+require('regenerator-runtime/runtime');
+
+module.exports = {
+  stories: [
+    '../stories/**/*.stories.mdx',
+    '../stories/**/*.stories.@(js|jsx|ts|tsx)',
+  ],
+  addons: [
+    '@storybook/addon-links',
```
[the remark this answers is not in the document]
felipeleusin: Removed all the additional addons, had to keep postcss since this is almost a mandatory one now for storybook.

packages/storybook-a11y-test/.storybook/main.js
```diff
@@ -0,0 +1,33 @@
...
+    '@storybook/addon-essentials',
+    '@storybook/addon-interactions',
+    '@storybook/addon-postcss',
+  ],
+  framework: '@storybook/react',
+  features: {
+    storyStoreV7: true,
+  },
+  babel: (config) => {
+    // This is to change babel-preset-env and we have to do this manually to prevent polluting the package.json
```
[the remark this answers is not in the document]
felipeleusin: I have no clue, I think they expect to have a `.babelrc`, 'browserlists' or some sort of config for the repo but since we don't want to apply this to the entire repo and only to storybook this is the best place to put this config.

packages/storybook-a11y-test/.storybook/preview.js
```diff
@@ -0,0 +1,9 @@
+export const parameters = {
```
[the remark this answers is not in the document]
felipeleusin: Removed

packages/storybook-a11y-test/README.md
```diff
@@ -46,39 +46,39 @@ For optimal test performance, break the build and accessibility testing steps in
...
+
+      // Optional: maximum time in milliseconds to wait for the browser instance to start.
+      // Defaults to 30000 (30 seconds). Pass 0 to disable timeout.
+      timeout: 30000,
+    });
+
+    if (results.length) {
+      console.error(`‼️  Accessibility violations found`);
+      console.log(results.join('\n'));
+      process.exit(1);
```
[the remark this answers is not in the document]
felipeleusin: That's a good catch, it will probably not run the teardown. Will switch to `process.exitCode = 1` here

packages/storybook-a11y-test/tests/index.test.ts
```diff
@@ -0,0 +1,81 @@
...
+const buildDir = path.join(__dirname, './fixtures/storybook');
+
+jest.setTimeout(180000);
+
+describe('can test a story', () => {
+  beforeAll(() => {
+    return new Promise((resolve, reject) => {
+      const icuDataPath = path.dirname(require.resolve('full-icu'));
+      exec(
+        `NODE_ICU_DATA=${icuDataPath} yarn run build-storybook`,
```
felipeleusin, line 16: Needed or it fails in the CI, unsure why 🤷🏼‍♂️

packages/storybook-a11y-test/src/index.ts
BPScott: cool, thanks for the context

packages/storybook-a11y-test/CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 <!-- ## Unreleased -->
 
+## Unreleased
+
+### Changed
+
+- Update storybook addon deps and remove use of fromId() in favor of loadStory(), introduces class based approach. [[#2194](https://github.com/Shopify/quilt/pull/2194)]
```
BPScott, line 14:
```suggestion
- Now requires storybook >6.4.0. Add support for storyStoreV7 mode. Reuse open browser instances instead of opening new browsers every time, which reduces memory pressure and makes the tests faster. [[#2194](https://github.com/Shopify/quilt/pull/2194)]
```

PR #2197, opened 2022-03-03
Automatically add STATSD_DEFAULT_TAGS to the globalTags

alexandcote:
## Description

Fixes (issue #)

## Type of change

## Checklist

### DIFF
--- a/packages/statsd/src/client.ts
+++ b/packages/statsd/src/client.ts
@@ -16,6 +16,16 @@ export interface Options extends ClientOptions {
   snakeCase?: boolean;
 }
 
+// eslint-disable-next-line no-process-env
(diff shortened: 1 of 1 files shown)


packages/statsd/src/client.ts
```diff
@@ -16,6 +16,16 @@ export interface Options extends ClientOptions {
...
 }
 
+// eslint-disable-next-line no-process-env
+const defaultStatsDTags = process.env.STATSD_DEFAULT_TAGS || '';
+
+const globalTags =
+  defaultStatsDTags === ''
+    ? {}
+    : Object.fromEntries(
+        defaultStatsDTags.split(',').map((tag) => tag.split(';')),
```
peter-hamilton, line 26:
```suggestion
        defaultStatsDTags.split(',').map((tag) => tag.split(':')),
```

PR #2198, opened 2022-03-03
Update react-intersection-observer README.md

matallo:
## Description

Fix component prop in README, based on usage of `onIntersectionChange`. Maybe there are more updates still pending.

## Type of change

## Checklist


BPScott (approved):
Yep, that's the correct prop name per [this](https://github.com/Shopify/quilt/blob/3299ef976e29b86331783dade2a4792d55878195/packages/react-intersection-observer/src/IntersectionObserver.tsx#L13)

PR #2200, opened 2022-03-04
Update ref to release to be version-bump

vsumner:
## Description
Update incorrect ref to `release` to be `version-bump` in documentation.
Fixes (issue #)

### DIFF
--- a/documentation/guides/release-and-deploy.md
+++ b/documentation/guides/release-and-deploy.md
@@ -44,7 +44,7 @@ git push origin main --follow-tags
 
 ### Step 1 - Publish a `beta` release for testing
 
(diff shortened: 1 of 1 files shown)


KevinMind:
Thanks for updating this!

(Could be just my machine but)I think we need a few more tweaks. the output from Lerna seems out of date as well, this could lead to confusion.

These are the options we get from lerna.

image

Currently the docs say to select "Custom" but there are 2 options that match Custom.
 I think we want to encourage "Custom Version" or "Custom Prerelease" to be clear.

image

It might actually make sense to prefer the prerelase option, you can enter "<my-feature>-beta" and it will auto target the version number and append the beta tag. (I feel like that could be automated as well)

PR #2203, opened 2022-03-07
Update `react-graphql-universal-provider` to Apollo 3

emiryy:
## Description

Resolves https://github.com/Shopify/web/issues/59426

Following the migration guide [here](https://www.apollographql.com/docs/react/migrating/apollo-client-3-migration/), this PR removes the following dependencies from the `react-graphql-universal-provider` package and replaces them with `@apollo/client`:

- `apollo-cache-inmemory`
- `apollo-client`
- `apollo-link`
- `apollo-link-context`

All existing imports are updated to point to the new paths from `@apollo/client`

## Type of change

## Checklist

### DIFF
--- a/packages/react-graphql-universal-provider/src/GraphQLUniversalProvider.tsx
+++ b/packages/react-graphql-universal-provider/src/GraphQLUniversalProvider.tsx
@@ -32,7 +32,7 @@ export function GraphQLUniversalProvider<
 
   const [client, ssrLink] = useLazyRef<
     [
-      import('@apollo/client').ApolloClient<any>,
+      ApolloClient<any>,
       ReturnType<typeof createSsrExtractableLink> | undefined,
     ]
   >(() => {
--- a/packages/react-graphql-universal-provider/src/GraphQLUniversalProvider.tsx
+++ b/packages/react-graphql-universal-provider/src/GraphQLUniversalProvider.tsx
@@ -31,10 +31,7 @@ export function GraphQLUniversalProvider<
   const requestID = useRequestHeader('X-Request-ID');
 
   const [client, ssrLink] = useLazyRef<
-    [
-      ApolloClient<any>,
-      ReturnType<typeof createSsrExtractableLink> | undefined,
-    ]
+    [ApolloClient<any>, ReturnType<typeof createSsrExtractableLink> | undefined]
   >(() => {
     const server = isServer();
 
--- a/packages/react-graphql-universal-provider/src/GraphQLUniversalProvider.tsx
+++ b/packages/react-graphql-universal-provider/src/GraphQLUniversalProvider.tsx
@@ -1,7 +1,6 @@
 import React from 'react';
-import {ApolloClient, ApolloClientOptions} from 'apollo-client';
-import {ApolloLink} from 'apollo-link';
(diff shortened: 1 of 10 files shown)


packages/react-graphql-universal-provider/src/tests/GraphQLUniversalProvider.test.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
-import {InMemoryCache} from 'apollo-cache-inmemory';
-import {ApolloLink} from 'apollo-link';
+import {ApolloLink} from '@apollo/client';
+import {InMemoryCache} from '@apollo/client/cache';
```
BPScott, line 3: It looks like this is available from the root entrypoint, no need to reach deep

packages/react-graphql-universal-provider/src/csrf-link.ts
```diff
@@ -1,4 +1,4 @@
-import {setContext} from 'apollo-link-context';
+import {setContext} from '@apollo/client/link/context';
```
BPScott, line 1: It looks like this is available from the root entrypoint, no need to reach deep

packages/react-graphql-universal-provider/src/GraphQLUniversalProvider.tsx
```diff
@@ -1,7 +1,6 @@
 import React from 'react';
-import {ApolloClient, ApolloClientOptions} from 'apollo-client';
-import {ApolloLink} from 'apollo-link';
-import {InMemoryCache, NormalizedCacheObject} from 'apollo-cache-inmemory';
+import {ApolloClient, ApolloClientOptions, ApolloLink} from '@apollo/client';
+import {InMemoryCache, NormalizedCacheObject} from '@apollo/client/cache';
```
BPScott, line 3: It looks like these are available from the root entrypoint, no need to reach deep

packages/react-graphql-universal-provider/src/GraphQLUniversalProvider.tsx
```diff
@@ -33,7 +32,7 @@ export function GraphQLUniversalProvider<
 
   const [client, ssrLink] = useLazyRef<
     [
-      import('apollo-client').ApolloClient<any>,
+      import('@apollo/client').ApolloClient<any>,
```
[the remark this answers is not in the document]
emiryy: The type error was related to a file I missed and didn't update the imports in.

packages/react-graphql-universal-provider/src/tests/GraphQLUniversalProvider.test.tsx
emiryy: We want to reach deep as client doesn't get exported here: https://github.com/apollographql/apollo-client/blob/main/src/index.ts

packages/react-graphql-universal-provider/src/GraphQLUniversalProvider.tsx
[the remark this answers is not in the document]
ryanwilsonperkin: My thought was to use this here:
```diff
- import('apollo-client').ApolloClient<any>,
+ ApolloClient<any>,
```
since that type should already be available in this context, does that not work?

PR #2204, opened 2022-03-08
adding section to readme for embedded app controller example

KenjiDuggan:
## Description

Fixes (issue #)

## Type of change

## Checklist


BPScott (changes requested):
The table of contents change seems incorrect, can you revert the indentation changes, so you're adding just the one heading.

The text in the new section seems reasonable.

PR #2212, opened 2022-03-09
Update graphql testing

dahukish:
## Description

Fixes (issue https://github.com/Shopify/web/issues/59428)

Bumps graphql-testing to apollo client v3. 

Bonus bumps apollo/client to v3 in the main package.json

## Type of change

## Checklist

### DIFF
--- a/packages/react-graphql/src/hooks/tests/query.test.tsx
+++ b/packages/react-graphql/src/hooks/tests/query.test.tsx
@@ -128,7 +128,8 @@ describe('useQuery', () => {
 
       // Once for initial render while loading, once for when the data loaded, and a final time
       // when we update the props and re-render the component.
-      expect(renderPropSpy).toHaveBeenCalledTimes(3);
+      // set this back to 3
+      expect(renderPropSpy).toHaveBeenCalledTimes(4);
 
       const [, firstLoadedCall, secondLoadedCall] = renderPropSpy.mock.calls;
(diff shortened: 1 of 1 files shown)


packages/graphql-testing/src/client.ts
```diff
@@ -1,67 +0,0 @@
-import {
```
emiryy, line 1: Non blocking, just curious why this file is no longer needed

packages/react-graphql/src/async/tests/component.test.tsx
```diff
@@ -143,7 +143,7 @@ describe('createAsyncQueryComponent()', () => {
 
       const client = createMockApolloClient();
       const watchQuerySpy = jest.spyOn(client, 'watchQuery');
-      watchQuerySpy.mockImplementation(() => ({subscribe() {}} as any));
+      watchQuerySpy.mockImplementation(() => ({subscribe() {} } as any));
```
emiryy, line 146: This seems to be extra spacing which will cause the linter to fail. See below on like 205 for same thing.

emiryy (approved):
Looks good to me!

packages/graphql-testing/src/client.ts
dahukish: This was added two years ago to stop certain events happening out of `react` `act`. We will need to reevaluate and figure out if this is even needed anymore. I am leaning towards not.

PR #2213, opened 2022-03-10
Remove resolutions on types/react, types/react-dom, get-port

BPScott:
## Description

These resolutions are not needed, there are no conflicting constraints.

### DIFF
--- a/packages/react-effect/package.json
+++ b/packages/react-effect/package.json
@@ -30,9 +30,6 @@
   "engines": {
     "node": ">=12.14.0"
   },
-  "devDependencies": {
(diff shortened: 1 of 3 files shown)


packages/react-effect/package.json
```diff
@@ -30,9 +30,6 @@
   "engines": {
     "node": ">=12.14.0"
   },
-  "devDependencies": {
```
BPScott, line 33: Dev dependency not neeeded because `@types/react-dom` is already in the root package.json's dev dependencies

royleejr (approved):
Code looks good.

PR #2223, opened 2022-03-17
Update how subscription interacts with hook state

dahukish:
Swap the responseId counter for loading state

### DIFF
--- a/packages/react-graphql/src/hooks/query.ts
+++ b/packages/react-graphql/src/hooks/query.ts
@@ -112,15 +112,14 @@ export default function useQuery<
     [queryObservable, client, serializedVariables],
   );
 
-  const [responseId, setResponseId] = useState(0);
-
+  const [isLoading, setIsLoading] = useState(true);
   useEffect(() => {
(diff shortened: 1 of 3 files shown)


packages/react-graphql/src/hooks/query.ts
```diff
@@ -184,9 +183,7 @@ export default function useQuery<
       networkStatus: result.networkStatus,
       loading: result.loading,
     };
-
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [responseId, skip, queryObservable, defaultResult, previousData]);
+  }, [isLoading, skip, queryObservable, defaultResult, previousData]);
```
dahukish, line 186: Using the responseId which was just a counter was causing and extra repaint. With this we still subscribe to the observer (which gives us the query when its loaded) so we can use this to bust the cache and get the correct amount of expected calls.

emiryy (approved):
Makes sense to me! 👍

dahukish:
Closing this in favour of using what we have for now.

PR #2224, opened 2022-03-18
Fix Debug current test with Loom + all tests + changed tests (debug current test command)

melodyhabbouche:
## Description

Screen Shot 2022-03-18 at 12 08 04
We were experiencing an issue when trying to debug an individual test with Jest.

Fixes https://github.com/Shopify/web/issues/62822

- [x] Fixed here using loom (see [comment](https://github.com/Shopify/web/issues/62822#issuecomment-1105621342) for lots of important context): 

Screen Shot 2022-04-25 at 10 31 52 (2)

- [ ] Follow up issue to add `--watchAll` to loom

Screen Shot 2022-04-25 at 11 03 30

## Type of change

## Checklist

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -6,26 +6,30 @@
       "type": "node",
       "request": "launch",
       "program": "${workspaceFolder}/node_modules/.bin/loom",
-      "args": ["test", "--debug"],
+      "args": ["test", "--debug", "--no-watch", "."],
       "console": "integratedTerminal",
       "internalConsoleOptions": "neverOpen",
-      "disableOptimisticBPs": true,
-      "windows": {
-        "program": "${workspaceFolder}/node_modules/jest/bin/loom"
-      }
+      "disableOptimisticBPs": true
+    },
+    {
+      "name": "Debug Changed Tests (loom)",
+      "type": "node",
+      "request": "launch",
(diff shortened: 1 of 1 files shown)


camd:
Hi @melodyhabbouche : Are we still planning to merge this?  Happy to 🎩 and review when it's ready.  :)

BPScott (commented):
This setup seems incorrect. I'd expect that we would be running this through loom rather than calling jest directly, as loom is what provides all the config for the project.

Something like:

```
"program": "${workspaceFolder}/node_modules/.bin/loom",
"args": ["test"],
```

melodyhabbouche:
Following up a conversation with @ryanwilsonperkin 

> This looks like a difference between --watch and --watchAll
> --watchAll says “run all the tests and rerun them if any change”
> --watch says “run only the tests that have changed
> When we previously called Jest directly we were passing it --watchAll but Loom doesn’t seem to have an option for that, it only has --watch and I’ll bet that it just passes that through to Jest.

It seems like we'll need to add `--watchAll` to loom aa Ryan suggested in order to for the `Debug all tests` to work as expected / same behaviour as before (instead of the image below).

Screen Shot 2022-04-25 at 11 03 30

BPScott (approved):

PR #2231, opened 2022-04-06
Update faker-js/faker to v6

BPScott:
## Description

Update faker-js/faker to v6 so we don't need to do weird hacks to provide types for the package.

I've updated calls to deprecated behaviour. Searching the CI test logs for 'faker' should result in no deprecation warnings logged to the console.

## Type of change

- graphql-fixtures: major


qt314 (approved):

ryanwilsonperkin (dismissed):

BPScott:
Rebased atop of #2277, which drops support for node 12.

ryanwilsonperkin (approved):

PR #2238, opened 2022-04-14
Relax version constraints for `graphql-tool-utilities`

nickpith:
## Description

When fixing CVE-2022-24785 in the Fraud Filter app, the `graphql-tool-utilities` package was locking down to a specific version of `apollo-codegen-core`. A quick look at the history and asking on Slack doesn't reveal any particular reason why we need to be doing this so I'm relaxing the constraint. Once `apollo-language-server` removes the `moment` dependency with https://github.com/apollographql/apollo-tooling/issues/2323 that I've submitted a PR for, this change will allow users to upgrade and remove `moment` as well without this package getting in the way.

## Type of change

- [x] `graphql-tool-utilities` Patch: Bug (non-breaking change which fixes an issue)

## Checklist


dahukish (approved):
:shipit:

PR #2239, opened 2022-04-16
React i18n fixes

GoodForOneFare:
## Description

Fixes (issue #)

## Type of change

## Checklist


packages/react-i18n/src/babel-plugin/shared.ts
```diff
@@ -29,7 +29,8 @@ export function findTranslationBuckets(rootDir) {
   // `find` is used here instead of Node's glob because it performs much faster
   // (20s vs 1s in web with ~750 translation folders and 21 langs)
   const files = execSync(
-    `find ${rootDir} -type d \\( -path ./node_modules -o -path ./build -o -path ./tmp -o -path ./.git -o -path ./public \\) -prune -o -name '*.json' -print | grep /${TRANSLATION_DIRECTORY_NAME}/`,
+    `find ${rootDir} -type d \\( -path ${rootDir}/node_modules -o -path ${rootDir}/build -o -path ${rootDir}/tmp -o -path ${rootDir}/.git -o -path ${rootDir}/public \\) -prune -o -name '*.json' -print | grep /${TRANSLATION_DIRECTORY_NAME}/`,
```
GoodForOneFare, line 32: The old version of this failed to ignore files because `rootDir` is a fully-qualified path (e.g., `/app/web/`), which `find` fails to match against relative paths like `./node_modules`.

The end result isn't so bad in CI because `build`, etc don't contain many matching paths.  It did match some node_modules paths, but libs like `@shopify/global-nav` never directly perform `useI18n()`, so the babel plugin parses them, but leaves them untouched.

packages/react-i18n/src/babel-plugin/shared.ts
```diff
@@ -29,7 +29,8 @@ export function findTranslationBuckets(rootDir) {
...
+    {maxBuffer: 1_000_000 * 10},
```
GoodForOneFare, line 33: This is a bit lazy, but `exec` still performs buffering, so also hits the same limits as `execSync`.  Switching to `spawn` does solve it, but `find`'s arguments are especially difficult to pass along because `spawn` quotes every argument, breaking stuff like `\\( `.  

I'm sure it's all solvable in a much cleaner way, but 10MB isn't much nowadays 🤷

packages/react-i18n/src/babel-plugin/generate-dictionaries.ts
```diff
@@ -70,5 +71,9 @@ async function readLocaleTranslations(
     filePath.endsWith(`/${locale}.json`),
   );
 
-  return translationPath ? fs.readJson(translationPath) : {};
+  return translationPath
+    ? JSON.parse(
+        stripJsonComments((await fs.readFile(translationPath)).toString()),
```
BPScott, line 76: I think if you specify an encoding then readFile returns a string

```
stripJsonComments(await fs.readFile(translationPath, 'utf8')),

**Review by BPScott — COMMENTED** [review]
Looking good thanks Gord!

Two small notes inline

File: packages/react-i18n/CHANGELOG.md [hunk 4]
```diff
@@ -5,7 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Added
+
+- Minor - `generateTranslationIndexes` now parses jsonc files (when they have a `.json` extension)
```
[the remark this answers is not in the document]
GoodForOneFare: I sure can!  Done!

packages/react-i18n/src/babel-plugin/generate-dictionaries.ts
GoodForOneFare: That's better, thanks 😄

BPScott (approved):
Magical, thanks!

PR #2248, opened 2022-04-27
React 18 - @shopify/react-testing

ryanwilsonperkin:
## Description

Base: #2241

Updates `@shopify/react-testing` to support React 18

Fixes (issue #)

## Type of change

## Checklist

### DIFF
--- a/packages/react-testing/src/root.tsx
+++ b/packages/react-testing/src/root.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import {render, unmountComponentAtNode} from 'react-dom';
+import {createRoot, Root as ReactRoot} from 'react-dom/client';
 import {act} from 'react-dom/test-utils';
 
 import {TestWrapper} from './TestWrapper';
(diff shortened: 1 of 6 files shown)


ryanwilsonperkin:
Tested here against `web` running React 17 using a beta release of this PR in order to show that this version could be installed into `web` without error prior to upgrading to React 18: https://buildkite.com/shopify/web-ci-builder/builds/487393

packages/react-testing/src/root.tsx
```diff
@@ -90,37 +92,59 @@ export class Root<Props> implements Node<Props> {
...
-    const afterResolve = () => {
-      updateWrapper();
-      this.acting = false;
-
-      return result;
-    };
-
-    const promise = act(() => {
-      result = action();
+    // act has two versions, act(() => void) or await act(async () => void)
```
melodyhabbouche, line 102: nit: prefer to use block comment here

PR #2261, opened 2022-05-02
Include Shopify/declarative-forms into Quilt

zeachco:
## Description

Adding a new library to help with form generation and control

## Type of change

New folder containing the code of [Shopify/declarative-forms](https://github.com/Shopify/declarative-forms)

## Checklist


packages/declarative-forms/src/debug/DebugNode.tsx
```diff
@@ -52,6 +53,7 @@ export function DebugNode({children, node}: DebugProps) {
     >
       {visible && (
         <>
+          {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events */}
```
zeachco, line 56: This is a debug visual element, a11y is not relevant here

packages/declarative-forms/src/utilities/hook.ts
```diff
@@ -68,25 +69,29 @@ export function useNode<
...
-            mapContextErrorsToValidationErrors(serverErrorsTarget);
+          const externalErrors = mapContextErrorsToValidationErrors(
+            serverErrorsTarget,
+          );
           node.setErrors(externalErrors);
         }
       });
     },
-    [node, state.remoteErrors],
+    [node, state.remoteErrors, cloneName],
```
zeachco, line 94: Added dependency as well, non changing value

zeachco:
since the future of quilt seems uncertain, keeping status quo for declarative-forms

PR #2266, opened 2022-05-05
Update react-async to support React 18

### LINKED ISSUE #2263: Update react-async package to work with React 18
`@shopify/react-async` has a typescript error, which causes builds to fail

```
$ loom build
🧵 build
[pre]
[19:31:18.821] starting step Run Typescript
[19:31:18.822] log from Run Typescript
executing command: tsc --build --pretty
[19:32:03.003] failed during step Run Typescript
TypeScript found type errors
packages/react-async/src/tests/component.test.tsx:821:51 - error TS2339: Property 'children' does not exist on type 'Readonly<{}>'.
821       return this.state.error ? null : this.props.children;
                                                      ~~~~~~~~
Found 1 error.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
yarn run build terminated with exit status 1
```

ryanwilsonperkin:
## Description

Fixes #2263

Updates the typing for component to explicitly include children since this no longer happens automatically anymore from `@types/react`

## Type of change

- [x] react-async Patch: Bug (non-breaking change which fixes an issue)

## Checklist


royleejr (approved):

BPScott:
> I think in general, I'd like to see anything compatible with React 17 ship on the main branch

I'm a big big fan of doing this too

PR #2267, opened 2022-05-10
Add `interpolate` option to `react-i18n` to support custom placeholder formats

emileber:
## Description

In order to support Shopify theme's locale files format, with placeholder using mustache style `{{key}}`, I've introduced an `interpolate` option that accepts a `RegExp`, just like the similar option of Underscore/Lodash's [`_.template` function](https://lodash.com/docs/4.17.15#template).

```tsx
const MUSTACHE_INTERPOLATION = /({{[\s]*([\S]+?)[\s]*}})/g;

function WithI18nComponent() {
  const [i18n] = useI18n({
    id: 'MyComponent',
    fallback: {hello: 'Hello {{name}} {{ surname }}!'}, // custom format
  });

  return (
    <div>{i18n.translate('hello', {name: 'foo', surname: 'bar'})}</div>
  );
}

const manager = new I18nManager({
  locale: 'en',
  // custom interpolation regular expression
  interpolate: MUSTACHE_INTERPOLATION,
});
const component = mount(
  <I18nContext.Provider value={manager}>
    <WithI18nComponent />
  </I18nContext.Provider>,
);

expect(component.text()).toBe('Hello foo bar!');
```

This option doesn't affect how the library works by default, using the original replacement `RegExp`.

I've decided to expose 3 common `RegExp` for the following formats:

- `DEFAULT_INTERPOLATION`: `{key}`
- `MUSTACHE_INTERPOLATION`: `{{key}}`
- `ERB_INTERPOLATION`: `<%= key %>`

Though any valid regular expression could be used.

## Type of change

## Checklist


packages/react-i18n/README.md
```diff
@@ -27,6 +27,7 @@ This library requires a provider component which supplies i18n details to the re
 - `currency`: the default currency to use for currency-aware formatting.
 - `pseudolocalize`: whether to perform [pseudolocalization](https://github.com/Shopify/pseudolocalization) on your translations.
 - `onError`: a callback to use when recoverable i18n-related errors happen. If not provided, these errors will be re-thrown wherever they occur. If it is provided and it does not re-throw the passed error, the translation or formatting that caused the error will return an empty string. This function will be called with the error object.
+- `interpolate`: a regular expression to be used for interpolation of custom variable placeholder formats.
```
emileber, line 30: The name is inspired from the same option of the Underscore/Lodash's `_.template` function.

packages/react-i18n/src/index.ts
```diff
@@ -10,7 +10,13 @@ export type {
 export {useI18n} from './hooks';
 export {withI18n} from './decorator';
 export type {WithI18nProps} from './decorator';
-export {memoizedNumberFormatter, translate} from './utilities';
+export {
+  memoizedNumberFormatter,
+  translate,
+  DEFAULT_INTERPOLATION,
+  MUSTACHE_INTERPOLATION,
+  ERB_INTERPOLATION,
```
emileber, line 18: Expose `RegExp` for common placeholder formats.

packages/react-i18n/src/tests/e2e/translations.test.tsx
```diff
@@ -57,4 +58,29 @@ describe('translations', () => {
...
+      });
+
+      return (
+        <div>{i18n.translate('hello', {name: 'foo', surname: 'bar'})}</div>
+      );
+    }
+
+    const manager = new I18nManager({
+      locale: 'en',
+      interpolate: MUSTACHE_INTERPOLATION,
```
emileber, line 76: Usually, an app will use a single format across all its translations, so passing the option to the manager in the provider makes enough sense.

packages/react-i18n/src/utilities/translate.tsx
```diff
@@ -193,104 +202,106 @@ function translateWithDictionary(
...
   str: string,
   replacements?: ComplexReplacementDictionary,
+  options?: UpdateStringWithReplacementsOptions,
 ): React.ReactElement<any>;
 function updateStringWithReplacements(
   str: string,
   replacements?: PrimitiveReplacementDictionary,
+  options?: UpdateStringWithReplacementsOptions,
 ): string;
 function updateStringWithReplacements(
```
emileber, line 242: I've simplified this function, so that there's only one path for every case now, instead of a special path for strings versus complex replacements.

packages/react-i18n/src/utilities/translate.tsx
```diff
@@ -193,104 +202,106 @@ function translateWithDictionary(
...
+  {interpolate}: UpdateStringWithReplacementsOptions = {},
 ): any {
-  const replaceFinder = /([^{]*)({([^}]*)})?/g;
-  const allReplacementsArePrimitives = Object.keys(replacements).every(
-    (key) => typeof replacements[key] !== 'object',
-  );
-
-  if (allReplacementsArePrimitives) {
-    return str.replace(REPLACE_REGEX, (match) => {
-      const replacement = match.substring(1, match.length - 1);
```
emileber, line 235: This line assumes the `RegExp` value and the placeholder format. With my suggested changes, the `RegExp` is now the only source of truth.

packages/react-i18n/src/utilities/translate.tsx
```diff
@@ -193,104 +202,106 @@ function translateWithDictionary(
...
+    },
+  );
 
-    replaceFinder.lastIndex = 0;
+  // Push the last part of the source string
+  pieces.push(str.substr(lastOffset));
 
-    return pieces;
-  }
+  return pieces.every(isString) ? pieces.join('') : pieces;
```
emileber, line 296: This line replaces the `allReplacementsArePrimitives` value and its related logic.

packages/react-i18n/src/utilities/translate.tsx
```diff
@@ -12,11 +12,14 @@ import {
...
 import {MissingTranslationError, MissingReplacementError} from '../errors';
 
-const REPLACE_REGEX = /{([^}]*)}/g;
+import {DEFAULT_INTERPOLATION} from './interpolate';
+
 const MISSING_TRANSLATION = Symbol('Missing translation');
 const PLURALIZATION_KEY_NAME = 'count';
 const SEPARATOR = '.';
 
+const isString = (value: any): value is string => typeof value === 'string';
```
emileber, line 21: Instead of having single-use variables like:

```ts
const isAllStrings = pieces.every((piece) => typeof piece === 'string');
if (isAllStrings) //...
```

Clear functional code is enough:

```ts
if (pieces.every(isString)) //...
```

packages/react-i18n/src/utilities/translate.tsx
```diff
@@ -193,104 +202,106 @@ function translateWithDictionary(
...
+  // Uses replace callback, but not its return value
+  str.replace(
+    replaceFinder,
+    (
+      match,
+      _fullPlaceholder: string,
+      replacementKey: string,
+      offset: number,
+    ) => {
+      if (!replacementKey) return '';
```
movermeyer, line 265: When would this code be hit?
When a user supplies a regex that is missing a capture group?
Should this be an error instead (Perhaps much earlier)?

packages/react-i18n/src/utilities/translate.tsx
```diff
@@ -193,104 +202,106 @@ function translateWithDictionary(
...
+
+  let matchIndex = 0;
+  let lastOffset = 0;
+
+  // Uses replace callback, but not its return value
+  str.replace(
+    replaceFinder,
+    (
+      match,
+      _fullPlaceholder: string,
```
movermeyer, line 261: Instead of `_fullPlaceholder`, could we remove the outer capture group for each regex? Does that make sense / why does `_fullPlaceholder` exist?

packages/react-i18n/src/utilities/translate.tsx
emileber: I didn't want to change the behavior of the lib, which was ignoring the missing keys within strings. Though I could definitely throw an helpful error.

packages/react-i18n/src/utilities/interpolate.ts
```diff
@@ -0,0 +1,17 @@
...
+ * Similar to Shopify themes' locale files.
+ * `{{key}}` will be replaced with `replacements[key]`.
+ */
+export const MUSTACHE_INTERPOLATION = /({{[\s]*([\S]+?)[\s]*}})/g;
+
+/**
+ * Similar to Ruby ERB templating system.
+ * `<%= key %>` will be replaced with `replacements[key]`.
+ */
+export const ERB_INTERPOLATION = /(<%=[\s]*([\S]+?)[\s]*%>)/g;
```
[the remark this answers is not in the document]
emileber: We officially only support the default format, but now offer a way to override it. If anyone decides to override the default format, it's their responsibility to make sure the format regex works well.

Though I still wanted to make sure our implementation wasn't assuming the format anywhere, which wasn't the case before (substring stuff here and there), by testing with different formats.

Since I took the time the write a couple regexes, I believe it's safer to expose them rather then to let devs figure it out every time.

movermeyer (approved):

packages/react-i18n/src/utilities/interpolate.ts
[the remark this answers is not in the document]
movermeyer: As a note for anyone else who might see this and want to use a custom regex, you'll also need to reach out to INTL Languages to help configure your Translation Platform `translation.yml` with a custom placeholder regex.

https://development.shopify.io/engineering/developing_at_Shopify/internationalization/translation_platform#Component_Scheme_Placeholders

PR #2269, opened 2022-05-12
Increase buffer for i18n dictionary generation again

GoodForOneFare:
## Description
Increase the buffer for the shell out to `find` when generating translation dictionaries.  I bumped it up a couple of weeks ago, but it still wasn't enough to contain web's translations 😬 

## Type of change

- [x] react-i18n Patch: Bug

## Checklist


vsumner (approved):

packages/react-i18n/src/babel-plugin/shared.ts
```diff
@@ -30,7 +30,7 @@ export function findTranslationBuckets(rootDir) {
   // (20s vs 1s in web with ~750 translation folders and 21 langs)
   const files = execSync(
     `find ${rootDir} -type d \\( -path ${rootDir}/node_modules -o -path ${rootDir}/build -o -path ${rootDir}/tmp -o -path ${rootDir}/.git -o -path ${rootDir}/public \\) -prune -o -name '*.json' -print | grep /${TRANSLATION_DIRECTORY_NAME}/`,
-    {maxBuffer: 1_000_000 * 10},
+    {maxBuffer: 2_000_000 * 10},
```
vsumner, line 33: Worth making configurable later?

GoodForOneFare: I think this'll move into web sooner than later.  IMO, we should completely delete this feature from Quilt.

vsumner (approved):

PR #2271, opened 2022-05-13
Move Atlas clients to atlas.shopifysvc.com

SagarSidhu:
## Description

Fixes https://github.com/Shopify/atlas/issues/811

This migration is related to project mobius. The purpose is outlined in the following document: https://docs.google.com/document/d/1QX61cLH5txKZS9ErbgpWj4MeHntOGkPCOl4R7wKYZFQ/edit

## Type of change

## Checklist


BPScott (approved):
Looks good.

Could you expand on the rational for why you consider this a breaking change? I imagine requests are still going to the same server and responding in the same way.

Do you think there there would be other issues with changing the hostname?

SagarSidhu:
> Looks good, merge when you're ready.
> 
> Could you expand on the rational for why you consider this a breaking change? I imagine requests are still going to the same server and responding in the same way.
> 
> Do you think there there would be other issues with changing the hostname?

I probably could have tagged this as a minor change as well. My consideration was that changing the URL could break things in other places, but this new endpoint has been tested and we are confident that it is hitting the same server and responding in the correct way.

emileber:
While this doesn't change the API, it does break integrations as we need to update our Content Security Policy to make it work.

znarf:
Small feedback in case that helps for the future. As users of `@shopify/address` this was a [breaking change](https://github.com/opencollective/opencollective-frontend/pull/7865) for us at Open Collective, breaking our Content Security Policy. We noticed a [major version](https://github.com/opencollective/opencollective-frontend/pull/7838), read the notes but could not see anything mentioning it.

PR #2282, opened 2022-05-19
Update graphql-config to version 4

vsumner:
## Description
This PR picks up from @dahukish previous [PR](https://github.com/Shopify/quilt/pull/2057) to upgrade graphql-config to version 4.

Note: `graphql-config` version 3 is still used by our eslint plugin. Until we drop `eslint-plugin-graphql` which is deprecated we will have 2 versions.

Fixes (issue #)

## Checklist


Urigo:
Let us know if you have any feedback on the new version or the migration process that we could help with!
cc @B2o5T @kamilkisiela

PR #2288, opened 2022-05-25
Add changesets

vsumner:
## Description

- [ ] Figure out dependency relationships and their definitions in changesets
- [ ] Update release documentation
- [ ] Write a real release version
- [ ] Add `/snapit`

Fixes (issue #)

## Type of change

## Checklist


BPScott:
> Is there a decision log as to why we are moving to changelog?

There is #2298. The short version is: easier releases because you don't have to work out versions of packages based on finding and reading reading PRs since last release (as that info is captured at the PR stage), more automation, fewer manual steps, allows for easier releases of betas, using their release action removes dependency on shipit"

> Are we capturing how we are to use this later on?

I shall be updating https://github.com/Shopify/quilt/blob/main/documentation/guides/release-and-deploy.md once I prove that this works. I expect the content to be similar to https://github.com/Shopify/polaris/blob/main/documentation/Releasing.md

> Is this addressing the checkmarks stated by @vsumner on the top?

- Figure out dependency relationships and their definitions in changesets
  - YES, this is resolved.
- Update release documentation
  - Not done yet, shall be in follow up once I prove this works per above
- Write a real release version
  - Cutting a release is part of my test. Needs to be done after this PR is merged so we can see what is triggered by the actions that this PR introduces.
- Add /snapit
  - Can be done in a follow up PR, once I prove releases from main are working 


My next steps:

- Merge this
- Confirm the action creates the "do a release PR"
- Confirm that making additional changes triggers updates to the release PR and that dependant packages are well handled
- Confirm cutting a release works
- Update documentation on release steps
- Add `/snapit` command, and test it

PR #2289, opened 2022-05-25
Add TypedDocumentNode support to graphql-typescript-definitions

BPScott:
## Description

We generate `*.graphql.d.ts` using `graphql-typescript-definitions`. These files claim graphql files export documents that follow the shape of `DocumentNode` from `graphql-typed` - exposing `__typeData` and `__typeVariables` keys. `@shopify/react-graphql`'s `useQuery` / `useMutation` use these keys to do type inference - telling you the shape of the data that is returned from the query, and what variables you can pass into a query/mutation.

We're looking at moving to use `@apollo/client` for data fetching, which does this type inference in a different way.

`@apollo/client` does type inference based upon `TypedDocumentNode` from `@graphql-typed-document-node/core` - using an `__apiType` key that is a function that accepts Variables, and returns a Result.

This PR adds a new option to `graphql-typescript-definitions`'s `export-format` - `documentWithTypedDocumentNode` that makes it so the document exported by `*.graphql.d.ts` files is a union of  `graphql-typed`'s `DocumentNode` and  `@graphql-typed-document-node/core`'s `TypedDocumentNode`. This means that when using this mode, graphql files can have their types inferred by `@apollo/client` in addition to `@shopify/react-graphql`.

---

Here's a comparison of output between "using export-format=document (the default)" and "using export-format=documentWithTypedDocumentNode" against a query called `MyQuery` when there are  variables. 

```diff
import { DocumentNode } from "graphql-typed";
+ import { TypedDocumentNode } from "@graphql-typed-document-node/core";

// ...

- declare const document: DocumentNode<MyQueryData, MyQueryData.Variables, MyQueryPartialData>
+ declare const document: DocumentNode<MyQueryData, MyQueryData.Variables, MyQueryPartialData> & TypedDocumentNode<MyQueryData, MyQueryData.Variables>;
```

Here's a comparison of output between "using export-format=document (the default)" and "using export-format=documentWithTypedDocumentNode" against a query called `MyQuery` when there are no variables. The usage of `{[key: string]: never;}` for when there are no variables matches the behaviour of [graphql-codegen's TypedDocumentNode plugin](https://www.graphql-code-generator.com/plugins/typed-document-node), in preparation of a possible future where we move to a community codegen solution. Honestly I think this makes more sense - saying Varibles is always an object, but it might be empty and have no valid keys feels a bit more consistent than saying "this is either totally not allowed, or an object that has at least one key". I considered changing this typing for DocumentNode but it has a bunch of knock-on effects in `@shopify/react-graphql` that I have zero interest in wrangling at the momnent if the plan is to move away from `@shopify/react-graphql` altogether.

```diff
import { DocumentNode } from "graphql-typed";
+ import { TypedDocumentNode } from "@graphql-typed-document-node/core";

// ...

- declare const document: DocumentNode<MyQueryData, never, MyQueryPartialData>
+ declare const document: DocumentNode<MyQueryData, never, MyQueryPartialData> & TypedDocumentNode<MyQueryData, {
+   [key: string]: never;
+ }>;
```

## Type of change

graphql-typescript-definitions - minor

### DIFF
--- a/packages/graphql-typescript-definitions/package.json
+++ b/packages/graphql-typescript-definitions/package.json
@@ -46,7 +46,7 @@
     "ts-node": "^8.0.3"
   },
   "peerDependencies": {
-    "@graphql-typed-document-node/core": "^1.0.0",
+    "@graphql-typed-document-node/core": "^3.1.1",
     "graphql-typed": ">=0.4.0 <2.0.0"
   },
   "peerDependenciesMeta": {


vsumner (approved):

packages/graphql-typescript-definitions/package.json
```diff
@@ -46,8 +46,14 @@
     "ts-node": "^8.0.3"
   },
   "peerDependencies": {
+    "@graphql-typed-document-node/core": "^1.0.0",
```
dahukish, line 49: nice!

PR #2290, opened 2022-05-26
[WIP] Upgrade learna to version 5

vsumner:
## Description

Fixes (issue #)

## Type of change

## Checklist


vsumner:
> Now you've forced a download of 14.19.3, I think the cache should be updated with its presence, so I thiiiink you could change this back to be `14` now and it'd pick up 14.19.3

Oh right! worth a try

BPScott (approved):
🎉

royleejr (approved):

PR #2293, opened 2022-05-30
Fix address mock country response

vsumner:
## Description
This PR fixes the country mock when the country code is canada. For example the canada `fr` locale contains translated province names.

Fixes (issue #)

## Type of change

- [X] `@shopify/address-mocks` Patch: Bug (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/packages/address-mocks/src/index.ts
+++ b/packages/address-mocks/src/index.ts
@@ -44,9 +44,7 @@ export function mockCountryRequests() {
       const countries = fixtures.countries[locale];
       if (operationName === 'countries') return countries;
 
-      const country = countries.data.countries.find(
-        ({code}) => code === countryCode,
-      );
+      const country = fixtures.country[locale].data.country;
       if (!country)
         return {
           errors: [
--- a/packages/address-mocks/src/index.ts
+++ b/packages/address-mocks/src/index.ts
@@ -44,7 +44,10 @@ export function mockCountryRequests() {
       const countries = fixtures.countries[locale];
       if (operationName === 'countries') return countries;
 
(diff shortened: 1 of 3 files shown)


packages/address-mocks/src/index.ts
```diff
@@ -44,9 +44,10 @@ export function mockCountryRequests() {
       const countries = fixtures.countries[locale];
       if (operationName === 'countries') return countries;
 
-      const country = countries.data.countries.find(
-        ({code}) => code === countryCode,
-      );
+      const country =
+        countryCode === 'CA'
```
emileber, line 48: The exception in the fixtures is that only canada is available in french.

Instead of hard-coding this exception, we could dump the french response of the atlas service as `countries_fr.ts` and then, we could remove all `country_ca_xy.ts` files as they would no longer be necessary.

emileber (approved):
I'm approving the changes as it does fix the regression where Canada in french was no longer available in the mocks (🖐️ my bad), though see my suggestion.

packages/address-mocks/src/index.ts
BPScott: This feels like a nicer fix than adding the exception!

BPScott (approved):
Let's ship this so we can unblock web.

@emileber would you be able to look at your suggested tidying in a follow-up PR?

packages/address-mocks/src/index.ts
emileber: > If we were to remove the counyry_* mocks then I think we'd need to adjust what gets reexported

Oh right, true, let's keep them then. It can still be useful when imported as-is.

PR #2295, opened 2022-05-31
Fix object replacement values being cast as strings

emileber:
## Description

New regression uncovered today:

```diff
With text content:
+  "To create this market, countries will be removed from your [object Object],, ,[object Object],, and ,[object Object] markets, and those markets will be deleted. Any settings for these markets will also be permanently deleted."
To contain string:
-  "countries will be removed from your Canada, United States, and United Kingdom markets"
```

I assumed all objects were going to be valid React elements [when I refactored the translation code](https://github.com/Shopify/quilt/pull/2267), but I was wrong as it could be an array of React nodes.

So I fixed the tests and the replacement logic, then **I made sure tests were passing with both the old and the new implementation**.

I also added a bunch of tests for the available format regexes, to make sure we do not break them if we ever update them.

## Type of change

- [x] `@shopify/react-i18n` Patch: Bug (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/packages/react-i18n/CHANGELOG.md
+++ b/packages/react-i18n/CHANGELOG.md
@@ -5,7 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-<!-- ## Unreleased -->
+## Unreleased
+
+### Fixed
+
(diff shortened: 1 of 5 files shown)


vsumner (approved):

emileber:
Is it me or is the CLA status glitched out?

BPScott:
> Is it me or is the CLA status glitched out?

Yeah it does that sometimes. I know you're a shopify employee so I'll use my admin-ness to merge this.

PR #2299, opened 2022-06-03
Added isAndroid method

aeperea:
## Description

Adds a `isAndroid` method that will help identify browser's OS better (the previous alternative was to compare the string of the os === 'Android'

## Type of change
- [x] @shopify/browser Minor: New feature (non-breaking change which adds functionality)

## Checklist


BPScott (approved):

BPScott:
Thanks!

PR #2301, opened 2022-06-07
Fix subtle test issues in react-form and react-form-state

ryanwilsonperkin:
These two test suites weren't always awaiting the promises that got enqueued as part of the actions they were triggering. This can lead to bugs when upgrading to React 18 because of the way that renders are changed and we need to use the correct process in order to ensure that we're ending up testing the right state of the component.

## Description

Fixes (issue #)

## Type of change

- ~[ ]  Patch: Bug (non-breaking change which fixes an issue)~
- ~[ ]  Minor: New feature (non-breaking change which adds functionality)~
- ~[ ]  Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)~

## Checklist

- ~[ ] I have added a changelog entry, prefixed by the type of change noted above (Documentation fix and Test update does not need a changelog as we do not publish new version)~
Not applicable, this only modifies tests

### DIFF
--- a/packages/react-form/src/hooks/tests/form.test.tsx
+++ b/packages/react-form/src/hooks/tests/form.test.tsx
@@ -56,7 +56,6 @@ describe('useForm', () => {
 
       changeTitle(wrapper, 'tortoritos, the chip for turtles!');
       hitSubmit(wrapper);
-      wrapper.forceUpdate();
 
       expect(wrapper).toContainReactComponent('p', {
         children: 'loading...',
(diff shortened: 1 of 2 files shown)


packages/react-form/src/hooks/tests/form.test.tsx
```diff
@@ -43,20 +43,27 @@ describe('useForm', () => {
...
-        <ProductForm
-          data={fakeProduct()}
-          onSubmit={() => new Promise(() => {})}
-        />,
+        <ProductForm data={fakeProduct()} onSubmit={() => submitPromise} />,
       );
 
       changeTitle(wrapper, 'tortoritos, the chip for turtles!');
       hitSubmit(wrapper);
+      wrapper.forceUpdate();
```
heathernfran, line 59: Would this test pass without `wrapper.forceUpdate()`?

heathernfran (approved):

PR #2302, opened 2022-06-08
Upgrade Apollo 2 to Apollo 3

vsumner:
## Description

This PR updates Apollo from version 2 to version 3.

Major version updates for `@shopify/graphql-persisted`, `@shopify/graphql-testing`, `@shopify/react-graphql` and `@shopify/react-graphql-universal-provider`.

See:

- Apollo 3 migration guide: https://www.apollographql.com/docs/react/migrating/apollo-client-3-migration/
- Apollo 3 release blog post: https://www.apollographql.com/blog/announcement/frontend/announcing-the-release-of-apollo-client-3-0/
- Apollo 3 changlog: https://github.com/apollographql/apollo-client/blob/main/CHANGELOG.md

When you update to new major versions, you also need to update `apollo-client` and other `apollo-*` packages to use the new `@apollo/client` package, per the above migration guide.

When using these Apollo 3, we suggest enabling `graphql-typescript-definitions`'s `--export-format=documentWithTypedDocumentNode` option, to allow strict type-checking when using Apollo's own hooks/components.

See changesets for cases where Apollo 2->3 behaviour changes may appear in your code.

### Breaking Changes

- `@shopfiy/graphql-testing`: Remove `unionOrIntersectionTypes` option from `createGraphQLFactory`'s options parameter as that concept has been replaced by `possibleTypes` in Apollo 3. Replace `createGraphQLFactory({unionOrIntersectionTypes: BLAH})` with `createGraphQLFactory({cacheOptions: {possibleTypes: BLAH}})`
- `@shopify/graphql-testing`: Removed the ability to filter by operationName. Replace  passing an `operationName` to filterable options (`graphql.operations.{all,first,last,nth}`),  with a `query` or `mutation` key instead. Replace `graphql.operations.all({operationName: 'BLAH'})` with `graphql.operations.all({query: queryReference}).

### Feature Changes

- `@shopfiy/graphql-testing`: Add `waitForQueryUpdates()` method onto graphql context.

### Known Issues

- When using server-side rendering queries from the extract passes after the first one are not being added to the Apollo cache correctly. This shall result in queries being made on the client instead of using a cache primed from a server-side render.


packages/graphql-testing/src/client.ts
```diff
@@ -1,67 +1,69 @@
...
+//   any
+// >['queryManager']['getCurrentQueryResult'];
+
+// type AnyFetchPolicy =
+//   | SubscriptionOptions['fetchPolicy']
+//   | WatchQueryOptions['fetchPolicy'];
 
 // no-cache and network-only fetch policies cause an initial async event to be
 // produced which is not properly caught inside of a react-testing act scope.
-const invalidFetchPolicies: AnyFetchPolicy[] = ['no-cache', 'network-only'];
```
vsumner, line 19: I temporarily switch the testing client for the real deal, I need to craft a test to figure out if this is still true.

packages/graphql-testing/src/client.ts
```diff
@@ -1,67 +1,69 @@
-import {
-  ApolloClient,
-  SubscriptionOptions,
-  WatchQueryOptions,
-  ApolloClientOptions,
-  ObservableQuery,
-} from 'apollo-client';
+// import {
```
vsumner, line 1: This needs to be removed. It's completely changed in Apollo.

packages/graphql-persisted/src/tests/apollo.test.ts
```diff
@@ -1,6 +1,5 @@
 import faker from '@faker-js/faker/locale/en';
-import gql from 'graphql-tag';
-import {ApolloLink} from 'apollo-link';
+import {ApolloLink, gql} from '@apollo/client';
```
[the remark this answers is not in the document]
vsumner: Done.

BPScott:
Removing `@shopify/react-testing` from the changeset as the only change in that package is a line in a readme gives us a much shorter "changed packages and their dependants are published" list. Double checking if the change in `graphql-typed` is needed would make it even shorter.

packages/graphql-testing/src/graphql-controller.ts
```diff
@@ -1,20 +1,19 @@
...
-import {TestingApolloClient} from './client';
 import {MockLink, InflightLink} from './links';
 import {Operations} from './operations';
 import {operationNameFromFindOptions} from './utilities';
 import {GraphQLMock, MockRequest, FindOptions} from './types';
 
 export interface Options {
-  unionOrIntersectionTypes?: any[];
-  cacheOptions?: ApolloReducerConfig;
+  possibleTypes?: InMemoryCacheConfig['possibleTypes'];
```
BPScott, line 15: Should we get rid of the possibleTypes option here, and just have cacheOptions, and down below do `const cache = new InMemoryCache(cacheOptions);`.

Then consumers would do createGraphQLFactory({cacheOptions: {possibleTypes: BLAH}})`

BPScott:
Rebased atop main. Tests failures are expected as we've merged #2434 into main, which we knew would cause problems in in the apollo3 branch.

mateus (approved):

atesgoral (approved):
🚀

PR #2305, opened 2022-06-13
Support for initial scope in react-i18n

jbalsas:
## Description

This is an exploratory PR that adds support for initial `scope` when using the `@shopify/react-i18n`.

Sparked by this [comment](https://github.com/Shopify/web/pull/64235#discussion_r880259566) by @mike-park:
 
> General point, I'm so not a fan of repeating our prefix n-times in translation. 
> 
> * This directly copies into the final build
> * If the component is renamed/copied, much more code needs adjusting
> 
> If prefer a local helper:
> 
> ```
> const translate = (key) => i18n.translate(`Prefix.${key}`);
> ```

This PR adds support for **passing a scope directly to the `useI18n` hook**:

```js
// Before this PR
const [i18n] = useI18n();
const translated = i18n.translate('MyComponent.key')

// After this PR
const [i18n] = useI18n('MyComponent');
const translated = i18n.translate('key'); 
```

Ideally, I also wanted to acheve:
- 🚫 Simplify most common use case
- 🚫 Allow different scoped translations within the same component

Since none of those are really possible without further modification, **I'm opening this as a draft PR to discuss wether:**
1. **There's value in this PR in itself**
2. **Would we want to pursue any additional changes to simplify usage of `react-i18n`**
3. **Would we want to add support for this in the `babel-plugin` itself**

#### Simplify most common use case

```js
const [{translate}] = useI18n('MyComponent');
```

This is in fact not possible because [I18n is in fact a class](https://github.com/Shopify/quilt/blob/main/packages/react-i18n/src/i18n.ts#L71), so we lose access to the instance when destructuring like that.

I think we could probably reexport bound functions instead of the I18n instance in our hook, to be able to achieve this, but I'm not 100% sure whether this is good or bad 😂 

#### Allow different scoped translations within the same component

```js
const [componentI18n] = useI18n('MyComponent');
const [buttonI18n] = useI18n('MyComponent.MyButton');
const [commonI18n) = useI18n('Common');
```

I was initially worried about the impact this could have on translations loading, but I think that could be okay since [they'd all be registering the same manager id](https://github.com/Shopify/quilt/blob/main/packages/react-i18n/src/hooks.tsx#L77)?

That being said, our babel plugin is rather explicit about [not supporting more than 1 usage of this hook per file](https://github.com/Shopify/quilt/blob/main/packages/react-i18n/src/babel-plugin/index.ts#L65), so this is not something that users can do either.

## Type of change

- [x] react-i18n Minor: New feature (non-breaking change which adds functionality)

## Checklist

### DIFF
--- a/packages/react-i18n/src/hooks.tsx
+++ b/packages/react-i18n/src/hooks.tsx
@@ -6,8 +6,21 @@ import {I18nManager, RegisterOptions} from './manager';
 import {I18nContext, I18nIdsContext, I18nParentContext} from './context';
 
 type Result = [I18n, React.ComponentType<{children: React.ReactNode}>];
+type Scope = string | string[];
 
-export function useI18n(options?: Partial<RegisterOptions>): Result {
+interface HookOptions extends Partial<RegisterOptions> {
+  scope?: Scope;
+}
+
+function isScope(options?: Scope | HookOptions): options is Scope {
+  return typeof options === 'string' || Array.isArray(options);
+}
+
+function normalizeOptions(options?: Scope | HookOptions): HookOptions | undefined {
+  return isScope(options) ? {scope: options} : options;
+}
+
+export function useI18n(options?: Scope | HookOptions): Result {
   const manager = React.useContext(I18nContext);
 
   if (manager == null) {
@@ -16,9 +29,11 @@ export function useI18n(options?: Partial<RegisterOptions>): Result {
     );
   }
 
-  const registerOptions = React.useRef(options);
+  const normalizedOptions = normalizeOptions(options);
 
-  if (shouldRegister(registerOptions.current) !== shouldRegister(options)) {
+  const registerOptions = React.useRef(normalizedOptions);
+
+  if (shouldRegister(registerOptions.current) !== shouldRegister(normalizedOptions)) {
     throw new Error(
       'You switched between providing registration options and not providing them, which is not supported.',
     );
@@ -28,16 +43,16 @@ export function useI18n(options?: Partial<RegisterOptions>): Result {
   // sure that they never switch from the case where `options == null` to `options != null`,
   // so we know that a given use of this hook will only ever hit one of these two cases.
   /* eslint-disable react-hooks/rules-of-hooks */
-  if (options == null) {
+  if (normalizedOptions == null) {
     return useSimpleI18n(manager);
   } else {
-    return useComplexI18n(options, manager);
+    return useComplexI18n(normalizedOptions, manager);
   }
   /* eslint-enable react-hooks/rules-of-hooks */
 }
 
 function useComplexI18n(
-  {id, fallback, translations}: Partial<RegisterOptions>,
+  {id, fallback, translations, scope}: HookOptions,
   manager: I18nManager,
 ): Result {
   const managerRef = React.useRef<I18nManager | null>(null);
@@ -67,7 +82,7 @@ function useComplexI18n(
     unsubscribeRef.current = manager.subscribe(
       ids.current,
       ({translations, loading}, details) => {
-        const newI18n = new I18n(translations, {...details, loading});
+        const newI18n = new I18n(translations, {...details, loading}, scope);
         i18nRef.current = newI18n;
         setI18n(newI18n);
       },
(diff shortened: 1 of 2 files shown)


jbalsas:
> I like your approach! I think it's worth mentioning in your PR description that scope can be string[] ... what is the meaning of that? ['foo', 'bar'] === {foo: {bar: {...}}?

Yes! That's what the underlying infrastructure already supports, scope can be either `string` or `string[]` where `foo.bar.baz` is equivalent to `['foo', 'bar', 'baz']` in providing access to `{foo: {bar: {baz: {...}}}`.

I'm not entirely sure what's the right use case for the array version (assuming something programatically), but thought I would keep it properly typed just in case.

BPScott:
A first glance suggests this is doing a similar thing to https://github.com/Shopify/quilt/pull/2151

Am i reading that right or is this a different feature?

jbalsas:
> A first glance suggests this is doing a similar thing to https://github.com/Shopify/quilt/pull/2151
> 
> Am i reading that right or is this a different feature?

Yeah, I think this is trying to achieve the same thing. My search powers totally failed me on this one 🙈 

Only difference I think is that I tried to use (or maybe abuse) the existing `scope` instead of adding a new `namespace` (which I also considered).

Considering it's been more than 5 months since the last comment there... is there any update you could provide either here or there? Has there anything changed in regards to our positioning on i18n? Are any of these PRs worth exploring, or is there a path for us to contribute in future approaches?

Thanks! 🙏

jbalsas:
Hey @GoodForOneFare, @meko-deng, @TheMallen, since you were originally also involved in [[FED] Add support to namespace on @shopify/react-i18n](https://github.com/Shopify/quilt/pull/2151) alongside @BPScott, would you mind taking a look see if we could provide some closure to both of this initiatives?

cc @pedrodurek, just fyi, but also in case you have something to add! ❤️

BPScott:
> Considering it's been more than 5 months since the last comment on the other issue... is there any update you could provide either here or there? Has there anything changed in regards to our positioning on i18n? Are any of these PRs worth exploring?, or is there a path for us to contribute in future approaches?

I don't think our position has changed much from [the other PR](https://github.com/Shopify/quilt/pull/2151#issuecomment-1026906735) - I don't think I like the idea of adding options that result in additional codepaths to react-i18n as it shall make any static analysis based changes harder in the future. I don't like the idea of adding a feature then rolling it back because it blockages us in the future. I'd prefer keeping consumers doing the current kinda clunky thing over having people have to adopt then drop a feature. That all is kinda FUDdy though, and I'm not living with this friction on a daily basis so I don't have great context on if it's worth it or not.

I think the crux of this issue is that quilt's ownership is somewhat up in the air - the Admin Web Foundations team owns it through grandfathered ownership but most of the current team hasn't worked on these packages. We don't have any time capacity to fully understand the state of the existing library and thus have a good opinion on any future direction, so we're biasing towards leaving it as it is for the most part.


https://github.com/Shopify/web/pull/66000 has been the main progress on how i18n is done, though I'm not sure if has a knock on effect on how you use react-i18n directly.

GoodForOneFare:
There are a couple of philosophies to run with here:
1. "Admin is fundamentally broken, and we should do whatever reduces development friction until a new stack is available"
2. "Admin's current stack is viable, and we should avoid digging holes deeper than they already are"

I lean towards _2_ being true, but I have no context on current performance explorations.  So that's not an informed opinion.  IMO, there's still low-hanging fruit to be grabbed via i18n improvements, and merging this would make that much harder.  

Even on the "fundamentally broken" path, it's unlikely that _everything_ would be thrown away, so this small change could have a large knock-on effect for future stack migrations.

jbalsas:
> Thank you for thinking about this @jbalsas and for considering adding behaviour to the shared library and I hope this rejection doesn't put you off working with us in the future.

No worries at all @BPScott, you won't get rid of me that easily! 😂 I totally understand needing to protect the infrastructure and prepare it for whatever you have outlined for it. 

To build upon something @GoodForOneFare mentioned though, I think that if friction at this level is not taken care of, its users will eventually find inventive solutions outside of your control (even forks) and potentially lose trust in the foundations and its ability to solve their higher-level problems. 

I wouldn't mind pitching in some work in the right direction if you can at some point share some outline/roadmap of things that you thing need to be done.

Thank you both for all your work to provide a solid and viable stack and your time analyzing this issue! 🤗

PR #2308, opened 2022-06-16
Fix issues with depending on not-current versions

BPScott:
## Description

We had some cases where packages had peer/dev dependencies on
not-the-current-version. This fixes them

- Remove address-mocks' outdated devDependency on jest-dom-mocks as it is
  already has a direct dependency on the current version
- Update graphql-fixtures and graphql-typescript-definitions' peer
  dependency on graphql-typed to include the current version

This fixes [warning output we were seeing when investigating implementing changesets](https://github.com/Shopify/quilt/runs/6785343134?check_suite_focus=true#step:4:7):

## Type of change

address-mocks / graphql-fixtures / graphql-typescript-definitions: patch

### DIFF
--- a/packages/graphql-fixtures/package.json
+++ b/packages/graphql-fixtures/package.json
@@ -30,7 +30,7 @@
     "graphql-tool-utilities": "^3.0.0"
   },
   "peerDependencies": {
-    "graphql-typed": ">=0.6.1 <2.0.0"
+    "graphql-typed": "^0.6.1 || ^0.7.0 || ^1.0.0 || ^2.0.0"
   },
   "files": [
     "build/",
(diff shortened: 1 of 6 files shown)


royleejr (approved):

packages/graphql-fixtures/package.json
```diff
@@ -30,7 +30,7 @@
     "graphql-tool-utilities": "^3.0.0"
   },
   "peerDependencies": {
-    "graphql-typed": ">=0.6.1 <2.0.0"
+    "graphql-typed": "^0.6.1 || ^0.7.0 || ^1.0.0 || ^2.0.0"
```
BPScott, line 33: The versions tab on https://www.npmjs.com/package/graphql-typed shows that 0.6.x an 0.7.x were the only pre-v1 releases greater that 0.6.1, so this maintains the existing set of range, while adding `^2.0.0`

PR #2310, opened 2022-06-17
Remove on_response hook preventing headers from being proxied to the client

joshuamsager:
## Description

React applications leveraging: `useHeader` to set response headers were not being propagated to the client. This is because configuring an `on_response` callback overwrites the default one configured [here](https://github.com/axsuul/rails-reverse-proxy/blob/master/lib/reverse_proxy/controller.rb#L9). This hook is responsible for proxying the headers along to the client.

This regression was first introduced [here](https://github.com/Shopify/quilt/pull/853).

## Type of change

- [x] quilt_rails Patch: Bug (non-breaking change which fixes an issue)

## Checklist

### DIFF
--- a/.changeset/little-eagles-glow.md
+++ b/.changeset/little-eagles-glow.md
@@ -1,5 +1,0 @@
----
-'@shopify/address-consts': patch
----
-
-No changes, this release is to test the new release process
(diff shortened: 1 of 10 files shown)


ryanwilsonperkin (approved):
Thanks for hunting this down and fixing it 👍🏻

Issue #2311, opened 2022-06-17
[graphql-fixtures] Seed collision for nested arrays on fill


Flufd:
## Overview

When using the graphQL filler to generate data with nested arrays, a `faker` seed collision occurs when generating the data for the elements in the arrays. 

This is because the `seedFromKeypath` generates a seed from the keypath of the node that is being filled, based on the collective sum of all of the characters that make up that keypath. 

Example:
Given this schema:
```gql
type Pet {
  id: ID!
}

type Person {
  pets: [Pet]!
}

type Query {
  people: [Person!]!
}
```

If we have two people with two pets each, the pets at `person[0].pet[1]` and `person[1].pet[0]` have the same seed generated, because their "sum" of the chars in the keypath is the same. Because of this, their IDs are the same.

Failing test here:
https://github.com/Shopify/quilt/compare/graphql-fixtures/seed-collision

This becomes a problematic issue when combined with the Apollo cache. By default, Apollo will normalise the items based on their `id` fields and so we get a cache collision and the test data we receive in a call to the Apollo client does not match the data we passed to the `createGraphQLFactory` result.

Potentially we could use some other stable method for generating the seed based on the keypath. Maybe some quick hash of the string? https://gist.github.com/hyamamoto/fd435505d29ebfa3d9716fd2be8d42f0

BPScott:
This crops up with a single entity too, you don't even need nesting for this to fail. These key paths will result in the same seed: `person[12]` and `person[21]`

PR #2312, opened 2022-06-17
[graphql-fixtures] Graphql fixtures/seed collision fix

Flufd:
## Description

Fixes https://github.com/Shopify/quilt/issues/2311

Uses a quick hash of the strings that make up a graphQL object's keypath to generate a random seed for that object, instead of a sum of all of the characters values.

I think this is a breaking change, only if people are asserting on the random values that are already generated.

TODO: changelog, once I know if this is considered breaking.

## Type of change

## Checklist

### DIFF
--- a/packages/graphql-fixtures/src/fill.ts
+++ b/packages/graphql-fixtures/src/fill.ts
@@ -495,14 +495,14 @@ function randomEnumValue(enumType: GraphQLEnumType) {
 }
 
 function seedFromKeypath(keypath: string[]) {
-  return keypath.reduce<number>((sum, key) => sum + seedFromKey(key), 0);
+  return keypath.reduce<number>((sum, key) => hashCode(key, sum), 0);
 }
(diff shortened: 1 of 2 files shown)


BPScott (commented):
Looking good! I've just added a commit to remove the need for an `as any` cast in the test.

I think this is a minor change. End users shouldn't be relying on the randomly returned filler data. If there's some data that they want to fix then they should specify it.

You can cut a beta release and test this in web by commenting `/snapit` on the PR.

Add a changeset (`yarn add changeset` and follow instructions). and I think we're good to go. See https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md for more info.

BPScott (changes requested):
Marking this as blocked, till we update web to not rely on this buggy behaviour that we're removing.

We'll also want to update this to be a major change in the changeset

Issue #1157, opened 2019-10-30
`refetch` from `useQuery` does not update data after a network error in React Native


josephroquedev:
## Overview

When using the `useQuery` hook in a React Native environment, I am experiencing an issue where `refetch` fails to update `data` and does not cause a re-render.

Reproduction:

1. Perform a query in a React Native environment, that succeeds.
2. Disconnect the data or turn off the server so the query will error out, then call `refetch` from the `useQuery` hook.
3. The query will fail and `data` will become `undefined`
4. Reconnect (or restart the server) and call `refetch`. If you watch the network, the query **will** succeed and will fetch the new data, but the `useQuery` hook and `data` will not be updated, and the component will not re-render.

Might be related to: https://github.com/Shopify/quilt/issues/821

There's also a similar `react-apollo` issue that was fixed in the past: https://github.com/apollographql/react-apollo/issues/2070 but forcing this version does not fix in this repo.

## Consuming repo

> What repo were you working in when this issue occurred?
https://github.com/shopify/arrive-client

warefhaque:
Having the exact same issue in https://github.com/Shopify/start-mobile. Has a workaround been found for this?

sbalay:
Bump

gavin-hinchley:
This issue affects `React` more generally also: `useQuery()` does not update after `refetch()` is called following `any graphql error`. 

This seems to be a bug within [@shopify/react-graphql -> useQuery()](https://github.com/Shopify/quilt/blob/main/packages/react-graphql/src/hooks/query.ts#L21) as `@apollo/react-hooks -> useQuery()` works as expected.

vsumner:
I've identified the problem and will have a fix shortly.

PR #2320, opened 2022-06-21
Fix refetch error and add tests

vsumner:
## Description
This PR updates `useQuery` hook to be able to recover from network errors. Previously if `refetch` encountered an error it would be able to fetch the request, but would not re-render. This PR address that bug and also updates `graphql-testing` with tests for error states and updates it's documentation.

Fixes (issue #) #1157

## Type of change

## Checklist

### DIFF
--- a/packages/react-graphql/src/hooks/query.ts
+++ b/packages/react-graphql/src/hooks/query.ts
@@ -119,17 +119,42 @@ export default function useQuery<
       return;
     }
 
+    let subscription: ZenObservable.Subscription | undefined;
+
     const invalidateCurrentResult = () => {
       setResponseId((x) => x + 1);
     };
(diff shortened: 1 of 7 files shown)


znja (approved):

packages/react-graphql/src/hooks/query.ts
```diff
@@ -119,17 +119,42 @@ export default function useQuery<
       return;
     }
 
+    let subscription: ZenObservable.Subscription | undefined;
```
BPScott, line 122: Belated review - It threw me that we don't import from zen-observable anywhere. I think we could lean on a bit of type inference to avoid having the direct dependency on zen-observable in the package.json:

```
let subscription: ReturnType<typeof queryObservable['subscribe']> | undefined;
```

PR #2324, opened 2022-06-21
Changeset updates

BPScott:
## Description

Changeset docs and tidying

- Add release/snapit documentation
- Remove js shipit stacks
- Update snapshot tag to be 'snapshot' insead of 'snapshot-release'
- Go back to using the default GITHUB_TOKEN as a personal access token
  is not worth it. Educating releasers that absent CI checks on version PRs is expected is easier than having to manage PAT tokens.

## Whats all this about skipping CI checks and PATs?

As noted in the release doc, pull requests / commits authenticated with the default automatically provided GITHUB_TOKEN secret do not trigger downstream actions. The "Version Packages" PR and the commits therein are created with this token, which means there our CI actions do not get triggered. These CI actions are marked as required, so you end up with these checks being marked as required but never sent:

image (1)

That's not ideal. I've tried to come up with a good way around this, and everything is worse than saying "yeah we know, use repo administrator privileges to force the merge even though those required checks are missing". Oh well.

### Alternative "Fix" 1: Use a Personal Access Token (PAT)

By uploading a Personal Access Token and using its credentials when pushing commits the it should be possible to cause the  pull_request events to trigger, and thus make the CI checks run and be associated with the commits.

However this means that now I have to manually cycle an access token whenever it expires. Having a long expiry is a bad idea as it means that if the token gets compromised then somebody can act as me without my knowledge. I'd have to do this in every repo we want to use changesets in (quilt, web-configs, sewing-kit, loom) and that overhead multiplies quickly. I don't want to be a single point of failure, and having to teach folks "every few months you need to update this token across many repos" is a terrible experience.

### Alternative "Fix 2: Use some common token 

Riffing on the PAT idea, what if there was an organisation level secret that we could use? That'd be great but it does not exist. I've spoke to code-scale about this - they're understandably against using service accounts (having been moving away from that idea for years).  There was some ideas about them building some token rotator that could update something in an org-level secret that many repos could have access too (kinda like the NPM_TOKEN), but that'll require a lot of working. It's a nice idea for the future but it isn't ready now.

### Alternative "Fix" 3: Trigger the CI actions manually

I thought it might be possible to trigger the CI actions manually. I couldn't get it work. I suspect that there's a difference between "this workflow is ran" and "this adds a check against a specific commit". I could do the former but not the latter.

### Alternative "Fix" 4: Remove those checks from being "required checks"

Uugh, no. I don't want to remove the notion of "these are required" on PRs made by humans. These being required makes sure that PRs can't be merged in the cases where webhooks are delayed and it takes a while for the CI actions are triggered. I could not see a way of marking "commits by this person can bypass the checks requirement". 

### Current state of the world

I think telling people "yes we know these checks shall be absent on the Version PR, but we think that is ok and you should use your admin-ness to merge this PR" is better than "having personal access tokens across multiple repositories" or "losing 'required' marks on PRs made by humans".


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -3,23 +3,5 @@
 Fixes (issue #)
 
 <!--
-Please include a summary of what you want to achieve in this pull request. Remember to indicate the affected package(s).
+Please include a summary of what you want to achieve in this pull request. Remember to add a changeset that indicates the affected package(s) and if they are major / minor / patch changes by using `yarn changeset`.
```
ryanwilsonperkin, line 6: `add a changeset`
I'd suggest link-ifying this section of text and have it point to the section in our release guidelines that talks about changesets. It'll be new tech to a bunch of folks, and knowing how developers can get nerd-sniped I'd expect that a bunch will bail out of the sentence at this point to go read about what changesets are. Linking them will help us control where they find that info instead of starting with the public docs on changesets.

documentation/guides/release-and-deploy.md
```diff
@@ -1,63 +1,30 @@
...
-```sh
-git checkout main && git pull
-```
+Upon merging PRs with a changeset entry, it shall create a "Version Packages" PR ([example](https://github.com/Shopify/quilt/pull/2309)) that shall contain any changeset changes since the last release.
 
-**Note🗒️** `git pull` should be used instead of `git pull origin main` to ensure that tags are pulled as well.
+To perform a release:
 
-## 2. Versioning and Tagging
+- Find the currently open "Version Packages" PR
```
ryanwilsonperkin, line 13:
```suggestion
- Find the [currently open "Version Packages" PR](https://github.com/Shopify/quilt/pulls?q=is%3Apr+is%3Aopen+author%3Aapp%2Fgithub-actions+%22Version+Packages%22)
```

documentation/guides/release-and-deploy.md
```diff
@@ -1,63 +1,30 @@
...
+- Note that by default all required CI checks are absent. This is expected as the pull request and commit are created using the default GITHUB_TOKEN access token, which [skips downstream events to prevent infinite loops](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#triggering-a-workflow-from-a-workflow).
+- Merge the PR by checking the `Merge without waiting for requirements to be met (administrators only)` checkbox and choosing `Squash and merge`. We consider skipping the required checks to be acceptable as we are confident that these automated updates of `package.json` and `CHANGELOG.md` would not cause CI to fail.
 
-- Begin the release process:
+The `Release` action shall run on the merge commit on the `main` branch, and shall publish the npm packages and create a GitHub tag and release for each package that is referenced in the PR. You can find the action log by looking at the release commit status on the merge commit.
 
-```sh
-yarn version-bump
-```
+## Performing a snapshot releaes
```
ryanwilsonperkin, line 19:
```suggestion
## Performing a snapshot release
```

documentation/guides/release-and-deploy.md
```diff
@@ -1,63 +1,30 @@
...
-- Hit the Deploy button on your Publish commit in Shipit to publish your beta release to npm
-
-### Step 2 - Consume the release
-
-- Add your release to a repository that uses the package you are testing:
-
-```sh
-yarn add --dev @shopify/my-package@0.29.10-my-feature-beta.1
-```
+This functionality is only available in PRs that point to branch in the `Shopify/quilt` repository - PRs from forks are not supported.
```
ryanwilsonperkin, line 30:
```suggestion
This functionality is only available in PRs that point to a branch in the `Shopify/quilt` repository - PRs from forks are not supported.
```

shipit.beta-js.yml
```diff
@@ -1,11 +0,0 @@
-ci:
```
ryanwilsonperkin, line 1: 😎 🔥

ryanwilsonperkin (approved):
Some minor nit picks but this looks great :)

royleejr (approved):

BPScott:
Turns out explicitly setting the GITHUB_TOKEN was important - https://github.com/Shopify/quilt/commit/6a91db721fbec7de60c7d7ab8432c09397b24a5c

PR #2325, opened 2022-06-22
Update snapit.yml to avoid Uncaught ReferenceError

aaronccasanova:


BPScott:
Thanks! I'm wrangling changeset followup / doc improvements in #2324 - I'll add this change to that PR

PR #2330, opened 2022-06-27
expose i18n getShortCurrencySymbol interface

brianshen1990:
## Description

Expose `getShortCurrencySymbol()` from i18n so that consumers can get the short symbol quickly instead of doing it themselves.

### DIFF
--- a/.changeset/spotty-buttons-poke.md
+++ b/.changeset/spotty-buttons-poke.md
@@ -1,0 +1,5 @@
+---
+'@shopify/react-i18n': patch
+---
+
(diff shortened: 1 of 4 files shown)


jkeohan (approved):
looks good...could we also have someone from the @Shopify/admin-web-foundations  review/approve?

BPScott (commented):
Code looks great. One note about the changeset version - I think that should be minor rather than patch as we're introducing new functionality rather than fixing a bug.

Once that is fixed then feel free to merge, and then we can get #2451 rebased

PR #2331, opened 2022-06-29
Introduce DateStyle.DateTime style for formatting date and time

depl0y:
### Fixes:
- https://github.com/Shopify/retail-international/issues/730
- https://github.com/Shopify/android-point-of-sale/issues/7356

### Why did you choose this approach?

This PR adds a way to format a date and time in one go. This allows projects using this library to format date/times as described in [this polaris doc](https://polaris.shopify.com/foundations/content/grammar-and-mechanics#date).

For the method signature, I tried to stay as close to the `formatDate` method as possible, but I did want to make a separation between the two (could use some guidance here).

### What reviewers should look at

- How is the changeset looking?
- The translations key is `date.humanize.datetime`, should this be in the humanize category or could I use the `lessThanOneYearAgo` translation? (that feels wrong, but contains the right text).
- Is it acceptable to have a separate method for this, or should this be included in the `formatDate` method, with a different `DateStyle` ?
- The `formatDateTime` method has a single `options` parameter which will be split up in time and date options in the method itself. This might need some more explanation in the method description?
- Wrote a couple of tests to ensure localisation of month names and date formats would work. Are these enough tests?

### DIFF
--- a/.changeset/dry-tools-wink.md
+++ b/.changeset/dry-tools-wink.md
@@ -1,0 +1,5 @@
+---
+'@shopify/react-i18n': minor
+---
+
+Add `DateStyle.DateTime`. to format dates in the style `Jun 12, 2022 at 10:34 pm`
--- a/packages/react-i18n/README.md
+++ b/packages/react-i18n/README.md
@@ -141,6 +141,7 @@ The provided `i18n` object exposes many useful methods for internationalizing yo
   - `DateStyle.Short`: e.g., `Dec 20, 2012`
   - `DateStyle.Humanize`: Adheres to [Polaris guidelines for dates with times](https://polaris.shopify.com/content/grammar-and-mechanics#section-dates-numbers-and-measurements), e.g., `Just now`, `3 minutes ago`, `4 hours ago`, `10:35 am`, `Yesterday at 10:35 am`, `Friday at 10:35 am`, or `Dec 20 at 10:35 am`, or `Dec 20, 2012`
   - `DateStyle.Time`: e.g., `11:00 AM`
+  - `DateStyle.DateTime`: Formats date and time separately and uses the translation string `date.humanize.lessThanOneYearAway` to format it according to this [Polaris guideline](https://polaris.shopify.com/foundations/content/grammar-and-mechanics#date), e.g. `Jun 12, 2022 at 10:34 pm`.
 - `weekStartDay()`: returns start day of the week according to the country.
 - `getCurrencySymbol()`: returns the currency symbol according to the currency code and locale.
 - `formatName()`: formats a name (first name and/or last name) according to the locale. e,g
--- a/packages/react-i18n/src/constants/index.ts
+++ b/packages/react-i18n/src/constants/index.ts
@@ -3,6 +3,7 @@ export enum DateStyle {
   Short = 'Short',
   Humanize = 'Humanize',
   Time = 'Time',
+  DateTime = 'DateTime',
 }
 
 export const dateStyle = {
@@ -25,6 +26,13 @@ export const dateStyle = {
   [DateStyle.Time]: {
     hour: '2-digit',
     minute: '2-digit',
+  },
+  [DateStyle.DateTime]: {
+    month: 'short',
+    day: 'numeric',
+    year: 'numeric',
+    hour: 'numeric',
+    minute: 'numeric',
   },
 } as const;
 
--- a/packages/react-i18n/src/i18n.ts
+++ b/packages/react-i18n/src/i18n.ts
@@ -308,9 +308,18 @@ export class I18n {
     const {style = undefined, ...formatOptions} = options || {};
 
     if (style) {
-      return style === DateStyle.Humanize
-        ? this.humanizeDate(date, {...formatOptions, timeZone})
-        : this.formatDate(date, {...formatOptions, ...dateStyle[style]});
+      switch (style) {
+        case DateStyle.Humanize:
+          return this.humanizeDate(date, {...formatOptions, timeZone});
+        case DateStyle.DateTime:
+          return this.formatDateTime(date, {
+            ...formatOptions,
+            timeZone,
+            ...dateStyle[style],
+          });
+        default:
+          return this.formatDate(date, {...formatOptions, ...dateStyle[style]});
+      }
     }
 
     return formatDate(date, locale, {...formatOptions, timeZone});
@@ -490,6 +499,25 @@ export class I18n {
     }
   }
 
+  private formatDateTime(
+    date: Date,
+    options: Intl.DateTimeFormatOptions,
+  ): string {
+    const {defaultTimezone} = this;
+    const {timeZone = defaultTimezone} = options;
+
+    return this.translate('date.humanize.lessThanOneYearAway', {
+      date: this.getDateFromDate(date, {
+        ...options,
+        timeZone,
+      }),
+      time: this.getTimeFromDate(date, {
+        ...options,
+        timeZone,
+      }),
+    });
+  }
+
   private humanizePastDate(date: Date, options?: Intl.DateTimeFormatOptions) {
     if (isLessThanOneMinuteAgo(date)) {
       return this.translate('date.humanize.lessThanOneMinuteAgo');
@@ -590,6 +618,34 @@ export class I18n {
     const zoneMatchGroup = /\s([\w()+\-:.]+$)/.exec(hourZone);
 
     return zoneMatchGroup ? zoneMatchGroup[1] : '';
+  }
+
+  private getDateFromDate(date: Date, options?: Intl.DateTimeFormatOptions) {
+    const {
+      localeMatcher,
+      formatMatcher,
+      weekday,
+      day,
+      month,
+      year,
+      era,
+      timeZone,
+      timeZoneName,
+    } = options || {};
+
+    const formattedDate = this.formatDate(date, {
+      localeMatcher,
+      formatMatcher,
+      weekday,
+      day,
+      month,
+      year,
+      era,
+      timeZone,
+      timeZoneName: timeZoneName === 'short' ? undefined : timeZoneName,
+    });
+
+    return formattedDate;
   }
 
   private getTimeFromDate(date: Date, options?: Intl.DateTimeFormatOptions) {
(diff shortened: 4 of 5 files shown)


packages/react-i18n/src/tests/i18n.test.ts
```diff
@@ -1277,6 +1277,163 @@ describe('I18n', () => {
...
+        minute: 'numeric',
+        hour12: false,
+      });
+
+      expect(translate).toHaveBeenCalledWith(
+        'date.humanize.datetime',
+        {
+          pseudotranslate: false,
+          replacements: {
+            date: 'Jan 1, 2018',
```
carly-lee, line 1427: As far as I know the date format in the UK is also `day-month-year`.

depl0y: Fixed the test to use the right locale

packages/react-i18n/src/i18n.ts
```diff
@@ -316,6 +316,45 @@ export class I18n {
...
+   *
+   * Polaris describes the format that should be used for date/time combinations here:
+   * https://polaris.shopify.com/foundations/content/grammar-and-mechanics#date
+   *
+   * @param {Date} date Date object to format
+   * @param {Intl.DateTimeFormatOptions} [options={}] Options for both date and time formatting
+   * @return {*}  {string} A formatted string
+   * @memberof I18n
+   */
+  formatDateTime(date: Date, options: Intl.DateTimeFormatOptions = {}): string {
```
trishrempel, line 331: As you mention, this approach requires people adding a new `date.humanize.datetime` format, which would require translations to connect the date and the time.

Since [INTL.DateTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat) provides the concatenation automatically if you specify a date and a time, I suggest instead that you continue to use `formatDate`, but add a new `DateStyle` enum [here](https://github.com/Shopify/quilt/blob/main/packages/react-i18n/src/constants/index.ts#L8), such as `DateStyle.LongDateTime` or `DateStyle.Full`, which would return a `DateStyle.Long` date with a time, like `Thursday, October 15, 2015 at 2:00 PM` in English.

It would be worth testing in different locales, but this might be a good format:
```
  [DateStyle.Full]: {
    weekday: 'long',
    month: 'long',
    day: 'numeric',
    year: 'numeric',
    hour: 'numeric',
    minute: 'numeric',
  },
```

It also looks like this might work, but the first is more explicit:
```
  [DateStyle.Full]: {
    dateStyle: 'full',
    timeStyle: 'short',
  },
```

depl0y: Updated it to be to style below, but as explained on Slack, it is still using a translation. This because the Intl libraries do not concat date and time correctly. They separate it using a comma, while Polaris describes that they should be separated by the word `at`.

In the latest version, it still uses a different translation key as the humanize methods already available in this library.

```
  [DateStyle.LongDateTime]: {
    month: 'short',
    day: 'numeric',
    year: 'numeric',
    hour: 'numeric',
    minute: 'numeric',
  },
```

packages/react-i18n/README.md
```diff
@@ -141,6 +141,7 @@ The provided `i18n` object exposes many useful methods for internationalizing yo
   - `DateStyle.Short`: e.g., `Dec 20, 2012`
   - `DateStyle.Humanize`: Adheres to [Polaris guidelines for dates with times](https://polaris.shopify.com/content/grammar-and-mechanics#section-dates-numbers-and-measurements), e.g., `Just now`, `3 minutes ago`, `4 hours ago`, `10:35 am`, `Yesterday at 10:35 am`, `Friday at 10:35 am`, or `Dec 20 at 10:35 am`, or `Dec 20, 2012`
   - `DateStyle.Time`: e.g., `11:00 AM`
+- `formatDateTime()`: formats a date and time according to the Polaris guidlines for [dates with times](https://polaris.shopify.com/foundations/content/grammar-and-mechanics#date).
```
trishrempel, line 144: Remember to update this to be another enum description like `DateStyle.ShortDateTime` and also include the key that needs to be translated to go with it.

packages/react-i18n/src/i18n.ts
```diff
@@ -490,6 +499,48 @@ export class I18n {
...
+      weekday,
+      era,
+      year,
+      month,
+      day,
+      timeZoneName,
+      ...timeOptions
+    } = options;
+
+    return this.translate('date.humanize.datetime', {
```
trishrempel, line 532: I suggest possibly reusing `date.humanize.lessThanOneYearAgo` - for existing users that already have this key translated, it will mean not having to add another key.

BPScott (commented):
Good stuff so far!

Something that I've not actually saw so far in this PR or the issues: Can somebody provide an example of the format that is desired in plain text?

---

Questions from the PR:


> How is the changeset looking?

Changesets are used in changelogs which are consumed by users of this project. They should describe how the public API has changed. "Introduce formatDateTime method" is not useful to consumers because it is a private method that they can't ever call directly. You should frame this and the README documentation based on how people interact with the library.

> The translations key is date.humanize.datetime, should this be in the humanize category or could I use the lessThanOneYearAgo translation? (that feels wrong, but contains the right text).

I think I'm leaning towards reusing the `lessThanOneYearAgo` label. It means consumers don't need to add a new translation key. @trishrempel is my spirit guide to when reusing translation keys is appropriate, I'd value their input on this.


> Is it acceptable to have a separate method for this, or should this be included in the formatDate method, with a different DateStyle ?

I don't think it is worth creating a second method for this - `formatDate` can already output times if you use `DateStyle.Humanize` - "Yesterday at 10:30pm" or `DateStyle.Time`. Let's lean into how we can already control formats - by adding a new DateStyle type.


> The formatDateTime method has a single options parameter which will be split up in time and date options in the method itself. This might need some more explanation in the method description?

It's going to be a private method, I wouldn't worry too much having exhaustive documentation for it

> Wrote a couple of tests to ensure localisation of month names and date formats would work. Are these enough tests?

It looks like you've got 5 tests that all do the same thing, with the exception of changing the locale/country. This could probably be made much shorter by removing the duplication by using an `it.each()`

packages/react-i18n/src/constants/index.ts
```diff
@@ -3,9 +3,17 @@ export enum DateStyle {
   Short = 'Short',
   Humanize = 'Humanize',
   Time = 'Time',
+  LongDateTime = 'LongDateTime',
 }
 
 export const dateStyle = {
+  [DateStyle.LongDateTime]: {
```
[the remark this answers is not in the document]
depl0y: What if I keep the weekday in there, and we name it `DateTime`. So that is in between long and short and maybe resonates what I am trying to do here. For my purposes I can just remove the weekday in the format I want?

As mentioned below for the `README.md`, I can include text that it uses that translation key and what the type of results are.

packages/react-i18n/README.md
depl0y: 👍 Sounds great! This was a relic from back in the day when I had it defined as a public method instead of a DateStyle 😉 .

packages/react-i18n/src/i18n.ts
depl0y: 👍 Agreed!

packages/react-i18n/src/i18n.ts
```diff
@@ -490,6 +499,48 @@ export class I18n {
...
+      date: this.formatDate(date, {
```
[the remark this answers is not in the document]
depl0y: Naming things is hard. The method actually removes time options from the options object and makes sure we just return the date elements. So maybe `getDateFromDate` sounds cryptic, but is exactly what it does? 😄

packages/react-i18n/src/tests/i18n.test.ts
```diff
@@ -1277,6 +1277,142 @@ describe('I18n', () => {
     });
   });
 
+  describe('#formatDateTime()', () => {
```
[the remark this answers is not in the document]
depl0y: Updated tests to fit into the other tests. Also uses a cleaner `each` way of testing.

packages/react-i18n/README.md
```diff
@@ -141,6 +141,7 @@ The provided `i18n` object exposes many useful methods for internationalizing yo
   - `DateStyle.Short`: e.g., `Dec 20, 2012`
   - `DateStyle.Humanize`: Adheres to [Polaris guidelines for dates with times](https://polaris.shopify.com/content/grammar-and-mechanics#section-dates-numbers-and-measurements), e.g., `Just now`, `3 minutes ago`, `4 hours ago`, `10:35 am`, `Yesterday at 10:35 am`, `Friday at 10:35 am`, or `Dec 20 at 10:35 am`, or `Dec 20, 2012`
   - `DateStyle.Time`: e.g., `11:00 AM`
+  - `DateStyle.DateTime`: Formats date and time separately and uses a translation string to format it correctly according to this [Polaris guideline](https://polaris.shopify.com/foundations/content/grammar-and-mechanics#date), e.g. `Jun 12, 2022 at 10:34 pm`.
```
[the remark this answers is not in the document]
depl0y: Great! I included your suggestion in this.

carly-lee (approved):

packages/react-i18n/src/constants/index.ts
```diff
@@ -3,9 +3,17 @@ export enum DateStyle {
   Short = 'Short',
   Humanize = 'Humanize',
   Time = 'Time',
+  DateTime = 'DateTime',
 }
 
 export const dateStyle = {
+  [DateStyle.DateTime]: {
```
BPScott, line 10: Silly nitpick: could we keep he ordering of the keys in this const and in DateStyle the same?

.changeset/dry-tools-wink.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/react-i18n': minor
+---
+
+Introduce DateStyle.DateTime.
```
BPScott, line 5: This will go in the changelog, can we provide a better description that mentions what DateStyle.DateTime does?

```suggestion
Add `DateStyle.DateTime`. to format dates in the style `Jun 12, 2022 at 10:34 pm`
```

Or something like that.

BPScott (approved):
Code changes look great to me! Two little nitpicks, then this is good to go!

packages/react-i18n/src/constants/index.ts
depl0y: Yes we should, updated the code to reflect this 👍

.changeset/dry-tools-wink.md
depl0y: Used your suggestion for this change, thanks!

PR #2333, opened 2022-06-29
`@shopify/react-graphql` - Prevent re render on error

vsumner:
## Description
This PR fixes a bug introduced in my previous PR.

Problem: On subscribing to the query observable the `onError` handler gets triggered again causing a new responseId to be set which in turn caused a re-render. This PR saves the previous error and compares them to ensure that they are not equal before setting a new response id.

Fixes (issue #)


packages/react-graphql/package.json
```diff
@@ -35,7 +35,8 @@
     "apollo-cache-inmemory": ">=1.0.0 <2.0.0",
     "apollo-client": ">=2.0.0 <3.0.0",
     "apollo-link": ">=1.0.0 <2.0.0",
-    "graphql-typed": "^2.0.0"
+    "graphql-typed": "^2.0.0",
+    "fast-deep-equal": "^3.1.3"
```
vsumner, line 39: This is used already in `quilt` so I'm not introducing a new dependency

PR #2337, opened 2022-06-30
Update node-mocks-http

atesgoral:
## Description

Updates node-mocks-http to pick up a fix to `response.getHeaders()` (it should be a shallow copy).

https://github.com/howardabrams/node-mocks-http/pull/217

### DIFF
--- a/.changeset/beige-steaks-divide.md
+++ b/.changeset/beige-steaks-divide.md
@@ -1,0 +1,11 @@
+---
+'@shopify/jest-koa-mocks': major
+---
+
+Update node-mocks-http, fixing response.getHeaders() behaviour
+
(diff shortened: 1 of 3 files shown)


BPScott (approved):
On the surface a version range bump from `^1.5.8` to `^1.11.0` doesn't sound like a major release. It's not bumping up a major version, and many consumers could already be using node-mocks-http v1.11.0 already, as that version satisfies both the old and new dependency ranges.

On slack you made the argument that the knock-on effects warrant a major bump - consumers could be relying on the broken behaviour present in v1.5.8 and thus produce tests would break in this update, and the update to ^v1.11.0 shall force consumers to address these mistakes - and version bumps that don't apply cleanly should be major version bumps.

PR #2340, opened 2022-07-04
Add DeepReadonly type

melnikov-s:
## Description

Added a `DeepReadonly` type to `useful-types`. Plan to use this type to make the cached response from `@shopify/react-graphql` readonly to catch any accidental mutations. This type can also be useful for other purposes in application code.

### DIFF
--- a/packages/useful-types/src/types.ts
+++ b/packages/useful-types/src/types.ts
@@ -86,3 +86,34 @@ export type PartialSome<T, K extends keyof T> = Omit<T, K> &
   Partial<Pick<T, K>>;
 export type RequireSome<T, K extends keyof T> = Omit<T, K> &
   Required<Pick<T, K>>;
+
+// Reference https://github.com/ts-essentials/ts-essentials/blob/5aa1f264e77fb36bb3f673c49f00927c7c181a7f/lib/types.ts
+type IsAny<T> = 0 extends 1 & T ? true : false;
+
+type IsUnknown<T> = IsAny<T> extends true
+  ? false
+  : unknown extends T
+  ? true
+  : false;
+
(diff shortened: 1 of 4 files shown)


packages/useful-types/src/types.ts
```diff
@@ -86,3 +86,34 @@ export type PartialSome<T, K extends keyof T> = Omit<T, K> &
...
+  ? ReadonlySet<DeepReadonly<U>>
+  : T extends WeakSet<infer U>
+  ? WeakSet<DeepReadonly<U>>
+  : T extends Promise<infer U>
+  ? Promise<DeepReadonly<U>>
+  : T extends {}
+  ? {readonly [K in keyof T]: DeepReadonly<T[K]>}
+  : IsUnknown<T> extends true
+  ? unknown
+  : Readonly<T>;
```
sslotsky, line 119: Given how big of a doozie this is, I'm _really_ glad we were able to lift it from `ts-essentials` instead of figuring it out on our own 😅

sslotsky (approved):
👏🏼 👏🏼 👏🏼 👏🏼 👏🏼 👏🏼 👏🏼 👏🏼 

Hi, I've been writing JS for 13+ years and I just mutated a cache by using an assignment operator instead of an equality check as the condition for a filter function. I APPROVE THIS MESSAGE!

BPScott (commented):
Coul you add some tests in `packages/useful-types/test-d/types.test-d.ts` please?

BPScott:
> I'm not sure how some of these tests can be exercised with tsd

I think using expectType would do the job for innerObj, map and set keys - I've just pushed up a commit. How does that look to you?

PR #2350, opened 2022-07-15
:wrench: @Shopify/react-html make script tag  more configurable

michaelnesen:
## Description

The `<Hml>` component in [@shopify/react-html](https://github.com/Shopify/quilt/tree/main/packages/react-html#html-) sets up a script tag with the [following default attributes](https://github.com/Shopify/quilt/blob/main/packages/react-html/src/server/components/Html.tsx#L121-L132). This component is only available from the server entrypoint of the [@shopify/react-html/server](https://github.com/Shopify/quilt/tree/main/packages/react-server) package and when using it we can't configure a script tag with the attributes that we need/don't need. 

Fixes (issue #)

This PR makes the script tag a little more configurable and allows you to disable [crossOrigin](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin) and add the [async attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#attributes) to the script. It does not change any default values of the script, it only adds additional props to be able to pass in and change the attributes.

### DIFF
--- a/.changeset/pretty-rings-destroy.md
+++ b/.changeset/pretty-rings-destroy.md
@@ -1,0 +1,5 @@
+---
+'@shopify/react-html': minor
+---
+
+Makes script tag more configurable in the <Html> component
--- a/packages/react-html/src/server/components/Html.tsx
+++ b/packages/react-html/src/server/components/Html.tsx
@@ -18,6 +18,8 @@ export interface Asset {
 
 export interface ScriptAsset extends Asset {
   type?: 'module' | 'nomodule' | 'script';
+  disableCrossOrigin?: boolean;
+  async?: boolean | undefined;
 }
 
 export interface InlineStyle {
@@ -107,25 +109,31 @@ export default function Html({
   });
 
   const blockingScriptsMarkup = blockingScripts.map((script) => {
+    const crossOrigin = script.disableCrossOrigin ? undefined : 'anonymous';
+
     return (
       <Script
         key={script.path}
         src={script.path}
         integrity={script.integrity}
         type={script.type}
-        crossOrigin="anonymous"
+        crossOrigin={crossOrigin}
+        async={script.async}
       />
     );
   });
 
   const deferredScriptsMarkup = scripts.map((script) => {
+    const crossOrigin = script.disableCrossOrigin ? undefined : 'anonymous';
+
     return (
       <Script
         key={script.path}
         src={script.path}
         integrity={script.integrity}
         type={script.type}
-        crossOrigin="anonymous"
+        crossOrigin={crossOrigin}
+        async={script.async}
         defer
       />
     );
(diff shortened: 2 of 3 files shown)


ryanwilsonperkin (commented):
The change to add `async` seems like a good improvement and one that we should adopt, there are others too such as `defer` which makes me wonder whether it might make more sense to extend React's built-in types for attributes of the HTML script tag.

The disabling for crossorigin though seems a bit clunky, and like a potential way to add insecure code. Could you tell us a bit more about why you're looking to disable anonymous crossorigin requests? We may be able to help identify a safer way of achieving your goal.

ryanwilsonperkin:
Another approach mentioned by the `react-devtools` package seems like it might be a better fit here and optionally included behind a check of `process.env.NODE_ENV === "production"`

> If you install react-devtools as a project dependency, you may also replace the <script> suggested above with a JavaScript import (import 'react-devtools'). It is important that this import comes before any other imports in your app (especially before react-dom). Make sure to remove the import before deploying to production, as it carries a large DevTools client with it. If you use Webpack and have control over its configuration, you could alternatively add 'react-devtools' as the first item in the entry array of the development-only configuration, and then you wouldn’t need to deal either with <script> tags or import statements.

ryanwilsonperkin:
> I tried this approach initially but didn't have any luck and was getting some errors 😕

Can you tell us a bit more about the errors you were facing with this? I'd like us to pursue this direction a bit more before committing to supporting this functionality to `react-html`

michaelnesen:
> Can you tell us a bit more about the errors you were facing with this? I'd like us to pursue this direction a bit more before committing to supporting this functionality to `react-html`

I was getting a `500` error coming from `react-dom` . In the docs it says it's important that the `react-devtools` import comes before any other imports in your app (especially before react-dom). I made sure to import it at the very top of our component tree but was still getting this error 🤔 

Screen Shot 2022-07-15 at 2 24 05 PM

michaelnesen:
@ryanwilsonperkin We have tried a few approaches before this one (injecting script into app, replacing the script with a `import react-devtools` , as well as adding it to our webpack config but nothing seems to work as we keep getting the same error. I believe it's because we are using `quilt_rails` and behind the scenes it is "magically" setting up our app and injecting/importing `react-devtools` after the app is loaded throws this error as the devtools needs to come before any other imports in your app basically.

michaelnesen:
@ryanwilsonperkin Appreciate the support, I actually just noticed there is a `headMarkup` [prop](https://github.com/Shopify/quilt/tree/main/packages/react-html#html-) on the `<Html />` component that lets you add additional JSX to be embedded in the head of the document 🙈 So either way regardless of the approach we take to connect `react-devtools` the change in this PR isn't necessary

PR #2352, opened 2022-07-18
[react-testing] Release act promises when root wrapper is destroyed

melnikov-s:
## Description

With React 18 a hanging promises returned from `act` can prevent other updates from occurring and since that happens on a global scope it will affect subsequent tests within the same test suite. This PR attempts to free up the act queue by resolving any  unfulfilled promises when the root wrapper is destroyed. 

Within our internal tests we typically have `destroyAll` within a `afterEach` which loops through all active wrappers and calls `destroy` on them. The change with this PR is that now `destoryAll` will have to be awaited so that any active promises holding up the `act` can be resolved.

This PR also adds a warning when attempting to call `wrapper.act` on an already destroyed root, this can happen if for example a `graphQL` context is shared between wrappers. Eventually we'd want to clean those up.

Note: this also eliminates the need to have something like
```
await act(async () => {
    await flushPromises();
  });
```
in a `afterEach`

### DIFF
--- a/.changeset/wise-apples-call.md
+++ b/.changeset/wise-apples-call.md
@@ -1,0 +1,7 @@
+---
+'@shopify/react-testing': major
+---
+
+WHAT: Release act promises when root wrapper is destroyed
+WHY: To prevent unresolved promises from failing subsequent act calls
+HOW: `destroyAll` in `afterEach` will now have to be awaited as well as `Root.prototype.destroy`
--- a/packages/react-testing/src/root.tsx
+++ b/packages/react-testing/src/root.tsx
@@ -74,6 +74,9 @@ export class Root<Props> implements Node<Props> {
   private reactRoot: ReactRoot | null = null;
   private root: Element<Props> | null = null;
   private acting = false;
+  private destroyed = false;
+  private actPromise!: Promise<void>;
+  private actPromiseResolver!: () => void;
 
   private render: Render;
   private resolveRoot: ResolveRoot;
@@ -88,10 +91,20 @@ export class Root<Props> implements Node<Props> {
   ) {
     this.render = render;
     this.resolveRoot = resolveRoot;
+    this.actPromise = new Promise((resolve) => {
+      this.actPromiseResolver = resolve;
+    });
     this.mount();
   }
 
   act<T>(action: () => T, {update = true} = {}): T {
+    if (this.destroyed) {
+      // eslint-disable-next-line no-console
+      console.error(
+        'Warning: attempting to perform an act on a destroyed root. This can lead to state changes not being applied',
+      );
+    }
+
     const updateWrapper = update ? this.update.bind(this) : noop;
     let result!: T;
 
@@ -123,7 +136,14 @@ export class Root<Props> implements Node<Props> {
       });
 
       if (isPromise(result)) {
-        return result as unknown as Promise<void>;
+        if (this.destroyed) {
+          return result as unknown as Promise<void>;
+        }
+
+        return Promise.race([
+          result,
+          this.actPromise,
+        ]) as unknown as Promise<void>;
       }
 
       return undefined as unknown as Promise<void>;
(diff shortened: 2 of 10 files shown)


melnikov-s:
Sorry for the delay, I ran these changes against web and everything passes without any additional changes in web. CI run here: https://buildkite.com/shopify/web-ci-builder/builds/522692 . I think we can keep it as a minor version bump.

packages/react-testing/src/root.tsx
```diff
@@ -88,10 +91,20 @@ export class Root<Props> implements Node<Props> {
...
     this.render = render;
     this.resolveRoot = resolveRoot;
+    this.actingPromise = new Promise((resolve) => {
+      this.actingPromiseResolver = resolve;
+    });
     this.mount();
   }
 
   act<T>(action: () => T, {update = true} = {}): T {
+    if (this.destroyed) {
```
ryanwilsonperkin, line 101: What purpose is the `destroyed` boolean serving that we don't currently get with the `withRoot` behaviour? It seems that whenever `destroyed` gets set to true, root will also be made unavailable, so we should be able to rely on that existing pattern. In most (all?) cases, act is already prefaced with a `withRoot` check

packages/react-testing/src/root.tsx
```diff
@@ -257,15 +277,20 @@ export class Root<Props> implements Node<Props> {
...
     }
-
     element.remove();
     connected.delete(this);
+    this.destroyed = true;
+    this.actingPromiseResolver();
+    // flush the micro task and then the macro task to wait until react commits all pending updates.
+    // flushing of the macro task is only required for React 17
+    await this.actingPromise;
+    await new Promise((resolve) => setTimeout(resolve, 0));
```
ryanwilsonperkin, line 293: Is this last line really needed? It'd seem that the whole original promise chain for `act` should be effectively "canceled" at this point since all of the races would now resolve.

ryanwilsonperkin (commented):
Excellent work, thank you for digging into this and I'm sorry it took me so long to get back to you with a review. I've left some comments here about the approach taken, I'm inclined to agree with @BPScott too that this should be a major release, since many folks will have a synchronous call to `destroy` in their cleanup code which would now be broken without an `await`.

packages/react-testing/src/root.tsx
melnikov-s: This is mostly for backwards compatibility. I wanted existing code which relied on being able to perform `act` on a destroyed root (see reply above) to function as it did before. 

As for re-using `withRoot` the issue with that is `this.root` being null is set as both the initial value and when it is destroyed and I need to make a distinction between un-initialized vs destroyed thus the new property.

packages/react-testing/src/root.tsx
melnikov-s: React 17 was failing with the added tests and that seemed to have resolved it. I'm going to take a deeper look to see what's actually going on there.

melnikov-s: Didn't look like it was needed, I've removed it from here and moved it to the test instead to prevent React 17 from failing

packages/react-testing/src/root.tsx
```diff
@@ -88,10 +91,20 @@ export class Root<Props> implements Node<Props> {
...
+    });
     this.mount();
   }
 
   act<T>(action: () => T, {update = true} = {}): T {
+    if (this.destroyed) {
+      // eslint-disable-next-line no-console
+      console.error(
+        'Warning: attempting to perform an act on a destroyed root. This can lead to state changes not being applied',
+      );
```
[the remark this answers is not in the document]
ryanwilsonperkin: Ah! Okay I see what you mean, that seems like an unfortunate limitation of the graphql-testing library, it'd be nice for us to have something like an "unwrap" functionality as well that could be configured to remove those wrappers, or turn them into no-ops here. We can use the warning for now, but I suspect we'll want another major release in the near future that turns that to an explicit error and encourages a pattern for not reusing root in graphql testing

.changeset/wise-apples-call.md
```diff
@@ -0,0 +1,7 @@
+---
+'@shopify/react-testing': major
+---
+
+WHAT: Release act promises when root wrapper is destroyed
+WHY: To prevent unresolved promises from failing subsequent act calls
+HOW: `destroyAll` in `afterEach` will now have to be awaited
```
ryanwilsonperkin, line 7: This should also call out that `destroy` is now async (not just `destroyAll`) since it's also a public API of Root

packages/react-testing/src/root.tsx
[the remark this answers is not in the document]
melnikov-s: Yeah good call! An unwrap on `graphql-testing` with an additional `beforeUnmount` lifecycle on `createMount` will do the trick here

PR #2356, opened 2022-07-20
Added types to triggerKeyPath

melnikov-s:
## Description

This PR makes the `triggerKeypath` method on `root` and `element` fully typed when `Props` are not `unknown`. Does this using typescript string literal types which were introduced in typescript 4.1.

As a result of this change, prettier also needed to be updated as the existing version was unable to parse string literal types. If preferred I can split that up into it's own PR. 

I've ran the typecheck with these new types for `web` and it only resulted in ~5~ 14 type errors, all of which appeared to be legitimate. 

type checking performance is a concern as this is a pretty complex type. I ran type-checking with `--force` and `--diagnostics` and both branches and here are the results after I addressed the failing types: 

Main: 
[typecheck-web-main.txt](https://github.com/Shopify/quilt/files/9183022/typecheck-web-main.txt)

With `triggerKeypath` typed:
[typecheck-with-triggerKeypath.txt](https://github.com/Shopify/quilt/files/9183028/typecheck-with-triggerKeypath.txt)

# Details

The overall goal is given a keypath string determine if it can lead to a function type. The keypath is a `.` separated string with each part representing a property that is to be looked up on the `Props` type. 

The first step is to split the keypath into a list of strings  but before that the keypath needs to be normalized as it can take many forms such as `propetry.0.anotherProp` or `property[0]anotherProp` or `property[0].anotherProp` etc, This is done in `type NormalizeKeypath`. Once normalized as a `.` separated value we use `type Split` to turn it into a list of property strings. 

The list of properties along with the `Props` is then used in `type ExtractProperty` to recursively iterate through each property on the `Props` and nested objects.  If the property does not exist on the object we short circuit with a `never` type.  Properties on array types are always assumed to be there as it cannot be determined if they will be defined or not at compile time. Once the final type is resolved form the `Props` and list of properties, we change the parameters on `triggerKeyPath` to be `never` if the property was not resolved to a function which is what creates the type error. Special exceptions are made for when the `Props` is `unknown` or when the path is a generic `string` type and not a string literal in which case we effectively turn off type checking and revert to the previous type behaviour. 

The final piece is dealing with unions  to ensure that a union type can still pass as a valid key path without the developer having to manually narrow the type or do casting. For example a keypath of `property.onAction` should be valid on:

```typescript
{
   property: {
     onAction: Function
   } | {
      foo: 'bar
   }
}
```

The idea here is that if the keypath can potentially lead to a valid function it should pass the types. This is achieved by stripping out any non-object types from each iteration of `type ExtractProperty` as well as merging object types into a single type using `type Merge<T>`

### DIFF
--- a/packages/react-testing/src/element.ts
+++ b/packages/react-testing/src/element.ts
@@ -10,9 +10,13 @@ import {
   PropsFor,
   UnknowablePropsFor,
   DebugOptions,
+  TriggerKeypathParams,
+  TriggerKeypathReturn,
+  KeyPathFunction,
+  ExtractKeypath,
 } from './types';
 
-type Root = import('./root').Root<any>;
+type Root = import('./root').Root<unknown>;
 
 interface Tree<Props> {
   tag: Tag;
@@ -193,17 +197,15 @@ export class Element<Props> implements Node<Props> {
   trigger<K extends FunctionKeys<Props>>(
     prop: K,
     ...args: DeepPartialArguments<Props[K]>
-  ): ReturnType<
-    NonNullable<
-      Props[K] extends ((...args: any[]) => any) | undefined ? Props[K] : never
-    >
-  > {
+  ): NonNullable<Props[K]> extends (...args: any[]) => any
+    ? ReturnType<NonNullable<Props[K]>>
+    : never {
     return this.root.act(() => {
       const propValue = this.props[prop];
 
       if (propValue == null) {
         throw new Error(
-          `Attempted to call prop ${prop} but it was not defined.`,
+          `Attempted to call prop ${String(prop)} but it was not defined.`,
         );
       }
 
@@ -211,8 +213,14 @@ export class Element<Props> implements Node<Props> {
     });
   }
 
-  triggerKeypath<T = unknown>(keypath: string, ...args: unknown[]): T {
+  triggerKeypath<
+    Path extends string,
+    ExtractedFunction extends KeyPathFunction = ExtractKeypath<Props, Path>,
+  >(
+    ...params: TriggerKeypathParams<Props, Path, ExtractedFunction>
+  ): TriggerKeypathReturn<Props, Path, ExtractedFunction> {
     return this.root.act(() => {
+      const [keypath, ...args] = params;
       const {props} = this;
       const parts = keypath.split(/[.[\]]/g).filter(Boolean);
 
--- a/packages/react-testing/src/root.tsx
+++ b/packages/react-testing/src/root.tsx
@@ -15,6 +15,10 @@ import {
   DeepPartialArguments,
   PropsFor,
   DebugOptions,
+  TriggerKeypathParams,
+  TriggerKeypathReturn,
+  KeyPathFunction,
+  ExtractKeypath,
 } from './types';
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires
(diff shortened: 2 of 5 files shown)


BPScott:
Can you move the prettier update into its own PR? Lets use the latest version 2.7.1 instead of only updating part of the way.

melnikov-s:
> Can you move the prettier update into its own PR? Lets use the latest version 2.7.1 instead of only updating part of the way.

https://github.com/Shopify/quilt/pull/2360

packages/react-testing/src/element.ts
```diff
@@ -193,26 +197,30 @@ export class Element<Props> implements Node<Props> {
...
       return (propValue as any)(...args);
     });
   }
 
-  triggerKeypath<T = unknown>(keypath: string, ...args: unknown[]): T {
+  triggerKeypath<
+    Path extends string,
+    ExtractedFunction extends KeyPathFunction = ExtractKeypath<Props, Path>,
+  >(
+    ...params: TriggerKeypathParams<Props, Path, ExtractedFunction>
```
BPScott, line 220: Having a `keypath` argument here ensured that users must specify at least one argument. By removing the keypath argument, you've made it so `element.triggerKeypath();` is considered valid code from a type checking perspective.

`element.triggerKeypath();` should raise a type error as the first value - the keypath is required. Please retain that behaviour.

packages/react-testing/src/root.tsx
```diff
@@ -201,16 +205,19 @@ export class Root<Props> implements Node<Props> {
   trigger<K extends FunctionKeys<Props>>(
     prop: K,
     ...args: DeepPartialArguments<Props[K]>
-  ): ReturnType<
-    NonNullable<
-      Props[K] extends ((...args: any[]) => any) | undefined ? Props[K] : never
-    >
-  > {
+  ): NonNullable<Props[K]> extends (...args: any[]) => any
```
BPScott, line 208: Same as prior comment - why has the typing to trigger needed to change?

packages/react-testing/src/root.tsx
```diff
@@ -201,16 +205,19 @@ export class Root<Props> implements Node<Props> {
...
+    ? ReturnType<NonNullable<Props[K]>>
+    : any {
     return this.withRoot((root) => root.trigger(prop, ...(args as any)));
   }
 
-  triggerKeypath<T = unknown>(keypath: string, ...args: unknown[]) {
-    return this.withRoot((root) => root.triggerKeypath<T>(keypath, ...args));
+  triggerKeypath<
```
BPScott, line 214: Same as previous comment - Having a `keypath` argument here ensured that users must specify at least one argument. By removing the keypath argument, you've made it so `element.triggerKeypath();` is considered valid code from a type checking perspective.

`element.triggerKeypath();` should raise a type error as the first value - the keypath is required. Please retain that behaviour.

packages/react-testing/src/tests/element.test.tsx
```diff
@@ -588,7 +588,7 @@ describe('Element', () => {
         defaultRoot,
       );
 
-      expect(() => element.triggerKeypath('actions')).toThrow(
+      expect(() => element.triggerKeypath('actions' as never)).toThrow(
```
BPScott, line 591: I think it'd be better to use a `@ts-expect-error` directive to describe why this line is not considered valid typescript rather than casting it to a nonsense value.

```suggestion
        // @ts-expect-error actions is not valid parameter, because it does not point to a function
      expect(() => element.triggerKeypath('actions')).toThrow(
```

---

In the test block above this we have:

```ts
expect(() => element.triggerKeypath('actions[1]onAction')).toThrow(
  /Attempted to access field keypath 'actions\.1', but it was not an object/,
);
```

I would expect that would also a require a similar treatment as it also throws an error, thanks to passing an invalid string parameter . `actions[1]onAction` and `actions` are both invalid strings to pass into `triggerKeypath` - I would expect both of them to raise a type error instead of only `actions`.

packages/react-testing/src/types.ts
```diff
@@ -118,12 +220,15 @@ export interface Node<Props> {
   trigger<K extends FunctionKeys<Props>>(
     prop: K,
     ...args: DeepPartialArguments<Props[K]>
-  ): ReturnType<
```
BPScott, line 121: Another case where we're modifying the trigger path. Rational has been reqested above.

packages/react-testing/src/element.ts
melnikov-s: It does raise an error as expected. `TriggerKeypathParams` is a tuple with a required value in position `0`

packages/react-testing/src/element.ts
```diff
@@ -193,26 +197,30 @@ export class Element<Props> implements Node<Props> {
...
     prop: K,
     ...args: DeepPartialArguments<Props[K]>
-  ): ReturnType<
-    NonNullable<
-      Props[K] extends ((...args: any[]) => any) | undefined ? Props[K] : never
-    >
-  > {
+  ): NonNullable<Props[K]> extends (...args: any[]) => any
+    ? ReturnType<NonNullable<Props[K]>>
+    : any {
```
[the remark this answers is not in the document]
melnikov-s: Yeah I should have pointed that out. I was getting some weird issues where ts was complaining about `trigger` even though it was never touched. Perhaps some kind of bug? Anyway changing it to `any` resolved that and doesn't change type safety as there's already a constraint to ensure that the string prop we pass in resolves to a function so it will never be resolved to `any` anyway. 

I tried changing it back to `never` just now and I no longer see the type error, so I have no idea what that was. I've reverted it back to `never` though.

melnikov-s:
Thanks Ben! 

Here's the PR for web to comply with the new types

https://github.com/Shopify/web/pull/71399

Once the new `react-testing` version is pushed to npm, I'll update `package.json` in web and get it out of draft.

PR #2359, opened 2022-07-25
[Admin FED Infra] Reduce console logs spam in quilt test runs

### LINKED ISSUE #2157: Reduce console.log spam in quilt test runs
## Overview

Running the unit tests is currently very noisy. There's a lot of console.log noise that you need to scroll through, which makes discovering failures harder than it needs to be. `CI=1 yarn run test` produces 1700 lines of noise, 200 of which are "PASS" lines. So about 7/8s of the output is noise.

## Task

Identify tests that produce noisy output and fix them:

- if the logspam is complaining about something then fix the issue it's complaining about
- if the logs are a required part of the test then mock console.log in the test so it doesn't output.

List of tests that are currently causing console.logs:

- [x] packages/logger/src/formatters/tests/ConsoleFormatter.test.ts
- [x] packages/react-async/src/tests/component.test.tsx
- [x] packages/react-async/src/tests/provider.test.tsx
- [x] packages/react-bugsnag/src/tests/Bugsnag.test.tsx
- [x] packages/react-csrf/src/tests/hooks.test.tsx
- [x] packages/react-form/src/hooks/list/tests/baselist.test.tsx
- [x] packages/react-form/src/hooks/list/tests/dynamiclist.test.tsx
- [x] packages/react-form-state/src/components/tests/List.test.tsx
- [x] packages/react-form-state/src/components/tests/Nested.test.tsx
- [x] packages/react-form-state/src/tests/FormState.test.tsx
- [x] packages/react-graphql/src/async/tests/component.test.tsx
- [x] packages/react-graphql-universal-provider/src/tests/GraphQLUniversalProvider.test.tsx
- [x] packages/react-html/src/server/components/tests/Html.test.tsx
- [x] packages/react-import-remote/src/tests/ImportRemote.test.tsx
- [x] packages/react-network/src/tests/NetworkUniversalProvider.test.tsx
- [x] packages/react-router/src/components/Router/tests/Router.test.tsx
- [x] packages/react-server/src/render/tests/render.test.tsx
- [x] packages/react-server/src/server/tests/server.test.tsx
- [x] packages/react-testing/src/tests/root.test.tsx
- [x] packages/storybook-a11y-test/tests/index.test.ts
- [x] packages/web-worker/src/tests/e2e.test.ts

melodyhabbouche:
## Description

Fixes #2157 

Compare how many logs there were [here](https://github.com/Shopify/quilt/runs/7505068849?check_suite_focus=true), over 1700 lines of extra console logs of noise, down to the 200 pass lines we need [here](https://github.com/Shopify/quilt/runs/7521561071?check_suite_focus=true). 

Amazing issue logged by @BPScott 🙌

### DIFF
--- a/packages/logger/src/formatters/tests/ConsoleFormatter.test.ts
+++ b/packages/logger/src/formatters/tests/ConsoleFormatter.test.ts
@@ -2,6 +2,22 @@ import {ConsoleFormatter} from '../ConsoleFormatter';
 import {LogLevel} from '../..';
 
 describe('ConsoleFormatter', () => {
+  let consoleErrorSpy: jest.SpyInstance;
+  let consoleWarnSpy: jest.SpyInstance;
+  let consoleLogSpy: jest.SpyInstance;
+
+  beforeEach(() => {
+    consoleErrorSpy = jest.spyOn(console, 'error').mockImplementation();
+    consoleWarnSpy = jest.spyOn(console, 'warn').mockImplementation();
+    consoleLogSpy = jest.spyOn(console, 'log').mockImplementation();
+  });
+
+  afterEach(() => {
+    consoleErrorSpy.mockRestore();
+    consoleWarnSpy.mockRestore();
+    consoleLogSpy.mockRestore();
+  });
+
   it('logs Critical entries to console.error', () => {
     const formatter = new ConsoleFormatter();
     const errorSpy = jest.spyOn(console, 'error');
--- a/packages/react-async/src/tests/component.test.tsx
+++ b/packages/react-async/src/tests/component.test.tsx
@@ -43,15 +43,19 @@ function Loading() {
 }
 
 describe('createAsyncComponent()', () => {
+  let consoleErrorSpy: jest.SpyInstance;
+
   beforeEach(() => {
     requestIdleCallback.mock();
     intersectionObserver.mock();
+    consoleErrorSpy = jest.spyOn(console, 'error').mockImplementation();
   });
 
   afterEach(() => {
     requestIdleCallback.cancelIdleCallbacks();
     requestIdleCallback.restore();
     intersectionObserver.restore();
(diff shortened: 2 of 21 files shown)


packages/logger/src/formatters/tests/ConsoleFormatter.test.ts
```diff
@@ -2,6 +2,22 @@ import {ConsoleFormatter} from '../ConsoleFormatter';
...
 
 describe('ConsoleFormatter', () => {
+  let consoleErrorSpy: jest.SpyInstance;
+  let consoleWarnSpy: jest.SpyInstance;
+  let consoleLogSpy: jest.SpyInstance;
+
+  beforeEach(() => {
+    consoleErrorSpy = jest.spyOn(console, 'error').mockImplementation();
+    consoleWarnSpy = jest.spyOn(console, 'warn').mockImplementation();
+    consoleLogSpy = jest.spyOn(console, 'log').mockImplementation();
```
melodyhabbouche, line 12: I wonder if there's a way to do this is one line?

packages/logger/src/formatters/tests/ConsoleFormatter.test.ts
```diff
@@ -2,6 +2,22 @@ import {ConsoleFormatter} from '../ConsoleFormatter';
...
+  });
```
melodyhabbouche, line 13: I wonder if there's a way to do this is one line?

BPScott: Maybe, but I think having an explicit line per function is very easy to read at a glance, I think trying to turn this into a one liner would impact that readability.

BPScott (approved):
This looks great to me!

Add a changeset - run `yarn changeset` and then this should be good to go

I like that we're only mocking what we need mock in these cases rather than blanket suppressing all console messages all the time.

PR #2360, opened 2022-07-25
Bumped prettier to latest (2.7.1)

melnikov-s:
## Description

Bumps prettier to latest version

Extracted from: https://github.com/Shopify/quilt/pull/2356


BPScott (approved):
I added a small commit to only disable the problematic eslint rule in intersection-observer.ts for just the lines where it is needed rather than the whole file

PR #2362, opened 2022-07-29
Fix loading projects with relative paths

morenobryan:
## Description

Fixes the loading of projects with a relative `schemaPath`.

The issue could be reproduced on `arrive-client` by running `yarn generate-gql-minis-types-only`. It accesses [.graphqlconfig.yml](https://github.com/Shopify/arrive-client/blob/master/src/mini-apps/.graphqlconfig.yml) which defines the project path with a relative path. That causes a path mismatch between `schemaPath` and the concatenation of `project.dirpath` and `project.schema`. 

Resolving the schema path there, as seen above in the same file, fixed the issue for me on `arrive-client`, and I believe it will cover other cases as well.


BPScott (approved):
Change looks good. I think this is a patch change as this is fixing a bug, rather than a minor - it fixes an issue rather than adds a new feature. I'll update that then merge.

Thanks for the PR!

Issue #2367, opened 2022-08-08
[graphql-config-utilities] Support multiple `schema` definitions (+ globs)


notjosh:
## Overview

`graphql-config` updated their spec for the `schema` field (since v3? v4?) to allow [multiple files and globs](https://www.graphql-config.com/docs/user/schema#multiple-files).

Our current model within `graphql-config-utilities` assumes a single schema file.

## Motivation

On Shop app, we're using [`graphql-typescript-definitions`](https://github.com/Shopify/quilt/tree/main/packages/graphql-typescript-definitions) to generate the types from schema files. We use multiple schema files, so currently we have to run a script to merge our schema files. I was investigating skipping this skip and keeping multiple schemas when I ran into this problem.

I currently have a pretty simple patch as a workaround for multiple files, but haven't extended it to globs. It works, but it's a pretty large patch, so I'd rather it were fixed upstream.

One alternative might be ditching `graphql-config-utilities` from within `graphql-typescript-definitions`, and just using `graphql-config` directly to do the config parsing. Superficially they have a fairly comparable API surface, so this shouldn't be too hard.

Similarly, we could just keep `graphql-config-utilities` as a thin wrapper around `graphql-config`, adding a couple of functions as enhancements if we're using it frequently across packages. This shouldn't impact existing packages, and reduces our code to maintain.

(Or maybe both? We can put `graphql-config-utilities` into a deprecated state, but update it for any current consumers) 

I can carve out some time for a PR if you have an opinions on the best direction to go here.

Thoughts?

---

## Checklist

- [x] Please delete the labels section before submitting your issue
- [x] I have described this issue in a way that is actionable (if possible)

BPScott:
[...]

PR #2374, opened 2022-08-15
[rfc] [graphql-typescript-definitions] Update to handle multiple schemas

notjosh:
## Description

Fixes https://github.com/Shopify/quilt/issues/2367

This PR does two things:

- updates `graphql-typescript-definitions` to handle multiple schemas (both via array or glob), via `graphql-config` calls
- moves `graphql-typescript-definitions` toward not needing to depend on `graphql-config-utilities`

Most of the references to `graphql-config-utilities` have been removed in favour of calling equivalent methods on `graphql-config`. Most of them are one-liner replacements, with a couple of small helpers added.

The remaining references to `graphql-config-utilities` are for backward compatibility, as our usage of legacy configs isn't 100% backward compatible with modern `graphql-config` - i.e. `documents` is now used in favour of `includes` to refer to operations/fragments, but `project.getDocuments()` won't fall back to referring to `includes` in legacy projects, so we need to call `getGraphQLProjectIncludedFilePaths()` and `getIncludesExcludesFromConfig()`.

❓ We could fully remove `graphql-config-utilities` by inlining+modernising the two functions we use. None of our products depend on `graphql-config-utilities` directly, so it could be removed from Quilt. Thoughts?

❓ We could remove support for legacy configs entirely, and force legacy projects to update their config. `graphql-config@3` is 2 years old now, so I don't think it's completely unreasonable. However, I think plenty of teams still use the legacy format, so it would be a barrier to adopting new versions. Thoughts?


packages/graphql-typescript-definitions/loom.config.ts
```diff
@@ -5,5 +5,9 @@ import {quiltPackage} from '../../config/loom';
 export default createPackage((pkg) => {
   pkg.entry({root: './src/index.ts'});
   pkg.binary({name: 'graphql-typescript-definitions', root: './src/cli.ts'});
-  pkg.use(quiltPackage());
+  pkg.use(
+    quiltPackage({
+      jestEnv: 'node',
```
notjosh, line 10: Works around this error: https://github.com/prisma/prisma/issues/8558

packages/graphql-typescript-definitions/src/index.ts
```diff
@@ -407,11 +399,16 @@ export class Builder extends EventEmitter {
...
-      this.emit('error', error);
+    } catch (error) {
+      if (error instanceof Error) {
+        const wrapped = new Error(
+          `Error in ${
+            file.path
+          }: ${error.message[0].toLowerCase()}${error.message.slice(1)}`,
+        );
+        this.emit('error', wrapped);
+      }
```
notjosh, line 410: TypeScript was angry, so this became a little more generic.

BPScott (commented):
Nice start! This suggests that migrating this to handle `documents` and multiple schemas in addition to `include`/`exclude` keys wouldn't be toooo complex.

I think there's some confusion over the notion of "legacy". graphql-config papers over the notion of legacy names (includes / excludes plural, replacing them with their singular versions), we don't need to do anything to handle them so there's nothing to drop support for.


We need to maintain support for using the `include` and `exclude` keys - as they are not legacy. We need to combine the "documents" and "include/exclude" arrays not pick one or the other. I've detailed more about this below.

I think this got confusing because our fixtures use the legacy names  (includes / excludes plural, schemaPath), but we lack any fixtures that use the modern names (include / exclude singular, schema), and thus you you conflated the idea of "include/exclude" as being a legacy concept when it is not.

BPScott:
> ❓ We could fully remove `graphql-config-utilities` by inlining+modernising the two functions we use. None of our products depend on `graphql-config-utilities` directly, so it could be removed from Quilt. Thoughts?

I like this long-term direction. In order to fully deprecate and remove graphql-config-utilities we'd need to also remove it from `graphql-fixtures` in quilt, and sewing-kit according to https://grokt.shopify.io/results?q=graphql-config-utilities+-r%3AShopify%2Fquilt
It feels like we're getting close to doing too many things at once - are we trying to re


> ❓ We could remove support for legacy configs entirely, and force legacy projects to update their config. `graphql-config@3` is 2 years old now, so I don't think it's completely unreasonable. However, I think plenty of teams still use the legacy format, so it would be a barrier to adopting new versions. Thoughts?

Per stuff above, we don't deal with the normalisation from legacy to modern config styles - graphql-config handles that. We still need to maintain support for the modern `include` / `exclude` keys

PR #2375, opened 2022-08-15
Expand tests for errors and allow errorPolicy to be set on useMutation

vsumner:
## Description
This PR will now correctly set the `errorPolicy` when it is passed to `useMutation`. I also expanded test coverage of `useMutation` and `useQuery` error states.

Fixes (issue #)


CameronGorrie (approved):
👍

PR #2381, opened 2022-08-19
[rfc] [graphql-typescript-definitions] Add configuration option for read-only types, and module imports

notjosh:
## Description

Hiya 👋

Over at Shop, we have a GraphQL project defined that points to a schema within `node_modules`. It's admittedly a bit of an edge case, but we want to be able to generate the operations and fragments in our project, but rely on the provided schema and types within the NPM module. (It is assumed the types in the NPM module has been generated via `graphql-typescript-definitions`, and are therefore compatible to be consumed in our generated documents).

This PR updates the configuration (with legacy support), so that a project can be configured like:

```javascript
// .graphqlrc.js
module.exports = {
  projects: {
    server: {
      schema: 'node_modules/some_module/src/schema.graphql',
      documents: 'src/**/*.graphql',
      extensions: {
      schemaTypes: {
        path: 'node_modules/some_module/src/types/',
        import: 'some_module/types',
        readOnly: true,
      }
    }
  }
}
```

The previous config option of `schemaTypesPath` has been turned into an object (`schemaTypes`), so we can add more config:

- `path`: equivalent to the old `schemaTypesPath`
- `import`: a fixed-string used in generated `.graphql.d.ts` files
- `readOnly`: to configure whether to write generated types into `schemaTypesPath`. This could maybe use some better naming, as we'll still write `.graphql.d.ts` document files, but not the `.ts` schema type files.

The configuration example above would, for example, take a file at `src/data/SomeQuery.graphql` and generate:

```typescript
import { DocumentNode } from "graphql-typed";
import { TypeA, TypeB, TypeC } from "some_module/types";
...
```

(and, of course, the contents of `node_modules/some_module/src/types/` would not be overwritten).

I'm open to suggestions on more general ways of solving this use case, but I think this approach offers value outside of just our case, i.e. users of Babel's `module-resolver` can alias their imports, so we don't see instances of `import { ... } from '../../../../../../../data/graqhql-types'` - instead it might look like `'import { ... } from 'data/graqhql-types/`.

Fixes (issue #)

### DIFF
--- a/packages/graphql-typescript-definitions/src/index.ts
+++ b/packages/graphql-typescript-definitions/src/index.ts
@@ -48,6 +48,7 @@ export {AbstractGraphQLFilesystem, EnumFormat, ExportFormat};
 export interface Options extends PrintDocumentOptions, PrintSchemaOptions {
   addTypename: boolean;
   schemaTypesPath: string;
+  schemaTypesReadOnly?: boolean;
   config?: GraphQLConfig;
   getProjectFilePaths?: (
     projectConfig: GraphQLProjectConfig,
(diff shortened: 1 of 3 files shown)


mateus:
Closing since it's been stale for many months. The Admin Web Platform team will explore using `graphql-codegen` in the future, and this could be something to think about then.

PR #2390, opened 2022-08-26
useSerialized now returns the latest value from the HtmlManager

emileber:
## Description

The current memoization of the serialized value inside of the `useSerialized` hook means that calling `setSerialization` will never reach the component unless it completely remounts.

The memoization doesn't look necessary as it's only preventing reading a value from a `Map` instance, which should not be more expensive than the memoization lookup anyway.

Then, removing this memoization ensures that the hook returns the latest value whenever it renders. Note that setting the serialization client-side doesn't trigger any re-render, as expected.

## Reasoning

Inside storybook stories, if we want to update the serialization, it's currently not possible to get the latest values. This change should at least make it possible to have specific stories for different serialization use-cases.

### DIFF
--- a/packages/react-html/src/serializer.tsx
+++ b/packages/react-html/src/serializer.tsx
@@ -18,10 +18,7 @@ export function useSerialized<T>(
   id: string,
 ): [T | undefined, React.ComponentType<SerializeProps<T>>] {
   const manager = React.useContext(HtmlContext);
-  const data = React.useMemo(
-    () => manager.getSerialization<T>(id),
-    [id, manager],
-  );
+  const value = manager.getSerialization<T>(id);
 
   const Serialize = React.useMemo(
     () =>
@@ -42,7 +39,7 @@ export function useSerialized<T>(
     [id],
   );
 
-  return [data, Serialize];
+  return [value, Serialize];
 }
 
 export function createSerializer<T>(id: string) {
--- a/packages/react-html/src/tests/serializer.test.tsx
+++ b/packages/react-html/src/tests/serializer.test.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import {extract} from '@shopify/react-effect/server';
+import {createMount} from '@shopify/react-testing';
 
 import {render, Html} from '../server';
 import {useSerialized, HtmlContext, HtmlManager} from '..';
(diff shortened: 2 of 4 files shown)


packages/react-html/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
```
emileber, line 3: It looks like the changelog process changed, but I can't find the new documentation about it.

packages/react-html/src/serializer.tsx
```diff
@@ -18,10 +18,7 @@ export function useSerialized<T>(
   id: string,
 ): [T | undefined, React.ComponentType<SerializeProps<T>>] {
   const manager = React.useContext(HtmlContext);
-  const data = React.useMemo(
```
emileber, line 21: This memoization prevents getting the latest value and it doesn't really optimize anything since a `Map` access time is probably inconsequential.

packages/react-html/CHANGELOG.md
ryanwilsonperkin: We're using `changesets` now, with some details here: https://github.com/Shopify/quilt/blob/main/documentation/release-and-deploy.md#creating-releases
We should improve our docs to make this easier to discover

tl;dr; is run `yarn changeset` and follow the prompts, then commit the changeset file. The rest will be handled by github actions.

packages/react-html/src/tests/serializer.test.tsx
```diff
@@ -23,4 +24,36 @@ describe('useSerialized', () => {
...
+
+    const wrapper = mount(<MockComponent />);
+
+    expect(manager.getSerialization(key)).toBeUndefined();
+    expect(wrapper.text()).toBe(defaultValue);
+
+    const newValue = 'new-value';
+    manager.setSerialization(key, newValue);
+    // forces re-render
+    wrapper.setProps({bar: true});
```
ryanwilsonperkin, line 53: Rather than change a prop to trigger re-render as a side effect, we should aim to use `act` here to show that setting the serialization will have the desired effect:

```suggestion
    wrapper.act(() => {
      manager.setSerialization(key, newValue);
    });
```

However I suspect that won't work, because I don't think we've configured the `manager` as as a mutable store of value to indicate to React that a new render is necessary.

What this means, is that we'll get some tearing where the serialization value can be updated, but it doesn't cause a render. Then when a render happens for other reasons, it will use the new updated value. Is that by intent?

emileber: You're right that it won't work in an `act` since it doesn't trigger anything by itself. 

Serialized values never change during the lifetime of the app, it's mostly useful in a testing environment.

By default, it should have worked, but the `useMemo` had the unintended effect of hiding any change to the serialization map in the manager. So any code using the manager _was_ getting the latest values, only React components using the `useSerialized` hook were not getting the latest values.

emileber:
While this PR wouldn't hurt, it's not worth the effort to merge it so I'm going to close it.

PR #2392, opened 2022-08-30
removeItems method for 'useDynamicList' hook in react-form

PatrickMcLennan:
## Description

[The current `removeItem` method implements JavaScripts .splice method](https://github.com/Shopify/quilt/blob/main/packages/react-form/src/hooks/list/hooks/reducer.ts#L214), which [mutates the original array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice) and is causing unwanted side-effects and [bugs](https://github.com/Shopify/partners/issues/41912) in work that I'm currently doing, [which requires me to remove multiple fields at once](https://github.com/Shopify/partners/blob/master/app/assets/javascripts/sections/PartnerLeads/components/PartnerLeadsForm/PartnerLeadsForm.tsx#L295).

I've created a `removeItems` method that accepts an array of indices to remove and negates the need for `removeItem` to be ran multiple times.  `removeItems` runs in a pure fashion that leaves the source array unchanged and provides an easier DX for developers who need this functionality.

Myself and another dev (cc: @dahukish) have [experienced some issues](https://shopify.slack.com/archives/C02E6ML0GS1/p1661368041736969) working locally on this repo with both spin and dev, and were unable to get prettier working properly.  It seems this CI is having issues with prettier and a `README` file that I've changed, and I'm having difficulties understanding what the formatting issues are.  The internal logic for this work is complete + tested, so it is ready for review, but if someone could assist with my prettier / README issues it would be much appreciated!

### DIFF
--- a/.changeset/twelve-scissors-approve.md
+++ b/.changeset/twelve-scissors-approve.md
@@ -1,0 +1,5 @@
+---
+'@shopify/react-form': minor
+---
+
+Created a 'removeItems' method for the 'useDynamicList' hook that accepts an array of indices to remove
--- a/packages/react-form/README.md
+++ b/packages/react-form/README.md
@@ -696,6 +696,7 @@ const {
   fields,
   addItem,
   removeItem,
+  removeItems,
   moveItem,
   reset,
   dirty,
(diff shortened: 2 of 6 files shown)


vsumner (approved):
LGTM.

PR #2403, opened 2022-09-05
Update Shopify GraphQL API for 2022

kwp-digital:
## Description
Update the supported Shopify API versions.

### DIFF
--- a/packages/koa-shopify-graphql-proxy/src/shopify-graphql-proxy.ts
+++ b/packages/koa-shopify-graphql-proxy/src/shopify-graphql-proxy.ts
@@ -5,10 +5,10 @@ export const PROXY_BASE_PATH = '/graphql';
 export const GRAPHQL_PATH_PREFIX = '/admin/api';
 
 export type ApiVersion =
-  | '2020-10'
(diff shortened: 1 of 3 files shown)


packages/koa-shopify-graphql-proxy/src/shopify-graphql-proxy.ts
```diff
@@ -5,10 +5,10 @@ export const PROXY_BASE_PATH = '/graphql';
 export const GRAPHQL_PATH_PREFIX = '/admin/api';
 
 export type ApiVersion =
-  | '2020-10'
-  | '2021-01'
-  | '2021-04'
-  | '2021-07'
```
BPScott, line 11: Removing existing values is a breaking API change. Please add these values, but don't remove any existing ones. If you do this then you can revert your changes to the test files too.

BPScott (changes requested):
This is currently a breaking change. If you want to add new API versions, please add new keys to these types instead of removing existing keys.

When you do, please also revert the changes to the test file as these would no longer be needed.

PR #2414, opened 2022-09-08
feat: support readonly typed schemas

lbenie:
## Description

Fixes (issue #)

- Added a new `readonly-types` option, to support readonly schemas


mateus:
Closing this.

Long term, we want to move `graphql-typescript-deifintions` towards using `@graphql-codegen` internally. I think that replacement work would supersede any work in this PR.

PR #2416, opened 2022-09-09
Reword react-testing changeset

BPScott:
## Description

I found the changeset added in #2352 a little difficult to read, so I had a go at rewriting it.

@melnikov-s does this still sound right to you? I think I'd like a better explanation of what the effect of unresolved act has on a test suite, but don't know enough about it. What previously happened when we had unresolved promises still lingering?

### DIFF
--- a/.changeset/wise-apples-call.md
+++ b/.changeset/wise-apples-call.md
@@ -2,7 +2,7 @@
 '@shopify/react-testing': major
 ---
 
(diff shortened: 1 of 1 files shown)


melodyhabbouche (approved):
🧹 ✨ thanks for keeping things clean!

melnikov-s (approved):
Much better, thanks!

melnikov-s:
> What previously happened when we had unresolved promises still lingering?

Subsequent tests (in the same test suite) would fail unless all async act promises were manually resolved now with this fix that's no longer a requirement.

Issue #2275, opened 2022-05-18
Update @shopify/react-router dependencies v6.3.0


devonpmack:
## Overview

Currently we're blocked from upgrading react router due to @shopify/react-router not supporting it. We should update the package dependencies.

## Motivation

ServicesDB is out of date on React router

## Checklist

- [x] Please delete the labels section before submitting your issue
- [x] I have described this issue in a way that is actionable (if possible)

PR #2422, opened 2022-09-21
Update react-router to v6

devonpmack:
## Description

Fixes https://github.com/Shopify/quilt/issues/2275

Updates React router to v6 to unblock the rest of Shopify on upgrading react-router and react-router-dom.


packages/react-router/src/index.ts
```diff
@@ -1,18 +1,12 @@
...
   useLocation,
-  useHistory,
+  useNavigate,
   generatePath,
 } from 'react-router-dom';
-export type {
-  SwitchProps,
-  RouteProps,
-  RouteComponentProps,
-  RouterChildContext,
```
devonpmack, line 16: I could not find equivalents for these types

packages/react-router/src/tests/utilities.ts
```diff
@@ -1,52 +0,0 @@
-import {RouterChildContext} from 'react-router-dom';
```
devonpmack, line 1: I could not find these being used, https://grokt.shopify.io/results?q=createDefaultHistory%28%29, and the API has significantly changed. I'm not sure how to port them.

I decided to remove them (if we can)

BPScott: The package exposes a single entrypoint, and the stuff in this folder is not referenced from it. It's good to burn it all

BPScott (commented):
(I'm going to take a while to get to the review bit, bare with me while I rant)


I think trying to reexport 90% of a public package, and then add on 3 extra things is silly. I think trying to make everybody use react-router's API through the reexports of `@shopify/react-router` just makes things more confusing. It makes teaching people "we use react-router as a router" harder. It makes reading upgrade notes / migration guides between major versions of react-router harder and keeping dependencies updated becomes pointless churn.

The web-foundations teams of yore went through a phase of trying to whitelabel everything - instead of using some public package you should use `@shopify/SOMETHING`! Which was a neat idea when the team was able to focus on keeping these foundational technologies up to date and innovate and we wanted to bet that we could do better than the JS ecosystem. In some places (like graphql tooling) that was true at the time. These days however the ecosystem has plugged ahead while much of quilt has stayed where it was.

These days I reject that idea of being a central authority. We shouldn't be embracing and extinguishing `react-router` by making people use `@shopify/react-router` instead. We should show that we have some useful complementary utilities rather than trying to abstract over react-router entirely.

---

This PR's contents is great, but by bumping the `react-router-dom` version number and adjusting the list of stuff we re-export from `react-router-dom`, we perpetuate the old status-quo of hiding the fact that `react-router` is doing all the heavy lifting. I think that abstraction ultimately hurts understanding and that consumers should be referencing `react-router-dom` directly in the majority of cases.

What would you think to a new approach? What if we stopped reexporting stuff from `react-router-dom`, and instead expected consumers to reference `react-router-dom` where possible.


What if in addition to the current state of this PR we also did the following:

- Remove `react-router-dom` as a direct dependency. Add it to the peerDependency and devDependency arrays instead
- Delete the `@types/react-router-dom` dependency as it now ships its own types
- Remove all the reexports from `react-router-dom` in `index.ts`.
- Update the README.md to remove references to reexports from react-router-dom, and make it clear that the 3 components are to be used in conjunction with `react-router-dom` and that other routing should look at its documentation.
- Update the changeset to to talk about how we no longer reexport a bunch of stuff from react-router and that consumers should access Route/Switch/etc from `react-router-dom` instead

julianguyen:
Are there specific or immediate blockers if this PR doesn’t get merged? Does this prevent current work from being completed?

BPScott:
I just remembered #1673 where I talked about the idea that `Link` isn't a useful component. It doesn't handle external urls - only internal urls, so the current behaviour of Link isn't ever usable in practice.

To have a Link component that plays nice with Polaris and handles external links is trivial to implement, and is demonstrated in [Polaris's documentation](https://polaris.shopify.com/components/app-provider#using-link-component). I don't think there's any value in having polaris-specific logic in this package.

I reckon we should delete the `Link` component too.

devonpmack:
@BPScott Are these flaky tests? Seems like they pass after re-running

image

BPScott:
react-testing being flakey is concerning, but it's unrelated to this PR. For this PR we can ignore that. I'm going to give this another once-over, but I think we're close to being good to go.


Ping @melnikov-s - as he added the test destruction stuff. Sergi could you take a look at what might be causing these stale promise tests to intermittently fail: ohttps://github.com/Shopify/quilt/actions/runs/3147770668/jobs/5128584382 as they were introduced in #2352.

melnikov-s:
> Ping @melnikov-s - as he added the test destruction stuff. Sergey could you take a look at what might be causing these stale promise tests to intermittently fail: https://github.com/Shopify/quilt/actions/runs/3147770668/jobs/5128584382 as they were introduced in https://github.com/Shopify/quilt/pull/2352. Test flake because of the testing framework itself feels very scary

I saw this before and thought I'd fixed it, guess not. It only happens with React 17 for some reason and it's not a result of library code in https://github.com/Shopify/quilt/pull/2352 but rather the test the was introduced in that PR. The tests creates a hanging promise which will fail consistently without the changes in https://github.com/Shopify/quilt/pull/2352 but for some reason fails intermittently and rarely with the changes and in React 17 only. It will be difficult to determine the source of the issue as I can't reproduce it locally. Knowing that it's the test we can maybe disable it when `process.env.REACT_VERSION` is `17`

devonpmack:
@BPScott Is this good to be merged? Just want to confirm!

PR #2435, opened 2022-10-07
[react-testing] Disable flakly test for React 17

melnikov-s:
## Description

Fixes (issue #)

Prevents a flakly tests from running in React 17. 

Related PR: https://github.com/Shopify/quilt/pull/2352


packages/react-testing/src/tests/e2e.test.tsx
```diff
@@ -173,31 +173,34 @@ describe('@shopify/react-testing', () => {
       );
     }
 
-    it('releases any stale promises when component is destroyed', async () => {
-      const wrapper = mount(<Counter />);
-      wrapper.act(() => new Promise(() => {}));
-      wrapper.act(() => new Promise(() => {}));
-      await wrapper.destroy();
+    // eslint-disable-next-line no-process-env
+    if (process.env.REACT_VERSION !== '17') {
```
egjiri, line 177: This seem reasonable although I don't know if anyone is aware of a better way of achieving this as I didn't see any precent of the use of `process.env.REACT_VERSION` in the codebase.

.changeset/five-birds-thank.md
```diff
@@ -0,0 +1,2 @@
+---
```
egjiri, line 1: What does this file do?

packages/react-testing/src/tests/e2e.test.tsx
BPScott: Rando idea to avoid nesting adjustments: 

```ts
const itExceptReact17 = process.env.REACT_VERSION !== '17' ? it : it.skip;

itExceptReact17('releases any stale promises when component is destroyed', async () => {
```

If the eventual plan is to remove the "stale promise releasing" behaviour in 17, and make it a react 18 only feature, as the behaviour in v17 is not stable for unknown reasons then I think I'd prefer it if we roll this change and that disabling in the same PR rather than have two separate PRs.

.changeset/five-birds-thank.md
ryanwilsonperkin: This is a [changeset](https://github.com/changesets/changesets) file, it's a new bit of tooling that we've adopted in the past couple of months that helps automatically generate changelog entries and version bumps.

packages/react-testing/src/tests/e2e.test.tsx
ryanwilsonperkin: Agreed, better to still have the `it` block but have it be skipped (shows up in test reporters that way)
Alternative approach (either is fine with me, I like this idea from the original [feature request](https://github.com/facebook/jest/issues/3652)):

```js
const itIf = condition => condition ? it : it.skip;

itIf(process.env.REACT_VERSION !== '17')('blah blah', () => {
  ...
});
```

melnikov-s: Sounds good @ryanwilsonperkin  , this PR was mostly meant to unblock tests while I work a fix for 17 queue emptying or disabling it all together. I had sometime this morning and I might have found a fix that I'll put up later.

packages/react-testing/src/tests/e2e.test.tsx
```diff
@@ -173,7 +175,8 @@ describe('@shopify/react-testing', () => {
       );
     }
 
-    it('releases any stale promises when component is destroyed', async () => {
+    // eslint-disable-next-line no-process-env
+    itIf(process.env.REACT_VERSION !== '17')('releases any stale promises when component is destroyed', async () => {
```
BPScott, line 179: I discovered that React exports a `version` variable that contains the what the name suggests - https://unpkg.com/browse/react@17.0.2/cjs/react.development.js#L2331

Perhaps we could use that? Something like:

```js
const isReact18OrMore = parseInt(version.split('.')[0], 10) >= 18;

// ...

itIf(isReact18OrMore)(...
```

ryanwilsonperkin (approved):

PR #2437, opened 2022-10-12
Set the correct initial state for the useMedia hook

### LINKED ISSUE #2436: useMedia returns the incorrect initial state
## Overview

The useMedia hook will return `false` every time it is initially called, and then it uses a `useEffect` to update to the correct value later. This can cause screen flickering and just incorrect logic overall.

## Consuming repo

[discovery-app](https://github.com/shopify/discovery-app)

### Area

- [x] Add any relevant `Area: <area>` labels to this issue

### Scope

  - [x] Tag it with the `Package: <package_name>` label.

---

## Checklist

jas7457:
## Description

Fixes #2436

This change updates useMedia to use React's lazy-initialization for useState. Instead of always returning `false` as the initial state, it will do an initial `matchMedia` call (if not on the server). This can help in prevent screen flickering for UIs that use this hook, and will prevent a render cycle as well.

### DIFF
--- a/packages/react-hooks/src/hooks/media.ts
+++ b/packages/react-hooks/src/hooks/media.ts
@@ -4,7 +4,13 @@ type EffectHook = typeof useEffect | typeof useLayoutEffect;
 
 function createUseMediaFactory(useEffectHook: EffectHook) {
   return (query: string) => {
-    const [match, setMatch] = useState(false);
+    const [match, setMatch] = useState(() => {
+      if (typeof window === 'undefined' || !window.matchMedia) {
+        return false;
+      }
+
+      return window.matchMedia(query).matches;
+    });
 
     useEffectHook(() => {
       if (!window || !window.matchMedia) {
--- a/.changeset/tall-news-help.md
+++ b/.changeset/tall-news-help.md
@@ -1,0 +1,5 @@
+---
+'@shopify/react-hooks': patch
+---
+
+Fixed useMedia to correctly send back the correct initial state of the query match


packages/react-hooks/src/hooks/media.ts
```diff
@@ -4,7 +4,13 @@ type EffectHook = typeof useEffect | typeof useLayoutEffect;
 
 function createUseMediaFactory(useEffectHook: EffectHook) {
   return (query: string) => {
-    const [match, setMatch] = useState(false);
+    const [match, setMatch] = useState(() => {
+      if (typeof window === 'undefined' || !window.matchMedia) {
```
lucabezerra, line 8: The useEffectHook below seems to be doing a couple more things than what was added in here (mainly adding the event listener). Do you think those steps should be added in here as well (and potentially extract the code somewhere so we don't have code duplication)?

jas7457: React lazy state initialization should not have any side effects such as adding event listeners, so we definitely don't want to pull that in here.

I'm not sure what we could extract to remove any duplication. Do you have an example of what you're thinking? Only thing I could see is something like this, but IMO this is much more difficult to reason about. Maybe you have another solution in mind?

```tsx
const getMatchMedia = useCallback((query: string) => {
  if (typeof window === 'undefined' || !window.matchMedia) {
    return;
  }
  return window.matchMedia(query);
}, []);

const [match, setMatch] = useState(
  () => getMatchMedia(query)?.matches ?? false,
);
useEffectHook(() => {
  const matchMedia = getMatchMedia(query);
  if (!matchMedia) {
    return;
  }
  const updateMatch = (event: MediaQueryListEvent) =>
    setMatch(event.matches);
  setMatch(matchMedia.matches);
  matchMedia.addListener(updateMatch);
  return () => {
    matchMedia.removeListener(updateMatch);
  };
}, [getMatchMedia, query]);

return match;
```

BPScott (changes requested):
How does this affect server-side-rendering? Will this still result in consistent first renders on the server and the client - and thus not cause "React hydration mismatch" errors that occur when then server provided HTML and the initial render do not match.

It seems like this would mean that on the sever we'd return `false` and then on the first client render we're potentially return `true` (depending on wha media state you're matching). In those cases you'd raise cause a React Hydration mismatch error and React would have to rerender the whole tree, throwing away all benefits of server-side rendering.


This is why we defer the first check - so that the initial render matches what the server provides and then there's a second rerender to account for any matchMedia matching.


Marking as request changes as I want a good answer to this use-case and I want to be confident that this won't trigger missmatch errors before merging this.

jas7457:
Closing for the concern of SSR. `useMediaLayout` prevents the jitter (although it doesn't prevent the component from needlessly rerendering, at least on the client side).

PR #2438, opened 2022-10-14
[react-testing] fixed react queue emptying for React 17

melnikov-s:
## Description

#2352 Introduced automatic act queue emptying when a destroying a root wrapper. It fixed this fully in React 18 but not for older versions of react. This resulted in flaky tests for quilt as well as potentially surprising behaviour for React 17 users upgrading to react-testing@5. 

This PR also adds queue emptying for older version of React.

### DIFF
--- a/packages/react-testing/src/compat.ts
+++ b/packages/react-testing/src/compat.ts
@@ -1,4 +1,5 @@
 import ReactDOM from 'react-dom';
+import React from 'react';
 import type {Root as ReactRoot} from 'react-dom/client';
 
 import {ReactInstance, Fiber} from './types';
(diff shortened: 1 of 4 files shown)


packages/react-testing/src/compat.ts
```diff
@@ -34,3 +35,32 @@ export function createRoot(element: HTMLElement): ReactRoot {
     return createRootShim(element);
   }
 }
+
+export const isLegacyReact = parseInt(React.version, 10) < 18;
+
+// https://github.com/facebook/react/blob/12adaffef7105e2714f82651ea51936c563fe15c/packages/shared/enqueueTask.js#L13
+let enqueueTaskImpl: any = null;
+
+export function enqueueTask(task: (v: any) => void) {
```
melnikov-s, line 44: This is how the queue gets emptied out in older version of react. It creates a macro task either with `setImmediate` or `postMessage` . We're probably ok doing only the `setImmediate` here but decided to copy the entire code to be sure and match what react is doing.

packages/react-testing/src/root.tsx
```diff
@@ -140,10 +141,14 @@ export class Root<Props> implements Node<Props> {
           return result as unknown as Promise<void>;
         }
 
-        return Promise.race([
+        const promise = Promise.race([
           result,
           this.actPromise,
         ]) as unknown as Promise<void>;
+
+        this.actPromises.push(promise);
```
melnikov-s, line 149: We collect all the promises we need to wait on as `Promise.race([Promise.resolve()]` does not resolve at the same time as `Promise.resolve()` (it adds a tick). So best to wait on the same promise that gets returned to `act`

BPScott (commented):
Heck yeah! Thank you so much for digging into this. I see no glaring issue with the code (not that I saw any problem with the prior code either).


Ryan ended up merging #2435 not spotting this PR was super close. Can you want to revert that commit so we can see the e2e test passes in react 17.

BPScott (approved):
Magic!

Merge when you're happy

PR #2441, opened 2022-10-18
[react-testing] do not use Promise.race to release stale act promises

melnikov-s:
## Description

Fixes (issue #)


packages/react-testing/src/root.tsx
```diff
@@ -140,10 +136,38 @@ export class Root<Props> implements Node<Props> {
...
 
-        return Promise.race([
-          result,
-          this.actPromise,
-        ]) as unknown as Promise<void>;
+        /* return a thenable which when invoked will add react's callback (to clear queue)
+         * into a set. If the result ever resolves we will remove the callback from the Set.
+         * If it doesn't we will call all callbacks in the Set when this root is destroyed which
+         * will unblock any stuck queues allowing subsequent tests to run
+         * Note: This can be cleanly achieved with a `Promise.race` but that adds an extra micro-task
```
melnikov-s, line 143: Added a comment about `Promise.race` in case anyone would be tempted to refactor this in the future 😓

packages/react-testing/src/root.tsx
```diff
@@ -274,21 +298,21 @@ export class Root<Props> implements Node<Props> {
...
+      mountedPromise = this.unmount();
     }
     element.remove();
     connected.delete(this);
     this.destroyed = true;
-    this.actPromiseResolver();
-    // flush the micro task to wait until react commits all pending updates.
-    await this.actPromise;
+    await mountedPromise;
+    this.actCallbacks.forEach((callback) => callback());
```
ryanwilsonperkin, line 315: Won't these need to be awaited as well? And shouldn't they be invoked prior to unmounting?

melnikov-s: > Won't these need to be awaited as well?

These callbacks will `popActScope` and `recursivelyFlushAsyncActWork` if needed. There's nothing to await there.

> And shouldn't they be invoked prior to unmounting?

Once the root is destroyed there should be no more assertions on it, we're calling these to flush the queue out and it's not really related to this instance of the wrapper itself. I tried moving them up but it fails the tests as it appears that the final act in the `unmount` needs to be called and awaited on first.

ryanwilsonperkin (approved):

BPScott (approved):
Fantastic sleuthing! And a huge thanks for doing the testing in web!

A test case that fails if the code tries to be refactored to use `Promise.race` would a great deterrent to stop people from speculative refactoring. That doesn't need block this PR though.


What's the relationship between this and #2438? I guess this PR superseeds that one aside from unskipping the test?


Merge when you're ready.

packages/react-testing/src/root.tsx
melnikov-s: Ah good call, I'll see if I can put a test case together later. It's going to quite tricky to extract a minimal test case from that `apollo-client` mess

melnikov-s:
> What's the relationship between this and https://github.com/Shopify/quilt/pull/2438? I guess this PR superseeds that one aside from unskipping the test?

I think I'm still going to need 

```
 if (isLegacyReact) {
      // flush macro task for react 17 only
      await new Promise((resolve) => enqueueTask(resolve));
    }
```

To get React 17 working

Issue #2303, opened 2022-06-09
[@shopify/react-form] Change submit to promise


Tamas-Kisss:
## Overview

I would like to raise to related topics here.

1. First is a bug in the `submit` type.
Currently the submit return type is `void`.
```
export interface Form<T extends FieldBag> {
  ...
  submit(event?: React.FormEvent): void;
  ...
}
```
but actually because it wraps an async function it is a promise
```
const submit: (event?: React.FormEvent<Element> | undefined) => Promise<void>
```

2. A feature request to take advantage of the `submit` as promise

In the current form to handle the errors in the submit only possible trough `submitErrors` (please correct me if I am wrong). If the user would like to do an extra action based on the error, it probably requires an extra useEffect, plus that will be out of the scope of submit. For example:

```
const {submit,  submitErrors} = form;

function onSubmit () {
 submit()
}

useEffect(() => {
  if (submitErrors === "something") {
      // Do some action (like open modal or call an api, etc),
  }
})
```

If the `submit` promise could throw the relevant error it would be possible to handle the error in the context of the submit without having an extra useEffect.

```
const {submit,  submitErrors} = form;

async function onSubmit () {
  try {
    await submit()
  } catch (error: SubmitErrors) {
      if (submitErrors === "something") {
          // Do some action (like open modal or call an api, etc),
      }
  }
}
```

It would require to throw an error whenever the "errors" are set in the useSubmit hook. For example in submit.ts#51:

```
      if (clientErrors.length > 0) {
        setErrors(clientErrors);
        + throw clientErrors
        return;
      }
```

## Type

- [ ] New feature
- [x] Changes to existing features

### Scope

- Is this issue related to a specific package?

  - [x] Tag it with the `Package: <package_name>` label.

---

## Checklist

- [ ] Please delete the labels section before submitting your issue
- [x] I have described this issue in a way that is actionable (if possible)

PR #2447, opened 2022-10-21
Fix a submit function type

stephen-last:
## Description

Contributes to: https://github.com/Shopify/quilt/issues/2303

`submit()` is [an async function](https://github.com/Shopify/quilt/blob/main/packages/react-form/src/hooks/submit.ts#L43) so returns a promise.

This PR updates the type definition for `submit()`.

### DIFF
--- a/packages/react-form/src/types.ts
+++ b/packages/react-form/src/types.ts
@@ -99,7 +99,7 @@ export interface Form<T extends FieldBag> {
   submitErrors: FormError[];
   validate(): FormError[];
   reset(): void;
-  submit(event?: React.FormEvent): void;
+  submit(event?: React.FormEvent): Promise<void>;
   makeClean(): void;
 }
 
--- a/.changeset/mighty-candles-film.md
+++ b/.changeset/mighty-candles-film.md
@@ -1,0 +1,5 @@
+---
+'@shopify/react-form': patch
+---
+
+fix the type for submit() as it returns a promise and not just void.


stephen-last:
> Amazing, thank you!
> 
> I don't really want to ask more, but would it be possible to return the error in the promise. Because devs would be able to use is like
> 
> ```
> const {error} = await submit(...)
> ```

I think we should have two PRs, one for the type fix, and a separate one for the new feature.

This way it'll be easier and quicker to get the fix merged in, and allow reviewers of the new feature to concentrate on just that.

I'm happy to work with you on the new feature PR.

If you're happy with that, and this first PR, give this an approval, and hopefully @lucabezerra (who was auto tagged for a review) will do as well when he has time.

stephen-last:
Also... I'm unsure about what to do for the failing `changeset` ci step. This is my first PR in this repo and I've not seen this before.

@lucabezerra are you able to provide some assistance with this?

packages/react-form/src/types.ts
```diff
@@ -99,7 +99,9 @@ export interface Form<T extends FieldBag> {
   submitErrors: FormError[];
   validate(): FormError[];
   reset(): void;
-  submit(event?: React.FormEvent): void;
+  submit(
+    event?: React.FormEvent<Element> | undefined,
```
lucabezerra, line 103: Do we really need both instances of `undefined` here? I believe the `event?` already implies that, while the one in the return type doesn't really add much more than `void` already does, but let me know if you think otherwise.

stephen-last: Good spot on the `event?` meaning `undefined`, I've removed that one.

I think the return type still needs `undefined` though. If you look at the function it has two return statements with no values ([here](https://github.com/Shopify/quilt/blob/main/packages/react-form/src/hooks/submit.ts#L53) and [here](https://github.com/Shopify/quilt/blob/main/packages/react-form/src/hooks/submit.ts#L60)) which will return `undefined`, but if neither of those `if` statements are entered there will be no return, so `void`.

egjiri: I don't think we need `undefined` here either.

If you let TypeScript infer the type as you can see in this simple example it will actually infer it as `void` because it was not explicitly `return undefined;` but was just `return;`
Screen Shot 2022-10-24 at 4 41 36 PM

If you slap an `async` in front of the function TypeScript then infer it to `Promise<void>`
Screen Shot 2022-10-24 at 4 44 50 PM

So in this case I would just update it to
```ts
submit(event?: React.FormEvent): Promise<void>;
```

BPScott: Plus one to this being `Promise<void>`. I also note that the default value of the generic you pass into React.FormEvent is `Element` so being explicit here serves no purpose, It is fine to keep `React.FormEvent` instead changing to `React.FormEvent<Element>` as they are functionally identical.

You mentioned in the PR description that this to match the [definition of submit](https://github.com/Shopify/quilt/blob/main/packages/react-form/src/hooks/submit.ts#L42). I note that hovering over `submit` gives the type inference of submit as `(event?: React.FormEvent) => Promise<void>`


Screen Shot 2022-10-24 at 14 24 00

stephen-last: Thanks for the great feedback @egjiri & @BPScott ... 👍 

I've updated it to `submit(event?: React.FormEvent): Promise<void>;`.

egjiri (approved):

lucabezerra (approved):
I'd just suggest squashing the commits into a single one to make the history a bit cleaner :)

Issue #2448, opened 2022-10-22
[@shopify/react-form] FormError interface should use string for the field


MauricioRobayo:
## Overview

Currently, the `FormError` interface from the `react-form` package has defined the field to be a `string array` or `null`:

https://github.com/Shopify/quilt/blob/9ac47c1f1f43ec693ec21d1e32204e5072fcd268/packages/react-form/src/types.ts#L114

It should be `string` or `null`:

```ts
export interface FormError {
  field?: string | null; // <<<<
  message: string;
}
```

## Consuming repo

> What repo were you working in when this issue occurred?

...

### Area

- [ ] Add any relevant `Area: <area>` labels to this issue

### Scope

- Is this issue related to a specific package?

  - [ ] Tag it with the `Package: <package_name>` label.

Tried modifying the labels but GitHub is not allowing me.

Package: @shopify/react-form

---

## Checklist

- [ ] Please delete the labels section before submitting your issue
- [ ] I have described this issue in a way that is actionable (if possible)

MauricioRobayo:
Not a bug. It was me not properly understanding the usage. Works as expected with a string array.

:man_facepalming:

PR #2450, opened 2022-10-23
[react-form] fix field type on FormError

MauricioRobayo:
## Description

Fixes (issue #2448)

The field type on the `FormError` interface is currently using `string[] | null`. It should be a `string | null` (not an string array).

### DIFF
--- a/packages/react-form/src/types.ts
+++ b/packages/react-form/src/types.ts
@@ -111,7 +111,7 @@ export interface FormWithDynamicLists<
 }
 
 export interface FormError {
-  field?: string[] | null;
+  field?: string | null;
   message: string;
 }
 


BPScott:
There are multiple tests in `packages/react-form/src/hooks/tests/form.test.tsx` that test where `errors.field` is an array and type-checking fails on these cases with this change.

We have hundreds of cases in Shopify's codebases where `field` is set to an array.

Can you provide a codepen that proves that a case where a string is valid? The nearest I can get is seeing that in `propagateErrors` we do `get(fieldBag, error.field)` and the [code/documentation](https://github.com/jonschlinkert/get-value) for `get-value` suggests that the second argument can be either a string or an array of strings, and that using an array leads to a slight performance improvement. So it seems at best this should be changed to `string | string[]` but I am not convinced we should widen this as they we have to teach people about the two format possibilities.

MauricioRobayo:
Hi @BPScott ! Thanks for the feedback!

I might have got this wrong.

I was rushing through a Shopify App project and used the `@shopify/react-form` package.

While using it with `@Shopify/polaris` components  ([TextField](https://polaris.shopify.com/components/text-field)) I'm almost certain that the error did not display if I used an string array for the fields, but will double-check and let you know if that was a mistake on my side, otherwise I'll provide a reproducible example.

I'll come back in short, probably on the weekend as I'm trying to meet deadline.

Thanks again for the feedback, and for the awesome Library!

PR #2453, opened 2022-10-25
Improve `useToggle` hooks to also provide tuple format

pedrodurek:
## Description
In all places across `shopify/web`, we rename fields when using object destructuring for the `useToggle` return, which makes it verbose and not much readable.
This PR aims to tackle the problem above by also providing a tuple format. So we can either destructure as an array or object.
```tsx
const [isOpened, hideModal, showModal, toggleModal] = useToggle(false);

// Or
const {
    value: isOpened,
    setTrue: hideModal,
    setFalse: showModal,
    toggle: toggleModal,
  } = useToggle(false);
```

I decided to place `toggle` function at the last position because it seems to be less used than `setTrue` and `setFalse`.

Since we're supporting both formats, there is no breaking changes.

### DIFF
--- a/packages/react-hooks/src/hooks/toggle.ts
+++ b/packages/react-hooks/src/hooks/toggle.ts
@@ -1,16 +1,31 @@
 import {useState, useCallback} from 'react';
 
+type ToggleReturn = {
+  value: boolean;
+  setTrue: () => void;
+  setFalse: () => void;
+  toggle: () => void;
+} & [boolean, () => void, () => void, () => void];
+
 /**
  * Returns a stateful value, and a set of memoized functions to toggle it,
- * set it to true and set it to false
+ * set it to true and set it to false.
+ *
+ * @param {boolean} initialState initial value state.
+ * @return [value, setTrue, setFalse, toggle] or {value, setTrue, setFalse, toggle}.
  */
 export function useToggle(initialState: boolean) {
   const [value, setState] = useState(initialState);
 
-  return {
-    value,
-    toggle: useCallback(() => setState((state) => !state), []),
-    setTrue: useCallback(() => setState(true), []),
-    setFalse: useCallback(() => setState(false), []),
-  };
+  const setTrue = useCallback(() => setState(true), []);
+  const setFalse = useCallback(() => setState(false), []);
+  const toggle = useCallback(() => setState((state) => !state), []);
+
+  const ret = [value, setTrue, setFalse, toggle] as ToggleReturn;
+  ret.value = value;
+  ret.setTrue = setTrue;
+  ret.setFalse = setFalse;
+  ret.toggle = toggle;
+
+  return ret;
 }
--- a/packages/react-hooks/README.md
+++ b/packages/react-hooks/README.md
@@ -242,24 +242,40 @@ function Score({value}) {
 
 ### `useToggle()`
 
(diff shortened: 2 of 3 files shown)


packages/react-hooks/src/hooks/toggle.ts
```diff
@@ -1,16 +1,31 @@
...
-  };
+  const setTrue = useCallback(() => setState(true), []);
+  const setFalse = useCallback(() => setState(false), []);
+  const toggle = useCallback(() => setState((state) => !state), []);
+
+  const ret = [value, setTrue, setFalse, toggle] as ToggleReturn;
+  ret.value = value;
+  ret.setTrue = setTrue;
+  ret.setFalse = setFalse;
+  ret.toggle = toggle;
```
pedrodurek, line 28: Since arrays are a special type of objects, this is totally doable.

packages/react-hooks/src/hooks/toggle.ts
```diff
@@ -1,16 +1,31 @@
 import {useState, useCallback} from 'react';
 
+type ToggleReturn = {
+  value: boolean;
+  setTrue: () => void;
+  setFalse: () => void;
+  toggle: () => void;
+} & [boolean, () => void, () => void, () => void];
```
meko-deng, line 8: Ah interesting! If I understand this correctly, although you're using an Intersection type here, the `ToggleReturn` type can only be either one of the two types you've defined here since there's no way to overlap the two. But this only really works because, like you mentioned, an array is a special type of object.. is this common practice?

packages/react-hooks/src/hooks/toggle.ts
meko-deng, line 28: could we just do the following?

```suggestion
  const ret = {value, setTrue, setFalse, toggle} as ToggleReturn;
```

orditeck (approved):
Great improvement, and technically very interesting. TIL!

packages/react-hooks/src/hooks/toggle.ts
pedrodurek: If we do that, it'll work as before (object format only). 

The goal here is to create a tuple containing `value`, `seTrue`, etc, and add the new properties (`value`, `setTrue`, `setFalse`, `toggle`) to this "special object".

BPScott:
Honestly I'm not a fan of this. Widening this shape means there's now two ways to do the same thing, and that means people will ask questions like "What's the difference between these two approaches, and which one is preferred". I think that the type signature of `const [isOpened, hideModal, showModal, toggleModal] = useToggle()` being `[boolean, () => void, () => void, () => void]` i unclear as it is not obvious which of those functions do what, I think it would be easy for consumers to get their ordering the wrong way around ("is setTrue or setFalse in the second position?").

I think it is better to have a single way of doing things, even if it is a little more verbose than the array-style notation.

mateus:
I second what @BPScott said. I am not a big fan of having two distinct return types, but my biggest issue with it is the fact that we can't name the items from the array `[boolean, function, function, function]`. We will get lots of questions about what is the purpose of each function, and it will be a responsibility of the consumer to properly name them. I'd prefer to avoid that.

BPScott:
That's 3 maintainers of quilt who prefer a single way of doing this rather than introducing a second way of doing the same thing. 
With that in mind, thank you for coming up with ideas to improve quilt but we're not going to accept this change as we feel that this approach introduces too much ambiguity.

PR #2454, opened 2022-10-26
Add a "resolveFetchMore" to fix the bug in fetchMore with Apollo 3

ryanwilsonperkin:


BPScott:
Thanks so much for digging into this Ryan. I've tested this in web and it helps!

I've rolled it into #2302 with the name `waitForQueryUpdates`, so I'm going to close this PR. [`dbeb32813`](https://github.com/Shopify/quilt/pull/2302/commits/dbeb32813479a054b735f516bfa10b117fc65d44) is the commit in the Apollo3 till the next rebase.

Issue #2304, opened 2022-06-11
[react-i18n] Improve formatCurrency `short` form symbol formatting


trishrempel:
## Overview

Related to #2265 and #1892 and https://github.com/Shopify/core-issues/issues/45203

Improve the `short` form of `formatCurrency` so that it more closely adheres to the [Polaris currency formatting guidelines](https://polaris.shopify.com/foundations/formatting-localized-currency), while also applying Shopify-defined exceptions for specific currencies where the longer form symbol should be displayed.

The `formatCurrencyShort` method calls `getShortCurrencySymbol`. This method calls `getCurrency`, which calls  [`Intl.NumberFormat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) with `style: 'currency'`.

If the formatted currency comes back with the first two characters of the currency and a symbol (e.g. `CA$`), the first two characters are stripped off (e.g. `$1.50` instead of `CA$1.50`).

However, we've been discovering some edge cases where a longer-form currency symbol is being returned when it shouldn't be:
- When formatting a `short` currency in `AUD` when the default is another currency, the currency symbol is `A$` instead of `$`.
- When formatting a `short` currency in `EGP`, the currency symbol is `EGP` instead of `E£`

#### Short currency symbol formatting improvement

By default , when `Intl.NumberFormat` has `style: 'currency'`, it uses the option `currencyDisplay: 'symbol'`.

However, the option `currencyDisplay: 'narrowSymbol'` might be more appropriate for more use cases, without the need to strip the first two characters of the symbol.

<details>
<summary>Example</summary>

```javascript
const number = 123456.789;

console.log(new Intl.NumberFormat('en', { style: 'currency', currency: 'AUD', currencyDisplay: 'symbol'}).format(number));
// AUD symbol: "A$123,456.79"

console.log(new Intl.NumberFormat('en', { style: 'currency', currency: 'AUD', currencyDisplay: 'narrowSymbol'}).format(number));
// AUD narrowSymbol: "$123,456.79"

console.log(new Intl.NumberFormat('en', { style: 'currency', currency: 'CAD', currencyDisplay: 'symbol'}).format(number));
// CAD symbol: "CA$123,456.79"

console.log(new Intl.NumberFormat('en', { style: 'currency', currency: 'CAD', currencyDisplay: 'narrowSymbol'}).format(number));
// CAD narrowSymbol: "$123,456.79"

[...]

PR #2456, opened 2022-11-01
Update `i18n.formatCurrency` to use `form: 'auto'` by default

trishrempel:
## Description

Related to https://github.com/Shopify/quilt/issues/2304

Currently, if you do not provide the `form` option to `i18n.formatCurrency`, then its behaviour reverts to the legacy (deprecated) `formatCurrency()`, which is a convenience function that simply _auto-assigns_ the `as` option to `currency` and calls `formatNumber()`. 

Note that this will resemble `form: 'short'`, but will sometimes extend the symbol with extra information depending on the browser's implementation of `Intl.NumberFormat` and the locale in use. For example, `formatCurrency(1.25, {currency: 'CAD'})` may return `$ 1.25`, or it might return `CA$ 1.25`.

There are many instances in Web that call `i18n.formatCurrency(amount)` without the `form` option, and so are not displaying currency formats according to [Polaris localized currency formatting guidelines](https://polaris.shopify.com/foundations/formatting-localized-currency).

This pull request changes the default behaviour to use `form: 'auto'` when the `form` option is not provided.


BPScott (approved):
Love this!

Big fan of making sure we have sensible defaults!

When you're happy go forth and squash and merge.

sunical:
@trishrempel, I didn't find the test coverage for this scenario: when `form` is omitted, default to `form: 'auto'`

packages/react-i18n/src/tests/i18n.test.ts
```diff
@@ -784,16 +784,48 @@ describe('I18n', () => {
   });
 
   describe('#formatCurrency()', () => {
+    const amount = 1234;
     const currency = 'USD';
 
-    it('formats the number as a currency', () => {
+    it('formats short when there is no currency in the options', () => {
```
trishrempel, line 790: @sunical These tests are copied from the `form:auto` tests just below this - they confirm that not passing the `form` option produces the same behaviour as `form:auto`.

PR #2458, opened 2022-11-03
Ordinal pluralization

movermeyer:
### What are you trying to accomplish?

The [react-i18n data format schema](https://github.com/Shopify/quilt/blob/main/packages/react-i18n/docs/react_i18n_schema.md#ordinal-pluralization) defines how to store ordinal pluralization values:

```jsonc
{
  "cars": {
    "ordinal": {
      // The special key "ordinal" creates an ordinal pluralization context
      "one": "This is my {ordinal}st car", // Used for 1, 21, 31, 41, 51, 61, 71, 81, 101, 1001, …
      "two": "This is my {ordinal}nd car", // Used for 2, 22, 32, 42, 52, 62, 72, 82, 102, 1002, …
      "few": "This is my {ordinal}rd car", // Used for 3, 23, 33, 43, 53, 63, 73, 83, 103, 1003, …
      "other": "This is my {ordinal}th car" // Used for 0, 4~18, 100, 1000, 10000, 100000, 1000000, …
    }
  }
}
```

But the `react-i18n` library doesn't have a mechanism to look them up.
This PR aims to add that mechanism.

### What approach did you choose and why?

Duplicated the logic for the cardinal pluralization, and adjusted it to use ordinal pluralization.

Made the choice to define a new special `ordinal` replacement.

#### Choice of `ordinal` name

This could be instead made to use the same `count` in the interpolation, but then the replacement name wouldn't match, which I thought would be strange.

`ruby-i18n` doesn't have ordinal pluralization lookups, so there's nothing to copy in this choice.

CLDR calls these ["lateral inheritance attributes"](https://www.unicode.org/reports/tr35/tr35.html#Lateral_Inheritance), and calls them `count` and `ordinal`, so that's another vote for `ordinal`. 

### What should reviewers focus on?

...

### The impact of these changes

Users will be able to provide translations for ordinal pluralization strings.

### Testing

...

### DIFF
--- a/packages/react-i18n/README.md
+++ b/packages/react-i18n/README.md
@@ -297,7 +297,11 @@ if (keyExists) {
 
 ##### Pluralization
 
-`@shopify/react-i18n` handles pluralization similarly to [Rails’ default i18n utility](https://guides.rubyonrails.org/i18n.html#pluralization) ([with some minor differences](docs/react_i18n_schema.md#pluralization-1)). The key is to provide the plural-dependent value as a `count` variable. `react-i18n` then looks up the plural form using [`Intl.PluralRules`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/PluralRules) and, within the keypath you have specified for the translation, will look up a nested translation matching the plural form:
+`@shopify/react-i18n` handles pluralization similarly to [Rails’ default i18n utility](https://guides.rubyonrails.org/i18n.html#pluralization) ([with some minor differences](docs/react_i18n_schema.md#pluralization-1)). `react-i18n` then looks up the plural form using [`Intl.PluralRules`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/PluralRules) and, within the keypath you have specified for the translation, will look up a nested translation matching the plural form.
+
+###### Cardinal Pluralization
+
+Cardinal pluralization lookups use the value of the `count` replacement to choose the correct string:
 
 ```ts
 // Assuming a dictionary like:
@@ -313,9 +317,6 @@ if (keyExists) {
 i18n.translate('MyComponent.searchResult', {count: searchResults});
 ```
 
-As noted above, this functionality depends on the `Intl.PluralRules` global. If this does not exist [for your environment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/PluralRules#Browser_compatibility), we recommend including the [`intl-pluralrules`](https://yarnpkg.com/en/package/intl-pluralrules) polyfill or included `import '@shopify/polyfills/intl';` from [`@shopify/polyfills`](https://github.com/Shopify/quilt/tree/main/packages/polyfills).
-We also recommend to have the `{count}` variable in all of your keys as some languages can use the key `"one"` when the count is `zero` for example. See MDN docs on [Localization and Plurals](https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_and_Plurals).
-
 By default, `{count}` will be automatically formatted as a number. If you want to format the variable differently, you can simply pass it in another variable.
 
 ```ts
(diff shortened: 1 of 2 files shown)


packages/react-i18n/README.md
```diff
@@ -335,6 +336,56 @@ i18n.translate('MyComponent.searchResult', {
...
+        "few": "This is the {formattedOrdinal}rd widget found",
+        "other": "This is the {formattedOrdinal}th widget found",
+      }
+    }
+  }
+}
+
+i18n.translate('MyComponent.searchResult', {
+  ordinal: searchResultOrdinal,
+  formattedOrdinal: i18n.formatNumber(searchResultOrdinal),
```
trishrempel, line 380: Since `translate` already formats the `ordinal` value as a number, I suggest using a different example, such as applying some visual formatting:

```suggestion
  formattedOrdinal: <b>{searchResultOrdinal}</b>,
```

trishrempel (approved):
Looks good! Just a couple of minor suggestions.

BPScott (approved):
Having "what replacement value is passed in - count or ordinal" controling what mode to be in seems decent and it follows the spec that was already written.

I suspect it is somewhat moot as [the ordinal stuff in the spec was authored in Nov 2020](https://github.com/Shopify/quilt/commit/e57631b5ee2bffd1824cbc7cfb1fdb38cf3d8a0d) and so it's probably too late to consider changing but when reviewing this I was thinking of an alternative approach where the key could be the deciding factor on if to use cardinal or ordinal numbering - perhaps keys with an `_ordinal` suffix could toggle the mode:

```json
{
  "MyComponent": {
    "searchResult_ordinal": {
      "one": "This is the {count}st widget found",
      "two": "This is the {count}nd widget found",
      "few": "This is the {count}rd widget found",
      "other": "This is the {count}th widget found"
    }
  }
}
```

```
i18n.translate('MyComponent.searchResult_ordinal', {count: searchResultOrdinal});
```

What you've got currently allows for a single key to have cardinal and ordinal translations strings (per the pluralisations test case) - which the by-key approach disallows - but I'm not sure if that's a case that would crop up much in reality, and even then having two keys doesn't seem that bad:

```js
// nested-style
{
  pluralizations: {
    one: 'I have {count} car!',
    other: 'I have {count} cars!',
    ordinal: {
      one: 'This is my {ordinal}st car!',
      two: 'This is my {ordinal}nd car!',
      few: 'This is my {ordinal}rd car!',
      other: 'This is my {ordinal}th car!',
    }
  }
}

// get regular count - "I have 1 car!"
i18n.translate('MyComponent.searchResult', {count: 1});
// get ordinal count - "This is my 1st car!"
i18n.translate('MyComponent.searchResult', {ordinal: 1});

//
// VS
//

// suffix style
{
  pluralizations: {
    one: 'I have {count} car!',
    other: 'I have {count} cars!',
  },
  pluralizations_ordinal: {
    one: 'This is my {count}st car!',
    two: 'This is my {count}nd car!',
    few: 'This is my {count}rd car!',
    other: 'This is my {count}th car!',
  }
}
```

// get regular count - "I have 1 car!"
i18n.translate('MyComponent.searchResult', {count: 1});
// get ordinal count - "This is my 1st car!"
i18n.translate('MyComponent.searchResult_ordinal', {count: 1});
```

---


That said, I suspect this isn't the place to ponder about the structures that could have been. We have the spec text and the code adheres to it and looks great!

Is it worth generating a snapshop release by commenting `/snapit`, and testing that snapshot in a consuming application? Personally this feels low-risk as unit tests can cover behaviour fully. If you don't consider that integration worth it then feel free to merge!

movermeyer:
> What you've got currently allows for a single key to have cardinal and ordinal translations strings (per the pluralisations test case)

Yup. Personally, I see it as a feature, not a bug (of course, I wrote the spec so I'm biased)

Aside: It's also how the Translation Platform distinguishes between cardinal and ordinal pluralization contexts. So that's at least one client that would have to change if we wanted to change it.

PR #2459, opened 2022-11-04
Ignore negative start values for lifecycle events

ryanwilsonperkin:
## Description

Some browsers such as Firefox have the potential to send negative time-to-first-byte values because of a bug in the browser: https://bugzilla.mozilla.org/show_bug.cgi?id=1429422#c12

This is causing the aggregations that we perform on our performance data to be skewed because we're getting nonsensical values, it doesn't make sense for us to be recording negative values for lifecycle events because they can't go backwards in time.

This change will drop those values, rather than report them. We'd rather drop them than clamp them to zero because zero also isn't a relevant indicator here.

### DIFF
--- a/.changeset/popular-pumas-marry.md
+++ b/.changeset/popular-pumas-marry.md
@@ -1,0 +1,5 @@
+---
+'@shopify/performance': patch
+---
+
+Ignore invalid negative values for lifecycle metrics like Time to First Byte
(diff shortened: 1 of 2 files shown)


CameronGorrie (approved):
Nice find and fix! I was going to comment that we should add a test but looks like there's problems with writing a [useful performance observer mock](https://github.com/Shopify/quilt/blob/1f76ed324172fe90048423e7a0503b762d7424af/packages/performance/src/tests/performance.test.ts#L8-L10).

PR #2463, opened 2022-11-10
Pass the calendar prop to `formatDate`

adam-felix-shopify:
## Description

[FormatJS's Intl.DateTimeFormat](https://formatjs.io/docs/polyfills/intl-datetimeformat/) only supports the Gregorian calendar.  These changes enforce that.

Fixes (or at least it should) crashing issues on React Native for the [mobile repo](https://github.com/Shopify/mobile)

### Mobile Tophat:

__Setup__:

1. Change an iOS device's region to Thailand
1. Change the device's language to Thai
1. In the mobile repo, run
```
yarn add @shopify/react-i18n@0.0.0-snapshot-20221111164351
```
1. After that successfully completes, run `dev up`
1. Run `dev react-native start --reset-cache`
    * __Note__:  `--reset-cache` may not be needed
1. Run the app (either by building in Xcode or `dev react-native run-ios`

__Test__:

1. Navigate to the online store section of the store tab
    * __Expected behaviour__:  No crashes occur

#### Video of this test:

__Before__:

https://user-images.githubusercontent.com/99921496/201183803-30c175b9-af29-41ac-8087-01ac8bb8fb57.mov

__After__:

https://user-images.githubusercontent.com/99921496/201183933-1b05abf3-14d8-4539-a9b0-a9a36d4532da.mov

### DIFF
--- a/packages/dates/src/utilities/formatDate.ts
+++ b/packages/dates/src/utilities/formatDate.ts
@@ -12,9 +12,11 @@ export function memoizedGetDateTimeFormat(
   return i;
 }
 
-const browserFeatureDetectionDate = Intl.DateTimeFormat('en', {
-  hour: 'numeric',
-});
+const browserFeatureDetectionDate = (options: FormatDateOptions) =>
+  Intl.DateTimeFormat('en', {
+    ...options,
+    hour: 'numeric',
+  });
 
 // hourCycle to Intl.DateTimeFormatOptions was added in TS 4.2, so we could
 // remove this, but that would require consumers to update to at least TS 4.2
@@ -29,16 +31,19 @@ interface ResolvedFormatDateOptions extends Intl.ResolvedDateTimeFormatOptions {
   hourCycle?: 'h11' | 'h12' | 'h23' | 'h24';
 }
 
-const resolvedOptions: ResolvedFormatDateOptions | undefined =
-  typeof browserFeatureDetectionDate.resolvedOptions === 'undefined'
+const resolveFormattingOptions: (
+  options: FormatDateOptions,
+) => ResolvedFormatDateOptions | undefined = (options) =>
+  typeof browserFeatureDetectionDate(options).resolvedOptions === 'undefined'
     ? undefined
-    : browserFeatureDetectionDate.resolvedOptions();
+    : browserFeatureDetectionDate(options).resolvedOptions();
 
 export function formatDate(
   date: Date,
   locales: string | string[],
   options: FormatDateOptions = {},
 ) {
+  const resolvedOptions = resolveFormattingOptions(options);
   const hourCycleRequired =
     resolvedOptions != null &&
     options.hour12 === false &&
--- a/packages/react-i18n/src/i18n.ts
+++ b/packages/react-i18n/src/i18n.ts
@@ -318,11 +318,17 @@ export class I18n {
             ...dateStyle[style],
           });
         default:
-          return this.formatDate(date, {...formatOptions, ...dateStyle[style]});
+          return this.formatDate(date, {
+            ...formatOptions,
+            ...dateStyle[style],
+          });
       }
     }
 
-    return formatDate(date, locale, {...formatOptions, timeZone});
+    return formatDate(date, locale, {
+      ...formatOptions,
+      timeZone,
+    });
   }
 
   ordinal(amount: number) {
@@ -627,7 +633,7 @@ export class I18n {
     date: Date,
     options?: Intl.DateTimeFormatOptions,
   ): string {
-    const {localeMatcher, formatMatcher, timeZone} = options || {};
+    const {localeMatcher, formatMatcher, timeZone, calendar} = options || {};
 
     const hourZone = this.formatDate(date, {
       localeMatcher,
@@ -636,6 +642,7 @@ export class I18n {
       hour12: false,
       timeZoneName: 'short',
       hour: 'numeric',
+      calendar,
     });
 
     const zoneMatchGroup = /\s([\w()+\-:.]+$)/.exec(hourZone);
@@ -654,6 +661,7 @@ export class I18n {
       era,
       timeZone,
       timeZoneName,
+      calendar,
     } = options || {};
 
     const formattedDate = this.formatDate(date, {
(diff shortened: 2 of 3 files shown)


packages/dates/src/utilities/formatDate.ts
```diff
@@ -14,6 +14,7 @@ export function memoizedGetDateTimeFormat(
 
 const browserFeatureDetectionDate = Intl.DateTimeFormat('en', {
   hour: 'numeric',
+  calendar: 'gregory',
```
atesgoral, line 17: It would be a nice to add comment about why this restriction is being added so that we can remember to remove it once there's wider support.

BPScott:
How come you're relying on the FormatJS's polyfill? Support for Intl.DateTimeFormat is available on [all modern platforms](https://caniuse.com/?search=datetimeformat). What browser engines are you targeting that need this polyfill?

adam-felix-shopify:
> How come you're relying on the FormatJS's polyfill? Support for Intl.DateTimeFormat is available on [all modern platforms](https://caniuse.com/?search=datetimeformat). What browser engines are you targeting that need this polyfill?

@BPScott:  I'm not sure.  We're using `react-i18n` with React Native.

BPScott (changes requested):
Looks like it came in when you adopted Hermes on 2022-09-27 in https://github.com/Shopify/mobile/pull/8541, and [Hermes iOS support for Intl gets added in react-native 0.70.0](https://github.com/facebook/react-native/blob/main/CHANGELOG.md#0700)

---

This change is a breaking change. It changes the date rendering behaviour for locales that do not use the gregorian calendar by default and that seems very suprising and unexpected. It also seems like a bad merchant experience as dates are no longer in the form that they are used to based upon their locale.

Running `console.log((new Intl.DateTimeFormat('th-TH')).resolvedOptions().calendar);` suggests that the default Thai calendar is the Buddhist calendar which [wikipedia](https://en.wikipedia.org/wiki/Buddhist_calendar) suggests is similar to the gergorian calendar except it uses a different year epoch. I suspect this change would stop your app from crashing thanks to your polyfill not supporting gregorian calendars but you, and every other app that consumes these packages would be left with Thai merchants wondering why their years are 543 years different to what they expect - the Gregorian year 2022 is represented as year 2565 in the Thai buddhist calendar.

Because of this change in merchant experience I do not think this is a fix that should be merged. This issue does not affect users in supported browsers. It only affect app authors who are using polyfills for Intl.DateTimeFormat because they use Hermes and react-native <0.70.0.

I think the right fix for this is in the mobile repo, which should remove their `Intl.DateTimeFormat` polyfill. It should either revert its Hermes support, or update to React Native 0.70.0 which adds support for these I18n APIs in iOS. A quick-fix while you update your react-native version would be to continue useing the snapshot you have created as part of this PR - assuming you consider this year change on iOS devices only acceptable.

adam-felix-shopify:
I updated the PR to pass calendar through to calls of `formatDate` and updated the browser related constant to be a function that takes options.

packages/dates/src/utilities/formatDate.ts
```diff
@@ -29,20 +31,21 @@ interface ResolvedFormatDateOptions extends Intl.ResolvedDateTimeFormatOptions {
...
   date: Date,
   locales: string | string[],
   options: FormatDateOptions = {},
 ) {
+  const ro = resolvedOptions(options);
   const hourCycleRequired =
-    resolvedOptions != null &&
-    options.hour12 === false &&
-    resolvedOptions.hourCycle != null;
+    ro != null && options.hour12 === false && ro.hourCycle != null;
```
gvarandas, line 48:
```suggestion
    resolvedOptions != null && options.hour12 === false && ro.hourCycle != null;
```

daehn (approved):
🎩 ✅  after the adjustments.

https://user-images.githubusercontent.com/4603353/201683109-d0171084-985d-48a9-a6d5-d9dc7f2b2254.mp4

packages/dates/src/utilities/formatDate.ts
```diff
@@ -12,9 +12,11 @@ export function memoizedGetDateTimeFormat(
   return i;
 }
 
-const browserFeatureDetectionDate = Intl.DateTimeFormat('en', {
-  hour: 'numeric',
-});
+const browserFeatureDetectionDate = (options: FormatDateOptions) =>
+  Intl.DateTimeFormat('en', {
```
BPScott, line 16: Why does this need to be changed? 

I feel pretty certain that `Intl.DateTimeFormat('en', {hour: 'numeric'}).resolvedOptions()` would always result in a gregorian calendar. We never call `.format` on the result of this, so it seems like it wouldn't ever be an issue in the Thai locale.


If my expectation is right and this is not needed, then nothing in this file needs to be changed.

packages/react-i18n/src/i18n.ts
```diff
@@ -318,11 +318,17 @@ export class I18n {
             ...dateStyle[style],
           });
         default:
-          return this.formatDate(date, {...formatOptions, ...dateStyle[style]});
+          return this.formatDate(date, {
```
BPScott, line 321: This seems to be just a whitespace change for the sake of it. Can we revert it to make the diff easier to review?

packages/react-i18n/src/i18n.ts
```diff
@@ -318,11 +318,17 @@ export class I18n {
...
+            ...formatOptions,
+            ...dateStyle[style],
+          });
       }
     }
 
-    return formatDate(date, locale, {...formatOptions, timeZone});
+    return formatDate(date, locale, {
```
BPScott, line 328: This seems to be just a whitespace change for the sake of it. Can we revert it to make the diff easier to review?

BPScott (changes requested):
I'm more amenable to letting folks override the calendar compared to the prior approach as it avoids the breaking change, thought this still feels like a fix for a temporary issue (till mobile apps adopt RN 0.70) rather than a long-term feature that we'd like to continue to support.

Having to explictly override the calendar for every formatDate call an application uses feels like a poor developer experience - it requires lots of updates to get things working, is error prone as it would be easy to miss setting this if a new call to formatDate is added (and once you forgot one, bang runtime errors in Thai), and once it is no longer needed you've got to go and remove every `calendar: 'gregory'` usage.


Would you be open to the idea of maintaining a fork of `@shopify/dates` that forces the usage of the gregorian calendar (i.e. what the original PR did)? That'd basically be changing this PR back to its original state, and using the snapshot package generated by `/snapit`. If at some point in the future we add other updates to `@shopify/dates` (unlikely, it is a very stable package) then it'd be a case of rebasing this PR and running snapit again.

Alternatively you could use [patch-package](https://www.npmjs.com/package/patch-package) to apply the "force the gregorian calendar" patch as part of installing `@shopify/dates` into the shopify/mobile codebase to make it so the calendar is forced to Gregorian within the context of mobile.

adam-felix-shopify:
We're going to go ahead with the patch approach and close this PR.

PR #2464, opened 2022-11-11
Bump @faker-js/faker to 7.x

notjosh:
## Description

:wave: Hello from Shop!

We're migrating over to v7 on Shop, and using `resolutions` + patch to get `graphql-fixtures` compatible with v7 (rationale: Faker is a large import over many files, which has an impact on Jest performance on large suites).

Consumers of `graphql-fixtures` can still `import {faker} from 'graphql-fixtures'` as before, which is (afaik?) the only public change this PR might have introduced (the rest are all within our tests?).

This PR is mostly just a large search/replace to swap out the `default` import for the named `faker` export. Nothing else seems too concerning in the v7 notes - it still supports Node 14, so nothing changes there.

(I'll add a changeset if we want to 🚢, otherwise we can hold off until later)

### DIFF
--- a/.changeset/quiet-impalas-retire.md
+++ b/.changeset/quiet-impalas-retire.md
@@ -1,0 +1,5 @@
+---
+'graphql-fixtures': major
+---
+
+Upgraded @faker-js/faker from v6 to v7
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "@babel/core": "^7.14.6",
     "@changesets/changelog-github": "^0.4.7",
     "@changesets/cli": "^2.25.0",
-    "@faker-js/faker": "^6.3.1",
+    "@faker-js/faker": "^7.6.0",
     "@shopify/app-bridge": "^2.0.3",
     "@shopify/babel-preset": "^24.1.4",
     "@shopify/browserslist-config": "^3.0.0",
--- a/packages/graphql-fixtures/README.md
+++ b/packages/graphql-fixtures/README.md
@@ -59,7 +59,7 @@ const data = fill(myQuery, data)({query: myQuery});
 
 #### Using faker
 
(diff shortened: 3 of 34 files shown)


ryanwilsonperkin:
Got this tested out against web and based on Ben's suggestion I applied a find-and-replace to swap out the default export, looks like it's mostly working now and I'm fairly confident that web won't face any major problems updating. I'll +1, @BPScott do you have any other concerns?

ryanwilsonperkin (approved):

.changeset/quiet-impalas-retire.md
```diff
@@ -0,0 +1,18 @@
...
+'@shopify/react-form-state': patch
+'@shopify/react-graphql': patch
+'@shopify/react-hydrate': patch
+'@shopify/react-i18n-universal-provider': patch
+'@shopify/react-performance': patch
+'@shopify/react-testing': patch
+'@shopify/react-universal-provider': patch
+---
+
+Upgraded @faker-js/faker from v6 to v7
```
BPScott, line 18: Changesets trigger releases of packages. Published packages do not contain test files (the `"!build/ts/**/tests/",` in the files array in the various package.json's excludes them). I think this changeset should only list packages where our changes would result in a change to the published package. If we only touch test files within a package then it can be omitted from the list.


It looks like the only package that touches files other than test files, is `graphql-fixtures`. All other packages can be removed from this list.

```suggestion
'graphql-fixtures': major
---

Upgraded `@faker-js/faker` from v6 to v7
```

BPScott (approved):
Code changes look great! One note inline about the changeset content - We can omit the packages where only test files have changed because only changing test files would not result in a change to the contents of the published packages (as we omit the test files from the published package contents)

BPScott:
No code related concerns from me. @notjosh could you adjust the changeset updating per my suggestion, and then you're free to merge when you're ready.

PR #2469, opened 2022-11-16
[react-form] - Support returning dirty nested fields from getDirtyValues

alexanderMontague:
## Description

- currently when you have a form consisting of nested fields, calling `getDirtyValues(fieldBag)` will only return the top level dirty fields
- since we support up to 2 nested fields in this library (field that is an object with other fields nested), we should probably return the values of these nested fields if they are dirty as well

Ex. given the following form:
```javascript
const { fields } = useForm({
    fields: {
      title: useField('some default title'),
      emails: {
        personal: useField('personal@email.com'),
        work: useField('work@email.com'),
      }
    },
    ...
  });
```

- if `title` and `emails.personal` were dirty, calling `getDirtyValues(fields)` would only yield `{title: 'my new title'}`
- with this change, nested fields would be handled, and the following example would return `{title: 'my new title', emails: { personal: 'personal-new@email.com'}}`

Fixes (issue #)


.changeset/popular-tomatoes-smash.md
```diff
@@ -0,0 +1,5 @@
+---
```
alexanderMontague, line 1: not sure if this is how I'm supposed to update the changeset...

packages/react-form/src/tests/utilities.test.ts
```diff
@@ -97,32 +97,121 @@ describe('getValues()', () => {
...
+          personal: {...mockField(formValue.emails[1].personal), dirty: true},
+        },
+        {
+          hobby: mockField(formValue.emails[2].hobby),
         },
       ],
     };
 
-    expect(getDirtyValues(fieldBag)).toStrictEqual({
+    expect(getDirtyValues(fieldBag as any)).toStrictEqual({
```
lucabezerra, line 199: Is there a reason for this `as any` cast here?

.changeset/popular-tomatoes-smash.md
```diff
@@ -0,0 +1,5 @@
...
+'@shopify/react-form': patch
+---
+
+Support returning dirty nested fields from `getDirtyValues`
```
lucabezerra, line 5: I feel like a clearer message would be something in the lines of "`getDirtyValues` detects dirty nested fields", since returning the dirty fields was already supported. Just semantics, but oh well... What do you think?

lucabezerra (commented):
Nice job,  just a few comments.

Also, are we sure (or do we even have a way of knowing whether) detecting these dirty nested fields will break someone's existing flows in any way (e.g. some automated flow that will now be prevented from continuing as normal)? Could this be a potentially breaking change? 🤔 

I find it unlikely, but just don't wanna let anything pass by unnoticed, since this is a public library.

packages/react-form/src/utilities.ts
```diff
@@ -174,15 +174,32 @@ export function getValues<T extends FieldBag>(fieldBag: T) {
...
         ...acc,
-        ...(field.dirty ? {[fieldName]: field.value} : {}),
+        ...(dirtyArray.length ? {[fieldName]: dirtyArray} : {}),
       };
-    },
-    {},
-  );
+    }
+
+    if (!isField(field)) {
```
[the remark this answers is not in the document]
alexanderMontague: Great questions and thanks for taking a look at this! To answer your review comment, I'm not sure if this would break anyones existing flow. This should still function exactly as the previous `getDirtyValues` util did, the previous tests pass and it seems to be backwards compatible. I'm open to adding more tests and I just saw the `/snapit` functionality, so open to taking a snapshot and testing in some projects where this util is used.

For this comment, you are correct that this test is functionally checking if the field is an object. I chose `!isField` as a field under the hood really is just an object haha, so any other means of checking if it is an object will yield true even when it is a field object and not just a plain object. Looking at the type definitions and testing with fields that are not just strings, anything passed into this helper should be a `FieldBag` (object collection) and if its not TS will be unhappy and it will break. Not sure how defensive we need to be here so welcome to ideas there too!

image

packages/react-form/src/tests/utilities.test.ts
alexanderMontague: The reason I casted this to any is to really test the nested functionality. Currently, the library only supports one deep nested fields (top level fields or an object containing other fields), so without this our types will complain 

image

Given the library only supports this one deep nesting, I was debating to whether even go the recursive approach since we could just brute force check a layer deep, but for JS usage and if this package ever gets updated to support deeper nested fields I went with this approach. Also open to ideas on the implementation given that context ^

.changeset/popular-tomatoes-smash.md
alexanderMontague: Nope, totally agree. I'm better with code than words haha

lucabezerra (approved):
As long as the test for the unsupported feature is removed (and, if you see fit, the changeset description is changed), it LGTM! Thanks for doing this :)

packages/react-form/src/utilities.ts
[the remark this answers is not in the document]
lucabezerra: Interesting, I had added a comment here and it didn't go along with the review 🤔 

I meant to say I was worried something that wasn't a `Field` but was something we were not expecting, like `null` (_because Javascript is weird_) could allow us to get into this `if` block. Upon further inspection, I realized my fear was unfounded, though.

PR #2474, opened 2022-11-23
Remove $ for better terminal pasting

JulienHe:
## Description
Removing the $ sign on the yarn installation to avoid to remove it when we are copy/pasting into our terminal


ryanwilsonperkin (approved):

BPScott:
If we want this convention then we should do it for all the readme files rather than just one.

I'm tempted to say a changeset isn't needed for this change - it can go out whenever we push a new release to these packages. We can skip adding a changeset to PR and add the `Skip changelog` label to the PR to make the check pass.

BPScott:
I've applied this fix to every README file in #2476.

PR #2477, opened 2022-11-30
Refactor graphql-testing e2e tests

BPScott:
## Description

Preparation for fixing a bug with how we report errors when mocks are not provided.

This PR makes it easier to test the error state, and adds a test that asserts the current behaviour when using a mock that was not defined (which shall fail in the apollo3 branch - a fix shall be in a follow-up PR)

Skipping changelog as this is a test-only change.


camd (approved):
This looks awesome!

PR #2480, opened 2022-11-30
Disable manifest caching in development

alexandcote:
## Description
This PR fixes an issue when sewing-kit only recompiles the client's assets, the manifest is cached on the server and the developer faced a script integrity issue because the new manifest file is not reloaded.

This PR adds a new option to disable the cache. This option should only be used in development.

### DIFF
--- a/packages/sewing-kit-koa/src/tests/manifests-read.test.ts
+++ b/packages/sewing-kit-koa/src/tests/manifests-read.test.ts
@@ -85,6 +85,17 @@ describe('Manifests read', () => {
     expect(readJson).toHaveBeenCalledTimes(1);
   });
 
+  it('reads manifests multiple times when caching is disabled', async () => {
+    const path = null;
+    const caching = false;
+    const manifests = new Manifests(path, caching);
+
+    await manifests.resolve('Chrome 71');
+    await manifests.resolve('Chrome 73');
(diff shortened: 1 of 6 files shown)


packages/sewing-kit-koa/src/tests/manifests-read.test.ts
```diff
@@ -85,6 +85,15 @@ describe('Manifests read', () => {
     expect(readJson).toHaveBeenCalledTimes(1);
   });
 
+  it('does not only reads manifests once when caching is disabled', async () => {
+    const manifests = new Manifests(null, false);
```
lucabezerra, line 89: Since `caching` is a new parameter and this is the only time parameters are passed to `Manifests` in this file, I think we'd benefit from naming them, even though it gets a bit verbose.

```suggestion
    const path = null;
    const caching = false; 
    const manifests = new Manifests(path, caching);
```

Also, I think this test's name could be improved. I imagine the idea was to directly negate the test above it, but I feel like it sounds a bit weird. What do you think about "reads manifests multiple times when caching is disabled"?

lucabezerra (approved):
I don't know this code very well, but the changes seem to make sense. Left a couple of comments on tiny improvements :)

PR #2488, opened 2022-12-05
Better measures with web vitals

ryanwilsonperkin:
## Description

In practice we've noticed that the First Contentful Paint metric that we record frequently appears to be wrong, showing an average value of several minutes on one of our largest users of the performance library. Looking to switch this over to the `web-vitals` implementation instead of our custom one on the expectation that it's solved more of these edge cases and may help provide us with a more accurate value.

In the process, seems worthwhile to switch over our recordings of First Input Delay (FID) and Time To First Byte (TTFB) as well for consistency, however a similar metric doesn't exist for First Paint, so I leave that as our custom implementation.


packages/performance/src/performance.ts
```diff
@@ -11,13 +11,7 @@ import {
   referenceTime,
   hasGlobal,
 } from './utilities';
-import {
-  Event,
-  EventType,
-  LifecycleEvent,
-  TimeToFirstPaintEvent,
-  TimeToFirstContentfulPaintEvent,
```
CameronGorrie, line 19: Can `TimeToFirstPaintEvent` and `TimeToFirstContentfulPaintEvent` be removed from the types file now? Or are there other references?

CameronGorrie (approved):
Did not 🎩 but this follows the same procedure as adding LCP and LGTM 👍

PR #2489, opened 2022-12-05
[statsd] Allow consumer to create child client

alexandcote:
## Description
- Update `hot-shots` to the latest version (9.3.0)
- Allow consumers to create a child client. This is super useful if you want a Singleton StatsD client in our application.


peter-hamilton (approved):
Nice. LGTM.

PR #2490, opened 2022-12-05
[koa-performance] Middleware now expect to receive a StatsDClient

alexandcote:
## Description
This PR changes the koa-performance middleware. The middleware now expects to receive a StatsDClient instead and will not create a new StatsDClient. This is super useful if you have buffering enabled on your StatsDClient.


packages/koa-performance/README.md
```diff
@@ -113,14 +126,10 @@ It takes options conforming to the following interface:
 
 ```ts
 interface Options {
-  // the prefix for metrics sent to StatsD
-  prefix: string;
+  // the StatsD Client instance
+  statsd: StatsDClient;
```
atesgoral, line 130: Shouldn't this be optional?

```suggestion
  statsd?: StatsDClient;
```

packages/koa-performance/README.md
```diff
@@ -113,14 +126,10 @@ It takes options conforming to the following interface:
...
   // whether the app is being run in development mode.
   development?: boolean;
-  // the host of the statsd server you want to send metrics to
-  statsdHost?: string;
-  // the port of the statsd server you want to send metrics to
-  statsdPort?: number;
```
atesgoral, line 123: Shouldn't these be left for the "old way"?

atesgoral: oh wait this is the README...

Issue #2497, opened 2022-12-07
[react-i18n] Support currency formatting for `ar-EG`


trishrempel:
## Overview

Support currency formatting for `ar-EG` and other similar locales.

When calling `i18n.formatCurrency` with the locale `ar-EG`, the following exception is thrown:
```
Number input in locale ar-EG is currently not supported.
```

## Motivation

> What inspired this enhancement? What makes you think this should be in `quilt`?

In `react-i18n`, `getCurrencySymbol` relies on calling [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) with the value `0`, and uses regular expressions to parse the currency symbol and position.

This is the result when I call Intl.NumberFormat with `ar-EG` vs. `ar`:
```js
console.log(new Intl.NumberFormat('ar-EG', { style: 'currency', currency: 'USD' }).format(number));
// output: "٠٫٠٠ US$"

console.log(new Intl.NumberFormat('ar', { style: 'currency', currency: 'USD' }).format(number));
// output: "US$ 0.00"
```

In Egyptian Arabic, the output is `٠٫٠٠ US$`, which doesn't match the regular expression, which is why it's throwing the exception it is.

We can possibly expand the regular expression to include  `٠` and treat it like `0`, so the regular expression can be successfully parsed. This will specifically solve the `ar-EG` use case, and any other locale that uses `٠` to represent `0`.

## Checklist

- [x] Please delete the labels section before submitting your issue
- [x] I have described this issue in a way that is actionable (if possible)

PR #2501, opened 2022-12-08
Improve currency formatting to better support `ar-EG` and similar locales

trishrempel:
## Description
PR based on https://github.com/Shopify/quilt/pull/2498.

Fixes (https://github.com/Shopify/quilt/issues/2497).

Improves currency formatting support for `ar-EG` and similar locales.

### Original problem

In `react-i18n`, `getCurrencySymbol` relies on calling [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) with the value `0`, and uses regular expressions to parse the currency symbol and position.

This is the result when I call Intl.NumberFormat with `ar-EG` vs. `ar`:
```js
console.log(new Intl.NumberFormat('ar-EG', { style: 'currency', currency: 'USD' }).format(number));
// output: "٠٫٠٠ US$"

console.log(new Intl.NumberFormat('ar', { style: 'currency', currency: 'USD' }).format(number));
// output: "US$ 0.00"
```

In Egyptian Arabic, the output is `٠٫٠٠ US$`, which doesn't match the regular expression, which is why it's throwing the exception it is.

### Force the latin numbering system

For `ar-EG`, `-1234.56` is formatted by [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat) like `؜-١٬٢٣٤٫٥٦ US$`.
When forcing a latin locale, `-1234.56` is instead formatted like `‎-US$ 1,234.56`.

This will provide greater consistency for `i18n.formatNumber()` and `i18n.formatCurrency()`, and it will ensure greater compatibility with `i18n.unformatNumber()` and `i18n.unformatCurrency()`.

### Use Unicode property escapes in number formatting regular expressions

[Unicode property escapes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Unicode_Property_Escapes) provide a broader and more comprehensive range of characters in specific categories for regular expressions, such as control formatting, punctuation, and dashes.

Using these will provide broader support for number formatting in different locales, and cover more use cases than we might initially anticipate.

### Preserve direction control characters for negative numbers

For locales like `ar-EG` and `he-IL`, Intl.NumberFormat returns direction control characters before the negative sign.
Previously, `formatCurrencyShort()` broke the order of these symbols for these locales.

Now, the order of the control characters and negative sign is preserved.

### Preserve original negative sign when formatting numbers

Previously, `formatCurrencyShort()` would replace negative signs encountered with the standard hyphen symbol ([-](https://www.compart.com/en/unicode/U+002D)), regardless of what Intl.NumberFormat returned for a given locale.

Now, `formatCurrencyShort()` preserves the original negative sign returned by Intl.NumberFormat, resulting in a more accurately localized currency string.

Additionally, `unformatNumber()` and `unformatCurrency()` were updated to scan for multiple kinds of negative signs, so they would be correctly interpreted and unformatted as negative numbers.

## Impact

By forcing latin number formatting, formatted numbers and currencies aren't as deeply localized as they otherwise could be for certain locales. However, the benefit is that the formatted numbers are more compatible with other `react-i18n` methods like `unformatNumber()` and `unformatCurrency()`, which is especially important for web input controls.

By preserving the original negative sign returned by `Intl.NumberFormat` when formatting currency, negative numbers could appear different in different locales. This could have a minor visual UX impact; for example, negative numbers lined up vertically against positive numbers may align differently.

### DIFF
--- a/.changeset/six-rice-rhyme.md
+++ b/.changeset/six-rice-rhyme.md
@@ -1,0 +1,5 @@
+---
+'@shopify/react-i18n': minor
+---
+
+Improve currency formatting by enforcing latin number formatting and preserving negative sign and directional control characters.
--- a/packages/react-i18n/README.md
+++ b/packages/react-i18n/README.md
@@ -126,9 +126,9 @@ export default withI18n()(NotFound);
 
 The provided `i18n` object exposes many useful methods for internationalizing your apps. You can see the full details in the [`i18n` source file](https://github.com/Shopify/quilt/blob/main/packages/react-i18n/src/i18n.ts), but you will commonly need the following:
 
-- `formatNumber()`: formats a number according to the locale. You can optionally pass an `as` option to format the number as a currency or percentage; in the case of currency, the `defaultCurrency` supplied to the i18n `I18nContext.Provider` component will be used where no custom currency code is passed.
+- `formatNumber()`: formats a number in the latin numbering system according to the locale. You can optionally pass an `as` option to format the number as a currency or percentage; in the case of currency, the `defaultCurrency` supplied to the i18n `I18nContext.Provider` component will be used where no custom currency code is passed.
 - `unformatNumber()`: converts a localized number string to a number string parseable by JavaScript. Example: `123.456,45 => 123456.45`
-- `formatCurrency()`: formats a number as a currency according to the locale. Its behaviour depends on the `form:` option.
+- `formatCurrency()`: formats a number as a currency in the latin numbering system according to the locale. Its behaviour depends on the `form:` option.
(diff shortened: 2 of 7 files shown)


movermeyer (approved):
Very nice!

BPScott (approved):
That PR explanation and rational is so deep I love it!

Not going to pretend I understand much of the implementation but I can't see anything sus, and the tests look fantastic!

Thanks for the contribution @trishrempel!

trishrempel:
The tests in Web against the snapshot didn't reveal anything concerning, so I'll merge now.

PR #2517, opened 2023-01-04
[react-form] export DynamicList interface

weslleyaraujo:
## Description

Export `DynamicList` on `types.ts` so that we can import as:

```tsx
import {DynamicList} from "@shopify/react-form"
```

instead of:

```tsx
import {DynamicList} from '@shopify/react-form/build/ts/hooks/list/dynamiclist';
```

### DIFF
--- a/.changeset/selfish-parents-tan.md
+++ b/.changeset/selfish-parents-tan.md
@@ -1,0 +1,5 @@
+---
+'@shopify/react-form': minor
+---
+
+Export DynamicList interface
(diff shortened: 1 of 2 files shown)


ryanwilsonperkin (approved):

BPScott:
Is `ReturnType<typeof useDynamicList>` insufficient?

weslleyaraujo:
> Is `ReturnType<typeof useDynamicList>` insufficient?
> 
> Alternatively, because in your example you're already accessing `DynamicListBag` you can get the type off that object shape too:
> 
> ```ts
> type DynamicList = DynamicListBag[keyof DynamicListBag];
> ```

Hi @BPScott!! I've tried that approach but it turns out that we will lose the ability to use the generic on `DynamicList`, i.e `DynamicList<ProductField>`

By using `DynamicListBag[keyof DynamicListBag]` we get a dynamic list of `unknown` 😕

PR #2518, opened 2023-01-04
Include link tags for CSS resource download timings

CameronGorrie:
## Description

Fixes https://github.com/Shopify/web/issues/80596

A .css file may have an initiatorType of "link" or "css" because it can either be fetched via a <link> tag or via an @import in a CSS file. While an initiatorType of "css" might actually be a foo.jpg image, because CSS fetched an image. Currently `@shopify/performance` only emits resource timing information when `initiatorType` is "css" so we're missing data for stylesheet download durations where the initiator is <link> (for Shopify/web this is almost everything).

This PR adds "link" to the watched resource types and ensures that the resource name being requested ends in `.css` before emitting the performance event.

More detail on [initiatorType](https://www.w3.org/TR/resource-timing/#dom-performanceresourcetiming-initiatortype) and since tests for this library are skipped I've got a console.log illustrating the new behaviour 🤷‍♂️ 
 
Screenshot 2023-01-04 at 11 13 40 AM

### DIFF
--- a/packages/performance/src/performance.ts
+++ b/packages/performance/src/performance.ts
@@ -107,7 +107,7 @@ export class Performance {
         }
 
         const type = getResourceTypeFromEntry(entry);
-        if (type == null) {
+        if (type === 'unsupported') {
           return;
         }
 
--- a/packages/performance/src/tests/utilities.test.ts
+++ b/packages/performance/src/tests/utilities.test.ts
@@ -3,12 +3,12 @@ import {EventType} from '../types';
 
 describe('utilities', () => {
   describe('getResourceTypeFromEntry()', () => {
-    it('returns null for unsupported initiator types', () => {
+    it('returns "unsupported" for unsupported initiator types', () => {
       expect(
         getResourceTypeFromEntry({
           initiatorType: 'audio',
         } as PerformanceResourceTiming),
-      ).toBeNull();
+      ).toBe('unsupported');
     });
 
     it('returns stylesheet event type when initiatorType is css', () => {
--- a/packages/performance/src/utilities.ts
+++ b/packages/performance/src/utilities.ts
@@ -182,13 +182,13 @@ export function getResourceTypeFromEntry({
 
 function getResourceTypeFromInitiator(
   initiatorType: string,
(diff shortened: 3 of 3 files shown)


packages/performance/src/performance.ts
```diff
@@ -105,12 +105,21 @@ export class Performance {
           return;
         }
 
+        let type: EventType.StyleDownload | EventType.ScriptDownload;
+        if (entry.initiatorType === 'link' && !entry.name.endsWith('.css')) {
```
ryanwilsonperkin, line 109: Are there other link-loaded resources that we might want to support in the future that this approach could make more difficult to track? For instance, if we were using a `<link rel="preload" ...>` to load an image asset or similar, then this early return approach would mean that the asset is never captured. The existing state of the code still fails this as it would mark the asset as a script, but I wonder if it might make more sense to flip this around so that we're deliberately checking if it's a CSS link to set the `type` rather than checking if it _isn't_ as a guard clause.


Does that make sense? My brain is still recuperating from the break 😄

CameronGorrie: The current library implementation is only concerned with the following [navigation event types](https://github.com/Shopify/quilt/pull/2518/files#diff-2fc98e52f9d7f0430bd4b1137c16926e157595d381234d81e1f118622adced8bL104) which is why I opted for the type condition so we don't emit metrics for images or [other content types that could be preloaded](https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/preload#what_types_of_content_can_be_preloaded). Otherwise we'd need a mapping of supported file types requested by the link to new EventTypes for image, video, document etc. I'm fine with making this a breaking change and adding this if you think that approach would be preferable.

Another, quicker fix would be to introduce a "default" case where the eventType is set to "unknown" for <link> resources that aren't CSS or Script type. What do you think?

ryanwilsonperkin: An unknown/other event type seems like a decent idea as well, since presumably those would otherwise get lumped into "script" at the moment which would be undesirable

ryanwilsonperkin (approved):

PR #2522, opened 2023-01-10
Support listing hidden zones when loading countries

cejaekl:
## Description

Related:
 - https://github.com/Shopify/address/issues/562#issuecomment-1372701071
 - https://github.com/Shopify/country_db/pull/1466
 - https://github.com/Shopify/atlas/pull/2099

As we work to roll out changes to the list of zones (provinces) for a country, we are adding the ability to define a zone before it becomes visible to buyers, thus giving us time to run a backfill adjusting shipping rates.  This change makes it possible to ask for the list of zones in a country both without (default) and with the hidden zones included.


packages/address/src/tests/mocks.test.ts
```diff
@@ -22,7 +22,7 @@ describe('mockCountryRequests', () => {
   });
 
   it('fetches a countries fixture', async () => {
-    const [firstCountry] = await loadCountries('ja');
+    const [firstCountry] = await loadCountries('ja', false);
```
trishrempel, line 25: As a follow-up, I suggest updating `packages/address-mocks` to handle `includeHiddenZones` in `mockCountryRequests()`, perhaps to return a dummy zone for one `en` country if `includeHiddenZones` is true.

royleejr (approved):
code changes look good to me!

packages/address/src/tests/AddressFormatter.test.ts
```diff
@@ -137,6 +137,13 @@ describe('AddressFormatter', () => {
 
       expect(fetch.calls()).toHaveLength(2);
     });
+
+    it('allows the optional parameter `includeHiddenZones:`.', async () => {
+      const addressFormatter = new AddressFormatter('en');
+      await addressFormatter.getCountries({includeHiddenZones: true});
+
+      expect(fetch.calls()).toHaveLength(1);
```
BPScott, line 145: This test doesn't seem very useful.  It should test that `getCountries` returns some new additional countries that were not available when you call getCountries without `includeHiddenZones`. It sounds like this can be done by following @trishrempel's suggestion to add support for `includeHiddenZones` to the address mocks.


We should also test that the the regular getCountries omits that hidden country.

packages/address/src/AddressFormatter.ts
```diff
@@ -29,12 +33,18 @@ export default class AddressFormatter {
     return loadCountry(this.locale, countryCode);
   }
 
-  async getCountries(): Promise<Country[]> {
-    const cachedCountries = ORDERED_COUNTRIES_CACHE.get(this.locale);
+  async getCountries(options?: LoadCountriesOptions): Promise<Country[]> {
+    const includeHiddenZones = options?.includeHiddenZones || false;
```
BPScott, line 37: Nitpick: You can do the destructuring in the parameter arguments. This would also allow you to remove the LoadCountriesOptions. as the booleanness of `includeHiddenZones` can be inferred


```suggestion
  async getCountries({includeHiddenZones = false} = {}): Promise<Country[]> {
```

packages/address/src/loader.ts
```diff
@@ -10,8 +10,11 @@ import {
 
 import query from './graphqlQuery';
 
-export const loadCountries: (locale: string) => Promise<Country[]> =
-  memoizeAsync(async (locale: string) => {
+export const loadCountries: (
+  locale: string,
+  includeHiddenZones: boolean,
```
BPScott, line 15: `loadCountries` is part of the public API of this package as it (and every other export in loader.ts) is exported by the index.ts file. Adding a new required argument to a public API function is a breaking change that would require a new major version.

Please make `includeHiddenZones` an optional parameter, with the default value of `false`. That way we maintain backwards compatibility (i.e. people doing `loadCountries('en')` won't get an error) and this is a minor, non-breaking change.

packages/address/src/tests/mocks.test.ts
BPScott: When we make the `loadCountries`'s second parameter optional per above, this can be reverted.

BPScott (changes requested):
Thanks for the PR!

In its current state it is a breaking change, as adding a required parameter to `loadCountries` changes the public API in a backwards incompatible manner. This can be fixed by making the `includeHiddenZones` parameter optional.

I'd also want to see tests that prove that including/omitting `includeHiddenZones` results in hidden zones being included/omitted in the result of `(new AddressFormatter('pt-br')).getCountries()`. Currently there is no test that shows using the parameter has any effect.

packages/address/src/tests/AddressFormatter.test.ts
cejaekl: That's a fair point.

I've revised the mocks to return an extra zone, `DH`, for country `IN` India, and put new tests in `AddressFormatter.test.ts` that check for its presence/absence:
 - returns hidden zones when `includeHiddenZones` is `true`
 - excludes hidden zones when `includeHiddenZones` is `false`
 - excludes hidden zones when `includeHiddenZones` is not passed

https://github.com/Shopify/quilt/pull/2522/commits/d17b6386b1437eb861b2b44e46a472652352edfa#diff-03ac05a5590924ec807cb0030e4ba5d5ac68662f4726dd8470be77887228a0a9

packages/address/src/loader.ts
cejaekl: Thank you for calling me out on this!  (My grasp of TypeScript isn't what it should be.)

Have I managed to get it right now, like this?
```patch
diff --git a/packages/address/src/loader.ts b/packages/address/src/loader.ts
index 8e9399945..816fbf63f 100644
--- a/packages/address/src/loader.ts
+++ b/packages/address/src/loader.ts
@@ -12,7 +12,7 @@ import query from './graphqlQuery';

 export const loadCountries: (
   locale: string,
-  includeHiddenZones: boolean,
+  includeHiddenZones?: boolean,
 ) => Promise<Country[]> = memoizeAsync(
   async (locale: string, includeHiddenZones: boolean) => {
     const response = await fetch(GRAPHQL_ENDPOINT, {
@@ -42,8 +42,13 @@ export const loadCountries: (
 export const loadCountry: (
   locale: string,
   countryCode: string,
+  includeHiddenZones?: boolean,
 ) => Promise<Country> = memoizeAsync(
-  async (locale: string, countryCode: string): Promise<Country> => {
+  async (
+    locale: string,
+    countryCode: string,
+    includeHiddenZones: boolean,
+  ): Promise<Country> => {
     const response = await fetch(GRAPHQL_ENDPOINT, {
       method: 'POST',
       headers: HEADERS,
@@ -53,6 +58,7 @@ export const loadCountry: (
         variables: {
           countryCode,
           locale: locale.replace(/-/, '_').toUpperCase(),
+          includeHiddenZones,
         },
       }),
     });
```

BPScott (approved):
Thanks for your patience, this fell off the radar.

I've took the liberty of pushing up a few small changes, rather than going back and forth with review feedback.

- I have added back the  country mocks in `address-mocks/src/fixtures/country-*`. This is because removing them is a breaking change because they're exported by the package. Consumers could do `import {fixtures} from '@shopify/address-mocks'; const x = fixtures.country.EN`. and removing these fixtures would break that behaviour. It looks like there's only one of `fixture.country` inside the Shopify org according to grokt, so I think this is a good bit of tidying, but lets do it in a separate PR, so we don't tie feature improvements and breaking changes together needlessly.
- I've renamed the key in the address mocks fixtures array from `includeHiddenZones` to `countriesIncludeHiddenZones`. As mentioned above this is part of the public API and I think this name is a bit more clear that it relates to a mock of the countries response.
- I've tweaked the types of `loadCountry`, `AddressFormatter.getCountry()` and `AddressFormatter.getCountries()`, so that their second parameter is an optional options object. Your change based on my suggestion was great, but then I realized that we might have additional options that we'd may want to set in the future - I note that `includeDeprecatedZones` is an option based upon the PRs in Atlas you linked to. Using an options object here will mean the list of parameters won't grow over time if we ever need to add more options.
- I've adjusted the changeset wording and added a changeset for `@shopify/address-mocks`.

---

There was one thing I was wondering about - I wonder if it'd be worth writing some logic into `address-mocks` `mockCountryRequests()` to inject that hidden `DH` region into India's zones programatically, instead of having to add 39k lines worth of new fixture data. Something like the following should avoid needing to add the `countries_*_includehiddenzones.ts` files:

```
const countries = fixtures.countries[locale];

if (includeHiddenZones) {
  const indiaIndex = countries.data.countries.findIndex(
    (country) => country.code === 'IN',
  );

  countries.data.countries[indiaIndex].zones.push({
    name: 'Dadra and Nagar Haveli and Daman and Diu',
    code: 'DH',
  });
}
```

Does that seem worthwhile or do you think it's important and useful to have the fixtures with the hidden zones available?

--
Thanks for the PR and your patience with waiting for my rereview. This good to go now! Squash and merge when you're happy!

PR #2525, opened 2023-01-17
`[react-form]` - Add `FormProvider` Context

Crunchyman-ralph:
## Description

Fixes (issue #)

- Add a `FormProvider`

### DIFF
--- a/.changeset/rare-steaks-exercise.md
+++ b/.changeset/rare-steaks-exercise.md
@@ -1,0 +1,5 @@
+---
+'@shopify/react-form': minor
+---
+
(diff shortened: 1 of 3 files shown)


Crunchyman-ralph:
PR summary definitely needs some polishing and I wasn't really sure how to run tests on the `@shopify/react-form` but the code works and I am currently using it on a personal project, so I decided to contribute to the actual package since I think this is something that is present in other react-form packages.

mateus:
We would suggest prop drilling to solve this problem, or introduce the React Context Provider pair directly to the application that you need this.

PR #2526, opened 2023-01-17
Remove the deprecated type attribute from <style> and <script>

atesgoral:
## Description

`type="text/css"` is deprecated for `<style>`:

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/style#attr-type

> This attribute should not be provided: if it is, the only permitted values are the empty string or a case-insensitive match for text/css.

`type="text/javascript"` is deprecated for `<script>`:

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script/type

> Authors are encouraged to omit the attribute if the script refers to JavaScript code rather than specify a MIME type.

Removing them from HTML generation.

### DIFF
--- a/packages/react-html/src/server/components/Script.tsx
+++ b/packages/react-html/src/server/components/Script.tsx
@@ -10,7 +10,7 @@ export function Script(props: Props) {
 
   const attributes = {
     ...otherProps,
-    type: !type || isNoModule ? 'text/javascript' : type,
+    type: isNoModule ? 'text/javascript' : type,
     defer: type === 'module' ? undefined : defer,
     noModule: isNoModule ? true : undefined,
   };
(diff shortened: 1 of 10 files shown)


packages/react-html/src/server/components/Script.tsx
```diff
@@ -10,7 +10,7 @@ export function Script(props: Props) {
 
   const attributes = {
     ...otherProps,
-    type: !type || isNoModule ? 'text/javascript' : type,
+    type: isNoModule ? 'text/javascript' : type,
```
atesgoral, line 13: I don't know why we're forcing `nomodule` to become `text/javascript`, but I won't touch this in this PR.

developit: Seems like this logic is just avoiding setting type on module scripts since they already have type=module?

developit (approved):
Yay!

ciuliano-shopify (approved):

Issue #2533, opened 2023-01-19
[graphql-testing] allow resolveAll() to resolve selectively by input values


IlanaB:
We ran into an issue testing a scenario where we have 2 outstanding queries to the same endpoint but with different inputs and need to test a scenario where one resolves while the other is still pending. There's no way to do this currently so I'd like to add an option to resolve by input in addition to the existing `query` option accepted by `resolveAll()`

              so it turns out that our existing test utilities don't provide any way to accurately test this. For reference, [this is the code for graphQL.resolveAll()](https://github.com/Shopify/quilt/blob/main/packages/graphql-testing/src/graphql-controller.ts#L78), where we can specifically resolve all requests for a given query but only for all outstanding requests based on query name rather than what inputs the query was called with.

Our specific simplified use case:
1. User increases the number input by 1: we send to `calculateRefund` query with quantity: 1
1. User increases the number input to 2: we send to `calculateRefund` query with quantity: 2
1. The first query resolves with `subtotalAmount: '$1'`. At this point we want to test that we did not set component state with `loading: false, subtotalAmount: '$1'` (so that user doesn't see this update) because we are still loading the calculated total with the quantity of 2. In this PR we prevent this in our code by discarding the response from older requests using the ref we created however there's no way in our mocks to have only one of the queries resolve while the other is already pending. I intend to open a PR to quilt to allow filtering by inputs but until that's done there's no way to do this without adding support for a callback or something in our production code just for the sake of testing.

_Originally posted by @IlanaB in https://github.com/Shopify/web/pull/80730#discussion_r1062778982_

PR #2534, opened 2023-01-19
graphql-testing resolveAll: allow variable specificity

IlanaB:
## Description

Fixes https://github.com/Shopify/quilt/issues/2533


IlanaB:
> @IlanaB What would you think about instead adding a `filter` option that accepts an `Operation` as a parameter and returns a `boolean`? A custom filter option could enable the use-case you're looking to solve for, and potentially eliminate the need to add more `FindOptions` in the future if other tests need to filter operations by additional conditions.
> 
> I created a rough draft PR here as an example: #2543

That's actually exactly the approach I was starting to entertain when I was thinking about this a few nights ago and I think I completely forgot about it. This is much better, thank you!!! Do you want to carry on with your PR then? Or do you prefer I pick that up?

Am I right in that the existing `FindOptions` aren't tested anywhere currently? I wanted to add a test for these changes but it looks like there's nothing existing to work off of in testing that selective operations resolve.

acmertz:
> That's actually exactly the approach I was starting to entertain when I was thinking about this a few nights ago and I think I completely forgot about it. This is much better, thank you!!! Do you want to carry on with your PR then? Or do you prefer I pick that up?

@IlanaB sure! I'm happy to pick it up. I'll add some tests and tag you as a reviewer so you're kept in the loop.

PR #2566, opened 2023-02-15
[StatsD] Support amount input on increment

RouzbehMajidi:
The base StatsD client we use (and also many other clients) support an optional input on `.increment` which increments the metric by the inputted value. This is especially helpful for bulk actions, currently we have to increment in a for loop. I think it's worth exposing this parameter as an optional input.

We've implemented this two different ways (each with pros and cons given the current setup)

### Method 1: Separate `incrementByAmount` method [9f10c4c80f1528fcc4968790a0f71059c0b3edaf]
This keeps things nice and clean input param wise, but can't help but think this action should be more of an overload than a separate method
```ts
incrementByAmount('metric', 5, ['tags'])
```
This is also inline with other similar packages

### Method 2: New optional param in `increment` [35bd84ebdc7525f1a1c90b3c4469abc21822dc84]
This allows for a cleaner interface, but with the downside of needing to also specify the optional `metricOptions` field if you want to specify `value`. Something like this
```ts
increment('metric', ['tags'], {}, 5)
```

Will add in the changelog once we decide on which way we want to go

### DIFF
--- a/packages/statsd/src/client.ts
+++ b/packages/statsd/src/client.ts
@@ -127,11 +127,16 @@ export class StatsDClient {
     });
   }
 
-  increment(stat: string | string[], tags?: Tags, options: MetricOptions = {}) {
+  increment(
+    stat: string | string[],
+    tags?: Tags,
+    options: MetricOptions = {},
+    value = 1,
+  ) {
     return new Promise<void>((resolve) => {
       this.statsd.increment(
         stat,
-        1,
+        value,
         options.sampleRate,
         this.normalizeTags(tags),
         this.createCallback(resolve),
--- a/.changeset/tender-lemons-film.md
+++ b/.changeset/tender-lemons-film.md
@@ -1,0 +1,5 @@
+---
+'@shopify/statsd': minor
+---
+
+Added support for incrementing metric by specific value
--- a/packages/statsd/README.md
+++ b/packages/statsd/README.md
@@ -77,6 +77,17 @@ statsdClient.increment(
 );
 ```
 
+Increment can also be supplied a value to increment the metric by.
+
+```javascript
+statsdClient.increment(
+  'myCounter',
+  ['navigation', 'complete', 'performance'], // user-defined tags to go with the data,
+  {}, // additional options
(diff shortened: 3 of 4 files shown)


packages/statsd/src/client.ts
```diff
@@ -139,6 +139,23 @@ export class StatsDClient {
     });
   }
 
+  incrementByAmount(
```
kvnify, line 142: So I think you can do this:
```typescript
  increment(stat: string | string[], tags?: Tags, options?: MetricOptions, value?: number);
  increment(stat: string | string[], tags?: Tags, options: MetricOptions = {}, value: number = 1) {
    return new Promise<void>((resolve) => {
      this.statsd.increment(
        stat,
        value,
        options.sampleRate,
        this.normalizeTags(tags),
        this.createCallback(resolve),
      );
    });
  }
```
To overload the method we have and not break the existing client usage.

RouzbehMajidi: I added [35bd84e](https://github.com/Shopify/quilt/pull/2566/commits/35bd84ebdc7525f1a1c90b3c4469abc21822dc84) with the changes. I think there's pros and cons to both, but the former (`incrementByAmount`) is a bit more explicit.

I personally prefer `increment` but just can't seem to get it clean enough, the current setup will require an empty metric options array if you choose to add a value in, i.e.

```ts
increment('metric', ['tags'], {}, 5)
```

Let me know what you think!

kvnify: I wish we'd used the approach of named parameters with a destructured object for these API for this very reason. It's much easier to evolve an API that way.

ryanwilsonperkin (commented):
Approach sounds good to me, it's too bad we didn't do this from the start to be able to have the `value` argument in the same place as for the other methods, but adding it on at the end optionally seems like a good way to do this so that we can support forwards.

Please generate a changeset file by running `yarn changeset add` and follow the prompts, then commit the generated changeset file which will allow us to release this new version

ryanwilsonperkin (approved):

PR #2576, opened 2023-02-22
Mklocke/add cls to performance

mklocke-shopify:
## Description

Fixes (issue #)


packages/koa-performance/src/middleware.ts
```diff
@@ -126,9 +126,14 @@ export function clientPerformanceMetrics({
...
 
+        const roundedValue =
+          event.type === EventType.CumulativeLayoutShift
+            ? value
+            : Math.round(value);
+
         metrics.push({
           name: eventMetricName(event),
-          value: Math.round(value),
+          value: roundedValue,
```
mklocke-shopify, line 136: The reason I added this is because CLS is calculated as a fraction (e.g. 0.1, 0.25) - the rounding here was rounding all CLS values down to 0.

This is not a super elegant way of handling it and I would happily change this if anyone has a better suggestion.

lucabezerra: I'm curious, would there be value in not rounding the other metrics too? Or to rounding them all to a set number of decimal places (say 2 or 3)?

mklocke-shopify: Good question! I think most (all?) of the other metrics are timestamps in nanoseconds? If so it probably wouldn't hurt to remove rounding or set to 2/3 decimal places as you suggest. I wonder why the rounding was added in here in the first place?

ryanwilsonperkin: If we need to do integers, it'd be more effective to multiply the CLS value by 100 and round

mklocke-shopify:
Sorry for the delay on this folks, I know its getting stale. I'll get to this again this week I hope.

packages/performance/src/types.ts
```diff
@@ -5,6 +5,7 @@ export enum EventType {
   TimeToLargestContentfulPaint = 'ttlcp',
   DomContentLoaded = 'dcl',
   FirstInputDelay = 'fid',
+  CumulativeLayoutShift = 'cls',
```
gbzodek, line 8: (passing by as I am interested by that PR)

It should also be added to the [README list of lifecycle events](https://github.com/Shopify/quilt/tree/main/packages/performance#events)

lucabezerra:
Do we care about commit organization in quilt? Wondering if it'd be valid to squash the commits into a single one before merging.

lucabezerra:
Is this PR moving forward or should we close it? I was hoping to stop getting daily notifications from it 😅

packages/performance/src/types.ts
mklocke-shopify: Thanks! Added here: 3f30837194f1cdaac00997629a93680f8b5fd1f6

mklocke-shopify:
I've squashed the commits and updated the readme. This is ready for review now.

If you have a few minutes @ryanwilsonperkin @lucabezerra please have another look and let me know if it looks okay, thanks!

lucabezerra (approved):
No worries, thanks for circling back to this!

From my end it looks ok, but since @ryanwilsonperkin flagged the issue, let's wait for him to say something :)

ryanwilsonperkin (approved):

PR #2577, opened 2023-02-23
Bugsnag: turn off trackInlineScripts

developit:
## Description

See [this document](https://docs.google.com/document/d/1btW6nWLW92R5fRDtMb6PRXe19Sj1aBkJiBh_HYtPKxw/edit) for rationale and more information.


developit:
To whoever might know: it looks like we're using `@bugsnag/js` here, and this option is only available in `@bugsnag/browser`. Are we adding plugins to Bugsnag manually?

BPScott (approved):

BPScott:
I'm tempted to say this a breaking change that users should opt into rather than getting for free - we're changing default configuration.

PR #2600, opened 2023-03-20
Enabling caching per manifest path

alexandcote:
## Description
Caching manifest base on the manifest path. If your application use two manifest, you can now still use caching to avoid reading the file system multiple time.


packages/sewing-kit-koa/src/middleware.ts
```diff
@@ -18,22 +18,28 @@ export interface Options {
...
   manifestPath?: string;
   caching?: boolean;
+  key?: symbol | string;
 }
 
 const ASSETS = Symbol('assets');
 
-export function getAssets(ctx: Context): Assets {
-  return (ctx.state as any)[ASSETS];
+export function getAssets(ctx: Context, key: string | symbol = ASSETS): Assets {
```
ryanwilsonperkin, line 26: This shouldn't allow arbitrary `key` access on context as it's a generic koa context object and this would allow you to accidentally overwrite or access values across middleware. If the goal is to allow multiple different types of asset access by key within the object then `ctx[ASSETS]` will need to be changed to a Map/Object instead that can maintain each of those. We can create a second `DEFAULT_ASSETS = Symbol('default_assets')` within that map to preserve the existing behaviour.

alexandcote:
Thanks for the review @ryanwilsonperkin 

Let me know if I understood correctly your suggestion 😄 

Here is some explanation of why consumers could need to have 2 manifests on the same server.

- Imagine a Web application which is compiled using `sewing-kit` containing the `main` entry point. The `main` entry point use React.
- Imagine a second entry point in the same project but this entry point need to use Preact instead of React. Webpack doesn't have a way at the moment to alias React with Preact only for one entry point.

The solution to that is to compile the second entry point using a separate webpack config. Because of that, we end up having 2 manifests.

Another potential solution could be to merge both manifests together BUT the task is much more complex. Each manifest is already consolidated because we have multiple builds (latest/baseline).

Let me know if you need more details 😄

packages/sewing-kit-koa/src/middleware.ts
```diff
@@ -18,22 +18,36 @@ export interface Options {
...
 const ASSETS = Symbol('assets');
+const DEFAULT_ASSETS = Symbol('default_assets');
 
-export function getAssets(ctx: Context): Assets {
-  return (ctx.state as any)[ASSETS];
+export function getAssets(
+  ctx: Context,
+  key: string | symbol = DEFAULT_ASSETS,
+): Assets {
+  return (ctx.state as any)[ASSETS]?.[key];
```
alexandcote, line 31: The return type doesn't reflect reality but it was already the case so I'm not sure if we should change the return type 🤔?

ryanwilsonperkin: Isn't it? We're using `setAssets` to set an `Assets` value in there, we're using `getAssets` to retrieve it so it should still be expected to return an `Assets` value. Unless you mean because it's potentially undefined?

PR #2602, opened 2023-03-21
[jest-koa-mocks] Using non-fully qualified URL

mateus:
## Description

Update `createContext()`  so that `url` options works the same way as it does in Koa, as well as the message URL [defined by Node.js](https://nodejs.org/api/http.html#messageurl), which is not a fully qualified URL.

> Request URL string. This contains only the URL that is present in the actual HTTP request. Take the following request:
>
> ``` 
> GET /status?name=ryan HTTP/1.1
> Accept: text/plain 
> ```

### DIFF
--- a/packages/jest-koa-mocks/src/create-mock-context/create-mock-context.ts
+++ b/packages/jest-koa-mocks/src/create-mock-context/create-mock-context.ts
@@ -69,8 +69,6 @@ export default function createContext<
     state,
   };
 
-  const protocolFallback = encrypted ? 'https' : 'http';
-
   const req = httpMocks.createRequest({
     url,
     method,
@@ -88,7 +86,7 @@ export default function createContext<
   req.socket = new stream.Duplex() as any;
   Object.defineProperty(req.socket, 'encrypted', {
     writable: false,
-    value: protocolFallback === 'https',
+    value: encrypted,
   });
 
   const res = httpMocks.createResponse();
--- a/packages/jest-koa-mocks/src/create-mock-context/create-mock-context.ts
+++ b/packages/jest-koa-mocks/src/create-mock-context/create-mock-context.ts
@@ -1,4 +1,3 @@
-import {URL} from 'url';
 import stream from 'stream';
 
 import httpMocks, {RequestMethod} from 'node-mocks-http';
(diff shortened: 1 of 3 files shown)


BPScott (approved):

packages/jest-koa-mocks/src/create-mock-context/create-mock-context.ts
```diff
@@ -70,17 +69,14 @@ export default function createContext<
     state,
   };
 
-  const protocolFallback = encrypted ? 'https' : 'http';
-  const urlObject = new URL(url, `${protocolFallback}://${host}`);
-
   const req = httpMocks.createRequest({
-    url: urlObject.toString(),
+    url,
```
falkenhawk, line 73: with this change handling fully-qualified URLs is broken.
```javascript
const context = createContext({ url: 'http://mystore.com/admin' });
console.log(context.host); // 'test.com'
```

What was wrong with using `url.URL` anyway? It works just fine with both fully and non-fully qualified URLs, falling back to `base` provided in its second argument 🤔

But if you think it's better how it is now for you, this should have been released as major version v6, since it's a breaking change. Just sayin' :)

falkenhawk: @mateus with this change handling fully-qualified URLs is broken.
```javascript
const context = createContext({ url: 'http://mystore.com/admin' });
console.log(context.host); // 'test.com'
```

What was wrong with using `url.URL` anyway? It works just fine with both fully and non-fully qualified URLs, falling back to `base` provided in its second argument 🤔

But if you think it's better how it is now for you, this should have been released as major version v6, since it's a breaking change. Just sayin' :)

falkenhawk, line 73: now it is necessary to process fully-qualified url on one's own before passing it to `createMockContext`, something what that function did internally and worked fine for both relative and absolute URLs before this PR :)
```javascript
const { host, protocol, pathname, searchParams } = new URL(url);

const ctx = createMockContext({
  url: pathname + searchParams,
  host,
  encrypted: protocol === 'https',
});
```

but I see the point, maybe that url: urlObject.toString() was wrong on line 77, since you expected only the relative part of the URL to be there. So, this could be used instead to ensure that:

```suggestion
    url: urlObject.pathname + urlObject.searchParams,
```

cheers 🍻

PR #1821, opened 2021-04-07
[react-testing] added TestInnerWrapper to cut down on element building

melnikov-s:
## Description

extracted commit from : https://github.com/Shopify/quilt/pull/1812

see: https://github.com/Shopify/quilt/pull/1812#issuecomment-814914125

Fixes (issue #)

## Type of change

- [x]  Patch: Bug (non-breaking change which fixes an issue)
- [ ]  Minor: New feature (non-breaking change which adds functionality)
- [ ]  Major: Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist

- [ ] I have added a changelog entry, prefixed by the type of change noted above (Documentation fix and Test update does not need a changelog as we do not publish new version)

PR #2612, opened 2023-03-23
[react-testing] Fixed a null exception if the inner root component is unmounted by the outer wrapper during an act

melnikov-s:
## Description

This fixes an edge-case that was introduced in https://github.com/Shopify/quilt/pull/1821

This is a rare occurrence but can happen if somehow the test wrapper unmounts the inner test root during an action. We found this while performing the react router 6 migration on `web`. Since we wrap our inner test root with a `<Route>` component at a given path a navigation within the test to a different route will unmount the component resulting in a cryptic error message:

`TypeError: Cannot use 'in' operator to search for '_reactInternalFiber' in null`

This PR prevents the root from updating when the inner wrapper is unmounted, once/if it is remounted subsequent updates will work as expected.

### DIFF
--- a/.changeset/witty-paws-chew.md
+++ b/.changeset/witty-paws-chew.md
@@ -1,0 +1,5 @@
+---
+'@shopify/react-testing': patch
+---
+
+Fixed a null exception if the inner root component is unmounted by the outer wrapper during an act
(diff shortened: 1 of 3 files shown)


iAmNathanJ (approved):
🚀

ryanwilsonperkin (approved):

PR #2613, opened 2023-03-23
Allow GraphQL v16 in quilt packages

mateus:
## Description

Currently, [graphql](https://www.npmjs.com/package/graphql) dependencies from some of the packages in Quilt are using a version range capped at v16, however, it's `graphql` now on version 16.0.0. This PR increases the range allowing usage of graphql v16.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "fs-extra": "^9.1.0",
     "get-port": "^5.0.0",
     "glob": "^7.1.6",
-    "graphql": "^16.6.0",
+    "graphql": "^15.8.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -61,7 +61,9 @@
     "fs-extra": "^9.1.0",
     "get-port": "^5.0.0",
     "glob": "^7.1.6",
-    "graphql": "^15.4.0",
+    "graphql": "^16.6.0",
+    "graphql15": "npm:graphql@^15.8.0",
+    "graphql16": "npm:graphql@^16.6.0",
```
mateus, line 66: Latest from each version

Issue #2620, opened 2023-03-30
Incorrect navigation_complete metrics due to counting non-Finished Navigations


ryanwilsonperkin:
## Overview

`@shopify/performance`'s `Navigation` object has a concept of a "Finished" result, as well as a "Cancelled" and "TimedOut". Any Navigations that get cancelled as a result of a new Navigation being started (eg. a user quickly clicking links before any one page can finish loading) will have the "Cancelled" status, whereas the "Finished" status has to be added deliberately, such as in the case of `@shopify/react-performance` where it can be marked finished by means of rendering a `PerformanceMark` with `Stage.Complete`.

When `@shopify/react-performance` sends these navigations in its report, it also includes any Navigations which were cancelled, and the backend doesn't differentiate on the result, it simply sends a `navigation_complete` metric for each Navigation that it receives.

This results in skewed metrics which count both the completed and the cancelled navigations, and makes the metrics unreliable. We believe that a better default is to only collect the Finished navigations in the report, but to give the consumer of the library an option to opt-in to other statuses as well.

This change bumps `@shopify/react-performance` to a new major version to indicate that if you want to keep the existing behaviour of collecting Cancelled and TimedOut navigations as well, that you'll have to deliberately opt-in.

## Consuming repo

Shopify's Web Admin

PR #2621, opened 2023-03-30
Only report finished navigations

ryanwilsonperkin:
## Description

Fixes https://github.com/Shopify/quilt/issues/2620
Changes the default behaviour of `usePerformanceReport` to only report on Navigations that have a result=Finished. Cancelled and TimedOut Navigations will no longer be sent meaning that metrics will be more accurate to their true intention of measuring completed navigations.

### DIFF
--- a/.changeset/forty-crews-fold.md
+++ b/.changeset/forty-crews-fold.md
@@ -1,0 +1,11 @@
+---
+'@shopify/react-performance': major
+---
+
+Default to only sending metrics for "Finished" navigations
+
+usePerformanceReport and PerformanceReport will now default to only sending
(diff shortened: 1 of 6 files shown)


ryanwilsonperkin:
Confirmed working in Spin

CameronGorrie (approved):
Nice sleuthing around for this fix! Is there a use for reporting on unfinished navigations? I'm wondering if the flag should be inverted e.g. `reportAllNavigations`

CameronGorrie:
> Do you feel strongly about changing it?

No I don't feel strongly about it. Go with your preference.

PR #2622, opened 2023-03-30
admin-graphql-api-utilities: Add `Gid` type

MrGVSV:
## Description

This PR adds a `Gid` type that can be used by consumers of `composeGid` and `composeGidFactory` to provide strongly-typed GID values. This makes it a lot less likely for one GID type to accidentally be used in place of another. For example, with the current API, this is perfectly fine:

```ts
async function getOrder(gid: string) {
 // ...
}

// Oops! This should say 'Order'!
const orderId = composeGid('Orde', 123);

// Runtime Error!
const order = await getOrder(orderId);
```

However, with this change we can easily make this catch the typo as a compile-time error:

```ts
async function getOrder(gid: Gid<'shopify', 'Order'>) {
 // ...
}

// Oops! This should say 'Order'!
const orderId = composeGid('Orde', 123);

// Compile-time Error!
const order = await getOrder(orderId);
```

Additionally, since `Gid` can implicitly cast down to `string`, this change should only affect those who opt into this behavior.

### Possible Additions

1. We could also introduce a Shopify-specific `Gid` type to reduce the amount of typing needed for users using this with the Shopify API:
    ```ts
    type ShopifyGid<T extends string> = Gid<'shopify', T>; 
    ```
2. We could also include type guards like `isGidFactory` and `isGid` to take in a string and check if it's the desired `Gid` type:
    <details>
    <summary>Code Sample</summary>
    
    ```ts
    export function isGidFactory<N extends string>(namespace: N) {
      return function isGid<T extends string = unknown>(
        gid: string,
        key?: T,
      ): gid is Gid<N, T> {
        if (!gid.startsWith(`gid://${namespace}/`)) {
          return false;
        }
    
        try {
          if (key !== undefined && parseGidType(gid) !== key) {
            return false;
          }
    
          parseGid(gid);
        } catch {
          return false;
        }

        return true;
      };
    }
    ```
    
    </details>

### DIFF
--- a/packages/admin-graphql-api-utilities/src/index.ts
+++ b/packages/admin-graphql-api-utilities/src/index.ts
@@ -1,6 +1,11 @@
 const GID_TYPE_REGEXP = /^gid:\/\/[\w-]+\/([\w-]+)\//;
 const GID_REGEXP = /\/(\w[\w-]*)(?:\?(.*))*$/;
 
+export type Gid<
+  Namespace extends string,
+  Type extends string,
+> = `gid://${Namespace}/${Type}/${number | string}`;
+
 interface ParsedGid {
   id: string;
   params: {[key: string]: string};
@@ -43,21 +48,21 @@ export function parseGidWithParams(gid: string): ParsedGid {
   throw new Error(`Invalid gid: ${gid}`);
 }
 
-export function composeGidFactory(namescape: string) {
-  return function composeGid(
-    key: string,
+export function composeGidFactory<N extends string>(namescape: N) {
+  return function composeGid<T extends string>(
+    key: T,
     id: number | string,
     params: {[key: string]: string} = {},
-  ): string {
+  ): Gid<N, T> {
     const gid = `gid://${namescape}/${key}/${id}`;
     const paramKeys = Object.keys(params);
 
     if (paramKeys.length === 0) {
-      return gid;
+      return gid as Gid<N, T>;
     }
 
     const paramString = new URLSearchParams(params).toString();
-    return `${gid}?${paramString}`;
+    return `${gid}?${paramString}` as Gid<N, T>;
   };
 }
 
--- a/packages/admin-graphql-api-utilities/README.md
+++ b/packages/admin-graphql-api-utilities/README.md
@@ -56,7 +56,7 @@ parseGidWithParams('gid://shopify/Customer/12345?sessionId=123&foo=bar');
 //   }
 ```
 
-### `function composeGidFactory(namespace: string): Function`
+### `function composeGidFactory<N extends string>(namespace: N): Function`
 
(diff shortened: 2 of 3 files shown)


packages/admin-graphql-api-utilities/src/index.ts
```diff
@@ -43,26 +50,51 @@ export function parseGidWithParams(gid: string): ParsedGid {
...
     const paramString = new URLSearchParams(params).toString();
-    return `${gid}?${paramString}`;
+    return `${gid}?${paramString}` as Gid<N, T>;
   };
 }
 
 export const composeGid = composeGidFactory('shopify');
 
+export function isGidFactory<N extends string>(namespace: N) {
+  return function isGid<T extends string = unknown>(
```
gardenm, line 74: Is the reason for defaulting this to `unknown` so that if you don't provide a key when calling `isGid` you don't have to provide this generic type? I played around to see if it'd be possible to get around that with function overloading (like one version of `isGid` without the key parameter and another with the key parameter but it's non-optional) ... trying to think of how to get rid of the `unknown` since that's tripping this up.

packages/admin-graphql-api-utilities/src/index.ts
```diff
@@ -43,26 +50,51 @@ export function parseGidWithParams(gid: string): ParsedGid {
...
+
+    try {
+      if (key !== undefined && parseGidType(gid) !== key) {
+        return false;
+      }
+    } catch {
+      return false;
+    }
+
+    // prepends forward slash to help identify invalid id
```
gardenm, line 90: I know that we don't need to extract the parameters here, but I wonder if it'd be better to just call `parseGidWithParams` rather than have to reproduce some of its internal logic (and think about keeping them in sync in the future).

packages/admin-graphql-api-utilities/src/index.ts
MrGVSV: Yeah that was the intention, but we could just make this default to `string`. Would that work?

packages/admin-graphql-api-utilities/src/index.ts
MrGVSV: If you check out the PR description, that's originally what I had haha. I chose to inline the part I needed, though, for two reasons:
1. To decouple the logic (if the parser functions want to do other stuff, they can without affecting this function)
2. To be slightly [more performant](https://stackoverflow.com/a/10940138/11621047) (which is probably not even a real concern here, but it's free so why not I guess haha)

Should I replace it with the parser function?

gardenm (approved):

BPScott (approved):
I've took the liberty of adjusting the changeset to be more descriptive. Thanks for the PR!

PR #2625, opened 2023-04-09
Delete packages/react-testing/src/tests/setup directory

MrHBS:
I believe this was added for loom but @BPScott nuked it.

## Description

Fixes (issue #)


BPScott:
This file is still needed. It is referenced by the [jest.config.js](https://github.com/Shopify/quilt/blob/a571510cb619f7c1974a22c4a05f1a910825704c/jest.config.js#L66). Removing just the file, as per this PR causes `yarn test` to refuse to run. Removing it this file file and the reference to it causes a boatload of ` Warning: The current testing environment is not configured to support act(...)` console output when running tests.

PR #2630, opened 2023-04-16
Export webpack-parts/loader.js as @shopify/web-worker/webpack-loader.js

vsumner:
## Description
Export `webpack-parts/loader.js` as `@shopify/web-worker/webpack-loader.js`.

Fixes (issue #)

### DIFF
--- a/packages/web-worker/package.json
+++ b/packages/web-worker/package.json
@@ -29,6 +29,10 @@
       "types": "./build/ts/webpack-parts/index.d.ts",
       "default": "./webpack.js"
     },
+    "./webpack-loader.js": {
+      "types": "./build/ts/webpack-parts/loader.d.ts",
+      "default": "./build/cjs/webpack-parts/loader.js"
+    },
     "./worker": {
       "types": "./build/ts/worker.d.ts",
       "esnext": "./worker.esnext",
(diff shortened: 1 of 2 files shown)


packages/web-worker/package.json
```diff
@@ -29,6 +32,10 @@
       "types": "./build/ts/webpack-parts/index.d.ts",
       "default": "./webpack.js"
     },
+    "./webpack-loader.js": {
+      "types": "./build/ts/webpack-parts/loader.d.ts",
+      "default": "./build/cjs/webpack-parts/loader.js"
```
lucabezerra, line 37: This being the only entry whose `default`  value comes from the `./build` folder, I'm assuming it's correct, but I just wanted to confirm it wasn't an accident :)

lucabezerra (approved):

BPScott (changes requested):
We shouldn't add the .js at the end of the exports names here. It should be extensionless - `webpack-loader`-  as you import `from @shopify/web-worker/webpack-loader` not `from @shopify/web-worker/webpack-loader.js`.


Added a comment inline about how to add the loader file as an entrypoint in the rollup config.


---

For i18n stuff we've been investigating an approach where instead of having a babel/swc plugin doing rewrites we've been looking into authoring a webpack loader that runs before babel-loader when processing ts files, and doing the file transformation in there (and an jest transform and esbuild loader that would do the same thing in those environments). At that point we wouldn't need babel or swc plugins at all. Would it be worth investigating a similar approach for the web-worker stuff?

PR #2637, opened 2023-04-25
:bug: Address React 18 effect bug for useMountedRef

### LINKED ISSUE #2500: [@shopify/react-form] `form.submitErrors` is not populated  when returning `{status: "fail"}`
## Overview

In a very simple example `submitErrors` is not populated.

## Consuming repo

Made a reproducable example. https://stackblitz.com/edit/vitejs-vite-qctwdk?file=src/App.tsx

---

## Checklist

QuintonC:
## Description

Fixes #2500

This addresses a bug in React 18 StrictMode environments where effects are run twice on initialization. Previously, the `useMountedRef` bug was initialized via `const mounted = useRef(true)`, but since the effect runs twice in development (including cleanups), the `ref` would be set to `false` and never re-initialized as `true`.

You can read more about the way effects are run in strict mode in the following places:
- [Updates to Strict Mode](https://react.dev/blog/2022/03/08/react-18-upgrade-guide#updates-to-strict-mode)
- [How to handle the Effect firing twice in development?](https://react.dev/learn/synchronizing-with-effects#how-to-handle-the-effect-firing-twice-in-development)

I'm uncertain where else this bug might be present (via the callbacks being run twice in React 18 w/ Strict Mode), but I do know that this is one instance of it that was causing bugs within `react-form` (see the linked issue that this resolves). I've tested this inside of a 1P app and can provide folks on the team with a demo/spin link for testing if desired.

### DIFF
--- a/packages/react-hooks/src/hooks/mounted-ref.ts
+++ b/packages/react-hooks/src/hooks/mounted-ref.ts
@@ -6,6 +6,8 @@ export function useMountedRef() {
   const mounted = useRef(true);
 
   useIsomorphicLayoutEffect(() => {
+    mounted.current = true;
+
     return () => {
       mounted.current = false;
     };
--- a/.changeset/wet-snakes-ring.md
+++ b/.changeset/wet-snakes-ring.md
@@ -1,0 +1,5 @@
+---
+'@shopify/react-hooks': patch
+---
+
+Addressed a bug with useMountedRef for React 18 Strict Mode in development where mounted.current would be false after the effect runs for the first time.
--- a/packages/react-hooks/src/hooks/tests/mounted-ref.test.tsx
+++ b/packages/react-hooks/src/hooks/tests/mounted-ref.test.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, {StrictMode, useState, useCallback} from 'react';
 import {mount} from '@shopify/react-testing';
 
 import {useMountedRef} from '../mounted-ref';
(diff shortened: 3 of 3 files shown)


packages/react-hooks/src/hooks/mounted-ref.ts
```diff
@@ -3,9 +3,11 @@ import {useRef} from 'react';
 import {useIsomorphicLayoutEffect} from './isomorphic-layout-effect';
 
 export function useMountedRef() {
-  const mounted = useRef(true);
+  const mounted = useRef(false);
```
BPScott, line 6: What's the thinking behind changing this default value.

My initial thoughts would be I think we want this to retain true as its default value. This would be the value that is used between the first render when useIsomorphicLayoutEffect is called.

On a server, useIsomorphicLayoutEffect uses a plain useEffect (as useLayoutEffect errors on the  server), which runs after initial rendering so there would be a moment when this would return false, which would be different to current behaviour.

QuintonC:
Updated a test instance which is using React 18 with Strict Mode enabled and getting the expected results within `react-form`.

BPScott (approved):
Great stuff! Thanks for spotting this and digging into the fix. "Only affects React 18 in StrictMode" is a fun edge-case.

Go hit squash and merge when you're happy.

BPScott:
I took the liberty of adding a test case that only fails in React18.

QuintonC:
> I took the liberty of adding a test case that only fails in React18.

Nice! 

Thank you for that and merging. My evening filled up yesterday and this got put to the side.

PR #2643, opened 2023-04-30
Numeric string fields

jayanth-kumar-morem:
## Description

1. Update the form state comparison by converting numeric string fields to numbers, enabling correct dirty state evaluation
2. Add isNumericString utility function to identify numeric strings and exclude empty strings
3. Add tests to ensure isNumericString() returns true for valid numeric strings
4. Add tests to ensure isNumericString() returns false for non-numeric strings
5. Add tests to ensure isNumericString() returns false for non-string values

Fixes (issue #)2629

### DIFF
--- a/packages/react-form/src/hooks/field/field.ts
+++ b/packages/react-form/src/hooks/field/field.ts
@@ -3,7 +3,7 @@ import {useCallback, useEffect, useMemo} from 'react';
 import isEqual from 'fast-deep-equal';
 
 import type {Validates, Field, DirtyStateComparator} from '../../types';
-import {normalizeValidation, isChangeEvent} from '../../utilities';
+import {normalizeValidation, isChangeEvent, isNumericString} from '../../utilities';
 
 import {
   updateAction,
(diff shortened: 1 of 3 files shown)


packages/react-form/src/hooks/field/field.ts
```diff
@@ -138,7 +138,9 @@ export function useField<Value = string>(
       return undefined;
     },
     // eslint-disable-next-line react-hooks/exhaustive-deps
-    [state.value, ...dependencies],
+    [state.value, ...dependencies].map(value =>
```
BPScott, line 141: Can you please add some unit tests that showcase the behaviour that would have broken without this change?

BPScott (changes requested):
Thank you for looking into this.

Are you able to add some tests to demonstrate the behaviour that was incorrect prior to this change and is now fixed?

Also, before this is merged we need to add a changeset entry using `yarn run changeset`.

packages/react-form/src/hooks/field/field.ts
jayanth-kumar-morem: Yes @BPScott , I have added tests in `utilities.test.ts`. I'll add the changeset entry now, thanks!

BPScott: The tests in `utilities.test.ts` test the new helper that you've added. They don't test this line we're commenting on. It would be totally possible to undo the change on this line and all the tests would continue to pass. Yet it's this line that you're fixing a bug on, the addition of utility functions is kinda incidental.


I would like to see an test of `useField` that uses a numeric string, that fails if this line is as it is on main, and passes when this line is changed, so that I can understand the expected behaviour that you want to see.

PR #2645, opened 2023-05-02
Removing onCompleted and onError from the useQuery types

alexandcote:
## Description
Our custom implementation of `useQuery` and `<Query />` doesn't support `onCompleted` and `onError`.

This PR update the types to reflect the difference between our API and Apollo API.

### DIFF
--- a/.changeset/breezy-dolphins-exercise.md
+++ b/.changeset/breezy-dolphins-exercise.md
@@ -1,0 +1,5 @@
+---
+'@shopify/react-graphql': patch
+---
+
+Remove `onCompleted` and `onError` from the types on `useQuery`'s options as they are not implemented
--- a/packages/react-graphql/src/hooks/types.ts
+++ b/packages/react-graphql/src/hooks/types.ts
@@ -13,7 +13,12 @@ import type {VariableOptions} from '../types';
 
 export type QueryHookOptions<Data = any, Variables = OperationVariables> = Omit<
   QueryComponentOptions<Data, Variables>,
-  'query' | 'partialRefetch' | 'children' | 'variables'
+  | 'query'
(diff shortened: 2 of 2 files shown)


BPScott:
Change looks reasonable though I'm concerned that folks might have been using these props without realising they do nothing, and thus removing them from the type will introduce a bunch of type errors in web. Could you please cut a snapshot and test if using it in web has an effect on running type-check?

alexandcote:
I see some instances: https://buildkite.com/shopify/web-ci-builder/builds/666387#0187e1dd-9a57-4e83-9a31-b20ab0060fd3

.changeset/breezy-dolphins-exercise.md
```diff
@@ -0,0 +1,7 @@
+---
+'@shopify/react-graphql': patch
+---
+
+Fixing an issue with useQuery / Query types
+
+`onCompleted` and `onError` callback are not supported by our custom implementation and should be exclude from the types
```
BPScott, line 7: Use the active present tense:

```suggestion
Remove `onCompleted` and `onError` from the types on `useQuery`'s options as they are not implemented
```

BPScott (commented):
Looks like the impact in web is pretty light and if we release this fixing web's CI would be pretty quick. I'm ok with claiming this is a patch release and we have to fix those places in web.

Code changes look good, I've added a suggestion for how to format the  changeset so something that would feel a bit nicer to read as a changelog entry.

Alex, can you own getting this updated within web, so we we don't have to do the tidying as part of some future update?

BPScott (approved):
LGTM. Can you own getting this updated in web (I've moved onto checkout things).

PR #2658, opened 2023-05-26
Enabling graphql-validate-fixtures compatibility with varying file types

johnboulder:
## Description
I apologize in advance if I'm not following some standard discussion-approach your team has laid out before submitting this PR. I had a hard time understanding if you guys were OK with outside contributions on some packages, so I just went ahead with making this.

### Background
The impression I'm getting is that `graphql-validate-fixtures` may be in somewhat of a maintenance state, presumably because it solves whatever problem it was created for or its usage may have diminished with the increased influence of graphql-tools. In either case, I think this package is kindof awesome and solves a big problem with the current graphql-mocking solutions. Without going into too much detail, using fixtures to mock graphql queries is a much more useful, and low-maintenance solution at scale in my opinion than the prevailing "dynamic mocking" approach taken by `graphql-tools`.

However, the problem I found with `graphql-validate-fixtures` is that it doesn't seem to work with the various filetypes that graphql schemas/queries/types can be declared in these days. Not to mention that it doesn't play well with fragments, or apollo-related stuff.

So this PR addresses those issues!

### What This PR Does
- **Maintains backward compatibility.**
- **Enables validation of queries declared in typescript (or any other valid file format recognized by the graphql community)** - This was just a matter of using `@graphql-tools/load-files` to read files instead of `fs-extra`. This does seem to be environment specific 
- **Enables validation where schemas are declared in any of the standard `graphql-config` file formats**
- **Enables usage of fragments**

### Known Issues:
- **I've only managed to correctly validate `@apollo/client` fixtures by running `validate-cli.ts` with the `ts-node` cli.** Without it, the script fails when it reads typescript files with the error. 
```bash
/Users/jstockwell/IdeaProjects/eg-platform-console/.offline/validate-cli.ts:1
import { evaluateFixtures } from '@openworld-ui/graphql-validate-fixtures';
^^^^^^

SyntaxError: Cannot use import statement outside a module
    at wrapSafe (internal/modules/cjs/loader.js:915:16)
    at Module._compile (internal/modules/cjs/loader.js:963:27)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1027:10)
    at Module.load (internal/modules/cjs/loader.js:863:32)
    at Function.Module._load (internal/modules/cjs/loader.js:708:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:60:12)
    at internal/main/run_main_module.js:17:47
```
I believe there's likely some way to resolve this through configuration or maybe running the script with node in a way that recognizes es6 syntax, but I've hit a wall with my knowledge in this area, so the simplest solution I could find was just running `validate-cli.ts` with `ts-node`
- **Projects that validate fixtures cannot contain a mix of both `@apollo/client` fixtures _and_ standard graphql fixtures.** It has to be one or the other. When using `@apollo/client` fixtures, the validation expects the `__typename` field on every child, so when it doesn't see the. I'm thinking the solution to this would probably be incorporating some way to differentiate between the fixtures.
- **Fragment validation does not account for the `__typename` field (if it exists).** This is likely necessary to ensure fragments cannot become stale. 

Any thoughts on this are welcome!


johnboulder:
Signed the CLA

johnboulder:
Just a note that I've found the following edge cases that are being validated incorrectly in the current version of this changeset.
- Inline-fragment missing a field
- Inline-fragment array missing a field
- Fragments/Inline-fragments throwing the wrong error when non-nullable fields are `undefined`. In this case, the error says the field should be the type or `null` which is wrong.
- Fragments nested in fragments are validated incorrectly 

Adding fixes for all these presently

Issue #2661, opened 2023-06-09
I propose adding a function to @shopify/address for retrieving province name


pop-web:
## Overview
I propose adding a function to @shopify/address for retrieving province name

I will create that addition.

...

## Type

- [x] New feature
- [ ] Changes to existing features

## Motivation

> What inspired this feature request? What problems were you facing,
> or what else makes you think this should be in `quilt`?

There is no way to retrieve the province name in @shopify/address. I have encountered several situations where I needed to retrieve it.

PR #2662, opened 2023-06-09
Add async function to get province name

pop-web:
## Description

Fixes (#2661)

I would like to add a function to @shopify/address for retrieving province names.

### DIFF
--- a/packages/address/src/tests/AddressFormatter.test.ts
+++ b/packages/address/src/tests/AddressFormatter.test.ts
@@ -195,7 +195,7 @@ describe('AddressFormatter', () => {
         'INVALID_CODE',
       );
 
-      expect(provinceName).toBe('');
+      expect(provinceName).toBeUndefined();
     });
   });
 


.changeset/tough-dancers-do.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/address': major
```
BPScott, line 2: This is not a breaking change so it does not need to be a major version number. This should be a `minor` release.

packages/address/src/AddressFormatter.ts
```diff
@@ -44,6 +44,19 @@ export default class AddressFormatter {
     return countries;
   }
 
+  async getProvinceName(
```
BPScott, line 47: Can we please call this `getZoneName`, to match the naming conventions that are used for this field in the backend. "Province" is a locale specific name, while zone is more agnostic (e.g. in the USA they're States not Provinces).

pop-web: @BPScott 
Thank you for your input, BPScott.
Your suggestion makes sense and I have changed the method name to getZoneName as you recommended.
Here is the commit:
https://github.com/Shopify/quilt/pull/2662/commits/fdf720d25ee74d770213f4a4d089d597878c024c

.changeset/tough-dancers-do.md
pop-web: @BPScott 
Thank you. I have made the changes.
https://github.com/Shopify/quilt/pull/2662/commits/5da45024046325ffd8e71940f5e1f7b2aa3ec0cb

packages/address/src/tests/AddressFormatter.test.ts
```diff
@@ -173,6 +173,28 @@ describe('AddressFormatter', () => {
...
 
+  describe('getProvinceName()', () => {
+    it('returns a provinceName', async () => {
+      const addressFormatter = new AddressFormatter('ja');
+      const provinceName = await addressFormatter.getProvinceName(
+        'JP',
+        'JP-01',
+      );
+
+      expect(provinceName).toBe('北海道');
```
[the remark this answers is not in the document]
pop-web: @znja 
Thank you for pointing that out.
Indeed, the test was necessary.
I have added the test.
https://github.com/Shopify/quilt/pull/2662/commits/e4022f41e703034f83342ec570fb124cb60224a4

packages/address/src/AddressFormatter.ts
```diff
@@ -44,6 +44,19 @@ export default class AddressFormatter {
     return countries;
   }
 
+  async getZoneName(
+    countryCode: string,
+    provinceCode: string,
+  ): Promise<string> {
+    const country = await this.getCountry(countryCode);
+    const countryZone = country.zones.find(
+      (item) => item.code === provinceCode,
```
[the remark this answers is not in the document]
pop-web: @ryanwilsonperkin 
Thank you. I have incorporated your suggestions and committed them.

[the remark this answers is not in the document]
pop-web: @ryanwilsonperkin 
Additionally, I have made changes to the zoneCode in other files as well.https://github.com/Shopify/quilt/pull/2662/commits/d5187221e99608536bee8d7cb0cf8b498e9bc650

pop-web:
@znja  @BPScott @ryanwilsonperkin

I have addressed the points you mentioned.
Would it be possible for you to approve this pull request?

packages/address/src/AddressFormatter.ts
```diff
@@ -44,6 +44,19 @@ export default class AddressFormatter {
...
 
+  async getZoneName(
+    countryCode: string,
+    zoneCode: string,
+  ): Promise<string> {
+    const country = await this.getCountry(countryCode);
+    const countryZone = country.zones.find(
+      (item) => item.code === zoneCode,
+    );
+    if (!countryZone?.name) return '';
```
ryanwilsonperkin, line 55: If no zone can be found at the given code, we should return `undefined` instead of an empty string, to make it easier for the consumer to switch on the type of it and ensure they're handling the case where this could be an invalid code.

pop-web: @ryanwilsonperkin 
Thank you for your feedback. I have made the suggested changes as per your instructions.
https://github.com/Shopify/quilt/pull/2662/commits/18e9932e0e3e8926b6aa832c21373cf59faa121c
https://github.com/Shopify/quilt/pull/2662/commits/9dfe2ec9b868676a1172612d39bb50e1ff04272f

PR #2670, opened 2023-07-12
added i18 abbreviate business name

mikegarfinkle:
## Description

Adds abbreviateBusinessName and to the react-i18n package. This allows client side functionality for business name abbreviation, which is already found in the backend ruby i18n library.

### DIFF
--- a/packages/react-i18n/src/i18n.ts
+++ b/packages/react-i18n/src/i18n.ts
@@ -46,6 +46,7 @@ import {
   memoizedPluralRules,
   convertFirstSpaceToNonBreakingSpace,
   tryAbbreviateName,
+  tryAbbreviateBusinessName,
   identifyScripts,
 } from './utilities';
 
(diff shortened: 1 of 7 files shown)


ryanwilsonperkin:
@mikegarfinkle thank you for the proposal, but as something that is particular to business names and isn't a generic i18n utility this should live directly in your application and not within this library.

We're actively working towards removing custom uses like this one, and don't want to add any more that would make the move off of react-i18n more difficult for consumers.

ryanwilsonperkin:
@mikegarfinkle and I chatted async and decided that a better fit for this, and the `abbreviateName` method, would be a separate Quilt package called "names" or similar, that would be responsible for these more business-logic-y things that we want to keep separate from standard I18n translation utilities which should mostly be thin wrappers around browser/node I18n APIs as well as utilities for loading locale JSON files

PR #2671, opened 2023-07-12
Create name package

mikegarfinkle:
## Description

Fixes (issue #)


packages/name/src/formatName.ts
```diff
@@ -0,0 +1,37 @@
+import {EASTERN_NAME_ORDER_FORMATTERS} from './constants';
+import {hasEasternNameOrderFormatter} from './utilities';
+
+export const formatName = Object.assign(
```
trishrempel, line 4: I think it would be better to have `formatName` as a method only and expose `hasEasternNameOrderFormatter` as another top-level method. Then you could move that method into its own file and maybe not have a separate file in `utilities`

packages/name/src/formatName.ts
```diff
@@ -0,0 +1,37 @@
...
+    if (customNameFormatter) {
+      return customNameFormatter(name.givenName, name.familyName, isFullName);
+    }
+    if (isFullName) {
+      return `${name.givenName} ${name.familyName}`;
+    }
+    return name.givenName;
+  },
+  {
+    hasEasternNameOrderFormatter: (locale: string) =>
```
trishrempel, line 34: I think that this name might be better as `hasEasternNameOrder` as an externally-exposed method

```suggestion
    hasEasternNameOrder: (locale: string) =>
```

packages/react-i18n/README.md
trishrempel, line 1: My preference would be to first deprecate these and point to the new package (e.g. ``@deprecated use @shopify/names `formatName()` instead``), and then remove them after we've had a chance to change the usages in Shopify's biggest repos.

I'm also ok with removing them right aways if the folks in Admin Web Foundations approve.

trishrempel (commented):
Looks good! I have some comments and suggestions.

packages/name/src/formatName.ts
trishrempel: I'm basing my thinking on the possibility that other repos might call this method independently outside of `formatName`. In our repos, there's only one reference I could find [here](https://github.com/Shopify/pos-channel/blob/5e04d06f5c3bd2fec254961873f7c38524217d4d/app/ui/sections/Staff/components/AccountInformation/AccountInformation.tsx#L118), where it's being used to decide what order the personal name fields should be displayed in.

I otherwise don't think it's necessary to strongly associate it with `formatName`; you can tell from context (and the README) that it pertains to personal name, not business name, for example.

packages/react-i18n/README.md
trishrempel: Another option would be to retain the name formatting methods in `react-i18n` and make them wrappers to the methods in `@shopify/name`. The benefit of this is that there are fewer changes to make to other repos, and the methods in `react-i18n` can use its internal `locale` value. The `react-i18n` library does use the `@shopify/dates` method this way, although it adds some functionality on top of it.

packages/name/src/formatName.ts
mikegarfinkle: oh I stand corrected! I see now that this is being used to render names in the correct order in cases where we don't use format name. thanks for clarifying! 

Do only "eastern" dialects have last name then first name ordering? do any not have this ordering? do any non-eastern dialects have this ordering? Given these questions... do you think `hasFamilyNameFirstPersonalNameOrdering` or something more along those lines would be better?

packages/react-i18n/README.md
mikegarfinkle: I like that last option... @ryanwilsonperkin would you be in agreement with the wrapping option too?

packages/name/src/utilities/getGraphemes.ts
```diff
@@ -0,0 +1,21 @@
...
+export function getGraphemes({
+  text,
+  locale,
+}: {
+  text?: string;
+  locale: Locale;
+}): string[] {
+  // returns undefined when Intl.Segmenter does not exist in our JS environment (such as in Firefox)
+  if (!text || !Intl.Segmenter) {
+    return undefined;
```
ryanwilsonperkin, line 12: Either the return type should specify `string[] | undefined` or this should be changed to return an empty array.

packages/react-i18n/README.md
ryanwilsonperkin: Sounds good to me, but you may have to do it across two PRs since it'll require `@shopify/react-i18n` to reference a package in its dependencies that doesn't yet exist on NPM

packages/name/src/tests/abbreviateBusinessName.test.ts
```diff
@@ -0,0 +1,18 @@
+import {tryAbbreviateBusinessName} from '../utilities';
+import {abbreviateBusinessName} from '..';
+
+describe('#abbreviateBusinessName()', () => {
+  it('returns input name if no abbreviation found', () => {
+    // no abbreviation as has space in last name
```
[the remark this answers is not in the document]
mikegarfinkle: this comment was copied erroneously, will delete

packages/name/src/hasFamilyNameFirstNameOrdering.ts
```diff
@@ -0,0 +1,7 @@
+import {FAMILY_NAME_FIRST_NAME_ORDERING} from './constants';
+
+export function hasFamilyNameFirstNameOrdering(locale: string) {
```
trishrempel, line 3: Aah, I just realized this should be named `hasFamilyNameGivenNameOrdering`, so that "given name" is aligned with the "given name" field we use in `formatName()`.

```suggestion
export function hasFamilyNameGivenNameOrdering(locale: string) {
```

packages/name/README.md
```diff
@@ -0,0 +1,52 @@
...
+- `abbreviateName({name: {givenName: 'John', familyName: 'Smith'}, locale: 'en'})` will return `JS`
+- `abbreviateName({name: {givenName: '健', familyName: '田中'}, locale: 'en'})` will return `田中`
+
+You may also pass an optional `idealMaxLength` parameter, which gives the maximum allowable abbreviation length when
+trying to abbreviate a name in the Korean language (default 3 characters). In Korean, if the first name is longer than
+this length, the method will instead return the first character of the first name.
+
+### abbreviateBusinessName
+
+Takes a name and returns a language appropriate abbreviated name, or will return the input name if it is unable to find
```
trishrempel, line 42:
```suggestion
Takes a business name and returns a language appropriate abbreviated name, or will return the input name if it is unable to find
```

packages/name/src/formatName.ts
```diff
@@ -0,0 +1,31 @@
...
+  if (!name.givenName) {
+    return name.familyName || '';
+  }
+  if (!name.familyName) {
+    return name.givenName;
+  }
+
+  const isFullName = Boolean(options && options.full);
+
+  const customNameFormatter = FAMILY_NAME_FIRST_NAME_ORDERING.get(locale);
```
trishrempel, line 22: This is a pre-existing problem, but I just realized that this method won't return the expected results if someone happens to use this library with a Japanese or Korean locale that isn't top-level (e.g. `ja-JP` or `ko-KR`). I'll create an issue for it, something to address in a future PR!

trishrempel (approved):
Looks good! Just some minor suggestions.

ryanwilsonperkin (approved):

Issue #2681, opened 2023-08-18
`@shopify/react-i18n` - Update name formatting functions to call equivalent functions in `@shopify/name`


trishrempel:
## Overview

Update `@shopify/react-i18n` so that the following methods call the equivalent functions from `@shopify/name`:
- `formatName()`
- `abbreviateName()`
- `hasEasternNameOrderFormatter()`

The following method was copied to `@shopify/name` but was not exposed, so it will remain duplicated for now until we can make sure they are not being used by any Shopify consuming code:
- `identifyScript`

## Motivation

> What inspired this enhancement? What makes you think this should be in `quilt`?

Reduce code duplication across Shopify packages.

---

## Checklist

- [x] Please delete the labels section before submitting your issue
- [x] I have described this issue in a way that is actionable (if possible)

PR #2682, opened 2023-08-21
wrapped name package methods in react i18n

mikegarfinkle:
## Description

Closes https://github.com/Shopify/quilt/issues/2681

Adds abbreviate business name method to react-i18n, removes all name related method source code from react-i18n and instead wraps imported methods from the `name` repo.


trishrempel (approved):
Looks good to me, thank you! 👍🏻

PR #2695, opened 2023-10-19
Install node-gyp

vsumner:
## Description
Not ideal, but it works for now. Hoping to move off packages that need `node-gyp` and `yarn` to something else in the future.

Fixes (issue #)


egjiri (approved):
Nice!

PR #2696, opened 2023-10-27
[graphql-typescript-definitions] Pass export-format cli option to Builder config

calirvine:
## Description

The `export-format` cli option was added in https://github.com/Shopify/quilt/commit/1c7dff61b7843627f13578012938f060446d44fc (updated in #2289 and #2559) but never actually passed to the `Builder` config. This means any projects that want to be typed for `@apollo/client` or other libraries using `TypedDocumentNode` can not use the cli for generating their types. This PR adds the value to the CLI Builder.

### DIFF
--- a/.changeset/silver-mails-fold.md
+++ b/.changeset/silver-mails-fold.md
@@ -1,0 +1,5 @@
+---
+'graphql-typescript-definitions': patch
+---
+
+Pass export-format CLI option to builder Config allowing cli users to change the type of generated Documents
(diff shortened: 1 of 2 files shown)


BPScott (approved):

BPScott:
Thanks for the spot! Approving, though as I'm no longer on the admin-web-platform team I no longer have permission to merge, you'll need somebody from that team to review and do the merge

vsumner (approved):

PR #2697, opened 2023-10-27
Expose matcher operations

mattkubej:
## Description

`graphql-testing/matchers` and `react-testing/matchers` are side-effect imports, which extend Jest's expect automatically for the consumer. This does provide the consumer control over how/when they would like to extend Jest's expect. This is specifically helpful with lazily loading these matchers.

This PR exposes the operations for each of the packages without extending Jest's expect to allow the consumer to do this themselves. `/matchers` now exports the matchers without extending Jest's expect, so the consumer can do it themselves. If a consumer wants to have all the matchers loaded via a side-effect import, then `/matchers-all` will provide that.


packages/graphql-testing/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@shopify/graphql-testing",
-  "version": "7.0.1",
+  "version": "7.0.2",
```
vsumner, line 3: Changesets will handle this for you

mattkubej: Thanks! I just noticed that in contribution guide and reverted :)

mattkubej:
No longer needed as an alternative has been found to lazy load side-effect imported matchers

PR #2701, opened 2023-12-07
Return early in GraphQL fill for primitive partial

Flufd:
## Description

Optimises the GraphQL filler by returning early in the `createValue` function if we have a concrete partial value. 

---

I was profiling some slower tests in admin and found that a lot of the time was spent calling faker's `seed` method. This is done with a stable seed for the particular branch/leaf node on the GraphQL document so that the random values for each node are stable. The re-seeding of the faker instance is slow, because it needs to reinitialise the [mersenne twister](https://github.com/faker-js/faker/blob/v7.3.0/src/modules/mersenne/twister.ts) that is used for RNG.

For large lists of known primitive values, such as currency or country codes, or other arrays of enume, which in the admin might be 200+ elements long for country codes, this becomes a performance issue. We are reseeding the faker instance for each element, even when not consuming it.

---

Another optimisation and some thoughts on random data in general:
https://github.com/Shopify/quilt/pull/2700


packages/graphql-fixtures/src/tests/fill.test.ts
```diff
@@ -426,6 +448,30 @@ describe('createFillers()', () => {
...
+      `);
+
+      const data = fillOperation(document, {
+        localizations: Array.from({length: 20}).map(() => ({
+          countryCodes: Array.from({length: 250}).map(() => 'CA'),
+          currencyCodes: Array.from({length: 150}).map(() => 'CAD'),
+        })),
+      });
+
+      expect(spy.mock.calls.length).toBeLessThan(100);
```
Flufd, line 473: Without this change, the seed is reset 8000+ times, with the change it is ~60

PR #2714, opened 2024-02-07
fixed japanese and korean name formatting

mikegarfinkle:
## Description

Before, the dictionary that was being used for last name / first name ordering was using `language` (ie. ja) instead of 'locale' (ie. ja-JP) for japanese and korean. This caused improper name ordering as we are passing `locale` to this function.

This PR makes the dictionary as indexed by `language` for dialects. We then parse `locale` into `language` before accessing this dictionary.


packages/name/src/tests/formatName.test.ts
trishrempel, line 1: I recommend adding at least one test that still tests a family name / given name ordering format assuming that a top-level language is being passed in (e.g. `ja` instead of `ja-JP`), just to confirm that it still works that way too.

trishrempel (approved):
Looks good; tophatted on the Orders page in web:
08-34-w25my-gg2p5

sjahu (approved):
latest commit lgtm, good catch on the `hasFamilyNameGivenNameOrdering` thing

trishrempel (approved):

PR #2721, opened 2024-02-13
Add new utility to check for Shopify POS App's user agent

chaitanyambilgikar:
## Description

Helps fix [this issue](https://github.com/Shopify/pos-next-react-native/issues/32638) in the pos-next-react-native repo

## TL;DR

Adds a new utility function that checks for the Shopify POS app's user agent

## Why

[Vault](https://vault.shopify.io/gsd/projects/37771)
[Technical design doc](https://docs.google.com/document/d/1F3N_UR8XTyEfLgusWf1Z5cEXHIa0ZaDEv3FDxX9LUvU/edit?usp=sharing)

The Shopify POS App uses web views for some of it's setup tasks. Some of these tasks open the `/settings` admin page in the web view. This page [currently has an enclosing modal](https://github.com/Shopify/web/blob/6cc03cb895883d78a4d996127db9542e1ee0ee19/app/foundation/Routes/components/Settings/Settings.tsx#L24) that does not work well in a web view.

You can see in from the code linked above that the modal is hidden if the `isNativeApp` utility from this package returns true. This is used when the setting section is shown from the Shopify Mobile app. We want a similar behavior for the Shopify POS App. Hence, this new utility will also be used to conditionally hide the enclosing modal when accessing the `/settings` section from the Shopify POS App

### DIFF
--- a/packages/browser/src/tests/browser.test.ts
+++ b/packages/browser/src/tests/browser.test.ts
@@ -48,8 +48,7 @@ describe('Browser', () => {
 
   describe('isShopifyPOSApp', () => {
     it('returns true for Shopify POS', () => {
-      const userAgent =
-        'com.jadedpixel.pos Shopify POS/5.11.0 (iPad; iOS 9.3; Scale/2.00)';
+      const userAgent = 'Shopify POS/5.11.0 (iPad; iOS 9.3; Scale/2.00)';
 
       expect(new Browser({userAgent}).isShopifyPOSApp).toBe(true);
     });


packages/browser/src/tests/browser.test.ts
```diff
@@ -46,6 +46,26 @@ describe('Browser', () => {
     });
   });
 
+  describe('isShopifyPOSApp', () => {
+    it('returns true for Shopify POS', () => {
+      const userAgent =
+        'com.jadedpixel.pos Shopify POS/5.11.0 (iPad; iOS 9.3; Scale/2.00)';
```
henriquenfaria, line 52: I'm not seeing `com.jadedpixel.pos` suffix on the user-agent coming from the POS app:

Screenshot 2024-02-13 at 3 15 28 PM


It appears that this suffix was included for first-party devices as part of the workaround found [here](https://github.com/Shopify/pos-next-react-native/pull/14837/files). Perhaps we could remove it from the tests, or even better, introduce an additional test to cover this case?

henriquenfaria (approved):

PR #2725, opened 2024-02-23
Migrate type tests to TSTyche

mrazauskas:
Thanks for using `jest-runner-tsd` and `tsd-lite` to test types. I ~am~ was maintaining `tsd-lite` and helping with the runner too.

Just wanted to let you know that `tsd-lite` got [deprecated](https://github.com/mrazauskas/tsd-lite/issues/364). It is replaced with TSTyche, a mighty type testing tool that I was developing for two years.

The problem is that architecture of `tsd` is very limiting. Without maintaining forked `tsd-lite` it was not possible to upgrade TypeScript in Jest repo. `tsd-lite` was temporary solution to get unblocked and to learn more about type testing.

I was contributing type tests to Jest repo for some time. There are around 1200 assertions. At this scale it was a pain debugging types. So I went on researching and building TSTyche. Similar to `tsd`, it compares types programmatically, but the rest is different.

I just try to explain that TSTyche is build to work at any scale and to last for year. That's why I would like to recommend to use it in your project.

Currently TSTyche is used by Immutable.js, Jest, Redwood and few more smaller project (including my own packages).

Repo: https://github.com/tstyche/tstyche
Documentation: https://tstyche.org

---

TSTyche has `test()` and `describe()` helpers which support run mode flags like `.only`, `.skip` or `.todo`. These are priceless if you have to debug something. By the way, additionally there are `expect.fail` `expect.only` or `expect.skip`.

`--only` and `--skip` can be set from the command line: `tstyche useful-types --only DeepPartial`. 

It has `expect` style assertions which look familiar. An expression or a type can be passed on both sides of an assertion. This means that TSTyche is able to compare two types directly. There is no need to create intermediate values.

Testing on several versions of TypeScript is supported: `tstyche --target 4.9,current`. This command would use TypeScript 4.9 and the one from `node_modules` (the `current`).

By default TSTyche uses `typescript` from `node_modules`. There is no need to install patched TypeScript package like `@tsd/typescript`. The install size of TSTyche is under 200kB. It has no dependencies.

TSTyche is fast. Just give it a spin.

And there is much more.

---

Curious what you think.

### DIFF
--- a/.changeset/empty-clocks-rhyme.md
+++ b/.changeset/empty-clocks-rhyme.md
@@ -1,0 +1,9 @@
+---
+'@shopify/graphql-testing': minor
+'@shopify/react-graphql': minor
+'@shopify/react-graphql-universal-provider': minor
+---
+
+- upgrade @apollo/client to the latest (3.8.10)
+- set `connectToDevTools` to false by default
+- allow consumers to pass `defaultOptions` into the options of
--- a/packages/graphql-testing/package.json
+++ b/packages/graphql-testing/package.json
@@ -31,12 +31,14 @@
     "node": "^14.17.0 || >=16.0.0"
   },
   "dependencies": {
-    "@apollo/client": "^3.5.8",
     "graphql": ">=14.5.0 <16.0.0",
     "jest-matcher-utils": "^26.6.2"
   },
   "devDependencies": {
     "graphql-typed": "^2.0.2"
+  },
+  "peerDependencies": {
+    "@apollo/client": ">=3.5.8 || <4.0.0"
   },
   "files": [
     "build/",
--- a/packages/graphql-testing/src/graphql-controller.ts
+++ b/packages/graphql-testing/src/graphql-controller.ts
@@ -1,4 +1,4 @@
-import type {InMemoryCacheConfig} from '@apollo/client';
+import type {DefaultOptions, InMemoryCacheConfig} from '@apollo/client';
 import {ApolloClient, ApolloLink, InMemoryCache} from '@apollo/client';
 
 import {MockLink, InflightLink} from './links';
@@ -9,6 +9,7 @@ import type {GraphQLMock, MockRequest, FindOptions} from './types';
 export interface Options {
   cacheOptions?: InMemoryCacheConfig;
   links?: ApolloLink[];
+  defaultOptions?: DefaultOptions;
 }
 
 interface ResolveAllFindOptions extends FindOptions {
@@ -29,7 +30,7 @@ export class GraphQL {
 
   constructor(
     mock: GraphQLMock = {},
-    {cacheOptions = {}, links = []}: Options = {},
+    {cacheOptions = {}, links = [], defaultOptions = {}}: Options = {},
   ) {
     const cache = new InMemoryCache(cacheOptions);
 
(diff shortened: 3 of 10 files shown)


packages/useful-types/test-d/types.test-d.ts
```diff
@@ -13,377 +8,288 @@ import type {
...
+  interface Base {
+    required: string;
+    optional?: string;
+    nested?: Base;
+  }
+
+  it('marks properties as optional recursively', () => {
+    expect<DeepPartial<Base>>().type.toMatch<{}>();
+    expect<DeepPartial<Base>>().type.toMatch<{optional?: string}>();
+    expect<DeepPartial<Base>>().type.toMatch<{required?: string}>();
```
mrazauskas, line 40: `.toMatch()` is used to check partial matching, otherwise `.toEqual()` is used. `.toEqual()` would not work here because the type is recursive.

packages/useful-types/test-d/types.test-d.ts
```diff
@@ -13,377 +8,288 @@ import type {
...
+      readonly preset?: {
+        readonly list?: ReadonlyArray<{
+          readonly options?: ReadonlyArray<{readonly name?: string}>;
+        }>;
+      };
+    }>();
+
+    expect<DeepPartial<State>>().type.toMatch<{
+      readonly preset?: {} | undefined;
+    }>();
```
mrazauskas, line 110: `.toMatch()` checks `readonly` modifiers strictly.

packages/useful-types/test-d/types.test-d.ts
```diff
@@ -13,377 +8,288 @@ import type {
...
+    expect<IfEmptyObject<boolean, true>>().type.not.toEqual<true>();
+  });
 });
 
-/**
- * should not be able to add undefined, need to fix this
- */
-expectNotAssignable<DeepPartial<RespectArrayElements>>({
-  // this should error, undefined is not a valid type for list
-  list: [undefined, 'stable', 2],
```
mrazauskas, line 143: Hard to say what was the problem and why "need to fix this" note is left here. As far as I can get, it all works as expected.

packages/admin-graphql-api-utilities/test-d/types.test-d.ts
```diff
@@ -1,29 +1,38 @@
...
+import {describe, it, expect} from 'tstyche';
 
 import type {Gid, ShopifyGid} from '../src';
 import {composeGid, composeGidFactory} from '../src';
 
-/**
- * Shopify GIDs
- */
+describe('composeGid', () => {
+  it('composes Gid using key and number id', () => {
```
mrazauskas, line 7: The diff is noisy, but it was so tempting to use `describe()` and `it()`.

mrazauskas:
Hm.. Not sure what to do with Node.js 14. TSTyche is using `AbortSignal.timeout()` which is available from Node.js 16.14. If Node.js 14 support is a hard requirement, I can try reimplement `AbortSignal.timeout()`. The rest should work.

mrazauskas:
There are only unrelated failures in e2e. The rest works as expected.

This PR is ready for review. Or at least some feed back (;

PR #2730, opened 2024-02-29
StatsD: consume STATSD_DEFAULT_TAGS env var

pedro-stanaka:
## Description

At Shopify we use the STATSD_DEFAULT_TAGS and consume this env var in most of our libraries, so far the quilt library has not consumed this environment variable. In this PR we introduce a change in the statsd client which will try to read globalTags from the environment variable, but will not override the existing values in case the user passed already tags in the constructor.

### DIFF
--- a/packages/statsd/src/client.ts
+++ b/packages/statsd/src/client.ts
@@ -60,6 +60,34 @@ export class StatsDClient {
       this.logger = logger;
     }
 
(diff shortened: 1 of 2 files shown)


ryanwilsonperkin:
The statsd client has a `globalTags` option for providing global tags when instantiating the client which is the best way to pass these. A `STATSD_DEFAULT_TAGS` env var isn't something that we would want the library itself to attempt to read from

PR #2732, opened 2024-03-04
Import faker from the locale-specific file

BPScott:
## Description

#2655 updated faker to v8. It also changed our imports of faker from `@faker-js/faker/locale/en` to `@faker-js/faker`.

However the reason why we imported from `@faker-js/faker/locale/en` instead of the base - because the root loads all locales which is about 6 times the size of just the english locale - [has not been addressed](https://fakerjs.dev/guide/localization.html#individual-localized-packages). 

Thus we should still continue to import faker from the locale-specific import as its root import still results in you loading all locales instead of english only.

This PR reverts the changes to our faker import paths - so we go back to importing from `@faker-js/faker/locale/en`

### Changeset notes

This PR touches files in many packages however all of them except for the `graphql-fixtures` are code that is only used within tests. Thus the only package with user-facing changes is `graphql-fixtures`, so that's the only package in the changeset.

### DIFF
--- a/.changeset/dry-buckets-juggle.md
+++ b/.changeset/dry-buckets-juggle.md
@@ -1,0 +1,5 @@
+---
+'graphql-fixtures': patch
+---
+
+Import from faker from `faker-js/faker/locale/en` to avoid the overhead of loading non-english locales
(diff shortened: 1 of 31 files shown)


mateus (approved):

BPScott:
Rebased to add the usages I missed in the initial commit, which required a little extra refactoring, CI is green now:

- packages/react-i18n-universal-provider/src/tests/I18nUniversalProvider.test.tsx
- packages/react-i18n-universal-provider/src/tests/utilities.test.ts

PR #2734, opened 2024-03-04
:label: [react-hooks] Introduce initializer function to useToggle

emileber:
## Description

`useToggle` is a convenience wrapper around `useState`, but it doesn't provide its [initializer function](https://react.dev/reference/react/useState#avoiding-recreating-the-initial-state) in the types. 

```ts
const {value, toggle} = useToggle(() => expensiveComputation());
```


james-a-c (approved):
lgtm - thanks for adding this!

PR #2737, opened 2024-03-08
Fix GraphQL type generation in ESM projects

lemonmade:
I was trying to convert our app to use `type: "module"` in its root package.json, and I found that there were a few dependencies in `graphql-typescript-definitions` that were not set up to work in ESM. `fs-extra` doesn’t actually provide named exports, and we were using `require()` to bring in `@babel/generator`, even in the ESM version of the build. This PR fixes both of those issues.

### DIFF
--- a/.changeset/long-donuts-enjoy.md
+++ b/.changeset/long-donuts-enjoy.md
@@ -1,0 +1,5 @@
+---
+'graphql-typescript-definitions': patch
+---
+
+Fix GraphQL type generation in ESM projects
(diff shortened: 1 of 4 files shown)


BPScott (approved):
I know you've got a version of this produced using patch-package that works. For extra safety can try the results of a `/snapit` build in your checkout-web PR prior to shipping this.

ryanwilsonperkin (approved):

lemonmade:
Thanks @BPScott, confirmed it does fix running the package in an ESM project.

PR #2739, opened 2024-03-15
Detect POS Go as a mobile device

chaitanyambilgikar:
## Description

Fixes [this issue in `pos-next-react-native`](https://github.com/Shopify/pos-next-react-native/issues/33275)

POS Go devices are not detected as mobile devices since the `getDevice().type` returns `undefined`. We're adding a method to check the device numbers which are part of the user agents. See parent issue for details.

### DIFF
--- a/.changeset/three-beers-sniff.md
+++ b/.changeset/three-beers-sniff.md
@@ -1,0 +1,5 @@
+---
+'@shopify/browser': patch
+---
+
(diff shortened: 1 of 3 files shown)


henriquenfaria (approved):
LGTM!

emiryy (approved):

PR #2755, opened 2024-04-11
✨ Change resolveAll method to wait for newly created gql calls

jas7457:
## Description

This changes the `resolveAll` method on the graphql object to wait for graphql to resolve, and if any new graphql calls get added while that settles, it will continuously wait until the stack is empty. I also added a `resolveNext()` method to do what `resolveAll()` used to do - just wait for a single gql call to finish.

This should negate the need to call graphQL.resolveAll() multiple times, here are [some examples](https://grokt.shopify.io/results?q=await%5CsgraphQL%5C.resolveAll%5C%28%5C%29%3B%5Cn%5Cs*await%5CsgraphQL%5C.resolveAll%5C%28%5C%29%3B).

### DIFF
--- a/.changeset/fair-pumas-try.md
+++ b/.changeset/fair-pumas-try.md
@@ -1,0 +1,5 @@
+---
+'@shopify/graphql-testing': major
+---
+
+Change graphQL.resolveAll() to additionally wait for any newly-pending gql calls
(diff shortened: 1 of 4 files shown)


jas7457:
I tested the snapshot against web and updated all of the tests [here](https://github.com/Shopify/web/pull/123866).

melnikov-s (approved):
🎉

PR #2758, opened 2024-04-16
Support webpack-virtual-modules 0.6.x

swissspidy:
## Description

Widen webpack-virtual-modules dependency range to allow for v0.6.x


swissspidy:
I have signed the CLA!

BPScott:
I ended up doing the same thing in #2767

PR #2764, opened 2024-04-23
feat: add money currency formatting package

IliasHad:
## Description

A new package to format Shopify price value to human readable value using [Shopify Currency formatting options](https://help.shopify.com/en/manual/markets/pricing/currency-formatting)


ryanwilsonperkin:
Hi @IliasHad, thanks for your contribution. At this time, this doesn't seem like the appropriate place to maintain this additional package. We'd suggest instead creating and using a package of your own on the NPM registry

PR #2770, opened 2024-05-03
Add `signupOnly` as part of the cache key

alex-amaral:
## Description

`signupOnly` parameter introduced in this [PR](https://github.com/Shopify/quilt/pull/2766) wasn't as part of the cache key. This means we could have a cross contamination between query results.

With this change we're including the flag as part of the cache key which would fix this issue.


ryanwilsonperkin (changes requested):
This library is intended to be a generic utility for getting access to loading a list of countries, what is `signupOnly` in this context and why would it be applicable to loading a list of countries?

ryanwilsonperkin:
> Hey @ryanwilsonperkin, this parameter is an extension of a variable that [Atlas API](https://github.com/Shopify/atlas/blob/1a2c4b41fd1fd42097e5c1ac8bef5c3683055099/app/graphql/schema.graphql#L1456) support and also [country_db](https://github.com/search?q=repo%3AShopify%2Fcountry_db+hide_from_signup%3A+true&type=code). I understand the point of this library intention to be an utility for the list of countries, but since this is not a required parameter and also provides the same utility that can be used across many clients, I wonder why it doesn't fit here.

It's encoding business-specific logic about what countries are supported for new shop signups within a reusable library that's primarily focussed on formatting of addresses according to locale-specific rules. I suppose in a broader sense I take issue with Atlas itself having a parameter for this and would prefer to see the core Shopify API expose this as a list of country codes which could then be used to index the list of countries retrieved from Atlas.

For example:

```
const countryData = await fetchFromAtlas();
const eligibleCountryCodes = await fetchFromCore();

return countryData.filter((country) => eligibleCountryCodes.includes(country.code));
```

This library currently uses Atlas as its source for loading country data, but it'd just as feasibly reference a JSON list, and would be odd for that list to include information about shop eligibility since it is so intimately tied to our business logic

PR #2783, opened 2024-06-05
Add support for omitting GraphQL source text

lemonmade:
## Description

This PR adds a new format for GraphQL documents that `graphql-mini-transforms` can output: `simple-persisted`. This is like the "simple" outputs I added in https://github.com/Shopify/quilt/pull/2597, but it removes the `source` field, so that the original document source is not present in the resulting build outputs. This will be used by Checkout, which already used the "simple" format, and which uses persisted queries in production as a performance optimization.

Previously, I added the "simple" format as a boolean `simple` option. Since this is a third format the library can output, I deprecated that `simple` option and introduced a new `format` option. The new option is implemented in a backwards-compatible way, so I am just marking this as a minor API change.

You might notice that I didn't add an associated Jest plugin for this new format. This was intentional: I suspect all clients will still want to include `source`s in development and test. In these environments, it's typically hard to coordinate server and client properly, and you often want the debuggability and mock-ability that having the original GraphQL document allows.

### DIFF
--- a/packages/graphql-mini-transforms/src/document.ts
+++ b/packages/graphql-mini-transforms/src/document.ts
@@ -12,6 +12,30 @@ import type {
 import {print, parse} from 'graphql';
 import type {DocumentNode, SimpleDocument} from 'graphql-typed';
 
+/**
+ * Controls the runtime code that will be generated for a GraphQL document.
+ */
+export type OutputFormat =
+  /**
+   * Outputs the JSON representation of a `DocumentNode`. Use this if you’re using
+   * a powerful GraphQL client like Apollo.
+   */
+  | 'document'
+  /**
+   * Outputs a simple representation of the document that includes just a hash of the
+   * document’s first operation (`id`), the operation’s name (`name`), and the operation’s
+   * source text as a string (`source`). Use this if you’re using a smaller GraphQL approach that
+   * relies on natively fetching the GraphQL query over http, like the one found in Checkout.
+   */
+  | 'simple'
+  /**
+   * Outputs the same simpler representation of `'simple'`, but with the `source` field removed.
+   * Use this if you’re using persisted queries, where the client only communicates the `id` of
+   * the document to the server. Persisted queries are only needed if the the `source` adds significant
+   * bundle size.
+   */
+  | 'simple-persisted';
+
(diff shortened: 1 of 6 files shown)


packages/graphql-mini-transforms/src/document.ts
```diff
@@ -81,6 +104,22 @@ export function toSimpleDocument<
...
+  switch (format) {
+    case 'document': {
+      return document;
+    }
+    case 'simple': {
+      return toSimpleDocument(document);
+    }
+    case 'simple-persisted': {
+      const simple = toSimpleDocument(document);
+      delete (simple as any).source;
```
BPScott, line 117: Creating the source key within `toSimpleDocument` then deleting it here feels a little suprising.

What if `toSimpleDocument`'s signature was changed to `toSimpleDocument(document, omitSource = false), and then it's up to toSimpleDocument to add/omit the source.


It looks like SimpleDocument has `source` as a required parameter. Will things get weird if the type says that a `source` key will be present, but at runtime the `source` key is absent? Would setting the source to be present but an empty string sidestep violating the type?

BPScott (commented):
Mostly looks good.

By omiting the `source` key we're violating the type of SimpleDocument. If we're doing that should we be updating SimpleDocument / creating a new type?

Perhaps we could sidestep that be setting the source as an empty string rather than removing that property entirely?

lemonmade:
@BPScott thanks for the review. I liked your suggestions, so I applied them mostly as you wrote — would love a second look before we lose you for awhile!

packages/graphql-mini-transforms/src/tests/webpack.test.ts
```diff
@@ -250,12 +250,31 @@ describe('graphql-mini-transforms/webpack', () => {
...
+      );
+      const persistedResult = await extractDocumentExport(
+        `query Shop { shop { id } }`,
+        createLoaderContext({query: {format: 'simple-persisted'}}),
+      );
+
+      const expected = {...simpleResult};
+      delete expected.source;
+
+      expect(persistedResult).not.toHaveProperty('source');
```
BPScott, line 274: This should fix the failing test:

```suggestion
      const expected = {...simpleResult, source: ''};
```

BPScott (approved):
Looking great to me!

Added a suggestion to fix the test failure.

vsumner (approved):

PR #2813, opened 2024-07-18
[react-hooks] Do not throw when using useMedia in SSR without providing an initialValue

melnikov-s:
## Description

Fixes (issue #)

We'll no longer throw if `initialValue` is not provided in SSR, though this will result in hydration mismatch and providing an `initialValue` should be considered required for SSR, but since there's no way to enforce this at type/build time we've opted to go with this approach instead.  

Related discussion can be found [here](https://github.com/Shopify/quilt/pull/2809#issuecomment-2237230632)


jas7457 (approved):
Thanks!

vividviolet (approved):

znja (approved):

Issue #2807, opened 2024-07-09
`@shopify/web-worker` does not support `chunkFilename` being a function


swissspidy:
## Overview

In webpack, `output.chunkFilename` can be [either a string or a function returning a string](https://webpack.js.org/configuration/output/#outputchunkfilename).

However, the webpack loader in `@shopify/web-worker` assumes it is only a string:

https://github.com/Shopify/quilt/blob/34ac07ad97377b4382fd75e063ee8ca93c8538b1/packages/web-worker/src/webpack-parts/loader.ts#L94-L96

https://github.com/Shopify/quilt/blob/34ac07ad97377b4382fd75e063ee8ca93c8538b1/packages/web-worker/src/webpack-parts/loader.ts#L151-L155

This leads to `TypeError: file.includes is not a function` errors when using something like

```js
output: {
	chunkFilename: (pathData) => {
			console.log(pathData.chunk.name);
			return '[name].js';
		},
}
```

To fix this, first check whether it's a function, and then maybe add a wrapper function that adds the suffix.

## Consuming repo

> What repo were you working in when this issue occurred?

https://github.com/swissspidy/media-experiments

### Area

- [ ] Add any relevant `Area: <area>` labels to this issue

### Scope

- Is this issue related to a specific package?

  - [ ] Tag it with the `Package: <package_name>` label.

---

## Checklist

- [ ] Please delete the labels section before submitting your issue
- [x] I have described this issue in a way that is actionable (if possible)

PR #2816, opened 2024-08-04
fix:loader-module-webpack

admirsaheta:
## Description

Fixes (issue #2807)

This PR addresses an issue with the webpack loader in ```@shopify/web-worker``` where ```output.chunkFilename``` could be either a string or a function returning a string. 
The loader previously assumed it was always a string, leading to ```TypeError: file.includes is not a function errors.```
The solution now checks if output.chunkFilename is a function and handles it correctly.

Changes Made
- Default Values: Default values [worker].js are provided for compiler.options.output.filename and compiler.options.output.chunkFilename in case they are undefined.
- Type Handling in addWorkerSubExtension:
- If file is a string, the function directly performs the replacement to add the .worker suffix.
- If file is a function, the function wraps it in another function that performs the replacement on its return value.

### DIFF
--- a/.changeset/spicy-panthers-prove.md
+++ b/.changeset/spicy-panthers-prove.md
@@ -1,0 +1,6 @@
+---
+'@shopify/web-worker': minor
+---
+
+- Allow `output.chunkFilename` to be a function or string.
+- `compiler.options.output.filename` and `compiler.options.output.chunkFilename` now default to `[name].js` when undefined.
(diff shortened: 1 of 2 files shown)


craigbrunner (approved):

admirsaheta:
Can we merge?  @craigbrunner

admirsaheta:
Done, also added a check to avoid this from happening @jesstelford 
image

PR #2819, opened 2024-08-09
Move @shopify/network to dependencies in react-performance

brendo:
## Description

The `@shopify/network` package is currently listed as a `devDependency` for the `@shopify/react-performance` package. I think this is incorrect as the package is referenced from from [performanceReport](https://github.com/Shopify/quilt/blob/75ea9fe52b65f001b238d3d21e76c3181badfe6c/packages/react-performance/src/performance-report.ts#L4) file.

I have moved the package from `devDependency` to `dependency` to resolve.

**That said**, if we look at the _usage_ of the package in the file, it's simply to reference two constants, so I wonder if we're more comfortable/pragmatic in simply inlining these definitions?

- `Header.ContentType`

https://github.com/Shopify/quilt/blob/75ea9fe52b65f001b238d3d21e76c3181badfe6c/packages/react-performance/src/performance-report.ts#L46

- `Method.Post`

https://github.com/Shopify/quilt/blob/75ea9fe52b65f001b238d3d21e76c3181badfe6c/packages/react-performance/src/performance-report.ts#L44

### DIFF
--- a/.changeset/empty-islands-grin.md
+++ b/.changeset/empty-islands-grin.md
@@ -1,0 +1,5 @@
+---
+'@shopify/react-performance': major
+---
+
+Move @shopify/network to a dependency, not devDependency
--- a/.changeset/empty-islands-grin.md
+++ b/.changeset/empty-islands-grin.md
@@ -1,5 +1,5 @@
 ---
-'@shopify/react-performance': major
+'@shopify/react-performance': patch
 ---
 
 Move @shopify/network to a dependency, not devDependency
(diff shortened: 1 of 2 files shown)


.changeset/empty-islands-grin.md
```diff
@@ -0,0 +1,5 @@
+---
+'@shopify/react-performance': major
```
brendo, line 2: I don't know if I did this right, would we consider this a major version?

brendo: Updated in 5b2c18e95 to `patch`.

BPScott (approved):
Change looks good for me, though I think you'll need to chase #help-admin-web-platform for a codeowner review to merge this

mattkubej (approved):

jesstelford:
+1 to inlining - the fewer dependencies the better 👍

PR #2820, opened 2024-08-21
Remove React 17 compatability from react-testing

alex-page:
## Description

When using `vitest` many errors are outputted to the console when using React 18. Since https://github.com/Shopify/quilt/pull/2787 shipped and we no longer support React 17 this PR cleans up some of the compat functions that are no longer necessary.


pandasurplus (approved):
Sweeeet!

vsumner (approved):

Issue #2822, opened 2024-08-21
Formatting percentage ranges


karrocon:
## Overview
We're using `@shopify/react-i18n` to format various text elements in our application. We have encountered a specific need to format percentage ranges in a concise and readable manner.

## Type

- [x] New feature
- [ ] Changes to existing features

## Motivation

> What inspired this feature request? What problems were you facing,
> or what else makes you think this should be in `quilt`?

We have a requirement to format ranges of tax rates as `5.25-10%`. Currently, we do not have a utility method in `@shopify/react-i18n` to handle this specific formatting. Adding this feature would streamline our code and ensure consistency across our application.

### Area

- [ ] Add any relevant `Area: <area>` labels to this issue

### Scope

- Is this issue related to a specific package?

  - [x] Tag it with the `Package: <package_name>` label.

- Is it related to a new package?
  - [ ] Tag it with the `New package` Label

---

## Checklist

- [x] Please delete the labels section before submitting your issue
- [x] I have described this issue in a way that is actionable (if possible)

karrocon:
Closing as we've agreed on resolving this locally for our use-case. Feel free to re-open this once there's a real use-case out there

karrocon:
Closing as we've agreed on resolving this locally for our use-case. Feel free to re-open this once there's a real use-case out there

PR #2823, opened 2024-08-21
feat: new utility method to format percentage range

karrocon:
## Description

Resolves https://github.com/Shopify/quilt/issues/2822

Add a new utility method to `@shopify/react-i18n` to format percentage ranges.


packages/react-i18n/src/tests/i18n.test.ts
```diff
@@ -1397,6 +1397,35 @@ describe('I18n', () => {
...
+
+    it('formats the range as an interval if minAmount is lower than maxAmount', () => {
+      const i18n = new I18n(defaultTranslations, defaultDetails);
+      const expected = `${Intl.NumberFormat(defaultDetails.locale).format(5)}–${Intl.NumberFormat(
+        defaultDetails.locale,
+        {
+          style: 'percent',
+        },
+      ).format(10)}`;
+      expect(i18n.formatPercentageRange(5, 10)).toBe(expected);
```
GoodForOneFare, line 1425: Is `5-1,000%` really the desired output here?

(I know this assertion follows the style of other tests, but I'd much prefer it if `expected` was a hardcoded value 🤷 )

GoodForOneFare (commented):
Code looks good, but the final result looks off to me.

packages/react-i18n/src/tests/i18n.test.ts
karrocon: Thanks, good catch. This is definitely not what we want. I've created a preliminary pull request to extend the `formatPercentage` method prior to continue with this work. I intend to use that in the `formatPercentageRange` method

https://github.com/Shopify/quilt/pull/2824

PR #2827, opened 2024-08-27
Deprecate graphql-validate-fixtures

jesstelford:
`graphql-validate-fixtures ` is no longer maintained or used within Shopify. This PR deprecates the package and removes its source.


GoodForOneFare (approved):
👍 Consider using services-db to archive `abando_ship` and `rnd-education-graphql-app`.  They both look unmaintained, and still consume this dependency.

PR #2830, opened 2024-08-27
Use absolute import path when importing apollo-codegen legacyIR

BPScott:
## Description

Fixes import error when using the mjs version of the graphql-tool-utilities package with node

## Grumble

The graphql packages should probably have always been commonjs only, but it's about 4 years too late for that.

### DIFF
--- a/.changeset/strange-dingos-confess.md
+++ b/.changeset/strange-dingos-confess.md
@@ -1,0 +1,5 @@
+---
+'graphql-tool-utilities': patch
+---
+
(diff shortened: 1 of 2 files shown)


BPScott:
Ah fun snapit seems bust. https://github.com/Shopify/quilt/pull/2821 seems like it might fix, though it's weird that type-check passes locally

alex-page (approved):

GoodForOneFare (approved):

PR #2832, opened 2024-08-30
fix: unknown type in ExecutionResult

nattydodd:
## Description

The Collective (`merchant-to-merchant`) team was trying to [upgrade our graphql dependency](https://github.com/Shopify/merchant-to-merchant/pull/15185) to v16+ and were not able to due to failing typechecks.
The change in [this PR](https://github.com/Shopify/quilt/pull/2778/files#diff-1fff474cb21f6f65d6c5d2be3e8164345ae406da083352db9da72ce2ba7f8389L9) from the `@shopify/graphql-testing` library was causing the typechecks to fail. The `ExecutionResult` type from graphql has a new definition is v16, which causes our `fillGraphQL` type to complain with this error:

| Failure point | Error message |
| ----- | ----- |
|  |  |

Here is the difference between the types in the two versions which caused the types to complain.

| graphql v15.8.0 | graphql v16.9.0 |
| ----- | ----- |
|  |  |

To fix the issue, we had to pass in `<{[key: string]: any}>` as the generic type for `ExecutionResult`, inside the `graphql-testing` package

### Tophatting
I tested this change in the merchant-to-merchant repo and we were able to successfully run our type-checks again. Thanks @BPScott for helping me out on this one 🙏

### DIFF
--- a/.changeset/rich-papayas-clean.md
+++ b/.changeset/rich-papayas-clean.md
@@ -1,0 +1,8 @@
+---
+'@shopify/graphql-testing': patch
+---
+
+MockGraphQLResponse should allow an ExecutionResult containing any data.
+GraphQL v15 defaulted ExecutionResult's data to be an object containing any keys
+but v16 narrowed that down to use `unknown` instead of `any`. This change
+explicitly uses the default behaviour of allowing any from v15.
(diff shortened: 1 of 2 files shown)


BPScott (commented):
Paired with Natalie on this fix.

Follow up to https://github.com/Shopify/quilt/pull/2778. When shipping that PR I was testing in checkout-web which uses a forked version of `graphql-testing` so I never spotted the change in default behaviour for ExecutionResult where in graphql v15 shape of the `TData` generic allowed an object with the values `any` but in v16 it that was restricted to `unknown`, and thus that we need to explicitly state `any`.

PR #2835, opened 2024-09-12
Typescript type fixes

jesstelford:
- Correctly type the mock function's return value
- Remove "@babel/traverse" dependency

### DIFF
--- a/packages/web-worker/src/babel-plugin.ts
+++ b/packages/web-worker/src/babel-plugin.ts
@@ -1,14 +1,17 @@
 import {resolve} from 'path';
 import {runInNewContext} from 'vm';
 
-import type {NodePath, Binding} from '@babel/traverse';
 import type {
   PluginObj,
+  BabelFile,
+  NodePath,
   types as babelTypes,
   template as babelTemplate,
 } from '@babel/core';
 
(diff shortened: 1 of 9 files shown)


packages/web-worker/src/babel-plugin.ts
```diff
@@ -1,15 +1,18 @@
...
   PluginObj,
+  BabelFile,
+  NodePath,
   types as babelTypes,
   template as babelTemplate,
 } from '@babel/core';
 
 import type {Options as LoaderOptions} from './webpack-parts/loader';
 
+type Binding = BabelFile['scope']['bindings'][string];
```
jesstelford, line 14: This is necessary to avoid an error under certain circumstances when the version of `@babel/traverse` is incompatible with the peer dependency `@babel/core`:

```
packages/web-worker/src/babel-plugin.ts(118,26): error TS2345: Argument of type 'Binding' is not assignable to parameter of type 'Binding'.
```

Investigating reveals `node_modules/@types/babel__traverse` to be version 7.11.1, but `node_modules/@types/babel__core/node_modules/@types/babel__traverse` is 7.20.1, which is where the conflicting versions of `Binding` come from.

PR #2842, opened 2025-01-08
Update and Map Deprecated Timezone

ginabak:
## Description

(similar to `shopify-internal`) 

📣 ❤️ co-authors with @trishrempel @christina-tran @jesstelford  ❤️ 

Preps https://github.com/Shopify/web/issues/146409
Preps https://github.com/Shopify/web/issues/150625
 
There are **[some bugsnag errors](https://app.bugsnag.com/shopify/shopify-web-client/timeline?filters%5Bevent.message%5D=Invalid%20time%20zone%20specified%3A%20Europe%2FKyiv&pivot_tab=browser.version)** for the error, `errorMessage:"Invalid time zone specified: Europe/Kyiv"`, although browser versions are part of this 🙄 , we decided that the best fix would be to allow for deprecated timezones as some of the browser versions are recent like `131` for chrome, seen **[here on BugSnag](https://app.bugsnag.com/shopify/shopify-web-client/timeline?filters[event.message]=Invalid%20time%20zone%20specified%3A%20Europe%2FKyiv&filters[browser.version]=131.0.6778&pivot_tab=event)**

This PR: 
- allows for `Europe/Kyiv` for older browsers 😬  (basically it's now backwards compatible), a thought though ... I'm curious if the deprecated `Europe/Kiev` is no longer allowed (which probably won't happen for a while  ... ) we'll have to delete this line 😄 
- maps the `Europe/Kyiv` timezone for react-i18n

### DIFF
--- a/packages/dates/src/utilities/formatDate.ts
+++ b/packages/dates/src/utilities/formatDate.ts
@@ -1,3 +1,5 @@
+import {mapDeprecatedTimezones} from '../map-deprecated-timezones';
+
 const intl = new Map<string, Intl.DateTimeFormat>();
 export function memoizedGetDateTimeFormat(
   locales?: string | string[],
@@ -58,7 +60,12 @@ export function formatDate(
     }).format(adjustedDate);
   }
 
-  return memoizedGetDateTimeFormat(locales, options).format(date);
+  return memoizedGetDateTimeFormat(locales, {
+    ...options,
+    timeZone: options.timeZone
+      ? mapDeprecatedTimezones(options.timeZone)
+      : undefined,
+  }).format(date);
 }
 
 export function dateTimeFormatCacheKey(
--- a/packages/dates/src/utilities/tests/formatDate.test.ts
+++ b/packages/dates/src/utilities/tests/formatDate.test.ts
@@ -1,0 +1,22 @@
+import {formatDate} from '../formatDate';
+
+jest.mock('../../map-deprecated-timezones', () => ({
+  mapDeprecatedTimezones: jest.fn(),
+}));
+
+const mapDeprecatedTimezones: jest.Mock = jest.requireMock(
+  '../../map-deprecated-timezones',
+).mapDeprecatedTimezones;
+
+describe('formatDate', () => {
+  it('maps the deprecated timezone', () => {
+    const date = new Date();
+    const locale = 'en-UA';
+    const options: Intl.DateTimeFormatOptions = {
+      timeZone: 'Europe/Kyiv',
+    };
+
+    formatDate(date, locale, options);
+    expect(mapDeprecatedTimezones).toHaveBeenCalledWith('Europe/Kyiv');
+  });
+});
--- a/packages/react-i18n/src/i18n.ts
+++ b/packages/react-i18n/src/i18n.ts
@@ -11,6 +11,7 @@ import {
   TimeUnit,
   isLessThanOneWeekAway,
   isLessThanOneYearAway,
+  mapDeprecatedTimezones,
 } from '@shopify/dates';
 import {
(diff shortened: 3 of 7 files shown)


packages/react-i18n/src/i18n.ts
```diff
@@ -825,4 +833,8 @@ export class I18n {
   private defaultOnError(error: I18nError) {
     throw error;
   }
+
+  private normalizeTimezone(timeZone?: string) {
+    return timeZone ? mapDeprecatedTimezones(timeZone) : undefined;
```
ginabak, line 838: We initially had a test added to this - but the tests were actually a false positive, testing date for date, the test for the mapping is actually seen in 

https://github.com/Shopify/quilt/blob/main/packages/dates/src/tests/map-deprecated-timezones.test.ts

jesstelford (approved):

packages/dates/src/utilities/formatDate.ts
```diff
@@ -58,7 +60,12 @@ export function formatDate(
     }).format(adjustedDate);
   }
 
-  return memoizedGetDateTimeFormat(locales, options).format(date);
+  return memoizedGetDateTimeFormat(locales, {
+    ...options,
+    timeZone: options.timeZone
+      ? mapDeprecatedTimezones(options.timeZone)
```
ginabak, line 66: HIII @jesstelford ❤️ SOOOOOOOOOO when we implemented the change inside `i18n` and tested out the snapshots ... we weren't able to see the changes actually fix the issue 😬 .... so we're trying it out directly in formatDate as parts of web directly call `formatDate` from `@shopify/dates` we shipped https://github.com/Shopify/web/pull/152962/files that actually fixes it in web, but if we wanted to do our "due diligence" for other repo's, thought we might as well try this LOL 😄

packages/dates/src/utilities/tests/formatDate.test.ts
```diff
@@ -0,0 +1,22 @@
...
+describe('formatDate', () => {
+  it('maps the deprecated timezone', () => {
+    const date = new Date();
+    const locale = 'en-UA';
+    const options: Intl.DateTimeFormatOptions = {
+      timeZone: 'Europe/Kyiv',
+    };
+
+    formatDate(date, locale, options);
+    expect(mapDeprecatedTimezones).toHaveBeenCalledWith('Europe/Kyiv');
```
ginabak, line 20: THIS TEST ISN'T GREAT ... but like at least we're testing that the function is actually being called, AND the fact that `quilt/packages/dates/src/tests/map-deprecated-timezones.test.ts` is testing that it's mapping / converting, feels safe to me - Ideally mocking out a real "bug" would be great here, but since we can't ... this is good enough imo 😄 ❤️

packages/dates/src/utilities/formatDate.ts
jesstelford: Interesting! During your testing, did you notice any _regressions_ with this change (trying some different mapped and non-mapped timezones)? If not, I agree we should ship this as a bug fix to ensure we've covered all our bases.

ginabak: HIII JESS!!! do you mean for the `i18n` testing? Or this current one? SADLY Npm hasn't finished propogating lol, so I'm just waiting to test 😢

ginabak:
WAS ABLE TO SUCCESSFULLY tophat in WEB that this actually fixes the deprecated timezone (the screenshot is the purchase order page, where we just manually added the `.formatDate` on the page, before the fix it was rendering the 500 ReferenceError and now it looks like: 

Screenshot 2025-01-09 at 12 56 26 PM

